PR #15, opened 2020-06-02
Stop using the deprecated 32-bit time interfaces when using kernel versions >= 5.6.0

VictorNogueiraRio:
Since Linux version 5.6.0, 32-bit time interfaces have been deprecated due to y2k38
https://www.kernel.org/doc/html/latest/core-api/timekeeping.html#deprecated-time-interfaces

### DIFF
--- a/lua/loslib.c
+++ b/lua/loslib.c
@@ -328,8 +328,8 @@ static int os_date (lua_State *L) {
 
 
 static int os_time (lua_State *L) {
-#ifndef _KERNEL
   time_t t;
+#ifndef _KERNEL
   if (lua_isnoneornil(L, 1))  /* called without args? */
     t = time(NULL);  /* get current time */
   else {
@@ -351,12 +351,11 @@ static int os_time (lua_State *L) {
   l_pushtime(L, t);
   return 1;
 #else
-  struct timespec ts;
   lua_Integer res;
 
-  getnstimeofday(&ts);
-  lua_pushinteger(L, (lua_Integer)ts.tv_sec);
-  res = ts.tv_sec * NSEC_PER_SEC + ts.tv_nsec;
+  luai_time(t);
+  lua_pushinteger(L, (lua_Integer)t.tv_sec);
+  res = t.tv_sec * NSEC_PER_SEC + t.tv_nsec;
   lua_pushinteger(L, res);
 
   return 2;
--- a/lua/luaconf.h
+++ b/lua/luaconf.h
@@ -839,13 +839,23 @@ extern void longjmp(luai_jmpbuf);
 #define LUAI_TRY(L,c,a)		if (setjmp(((c)->b)) == 0) { a }
 
 /* time.h */
-#include <linux/time.h>
 #include <linux/ktime.h>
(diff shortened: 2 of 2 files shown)


lua/loslib.c
```diff
@@ -351,10 +351,17 @@ static int os_time (lua_State *L) {
   l_pushtime(L, t);
   return 1;
 #else
-  struct timespec ts;
   lua_Integer res;
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5,6,0)
```
lneto, line 355: please, move this logics to luaconf.h..

lua/loslib.c
```diff
@@ -351,10 +351,17 @@ static int os_time (lua_State *L) {
...
+  struct timespec ts;
 
   getnstimeofday(&ts);
+#else
+  struct timespec64 ts;
+
+  ktime_get_real_ts64(&ts);
+#endif
+
```
lneto, line 364: s/\n//

lua/luaconf.h
```diff
@@ -839,13 +839,28 @@ extern void longjmp(luai_jmpbuf);
...
-  ((void) p);
+
   getnstimeofday(&t);
+#else
+  struct timespec64 t;
+
+  ktime_get_real_ts64(&t);
+#endif
+
+  ((void) p);
```
lneto, line 863: please, move to the top..

lua/luaconf.h
```diff
@@ -839,13 +839,28 @@ extern void longjmp(luai_jmpbuf);
...
+#include <linux/time.h>
+#else
+#ifdef LUA_32BITS
+#error "kernel versions from 5.6.0 onwards do not support 32 bit time values"
+#endif
+#include <linux/time64.h>
+#endif
 static inline int time(void *p)
 {
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5,6,0)
```
lneto, line 853: please, don't duplicate this logic.. let's just create an abstraction layer to time..

lua/loslib.c
```diff
@@ -351,10 +351,10 @@ static int os_time (lua_State *L) {
   l_pushtime(L, t);
   return 1;
 #else
-  struct timespec ts;
+  time_t ts;
```
lneto, line 354: we don't need to duplicate this line; just move the ifndef one line below..

lua/luaconf.h
```diff
@@ -839,13 +839,24 @@ extern void longjmp(luai_jmpbuf);
...
 
 /* time.h */
-#include <linux/time.h>
 #include <linux/ktime.h>
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5,6,0)
+#include <linux/time.h>
+typedef struct timespec time_t;
+#define luai_time(t)		getnstimeofday(&t)
+#else
+#ifdef LUA_32BITS
```
lneto, line 848: #elif?

VictorNogueiraRio: I don't believe that will work.
Because the error generation only occurs when `LUA_32BITS `is defined, however the other definitions inside the `else` don't depend on that

PR #51, opened 2023-07-03
add luarcu

lneto:


lib/luarcu.c
```diff
@@ -0,0 +1,358 @@
...
+	luarcu_entry_t *entry;
+
+	keylen++; /* null-terminated */
+	entry_key = (char *)kmalloc(sizeof(const char) * keylen, GFP_ATOMIC);
+	if (entry_key == NULL)
+		return -1;
+
+	entry = luarcu_alloc(entry_key, data, len);
+	if (entry == NULL) {
+		kfree(key);
```
jpereira: Missing `kfree(entry_key)` here.

jpereira (changes requested):

lib/luarcu.c
```diff
@@ -0,0 +1,358 @@
...
+	unsigned int index = luarcu_hash(table, key, keylen);
+
+	rcu_read_lock();
+	entry = luarcu_lookup(table, index, key, keylen);
+	if (entry == NULL) {
+		lua_pushnil(L);
+		goto out;
+	}
+
+	bufflen = entry->len;
```
jpereira, line 230: `bufflen` is unnecessary. maybe just use `entry-len`

jpereira (changes requested):

PR #65, opened 2023-12-03
Support for version kernel 6.5.x

VitorMob:


lib/luasocket.c
```diff
@@ -333,8 +333,12 @@ static const lunatik_reg_t luasocket_msg[] = {
 	{"NOSIGNAL", MSG_NOSIGNAL},
 	{"MORE", MSG_MORE},
 	{"WAITFORONE", MSG_WAITFORONE},
-	{"SENDPAGE_NOPOLICY", MSG_SENDPAGE_NOPOLICY},
+	#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 4, 0)
+		
+	#else
 	{"SENDPAGE_NOTLAST", MSG_SENDPAGE_NOTLAST},
+	#endif
+	{"SENDPAGE_NOPOLICY", MSG_SENDPAGE_NOPOLICY},
```
lneto, line 341: we should not change the order to facilitate keep up with the Linux internal defines. We do not ident `#if-else-endif` in this case. Also, I see no point on the empty `#else`. I think this code should look like this:

```
	{"SENDPAGE_NOPOLICY", MSG_SENDPAGE_NOPOLICY},
#if LINUX_VERSION_CODE < KERNEL_VERSION(6, 4, 0)
	{"SENDPAGE_NOTLAST", MSG_SENDPAGE_NOTLAST},
#endif
```

lneto (changes requested):
thanks for the PR; I've made some comments.

lneto:
replaced by https://github.com/luainkernel/lunatik/pull/68

PR #101, opened 2024-03-22
prevent running already loaded script

glk0:


lunatik.lua
```diff
@@ -52,6 +52,9 @@ driver.__runtimes = {}
 driver.__threads = {}
 
 function driver:run(script, ...)
+	if self.__runtimes [script] then
```
lneto, line 55: please mind the code style on braces and parentheses.. e.g., `s/self.__runtimes [script]/self.__runtimes[script]/`

lunatik.lua
```diff
@@ -52,6 +52,9 @@ driver.__runtimes = {}
...
+		error (string.format ("%s is already running", script))
```
lneto, line 56: would it print the error message on dmesg or the command line?

lneto (changes requested):

lunatik.lua
```diff
@@ -52,6 +52,9 @@ driver.__runtimes = {}
 driver.__threads = {}
 
 function driver:run(script, ...)
+	if self.__runtimes[script] then
+		return string.format ("%s is already running", script)
```
lneto, line 56: you're still using spaces between function name and parenthesis; please mind it as whole in your patches (not only in the indicated lines).

what's the result of this line? is it printing the error on the command line (or in dmesg)?

lneto (changes requested):
please, squash your commits..

lunatik.lua
glk0: Updated it. The messages is printed on the command line, not in dmesg.

lneto (approved):

lneto:
> Any other suggestions, or is it ready to merge ?

it's approved, so ready to merge ;-).. I just want to give it a try before merging it..

lneto:
@glk0, I finally gave it a try.. we shouldn't show the line error in this case, such as..

```
$ sudo lunatik run passwd
/lib/modules/lua/lunatik.lua:56: passwd is already running
```

instead, it should print..

```
$ sudo lunatik run passwd
'passwd' is already running
```

and `lunatik` command should return something different from `0`.. the same should apply to other `lunatik` errors in the command line.. one example is `lunatik stop` that is failing silently.. perhaps we should create an issue for that..

I'm merging it anyway because it's better that what we have now ;-)..

thanks..

PR #104, opened 2024-05-01
add luaxdp

lneto:


lneto:
@sheharyaar please add your findings about bpf macros here as a comment/suggestion.. thanks!

sheharyaar:
We can use __bpf_kfunc_start_defs() and the corresponding end function instead of __diag_push() and others.  
This macro is introduced in kernel 6.7 : https://elixir.bootlin.com/linux/v6.7/source/include/linux/btf.h#L87

lneto:
> We can use __bpf_kfunc_start_defs() and the corresponding end function instead of __diag_push() and others. This macro is introduced in kernel 6.7 : https://elixir.bootlin.com/linux/v6.7/source/include/linux/btf.h#L87

next time; please comment inline ;-) thank you!

sheharyaar:
`LUNATIK_NEWLIB` does not use lunatik_class. I am confused about classes in lunatik. The documentation doesn't have notes regarding this.

lneto:
> `LUNATIK_NEWLIB` does not use lunatik_class. I am confused about classes in lunatik.

yup, we don't actually have XDP objects by now.. I think on the next increment we will have ;-)..

> The documentation doesn't have notes regarding this.

Indeed.. it's kinda a "hidden" facility right now, but I think it would be a good task for you to handle to document this API.. what do you think @sheharyaar ?

sheharyaar:
Yes, I will handle the documentation for this

lneto:
> Yes, I will handle the documentation for this

please, create an issue for that then =)

sheharyaar (approved):
LGTM.

Issue #105, opened 2024-05-02
[GSoC 2024] update README.md with the base API for luanetfilter 


lneto:
Add the documentation for the base API for luanetfilter; that is, only for the functions that are strictly necessary for the base functionality of calling Lua from Netfilter rules. It shouldn't contain packet handling yet; just callbacks registering and actions (e.g., `PASS`, `DROP`). Thus, you can create issues, after discussing this specification, for implementing this foundation.

sheharyaar:
For xtables `match` and `target`, we need to implement the following :
- register_pernet_subsys()
- xt_(un)register_match(es)()
- xt_(un)register_target(s)()

For netfilter hooks, we need to implement :
- struct nf_hook_ops
- Netfilter verdict helpers (NF_ACCEPT, NF_DROP, NF_STOLEN, NF_REPEAT,
NF_STOP)
- Implement nf_(un)register_hook(s)

This will provide the base functionality of registering callbacks and actions.
Future work will require implementing user-space libraries, skb handlers, packet handling, conntrack, luasocket port, netlink support

lneto:
@sheharyaar this issue is to create the proper documentation on README.md for the base API you expose to the Lua scripts, you can use any Lunatik library documentation as reference (e.g., [device](https://github.com/luainkernel/lunatik#device)). I would expect you to create a PR specifying such base/minimal API for the hooks.

sheharyaar:
Can be closed as completed

PR #106, opened 2024-05-02
update README with luanetfilter base API

sheharyaar:
Update base API details for basic registering and callbacks of netfilter modules using Lunatik.
See #105

### DIFF
--- a/README.md
+++ b/README.md
@@ -933,6 +933,150 @@ _syscall.number()_ returns the system call number referenced by the given `name`
 
 The `syscall.table` library provides support for translating system call names to addresses (light userdata).
 
+### netfilter
+
+The `netfilter` library provides support for the kernel [Netfilter](https://netfilter.org/) subsystem. This includes support for xtables (`match`, `target`) and netfilter hooks.
+
+#### `netfilter.xtable(match, target)`
+
+_netfilter.xtable(ops)_ registers a xtable with the specified match and target operations. The `match` and `target` are tables with the following fields:
+
+* `name`: the name of the xtable.
+* `revision`: the revision of the xtable.
+* `match` or `target`: the callback function that is called when the xtable is matched or targeted. The match callback should return either `NF_DROP` or `NF_ACCEPT` and the target callback should return one of the following values:
+  * `NF_ACCEPT`: accept the packet.
+  * `NF_DROP`: drop the packet.
+  * `NF_STOLEN`: the packet is stolen by the hook.
+  * `NF_QUEUE`: the packet is queued.
+  * `NF_REPEAT`: the packet is repeated.
+  * `NF_STOP`: stop the traversal.
+* `checkentry`: the callback function that is called when the xtable is checked.
+* `destroy`: the callback function that is called when the xtable is destroyed.
+* `family`: the protocol family. This is one of :
+  * `NFPROTO_UNSPEC` (unspecified).
+  * `NFPROTO_INET` (IPv4).
+  * `NFPROTO_IPV4` (IPv4).
+  * `NFPROTO_ARP` (ARP).
+  * `NFPROTO_NETDEV` (network device).
+  * `NFPROTO_BRIDGE` (bridge).
+  * `NFPROTO_IPV6` (IPv6).
+* `proto`: the protocol. This is one of (See [socket-ipproto](#socketipproto)) :
+  * TCP
+  * UDP
+  * ICMP
+  * ICMPv6
+  * DCCP
+  * SCTP
+  * UDPLITE
+  * RAW
+
+The userspace xtable library is used to provide help message, parse logic, sanity check, pass program specific data to the kernel space and to save the iptables state in the userspace.
+
+The arguments to the match and target callbacks have `skb` and `params`. The `skb` is the packet buffer and `params` is the program specific data passed to the kernel space from the xtables userspace library. The `skb` is a `sk_buff` structure which contains the packet data. The `params` is a `struct xt_entry_match` or `struct xt_entry_target` structure which contains the program specific data passed to the kernel space. 
+
+For the checkentry callback, the argument is a `struct xt_mtchk_param` or `struct xt_tgcheck_param` and for the destroy callback, the argument is a `struct xt_mtdtor_param` or `struct xt_tgdtor_param`.
+
+#### Example Script
+
+```lua
+local nf = require("netfilter")
+local action = nf.action
+local xt = nf.xt
+local socket = require("socket")
+
+function drop_packet(skb, params)
+	print("drop_packet called")
+	return action.DROP
+end
+
+function ignore_packet(skb, params)
+	print("ignore_packet called")
+	return xt.CONTINUE
+end
+
+local ipproto = socket.ipproto
+
+local match_ops = {
+	name= "mymatch",
+	revision=1,
+	match= drop_packet,
+	checkentry=nil,
+	destroy=nil,
+	family= nf.NFPROTO_INET,
+	proto= ipproto.TCP
+}
+
+local target_ops = {
+	name= "mytarget",
+	revision=1,
+	target= ignore_packet,
+	checkentry=nil,
+	destroy=nil,
+	family= nf.NFPROTO_INET,
+	proto= ipproto.TCP
+}
+
+nf.xtable{
+	match= match_ops,
+	target= target_ops
+}
+```
+
+#### `netfilter.hook(ops)`
+
+_netfilter.hook(ops)_ registers a netfilter hook with the specified `ops`. The `ops` is a table with the following fields:
+
+* `hook`: a callback function that returns a value one of the following values:
+  * `NF_ACCEPT`: accept the packet.
+  * `NF_DROP`: drop the packet.
+  * `NF_STOLEN`: the packet is stolen by the hook.
+  * `NF_QUEUE`: the packet is queued.
+  * `NF_REPEAT`: the packet is repeated.
+  * `NF_STOP`: stop the traversal.
+
+* `priority`: the priority of the hook.
+
+* `pf`: the protocol family. This is one of :
+  * `NFPROTO_UNSPEC` (unspecified).
+  * `NFPROTO_INET` (IPv4).
+  * `NFPROTO_IPV4` (IPv4).
+  * `NFPROTO_ARP` (ARP).
+  * `NFPROTO_NETDEV` (network device).
+  * `NFPROTO_BRIDGE` (bridge).
+  * `NFPROTO_IPV6` (IPv6).
+
+* `hooknum`: the hook number. This is one of :
+  * `NF_INET_PRE_ROUTING` (pre-routing).
+  * `NF_INET_LOCAL_IN` (local input).
+  * `NF_INET_FORWARD` (forward).
+  * `NF_INET_LOCAL_OUT` (local output).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -933,7 +933,62 @@ _syscall.number()_ returns the system call number referenced by the given `name`
 
 The `syscall.table` library provides support for translating system call names to addresses (light userdata).
 
-# Examples
```
lneto, line 936: Why removing this line?

lneto:
@sheharyaar please provide an example of how you are planning to use this API; it might be a simple script for dropping everything.

lneto:
> Added a sample example, this example may change a bit in the future, depending on the work progress.

no worries; the goal of this task is only to define a base API for starting the implementation.. the plan is to develop it incrementally and improve it over time ;-)

README.md
```diff
@@ -933,6 +933,92 @@ _syscall.number()_ returns the system call number referenced by the given `name`
...
+_netfilter.hooks(hooks)_ registers multiple `hooks`. The `hooks` is a table that contains the ops as described above.
+
+#### Example Script
+
+```lua
+local nf = require("netfilter")
+
+function drop_packet(skb)
+	print("drop_packet called")
+	return false
```
lneto, line 1000: shouldn't use `NF_DROP` here?

sheharyaar: NF_DROP can be used, but it may give the idea that `NF_*` values are to be used, however, only boolean values can be used (NF_DROP and NF_ACCEPT qualify this).

README.md
```diff
@@ -933,6 +933,92 @@ _syscall.number()_ returns the system call number referenced by the given `name`
 
 The `syscall.table` library provides support for translating system call names to addresses (light userdata).
 
+### netfilter
+
+The `netfilter` library provides support for the kernel [Netfilter](https://netfilter.org/) subsystem. This includes support for xtables (`match`, `target`) and netfilter hooks.
+
+#### `netfilter.xtable(name)`
+
+_netfilter.xtable(name)_ creates a new `netfilter.xtable` object for the specified `name`. The 'name' must be unique and must not be used by any other xtable. This name will be used to create userspace and kernel objects for the xtable when `match` and `target` objects are added to the xtable.
```
lneto, line 942: can you clarify how xtable will be used from the userspace?

README.md
```diff
@@ -933,6 +933,92 @@ _syscall.number()_ returns the system call number referenced by the given `name`
...
+
+#### `netfilter.match(match_func), nf:match(match_func)`
+
+_netfilter.match(match_func)_ registers the specified `match_func`. The `match_func` is a function that will be called when a packet is processed by the xtable. The function must return boolean value indicating whether the packet should be accepted or dropped.
+
+#### `netfilter.matches(matches)`
```
lneto, line 948: do we really need this? if yes, why?

README.md
```diff
@@ -933,6 +933,92 @@ _syscall.number()_ returns the system call number referenced by the given `name`
...
+
+_netfilter.matches(matches)_ registers multiple `matches`. The `matches` is a table that contains the name of the filter object and the corresponding match functions that will be called when a packet is processed by the xtable. The function must return boolean value indicating whether the packet should be accepted or dropped.
+
+#### `netfilter.target(target_func), nf:target(target_func)`
+
+_netfilter.target(target_func)_ registers the specified `target_func`. The `target_func` is a function that will be called when a packet is processed by the xtable. The function must return a boolean value indicating whether the packet should be accepted or dropped.
+
+#### `netfilter.targets(targets)`
```
lneto, line 956: do we really need this? if yes, why?

README.md
```diff
@@ -933,6 +933,92 @@ _syscall.number()_ returns the system call number referenced by the given `name`
...
+  * `NFPROTO_IPV6` (IPv6).
+* `hooknum`: the hook number. This is one of :
+  * `NF_INET_PRE_ROUTING` (pre-routing).
+  * `NF_INET_LOCAL_IN` (local input).
+  * `NF_INET_FORWARD` (forward).
+  * `NF_INET_LOCAL_OUT` (local output).
+  * `NF_INET_POST_ROUTING` (post-routing).
+  * `NF_INET_NUMHOOKS` (number of hooks).
+
+#### `netfilter.hooks(hooks)`
```
lneto, line 989: do we really need this? if yes, why?

README.md
```diff
@@ -933,6 +933,92 @@ _syscall.number()_ returns the system call number referenced by the given `name`
...
+
+#### `netfilter.target(target_func), nf:target(target_func)`
+
+_netfilter.target(target_func)_ registers the specified `target_func`. The `target_func` is a function that will be called when a packet is processed by the xtable. The function must return a boolean value indicating whether the packet should be accepted or dropped.
+
+#### `netfilter.targets(targets)`
+
+_netfilter.targets(targets)_ registers multiple `targets`. The `targets` is a table that contains the name of the filter object and the corresponding target functions that will be called when a packet is processed by the xtable. The function must return a boolean value indicating whether the packet should be accepted or dropped.
+
+#### `netfilter.hook(ops)`
```
lneto, line 960: it's not clear to me how it would be used; can you exemplify it as well?

README.md
```diff
@@ -933,6 +933,92 @@ _syscall.number()_ returns the system call number referenced by the given `name`
...
+
+The `netfilter` library provides support for the kernel [Netfilter](https://netfilter.org/) subsystem. This includes support for xtables (`match`, `target`) and netfilter hooks.
+
+#### `netfilter.xtable(name)`
+
+_netfilter.xtable(name)_ creates a new `netfilter.xtable` object for the specified `name`. The 'name' must be unique and must not be used by any other xtable. This name will be used to create userspace and kernel objects for the xtable when `match` and `target` objects are added to the xtable.
+
+#### `netfilter.match(match_func), nf:match(match_func)`
+
+_netfilter.match(match_func)_ registers the specified `match_func`. The `match_func` is a function that will be called when a packet is processed by the xtable. The function must return boolean value indicating whether the packet should be accepted or dropped.
```
lneto, line 946: it actually should return any netfilter action (e.g., `NF_STOLEN`) instead only drop/accept, right?

README.md
lneto: why it can't be used? if I recall correctly nflua uses it.

sheharyaar: Yes, `NF_*` can be used for `target` but not for `match` (since callback should return a bool) , so I will create a different function for target callback. nflua also uses this for targets only.

README.md
sheharyaar: `xt_register_match` takes `struct xt_match` as an argument :

```c
          /* Return true or false: return FALSE and set *hotdrop = 1 to
             force immediate packet drop. */
          /* Arguments changed since 2.6.9, as this must now handle
             non-linear skb, using skb_header_pointer and
             skb_ip_make_writable. */
          bool (*match)(const struct sk_buff *skb,
                        struct xt_action_param *); 
```

This is the callback signature from `struct xt_match`

README.md
sheharyaar: To allow registration of multiple matches, the kernel has the corresponding function : `xt_register_matches`. Making it part of the base API depends on the community usage.

sheharyaar: We may not need this as the registration can be done in a loop too, this is what the function does internally.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "C_Cpp.default.intelliSenseMode": "linux-gcc-x64"
+}
```
lneto, line 3: please be careful with the files you add to your commits.. it should be a good opportunity to check if you are using hard tabs on your editor..

lneto:
@sheharyaar btw, these commits should be squashed.. no need to split those changes in separate commits.. you may want to use `--fixup` while you are still make changes and then `rebase -i --autosquash` in the end (or time to time)..

sheharyaar:
Changes: added some extra parameters for match and target, these parameters will be refined while working on the APIs.

README.md
```diff
@@ -933,6 +933,149 @@ _syscall.number()_ returns the system call number referenced by the given `name`
...
+	name= "mytarget",
+	revision=1,
+	target= ignore_packet,
+	checkentry=nil,
+	destroy=nil,
+	family= nf.NFPROTO_INET,
+	proto= nf.IPPROTO_TCP
+}
+
+local ok = xtable:match{
```
lneto, line 1014: I think we should use Lua error handling instead of returning error conditions.. unless you have a good reason for it..

lneto, line 1014: where did you instantiate this `xtable` object? it's not clear.. perhaps it should be something like that..

```
nf.xtable{
```

sheharyaar: Do you mean using pcall ?

sheharyaar: Ah, yes, I forgot it among the changes.

README.md
```diff
@@ -933,6 +933,149 @@ _syscall.number()_ returns the system call number referenced by the given `name`
...
+For the checkentry callback, the argument is a `struct xt_mtchk_param` or `struct xt_tgcheck_param` and for the destroy callback, the argument is a `struct xt_mtdtor_param` or `struct xt_tgdtor_param`.
+
+#### Example Script
+
+```lua
+local nf = require("netfilter")
+
+function drop_packet(skb, params)
+	print("drop_packet called")
+	return nf.NF_DROP
```
lneto, line 986: probably will be simpler to use the Lunatik flags facility for your definitions.. e.g.,

```
local action = nf.action
...
function drop_packet(skb, params)
	print("drop_packet called")
	return action.DROP
```

if you think `action` is not a good name, you can call anything else..

README.md
```diff
@@ -933,6 +933,149 @@ _syscall.number()_ returns the system call number referenced by the given `name`
...
+end
+
+function ignore_packet(skb, params)
+	print("ignore_packet called")
+	return nf.XT_CONTINUE
```
lneto, line 991: same here.. e.g.,
```
local xt = nf.xt
...
return xt.CONTINUE
```

README.md
sheharyaar: Looks good

README.md
```diff
@@ -933,6 +933,149 @@ _syscall.number()_ returns the system call number referenced by the given `name`
...
+end
+
+local match_ops = {
+	name= "mymatch",
+	revision=1,
+	match= drop_packet,
+	checkentry=nil,
+	destroy=nil,
+	family= nf.NFPROTO_INET,
+	proto= nf.IPPROTO_TCP
```
lneto, line 1001: `IPPROTO_TCP` doesn't belong to NF, you might want to add this to lualinux or create another binding for network definitions.. not sure if socket is already exposing this.. if so, perhaps we should move from there..

sheharyaar: Understood

lneto:
@sheharyaar perhaps a good introductory task is to create the skeleton luanetfilter module with exposing only some netfilter definitions using the Lunatik flags support.. it will be a good opportunity to figure out how to use such facilities and how to integrate a new module in the framework..

sheharyaar:
I agree, I will study the current features available in lunatik and how we can extend it, depending on the needs felt in this module.

README.md
lneto: you actually don't need to use `pcall` here because the driver will do it for you.. but yes, using `pcall` on the caller and `error()` on callee

sheharyaar:
This is done in PR #151.

Issue #107, opened 2024-05-03
[GSoC 2024] xtable API for luanetfilter


sheharyaar:
Issue to track xtable APIs and features for luanetfilter module.
Design issue : #105

sheharyaar:
Deliverables checkilst :
- [ ] xtables registration with the kernel
- [ ] userspace library generation and adding API functions for the userspace library
- [ ] adding skbuff helpers for interaction with the packet (porting luapacket)
- [ ] adding helpers for xtable `match/target` params, `mtchk/tgchk` params, `mtdtor/tgdtor` params for xtable callbacks

jperon:
Hello,
following this issue with much interest, I was wondering what would be the nftables equivalent of following command (from Readme.md about dnsblock):

```sh
sudo iptables -A OUTPUT -m dnsblock -j DROP
```

Thank you very much for the hard work!

sheharyaar:
Hi, thanks for the compliment. I am currently working on nftable implementation or the same. I will update here once that is done :)

jperon:
Many thanks for your quick reply; can’t wait for the update!

jperon:
Not sure it’s the right place for questions, but I had the following one: does the `skb` argument of `match` function passed to `xtable.match` contain the whole ip packet, or only the header part?

sheharyaar:
The whole packet starting from the ip header.

sheharyaar:
In the `examples/dnsblock/dnsblock.lua`, we extract the dns name because the skb includes the entire packet (https://github.com/luainkernel/lunatik/blob/master/examples/dnsblock/dnsblock.lua#L51). In this case `thoff` is the transport header offset (ipv4 header offsert) and the dns (udp data) is 20 bytes from the thoff.

jperon:
Thanks again for the reply. My question came because, trying to develop a SNI filter like https://github.com/luainkernel/lunatik/blob/master/examples/filter/sni.lua, but using `xtables` instead of `xdp`, I always get `#skt == 52` with this rule: `iptables  -A OUTPUT -p tcp --dport 443 -m sni -j REJECT`, event when the `Total length` of the ip packet (obtained with `short(2)`) is bigger. And in fact, if I try `skt:getbyte(53)`, I get this error in kernel logs: `luaxtable: match error: bad argument #2 to '?' (out of bounds)`.

lneto:
[...]

PR #111, opened 2024-05-11
[GSoC 2024] luanetfilter skeleton with flags

sheharyaar:
PR to add luanetfilter skeleton with netfilter flags
Issue : https://github.com/luainkernel/lunatik/issues/107


lib/luanetfilter.c
```diff
@@ -0,0 +1,113 @@
...
+* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+#define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+#include <linux/module.h>
+#include <linux/version.h>
+#include <linux/printk.h>
+#include <linux/types.h>
+#include <linux/list.h>
```
lneto, line 29: please remove unused headers

lib/luanetfilter.c
```diff
@@ -0,0 +1,113 @@
...
+#include <linux/netfilter.h>
+#include <linux/netfilter/x_tables.h>
+#include <net/net_namespace.h>
+
+#include <lua.h>
+#include <lauxlib.h>
+#include <lualib.h>
```
lneto, line 36: like this as well..

lib/luanetfilter.c
```diff
@@ -0,0 +1,113 @@
...
+static struct luanetfilter_xtable_s {
+} luanetfilter_xtable_t ;
+
+static void luanetfilter_release(void *private)
+{
+}
+
+static int luanetfilter_xtable(lua_State *L)
+{
+	return 0;
```
lneto, line 49: as you are defining the class, I would create and push the object here..

lib/luanetfilter.c
```diff
@@ -0,0 +1,113 @@
...
+}
+
+static const lunatik_reg_t netfilter_protocol[] = {
```
lneto, line 52: let's prefix everything with `luanetfilter_` instead of `netfilter`, so we know easily that the symbol is ours.. if you think `luanetfilter` is too verbose, I'm not against name it `luanf`.. mixed feelings though ;-)

lib/luanetfilter.c
sheharyaar: This will be done in the next PR with the basic functionilities

lib/luanetfilter.c
sheharyaar: luanetfilter seems better than luanf

lib/luanetfilter.c
lneto: So, I would add class only in the next PR.. or simply create a dummy object of that class here..

lunatik
```diff
@@ -26,7 +26,7 @@ local lunatik = {
 	copyright = "Copyright (C) 2023-2024 ring-0 Ltda.",
 	device = "/dev/lunatik",
 	modules = {"lunatik", "luadevice", "lualinux", "luanotifier", "luasocket", "luarcu",
-		"luathread", "luafib", "luadata", "luaprobe", "luasyscall", "luaxdp", "lunatik_run"}
+		"luathread", "luafib", "luadata", "luaprobe", "luasyscall", "luaxdp", "lunatik_run","luanetfilter"}
```
lneto, line 29: `lunatik_run` should be the last module..

lib/luanetfilter.c
```diff
@@ -0,0 +1,105 @@
...
+#include <linux/netfilter.h>
+#include <linux/netfilter/x_tables.h>
+
+#include <lunatik.h>
+
+static struct luanetfilter_xtable_s {
+} luanetfilter_xtable_t ;
+
+static void luanetfilter_release(void *private)
+{
```
lneto, line 36: `(void)private`

sheharyaar: should be (void *private), right ?

lneto: sorry, my bad.. I was referring to https://github.com/luainkernel/lunatik/blob/master/lib/luaskel.c#L49, it wasn't need it this case..

lib/luanetfilter.c
```diff
@@ -0,0 +1,122 @@
...
+	luanetfilter_xtable_t *luaxtable;
+
+	object = lunatik_newobject(L, &luanetfilter_class, sizeof(luanetfilter_xtable_t));
+	luaxtable = (luanetfilter_xtable_t *)object->private;
+
+	memset(luaxtable, 0, sizeof(luanetfilter_xtable_t));
+
+	luaxtable->runtime = lunatik_toruntime(L);
+	lunatik_getobject(luaxtable->runtime);
+
```
lneto, line 103: sorry to be picky here, but we avoid this new line before `return` on the other libraries.. but I will leave the decision to you.. (just saying =)..

lneto (commented):
it's looking good.. quite minor comments only

lib/luanetfilter.c
```diff
@@ -0,0 +1,118 @@
...
+* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+#define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+#include <linux/module.h>
+#include <linux/version.h>
+#include <linux/netfilter.h>
+#include <linux/netfilter/x_tables.h>
+
```
lneto, line 29: we should have
```
#include <lua.h>
#include <lauxlib.h>
```

as we use `lua_State` and `luaL_Reg`, I was referring to `lualib.h` only before..

lneto (approved):

PR #112, opened 2024-05-11
add rm, install and mkdir variables to makefile

sheharyaar:
When running `make clean` and other rules that use `rm` to remove fils/folder, these give out errors if the reffered file/folder does not exist. Hence this PR adds `-f` to ignore nonexistent files and arguments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,9 @@ SCRIPTS_INSTALL_PATH = /lib/modules/lua
 LUNATIK_INSTALL_PATH = /usr/local/sbin
 LUA_API = lua/lua.h lua/lauxlib.h lua/lualib.h
 KDIR ?= ${MODULES_INSTALL_PATH}/build
+RM = rm -f
+MKDIR = mkdir -p -m 0755
+INSTALL = install -o root -g root
 
 all: lunatik_sym.h
 	make -C ${KDIR} M=${PWD} CONFIG_LUNATIK=m	\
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,7 +14,7 @@ all: lunatik_sym.h
 
 clean:
 	make -C ${KDIR} M=${PWD} clean
-	rm lunatik_sym.h
+	rm -f lunatik_sym.h
```
lneto, line 17: perhaps a better idea is to put this in a var.. e.g.,
```
RM = rm -f
```

and while there do the same for mkdir, and install.. just a suggestion, I can merge as is if you disagree..

sheharyaar: Yes, that is much better

sheharyaar: I will make the changes

sheharyaar:
`install` hade params with values so I omitted it.

Makefile
```diff
@@ -3,6 +3,8 @@ SCRIPTS_INSTALL_PATH = /lib/modules/lua
 LUNATIK_INSTALL_PATH = /usr/local/sbin
 LUA_API = lua/lua.h lua/lauxlib.h lua/lualib.h
 KDIR ?= ${MODULES_INSTALL_PATH}/build
+RM = rm -f
+MKDIR = mkdir -p
```
lneto, line 7: Please, split this in another commit or update the commit message to reflect all your changes..

Also, I think you could add the permission to this var as well..

lneto:
> `install` had params with values so I omitted it.

perhaps.. 
```
...
${INSTALL} -m 0755 ...
...
${INSTALL} -m 0644 ...
```

what do you think?

sheharyaar:
Yes, this can be done.

sheharyaar:
Done

lneto (approved):

PR #113, opened 2024-05-15
[GSoC 2024] add xtable opts parsing and basic registration

sheharyaar:


lib/luanetfilter.c
```diff
@@ -99,18 +190,33 @@ static int luanetfilter_xtable(lua_State *L)
...
+
+	lunatik_registerobject(L, 1, object); /* xtable filter */
 	return 1;
 }
 
 LUNATIK_NEWLIB(netfilter, luanetfilter_lib, &luanetfilter_class, luanetfilter_flags);
 
 static int __init luanetfilter_init(void)
 {
+	pr_err("[luanetfilter] init module");
```
lneto, line 213: let's not merge debug printing

lib/luanetfilter.c
```diff
@@ -99,18 +190,33 @@ static int luanetfilter_xtable(lua_State *L)
 
 	luaxtable->runtime = lunatik_toruntime(L);
 	lunatik_getobject(luaxtable->runtime);
+
+	luaL_checktype(L, 1, LUA_TTABLE);
+
+	/* handle match and target */
+	lua_getfield(L, 1, "match");
+	if(handle_xtable_reg(L, luaxtable, LUANETFILTER_MATCH) != 0 )
+		luaL_error(L, "[luanetfilter] handle match error - %s\n", lua_tostring(L, -1));
```
lneto, line 199: you don't need to prefix the error message..

lib/luanetfilter.c
```diff
@@ -43,6 +73,67 @@ static void luanetfilter_release(void *private)
...
+	if (mode == LUANETFILTER_MATCH)
+		luanetfilter_xtable_op_func(L, base, "match");
+	else
+		luanetfilter_xtable_op_func(L, base, "target");
+
+	/* Print for debugging */
+	pr_err("[luanetfilter] target name : %s\n", name);
+	pr_err("[luanetfilter] target revision : %u\n", revision);
+	pr_err("[luanetfilter] target family : %d\n", family);
+	pr_err("[luanetfilter] target proto : %d\n", proto);
```
lneto, line 123: let's no add this to PRs

lib/luanetfilter.c
```diff
@@ -43,6 +73,67 @@ static void luanetfilter_release(void *private)
...
+	luanetfilter_xtable_op(L, base, "proto", lua_isnumber);
+	proto = (unsigned short)lua_tointeger(L, -1);
+
+	luanetfilter_xtable_op_func(L, base, "checkentry");
+	luanetfilter_xtable_op_func(L, base, "destroy");
+
+	if (mode == LUANETFILTER_MATCH)
+		luanetfilter_xtable_op_func(L, base, "match");
+	else
+		luanetfilter_xtable_op_func(L, base, "target");
```
lneto, line 117:
```
luanetfilter_xtable_op_func(L, base, mode == LUANETFILTER_MATCH ? "match" : "target");
```

lib/luanetfilter.c
```diff
@@ -43,6 +73,67 @@ static void luanetfilter_release(void *private)
 	lunatik_putobject(luaxtable->runtime);
 }
 
+/* Register the target with the netfilter module. The caller must push the table on the stack.
+ * Pops the table from the stack both on success and failure.
+ * Pushes the error string on the stack on failure and returns -1.
+ * */
+static int handle_xtable_reg(lua_State *L, luanetfilter_xtable_t *luaxtable, LUANETFILTER_MODE mode)
```
lneto, line 80: let's prefix everything with `luanetfilter_`, e.g., `luanetfilter_xtable_register`

sheharyaar: Got it

lib/luanetfilter.c
sheharyaar: yeah, I will remove these comments before getting the PR merged.

lib/luanetfilter.c
```diff
@@ -99,18 +190,33 @@ static int luanetfilter_xtable(lua_State *L)
...
+
+	luaL_checktype(L, 1, LUA_TTABLE);
+
+	/* handle match and target */
+	lua_getfield(L, 1, "match");
+	if(handle_xtable_reg(L, luaxtable, LUANETFILTER_MATCH) != 0 )
+		luaL_error(L, "[luanetfilter] handle match error - %s\n", lua_tostring(L, -1));
+
+	lua_getfield(L, 1, "target");
+	if(handle_xtable_reg(L, luaxtable, LUANETFILTER_TARGET) != 0 )
```
lneto, line 202: it's not clear to me why you are using this enum instead of just passing the string parameter around.. e.g.,
```
luanetfilter_xtable_register(L, luaxtable, "target");
```

lib/luanetfilter.c
```diff
@@ -43,6 +73,67 @@ static void luanetfilter_release(void *private)
...
+ * */
+static int handle_xtable_reg(lua_State *L, luanetfilter_xtable_t *luaxtable, LUANETFILTER_MODE mode)
+{
+	int base = lua_gettop(L);
+	int type;
+
+	const char *name = NULL;
+	u_int8_t revision = 0;
+	unsigned short family = 0;
+	unsigned short proto = 0;
```
lneto, line 88: why filling this instead of the actual struct?

lib/luanetfilter.c
```diff
@@ -43,6 +73,67 @@ static void luanetfilter_release(void *private)
...
+	if (type == LUA_TNIL){
+		pr_err("[luanetfilter] no ops found, assuming nil\n");
+		return 0;
+	} else if (type != LUA_TTABLE){
+		lua_pushstring(L, "invalid ops type, expected table or nil");
+		goto err;
+	}
+
+	luanetfilter_xtable_op(L, base, "name", lua_isstring);
+	name = lua_tostring(L, -1);
```
lneto, line 100: you can't persist strings in this way.. take a look at the manual..

sheharyaar: I had this just for debugging, the storage to registry will come after this.

lib/luanetfilter.c
```diff
@@ -35,6 +36,35 @@ typedef struct luanetfilter_xtable_s {
...
 
+typedef enum {
+	LUANETFILTER_MATCH = 0,
+	LUANETFILTER_TARGET,
+} LUANETFILTER_MODE;
+
+#define luanetfilter_xtable_op(L, base, op, cmp)	\
+do {							\
+	lua_getfield(L, base, op);			\
+	if(!cmp(L, -1)){				\
```
lneto, line 47: please, mind the code style.. `) {` instead of `){`

lib/luanetfilter.c
sheharyaar: In the new commits the if..else block need the enums, otherwise I can use bool too, if that is better ?

lib/luanetfilter.c
```diff
@@ -35,6 +36,30 @@ typedef struct luanetfilter_xtable_s {
...
 
+typedef enum {
+	LUANETFILTER_MATCH = 0,
+	LUANETFILTER_TARGET,
+} LUANETFILTER_MODE;
+
+#define luanetfilter_xtable_op(L, base, op, cmp)	\
+do {							\
+	lua_getfield(L, base, op);			\
+	if(!cmp(L, -1)) {				\
```
lneto, line 47: please, mind the code style.. `if (`, not `if(`

lib/luanetfilter.c
```diff
@@ -35,15 +36,115 @@ typedef struct luanetfilter_xtable_s {
...
+/* Verifies the type of "opt" in the table at "idx" and returns true if the value can be extracted.
+ * If the value is not required, it returns false.
+ * If the value is required and the type is invalid, it raises an error.
+ * */
+static bool luanetfilter_xtable_opt(lua_State *L, int idx, const char *opt, int type, bool req)
 {
-	luanetfilter_xtable_t *luaxtable = (luanetfilter_xtable_t *)private;
-	lunatik_putobject(luaxtable->runtime);
+	lua_getfield(L, idx, opt);
+	int type_ = lua_type(L, -1);
```
lneto, line 106: it looks like you are kinda reimplementing `luaL_check*` APIs.. please take a look at the reference manual and verify if none suits to you..

lib/luanetfilter.c
```diff
@@ -89,19 +191,27 @@ static const lunatik_class_t luanetfilter_class = {
...
 {
-	lunatik_object_t *object;
-	luanetfilter_xtable_t *luaxtable;
-
-	object = lunatik_newobject(L, &luanetfilter_class, sizeof(luanetfilter_xtable_t));
-	luaxtable = (luanetfilter_xtable_t *)object->private;
+	if (match) {
+		xt_unregister_match(match);
+		lunatik_free(match);
+		match = NULL;
```
lneto, line 199: you can move this to a "template" as well.. `luanetfilter_xtable_unreg`

lib/luanetfilter.c
sheharyaar: I needed this because some parameters are optional and some are mandatory, so for optional parameters we can use a default value.

So, if the check is successful -> returns true -> take the value from the stack
if the check is unsuccessful -> it is required -> raises luaL_error
if the check is unsuccessful -> it is _not_ required -> returns false -> pick a default value

lib/luanetfilter.c
```diff
@@ -89,19 +191,27 @@ static const lunatik_class_t luanetfilter_class = {
...
 
-static int luanetfilter_xtable(lua_State *L)
+static void luanetfilter_release(void *private)
 {
-	lunatik_object_t *object;
-	luanetfilter_xtable_t *luaxtable;
-
-	object = lunatik_newobject(L, &luanetfilter_class, sizeof(luanetfilter_xtable_t));
-	luaxtable = (luanetfilter_xtable_t *)object->private;
+	if (match) {
```
lneto, line 196: these globals (match and target) are temporary while we don't have the RCU table, right?

lib/luanetfilter.c
sheharyaar: This will not be required in the future versions when we introduce rcu hashtable, at that time we get the object from the registry and free it.

lib/luanetfilter.c
sheharyaar: Yes, exactly

lib/luanetfilter.c
lneto: can't you use [luaL_opt*](https://www.lua.org/manual/5.4/manual.html#luaL_opt)?

lib/luanetfilter.c
lneto: I don't think it should be the case here.. you have the object and it should hold references to match and target structures.. btw, I would do this change promptly.. that is, move `match` and `target` structures to `luanetfilter_xtable_t`.. and having a global `luanetfilter` object while you don't have the RCU table

lib/luanetfilter.c
lneto: we might at least use their implementation as a base to implement this function.. replicating the same idea but for table fields..

sheharyaar: For datatypes other than functions, I have proceeded as follows. Is this fine ?? : 
```c
	lua_getfield(L, 1, "name");				   		\
	const char *name = luaL_checkstring(L, -1);				\
										\
	lua_getfield(L, 1, "revision");				   		\
	uint8_t revision = (uint8_t)luaL_checkinteger(L, -1);			\
										\
	lua_getfield(L, 1, "family");				   		\
	unsigned short family = (unsigned short)luaL_checkinteger(L, -1);	\
                                                                                \
	/*optional parameter*/							\
	lua_getfield(L, 1, "proto");				   		\
	unsigned short proto = (unsigned short)luaL_optinteger(L, -1, 0);	\
```

For functions I use the following (the functions can be nil and no default value is needed, but they should be valid) :
```c
#define luanetfilter_xtable_func(L, idx, name)					\
do {  								          	\
	lua_getfield(L, idx, name);				   	  	\
	if(!lua_isnoneornil(L, -1)) luaL_checktype(L, -1, LUA_TFUNCTION); 	\
} while(0)

...

luanetfilter_xtable_func(L, 1, "checkentry");				\
luanetfilter_xtable_func(L, 1, "destroy");				\
```

lib/luanetfilter.c
lneto: > Can I proceed like this ?

yup.. I think it's a good approach, I would just get rid of the bitfield.. it doesn't make much sense here, IMHO.. as it's not a packed structure, we will have padding there anyway.. also, I would have typedef for this enum and name it `luanetfilter_type_t` to simplify the namespace..

I really like this.. so we can have a the same `:unregister()` method for any kind of hook.. good one ;-)

lib/luanetfilter.c
lneto: > For datatypes other than functions, I have proceeded as follows. Is this fine ?? :
> ...

the only issue I see with this approach is the error messages will be a little misleading.. it will refer to a unknown argument.. I think we can leave this as is and add a TODO comment to create a `lunatik_checkfield()` API to check the type of a table field and raise a proper error message.. something like [this](https://github.com/luainkernel/lunatik/blob/master/lib/luadevice.c#L289).. 

> For functions I use the following (the functions can be nil and no default value is needed, but they should be valid) :
> ...

why does the callback can be nil? what's the use case for that?

sheharyaar: If the user doesn't want to have a checkentry ir destroy function, it can be nil.

lneto: I would simplify by now and force the user to define at least a `nop` callback as we do for the device library.. if we really have practical use case, we can always do this in a higher layer implementing it in Lua itself.. let's keep the binding as simpler as possible ;-)

sheharyaar: Ok, understood

lib/luanetfilter.c
sheharyaar: Yeah, I thought of having `unregister()` method for these, that's why I came up with this

lib/luanetfilter.c
```diff
@@ -24,26 +24,139 @@
...
 #include <linux/version.h>
+#include <linux/types.h>
 #include <linux/netfilter.h>
 #include <linux/netfilter/x_tables.h>
 
 #include <lua.h>
 #include <lauxlib.h>
 #include <lunatik.h>
 
+typedef enum luanetfilter_hook_type {
```
lneto, line 35: s/luanetfilter_hook_type/luanetfilter_type_e/

lib/luanetfilter.c
```diff
@@ -24,26 +24,136 @@
...
+										\
+	luanetfilter_xtable_opt(L, 1, "family", LUA_TNUMBER, true);		\
+	unsigned short family = (unsigned short)lua_tointeger(L, -1);		\
+										\
+	unsigned proto = luanetfilter_xtable_opt(L, 1, "proto", LUA_TNUMBER, false) ? \
+			 (unsigned short)lua_tointeger(L, -1) : 0;		\
+										\
+	luanetfilter_xtable_opt(L, 1, "checkentry", LUA_TFUNCTION, true);	\
+	luanetfilter_xtable_opt(L, 1, "destroy", LUA_TFUNCTION, true);		\
+	luanetfilter_xtable_opt(L, 1, #mode, LUA_TFUNCTION, true);		\
```
lneto, line 100: I would move the code that doesn't depend on `##mode` to an auxiliary function.. I think it will clean up the code a bit and make it more readable..

lib/luaxtable.c
```diff
@@ -0,0 +1,241 @@
...
+#include <linux/netfilter.h>
+#include <linux/netfilter/x_tables.h>
+
+#include <lua.h>
+#include <lauxlib.h>
+#include <lunatik.h>
+
+typedef enum luaxtable_type_e {
+	luaxtable_target_t,
+	luaxtable_match_t,
```
lneto, line 37:
```suggestion
	LUAXTABLE_TMATCH,
	LUAXTABLE_TTARGET,
```

sheharyaar: I would have to capitalise the macro for this, hence I kept it lowercase..

lneto: just the enum values to keep the code style.. ;-)

sheharyaar: Done

lib/luaxtable.c
```diff
@@ -0,0 +1,241 @@
...
+									\
+	unsigned proto = luaxtable_reg_opt(L, 1, "proto", LUA_TNUMBER, false) ? \
+			 (unsigned short)lua_tointeger(L, -1) : 0;	\
+									\
+	luaxtable_reg_opt(L, 1, "checkentry", LUA_TFUNCTION, true);	\
+	luaxtable_reg_opt(L, 1, "destroy", LUA_TFUNCTION, true);	\
+	luaxtable_reg_opt(L, 1, #mode, LUA_TFUNCTION, true);		\
+									\
+	struct xt_##mode *mode = lunatik_checkalloc(L, sizeof(struct xt_##mode));	\
+	memcpy((void *)mode->name, name, XT_EXTENSION_MAXNAMELEN);	\
```
lneto, line 101: are you sure this doesn't need to be null-terminated?

lib/luaxtable.c
```diff
@@ -0,0 +1,241 @@
...
+		}
+		luaL_error(L, "invalid type for %s, expected %s", opt, lua_typename(L, type));
+	}
+
+	/* true if the value can be extracted (for optional parameters) */
+	return true;
+};
+
+#define LUAXTABLE_REG(mode, ttype)						\
+static int luaxtable_reg_##mode(lua_State *L) {				\
```
lneto, line 78:
```suggestion
#define LUAXTABLE_NEWHOOK(hook, HOOK)						\
static int luaxtable_new##mode(lua_State *L) {				\
```

lib/luaxtable.c
```diff
@@ -0,0 +1,241 @@
...
+	union {
+		struct xt_match *match;
+		struct xt_target *target;
+	} hook;
+	luaxtable_type_t type;
+} luaxtable_t;
+
+static void luaxtable_release(void *private);
+static int luaxtable_reg_match(lua_State *L);
+static int luaxtable_reg_target(lua_State *L);
```
lneto, line 51: it's weird to create the implementation of these functions using a macro and having a predefined prototype here.. I would just move the macro declaration to before [this line](https://github.com/luainkernel/lunatik/pull/113/files#diff-292f6a11c6f52f91544b0b47343e04b5d6516316ca5a075844659348899709ffR183)

lib/luaxtable.c
```diff
@@ -0,0 +1,241 @@
...
+
+/* Verifies the type of "opt" in the table at "idx" and returns true if the value can be extracted.
+ * If the value is not required, it returns false.
+ * If the value is required and the type is invalid, it raises an error.
+ *
```
lneto, line 56: I don't think we need this comment here..

lib/luaxtable.c
```diff
@@ -0,0 +1,241 @@
...
+#include <lua.h>
+#include <lauxlib.h>
+#include <lunatik.h>
+
+typedef enum luaxtable_type_e {
+	LUAXTABLE_TMATCH,
+	LUAXTABLE_TTARGET,
+} luaxtable_type_t;
+
+typedef struct luaxtable_xtable_s {
```
lneto, line 40:
```suggestion
typedef struct luaxtable_s {
```

lib/luaxtable.c
```diff
@@ -0,0 +1,241 @@
...
+		if (type_ <= 0 && !req) {
+			/* the value was not required hence, should not try to extract it */
+			return false;
+		}
+		luaL_error(L, "invalid type for %s, expected %s", opt, lua_typename(L, type));
+	}
+
+	/* true if the value can be extracted (for optional parameters) */
+	return true;
+};
```
lneto, line 75:
```suggestion
static inline void luaxtable_checkopt(lua_State *L, int idx, const char *opt, int type)
{
	lua_getfield(L, idx, opt);
	if (lua_type(L, -1) != type)
		luaL_error(L, "invalid type for %s, expected %s", opt, lua_typename(L, type));
}
```

lib/luaxtable.c
```diff
@@ -0,0 +1,241 @@
...
+	const char *name = lua_tostring(L, -1);				\
+									\
+	luaxtable_reg_opt(L, 1, "revision", LUA_TNUMBER, true);		\
+	uint8_t revision = (uint8_t)lua_tointeger(L, -1);		\
+									\
+	luaxtable_reg_opt(L, 1, "family", LUA_TNUMBER, true);		\
+	unsigned short family = (unsigned short)lua_tointeger(L, -1);	\
+									\
+	unsigned proto = luaxtable_reg_opt(L, 1, "proto", LUA_TNUMBER, false) ? \
+			 (unsigned short)lua_tointeger(L, -1) : 0;	\
```
lneto, line 94: if you are going to convert anything to integer here.. you don't need to check its type.. just do `getfield` and `tointeger`.. nil or none will be converted to `0` anyway

lib/luaxtable.c
```diff
@@ -0,0 +1,241 @@
...
+{
+	lua_getfield(L, idx, opt);
+	int type_ = lua_type(L, -1);
+	if (type_ != type) {
+		// LUA_TNIL is 0 and LUA_TNONE is -1
+		if (type_ <= 0 && !req) {
+			/* the value was not required hence, should not try to extract it */
+			return false;
+		}
+		luaL_error(L, "invalid type for %s, expected %s", opt, lua_typename(L, type));
```
lneto, line 70: let's standardize the error messages for the framework.. please, base on [this](https://github.com/luainkernel/lunatik/blob/master/lib/luadevice.c#L290)

lib/luaxtable.c
```diff
@@ -0,0 +1,241 @@
...
+									\
+	luaxtable_reg_opt(L, 1, "family", LUA_TNUMBER, true);		\
+	unsigned short family = (unsigned short)lua_tointeger(L, -1);	\
+									\
+	unsigned proto = luaxtable_reg_opt(L, 1, "proto", LUA_TNUMBER, false) ? \
+			 (unsigned short)lua_tointeger(L, -1) : 0;	\
+									\
+	luaxtable_reg_opt(L, 1, "checkentry", LUA_TFUNCTION, true);	\
+	luaxtable_reg_opt(L, 1, "destroy", LUA_TFUNCTION, true);	\
+	luaxtable_reg_opt(L, 1, #mode, LUA_TFUNCTION, true);		\
```
lneto, line 98: are you not using these callbacks anywhere, right? if so, I think this code doesn't belong to this patch..

lib/luaxtable.c
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	/* true if the value can be extracted (for optional parameters) */
+	return true;
+};
+
+#define LUAXTABLE_REG(mode, ttype)						\
+static int luaxtable_reg_##mode(lua_State *L) {				\
+	lunatik_object_t *object;					\
+	luaxtable_t *xtable_##mode;					\
```
lneto, line 80: we don't need to name the variable based on the "mode".. 
```suggestion
	luaxtable_t *xtable;					\
```

lib/luaxtable.c
```diff
@@ -0,0 +1,241 @@
...
+	}								\
+									\
+	object = lunatik_newobject(L, &luaxtable_class, sizeof(luaxtable_t));	\
+	xtable_##mode = (luaxtable_t *)object->private;			\
+									\
+	xtable_##mode->runtime = lunatik_toruntime(L);			\
+	xtable_##mode->hook.mode = mode;				\
+	xtable_##mode->type = ttype;					\
+	lunatik_getobject(xtable_##mode->runtime);			\
+	lunatik_registerobject(L, 1, object);				\
```
lneto, line 122: we can move this to an auxiliary function.. e.g., `luaxtable_newhook`

lib/luaxtable.c
```diff
@@ -0,0 +1,241 @@
...
+		lunatik_free(xtable->hook.match);
+		xtable->hook.match = NULL;
+
+		break;
+	case LUAXTABLE_TTARGET:
+		xt_unregister_target(xtable->hook.target);
+		lunatik_free(xtable->hook.target);
+		xtable->hook.target = NULL;
+
+		break;
```
lneto, line 217: this can be moved to a "template" as well

lib/luaxtable.c
```diff
@@ -0,0 +1,241 @@
...
+	lunatik_object_t *object;					\
+	luaxtable_t *xtable_##mode;					\
+									\
+	luaL_checktype(L, 1, LUA_TTABLE);				\
+									\
+	luaxtable_reg_opt(L, 1, "name", LUA_TSTRING, true);		\
+	const char *name = lua_tostring(L, -1);				\
+									\
+	luaxtable_reg_opt(L, 1, "revision", LUA_TNUMBER, true);		\
+	uint8_t revision = (uint8_t)lua_tointeger(L, -1);		\
```
lneto, line 88: why` proto` is optional but `revision` and `family` aren't?

lneto (changes requested):
please, split your commit in one to rename the file to luaxtable and another for your actual changes in the code, at least..

lib/luaxtable.c
sheharyaar: Yes these are being set in the program

lib/luaxtable.c
sheharyaar: This uses mode/hook, so we would again have a new macro or the auxiliary function will need to handle both cases under if..else or switch..case block

lib/luaxtable.c
```diff
@@ -0,0 +1,237 @@
...
+{
+	int tname;
+	lua_getfield(L, idx, opt);
+	if ((tname = lua_type(L, -1)) != type)
+		luaL_error(L, "bad field '%s' (%s expected, got %s)", opt, \
+			lua_typename(L, type), lua_typename(L, tname));
+}
+
+static inline luaxtable_t *luaxtable_newhook(lua_State *L, const lunatik_class_t *cls, int hook){
+	lunatik_object_t *object = lunatik_newobject(L, cls, sizeof(luaxtable_t));	\
```
lneto, line 65: please, mind the code style..

lib/luaxtable.c
```diff
@@ -0,0 +1,224 @@
...
+}
+
+static void __exit luaxtable_exit(void)
+{
+}
+
+module_init(luaxtable_init);
+module_exit(luaxtable_exit);
+MODULE_LICENSE("Dual MIT/GPL");
+MODULE_AUTHOR("Mohammad Shehar Yaar Tausif <sheharyaar48@gmail.com>");
```
lneto, line 224: quite minor, but we use `\n` at the end of files on the code base..

lib/luaxtable.c
```diff
@@ -0,0 +1,224 @@
...
+
+#define LUAXTABLE_RELEASE_HOOK(table, hook)	\
+do {						\
+	xt_unregister_##hook(table->hook);	\
+	lunatik_free(table->hook);		\
+	table->hook = NULL;			\
+} while (0)
+
+static int luaxtable_newmatch(lua_State *L);
+static int luaxtable_newtarget(lua_State *L);
```
lneto, line 114:
```suggestion
LUAXTABLE_NEWHOOK(match, LUAXTABLE_TMATCH);
LUAXTABLE_NEWHOOK(target, LUAXTABLE_TTARGET);
```

lneto: you can also move `luaxtable_*_check` and `luaxtable_*_destroy` to above https://github.com/luainkernel/lunatik/pull/113/files#diff-292f6a11c6f52f91544b0b47343e04b5d6516316ca5a075844659348899709ffR63

lib/luaxtable.c
```diff
@@ -0,0 +1,224 @@
...
+	xtable->hook = hook;						\
+	return 1;							\
+}
+
+#define LUAXTABLE_RELEASE_HOOK(table, hook)	\
+do {						\
+	xt_unregister_##hook(table->hook);	\
+	lunatik_free(table->hook);		\
+	table->hook = NULL;			\
+} while (0)
```
lneto, line 111:
```suggestion
#define LUAXTABLE_RELEASER(hook)                                \
static inline void luaxtable_release##hook(luaxtable_t *xtable) \
{                                                               \
        xt_unregister_##hook(xtable->hook);                     \
        lunatik_free(xtable->hook);                             \
        xtable->hook = NULL;                                    \
}```

File: lib/luaxtable.c [hunk 34]
```diff
@@ -0,0 +1,224 @@
...
+		struct xt_target *target;
+	};
+	luaxtable_type_t type;
+} luaxtable_t;
+
+static void luaxtable_release(void *private);
+
+static inline luaxtable_t *luaxtable_newhook(lua_State *L, const lunatik_class_t *cls, int hook)
+{
+	lunatik_object_t *object = lunatik_newobject(L, cls, sizeof(luaxtable_t));	\
```
lneto, line 53: please mind this sort of `\` across your patches.. let's keep it clean ;-)

lib/luaxtable.c
```diff
@@ -0,0 +1,224 @@
...
+	union {
+		struct xt_match *match;
+		struct xt_target *target;
+	};
+	luaxtable_type_t type;
+} luaxtable_t;
+
+static void luaxtable_release(void *private);
+
+static inline luaxtable_t *luaxtable_newhook(lua_State *L, const lunatik_class_t *cls, int hook)
```
lneto, line 51:
```suggestion
static inline luaxtable_t *luaxtable_new(lua_State *L, const lunatik_class_t *cls, int hook)
```

lib/luaxtable.c
```diff
@@ -0,0 +1,224 @@
...
+	hook->revision = revision;					\
+	hook->family = family;						\
+	hook->proto = proto;						\
+	hook->hook = luaxtable_##hook;					\
+	hook->checkentry = luaxtable_##hook##_check;			\
+	hook->destroy = luaxtable_##hook##_destroy;			\
+	hook->me = THIS_MODULE;						\
+									\
+	if (xt_register_##hook(hook) != 0) {				\
+		lunatik_free(hook);					\
```
lneto, line 97: I would let it to be freed on your release() callback..

lneto: if not null there, then we free obj->hook

lib/luaxtable.c
```diff
@@ -0,0 +1,224 @@
...
+									\
+	lunatik_checkfield(L, 1, "name", LUA_TSTRING);			\
+	const char *name = lua_tostring(L, -1);				\
+									\
+	if (strlen(name) >= XT_EXTENSION_MAXNAMELEN) {			\
+		luaL_error(L, "name too long");				\
+	}								\
+									\
+	lunatik_checkfield(L, 1, "revision", LUA_TNUMBER);		\
+	uint8_t revision = (uint8_t)lua_tointeger(L, -1);		\
```
lneto, line 75: I would create a new macro to checkfield and assign it to the struct.. it will clean up a bit the code.. e.g.,

```
	lunatik_checkfield(L, 1, ##field, LUA_TNUMBER);
	hook->field = lua_tointeger(L, -1);
	lua_pop(L, 1);
```

lib/luaxtable.c
```diff
@@ -0,0 +1,224 @@
...
+	lunatik_checkfield(L, 1, "family", LUA_TNUMBER);		\
+	unsigned short family = (unsigned short)lua_tointeger(L, -1);	\
+									\
+	unsigned proto = (unsigned short)lua_tointeger(L, -1);		\
+									\
+	lunatik_checkfield(L, 1, "checkentry", LUA_TFUNCTION);		\
+	lunatik_checkfield(L, 1, "destroy", LUA_TFUNCTION);		\
+	lunatik_checkfield(L, 1, #hook, LUA_TFUNCTION);			\
+									\
+	struct xt_##hook *hook = lunatik_checkalloc(L, sizeof(struct xt_##hook));	\
```
lneto, line 86: I would alloc this before checking fields and let it be freed on release() in case of error.. it will simplify this macro..

lib/luaxtable.c
```diff
@@ -0,0 +1,234 @@
...
+	union {
+		struct xt_match *match;
+		struct xt_target *target;
+	};
+	luaxtable_type_t type;
+} luaxtable_t;
+
+static void luaxtable_release(void *private);
+
+#define luaxtable_setinteger(L,hook, field) 			\
```
lneto, line 51: please, mind the spaces between the args

lib/luaxtable.c
```diff
@@ -0,0 +1,234 @@
...
+{
+        const char *name;
+        size_t len;
+
+        lunatik_checkfield(L, idx, "name", LUA_TSTRING);
+        name = lua_tolstring(L, -1, &len);
+        if (len >= XT_EXTENSION_MAXNAMELEN)
+                luaL_error(L, "name is too long");
+        return name;
+}
```
lneto, line 67:
```suggestion
#define luaxtable_setstring(L, int idx, hook, field, maxlen)            \
do {                                                                    \
        const char *str;                                                \
        size_t len;                                                     \
                                                                        \
        lunatik_checkfield(L, idx, #field, LUA_TSTRING);                \
        str = lua_tolstring(L, -1, &len);                               \
        if (len >= maxlen)                                              \
                luaL_error(L, "'%s' is too long", #field);              \
        strncpy(hook->##field, str, maxlen);                            \
        lua_pop(L, 1);                                                  \
} while (0)
```

lib/luaxtable.c
```diff
@@ -0,0 +1,234 @@
...
+	};
+	luaxtable_type_t type;
+} luaxtable_t;
+
+static void luaxtable_release(void *private);
+
+#define luaxtable_setinteger(L,hook, field) 			\
+do {								\
+	lunatik_checkfield(L, 1, #field, LUA_TNUMBER);		\
+	hook->field = lua_tointeger(L, -1);			\
```
lneto, line 54:
```
lua_pop(L, 1);
```

lib/luaxtable.c
```diff
@@ -0,0 +1,234 @@
...
+	luaxtable_setinteger(L, hook, revision);			\
+	luaxtable_setinteger(L, hook, family);				\
+	luaxtable_setinteger(L, hook, proto);				\
+									\
+	hook->hook = luaxtable_##hook;					\
+	hook->checkentry = luaxtable_##hook##_check;			\
+	hook->destroy = luaxtable_##hook##_destroy;			\
+									\
+	if (xt_register_##hook(hook) != 0) {				\
+		luaL_error(L, "unable to register" #hook);		\
```
lneto, line 101: don't you need a space here? otherwise it will print "unable to registermatch", right?

lib/luaxtable.c
```diff
@@ -0,0 +1,234 @@
...
+		struct xt_target *target;
+	};
+	luaxtable_type_t type;
+} luaxtable_t;
+
+static void luaxtable_release(void *private);
+
+#define luaxtable_setinteger(L,hook, field) 			\
+do {								\
+	lunatik_checkfield(L, 1, #field, LUA_TNUMBER);		\
```
lneto, line 53: let's pass stack index as argument as well..

lib/luaxtable.c
```diff
@@ -0,0 +1,234 @@
...
+	hook->hook = luaxtable_##hook;					\
+	hook->checkentry = luaxtable_##hook##_check;			\
+	hook->destroy = luaxtable_##hook##_destroy;			\
+									\
+	if (xt_register_##hook(hook) != 0) {				\
+		luaL_error(L, "unable to register" #hook);		\
+	}								\
+									\
+	luaxtable_t *xtable = luaxtable_new(L, &luaxtable_class, HOOK); \
+	xtable->hook = hook;						\
```
lneto, line 105: for having "release()" being called in case of errors, you would need to call  "lunatik_newobject()" before calling "luaL_error()", because it's the one that pushes the object with the associated "release()" callback in the Lua stack.. perhaps you can move it to the beginning of this macro..
```
	luaL_checktype(L, 1, LUA_TTABLE);				\
									\
	lunatik_object_t *object = lunatik_newobject(L, &luaxtable_class, sizeof(luaxtable_t));	\
	/* errors after this line will eventually trigger "object->release()" */
	luaxtable_t *xtable = (luaxtable_t *)object->private;				\
									\
	struct xt_##hook *hook = lunatik_checkalloc(L, sizeof(struct xt_##hook));	\
```

you would probably need to check if it's safe to unregister a not registered hook.. if it's not, you need have some guard for that on your release callback.. because this function might fail before or after registering..

lib/luaxtable.c
lneto: we might want to print the xt_register_* error code as well..

lib/luaxtable.c
lneto: to be clear, my point here is that it's better to add the prototypes of the "static/fixed" functions than the "dynamic/generated by templates" ones.. you can refer to luanotifier

lib/luaxtable.c
```diff
@@ -0,0 +1,234 @@
...
+        size_t len;
+
+        lunatik_checkfield(L, idx, "name", LUA_TSTRING);
+        name = lua_tolstring(L, -1, &len);
+        if (len >= XT_EXTENSION_MAXNAMELEN)
+                luaL_error(L, "name is too long");
+        return name;
+}
+
+static inline luaxtable_t *luaxtable_new(lua_State *L, const lunatik_class_t *cls, int hook)
```
lneto, line 69: you don't need to pass class around.. you can assume one class per file.. also, let's always use parametrize the stack index..

sheharyaar:
Done, rearranged the functions orders, to make it align with the requirements.

lib/luaxtable.c
```diff
@@ -0,0 +1,234 @@
...
+#define LUAXTABLE_NEWHOOK(hook, HOOK)					\
+static int luaxtable_new##hook(lua_State *L) 				\
+{									\
+	luaL_checktype(L, 1, LUA_TTABLE);				\
+									\
+	lunatik_object_t *object = luaxtable_new(L, 1, HOOK); 		\
+	luaxtable_t *xtable = (luaxtable_t *)object->private;		\
+	xtable->hook = NULL;						\
+									\
+	struct xt_##hook *hook = lunatik_checkalloc(L, sizeof(struct xt_##hook));	\
```
lneto, line 132: if setstring() fails below, hook will leak

lib/luaxtable.c
```diff
@@ -0,0 +1,234 @@
...
+	luaxtable_setstring(L, 1, hook, name, XT_EXTENSION_MAXNAMELEN);	\
+	luaxtable_setinteger(L, 1, hook, revision);			\
+	luaxtable_setinteger(L, 1, hook, family);			\
+	luaxtable_setinteger(L, 1, hook, proto);			\
+									\
+	hook->hook = luaxtable_##hook;					\
+	hook->checkentry = luaxtable_##hook##_check;			\
+	hook->destroy = luaxtable_##hook##_destroy;			\
+									\
+	if (xt_register_##hook(hook) != 0) {				\
```
lneto, line 144: please mind the code style, we are not using `{ ... }` for single-liners

lib/luaxtable.c
```diff
@@ -0,0 +1,234 @@
...
+	lunatik_registerobject(L, idx, object);
+	return object;
+}
+
+#define LUAXTABLE_NEWHOOK(hook, HOOK)					\
+static int luaxtable_new##hook(lua_State *L) 				\
+{									\
+	luaL_checktype(L, 1, LUA_TTABLE);				\
+									\
+	lunatik_object_t *object = luaxtable_new(L, 1, HOOK); 		\
```
lneto, line 128: I don't think we should register the object here; if we raise an error afterwards, the object won't be collected because we hold a reference.. registering the object should be the last thing to do on this macro.. we probably need to split (or get rid of) this luaxtable_new() function..

lneto: btw, it would be good to run some tests forcing the error condition, putting some prints on release and also reloading lunatik (to observe if it will hang)..

lib/luaxtable.c
```diff
@@ -0,0 +1,232 @@
...
+do {								\
+	const char *str;					\
+	size_t len;						\
+								\
+	lunatik_checkfield(L, idx, #field, LUA_TSTRING);	\
+	str = lua_tolstring(L, -1, &len);			\
+	if (len >= maxlen)					\
+		luaL_error(L, "'%s' is too long", #field);	\
+	strncpy((char *)hook->field, str, maxlen);		\
+	lua_pop(L, 1);						\
```
lneto, line 109:
```suggestion
	size_t len;						\
	lunatik_checkfield(L, idx, #field, LUA_TSTRING);	\
	const char *str = lua_tolstring(L, -1, &len);		\
	if (len >= maxlen)					\
		luaL_error(L, "'%s' is too long", #field);	\
	strncpy((char *)hook->field, str, maxlen);		\
	lua_pop(L, 1);						\
```

as you are using C99 style declarations (scattered around the block, not only at the beginning), you can do it here as well.. I'm not against it.. but I think we should apply to the other modules to standardize the code base.. perhaps we should create a ticket for this..

we should use `maxlen - 1` (for `\0`) on strncpy, right? Perhaps a better idea is to pass `XT_EXTENSION_MAXNAMELEN - 1` to this function and do `if (len > maxlen)`... it might be clearer.. so we use `len` as in the [libc semantics](https://pubs.opengroup.org/onlinepubs/9699919799/functions/strlen.html)..

lib/luaxtable.c
```diff
@@ -0,0 +1,232 @@
...
+									\
+	hook->hook = luaxtable_##hook;					\
+	hook->checkentry = luaxtable_##hook##_check;			\
+	hook->destroy = luaxtable_##hook##_destroy;			\
+									\
+	if (xt_register_##hook(hook) != 0)				\
+		luaL_error(L, "unable to register " #hook);		\
+									\
+	xtable->registered = true;					\
+	xtable->runtime = lunatik_toruntime(L);				\
```
lneto, line 146:
```suggestion
	xtable->runtime = lunatik_toruntime(L);				\
```
we can actual use `xtable->runtime` for this purpose.. now I see.. =)..

lib/luaxtable.c
```diff
@@ -0,0 +1,232 @@
...
+	case LUAXTABLE_TMATCH:
+		luaxtable_releasematch(xtable);
+		break;
+	case LUAXTABLE_TTARGET:
+		luaxtable_releasetarget(xtable);
+		break;
+	}
+
+	if (xtable->runtime)
+		lunatik_putobject(xtable->runtime);
```
lneto, line 214:
```suggestion
	if (xtable->runtime) {
		switch (xtable->type) {
		case LUAXTABLE_TMATCH:
			xt_unregister_match(&xtable->match);
			break;
		case LUAXTABLE_TTARGET:
			xt_unregister_target(&xtable->target);
			break;
		}
		lunatik_putobject(xtable->runtime);
		xtable->runtime = NULL;
	}
```

lib/luaxtable.c
```diff
@@ -0,0 +1,232 @@
...
+	lunatik_registerobject(L, 1, object);				\
+	return 1;							\
+}
+
+#define LUAXTABLE_RELEASER(hook)                                \
+static inline void luaxtable_release##hook(luaxtable_t *xtable) \
+{                                                       	\
+	if(xtable->registered)					\
+		xt_unregister_##hook(&xtable->hook);     	\
+}
```
lneto, line 157: now we don't have much to do here, I think this template is overkill.. I would get rid of this..

lib/luaxtable.c
```diff
@@ -0,0 +1,232 @@
...
+	lua_pop(L, 1);						\
+} while (0)
+
+static inline lunatik_object_t *luaxtable_new(lua_State *L, int hook)
+{
+	lunatik_object_t *object = lunatik_newobject(L, &luaxtable_class , sizeof(luaxtable_t));
+	luaxtable_t *xtable = (luaxtable_t *)object->private;
+
+	xtable->type = hook;
+	xtable->registered = false;
```
lneto, line 118:
```suggestion
	xtable->runtime = NULL;
```

lib/luaxtable.c
```diff
@@ -0,0 +1,232 @@
...
+} luaxtable_type_t;
+
+typedef struct luaxtable_s {
+	lunatik_object_t *runtime;
+	union {
+		struct xt_match match;
+		struct xt_target target;
+	};
+	luaxtable_type_t type;
+	bool registered;
```
lneto, line 47:
```suggestion
```

lneto (changes requested):

lib/luaxtable.c
```diff
@@ -0,0 +1,219 @@
...
+	hook->hook = luaxtable_##hook;					\
+	hook->checkentry = luaxtable_##hook##_check;			\
+	hook->destroy = luaxtable_##hook##_destroy;			\
+									\
+	if (xt_register_##hook(hook) != 0)				\
+		luaL_error(L, "unable to register " #hook);		\
+									\
+	xtable->runtime = lunatik_toruntime(L);				\
+	lunatik_getobject(xtable->runtime);				\
+	lunatik_registerobject(L, 1, object);				\
```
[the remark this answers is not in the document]
lneto: quite minor, though..

lib/luaxtable.c
```diff
@@ -0,0 +1,223 @@
...
+static const luaL_Reg luaxtable_lib[] = {
+	{"match", luaxtable_newmatch},
+	{"target", luaxtable_newtarget},
+	{NULL, NULL}
+};
+
+static void luaxtable_release(void *private)
+{
+	luaxtable_t *xtable = (luaxtable_t *)private;
+	if (!xtable->runtime) return;
```
lneto, line 193:
```suggestion

	if (!xtable->runtime)
		return;
```

lib/luaxtable.c
```diff
@@ -0,0 +1,223 @@
...
+{
+	xtable->runtime = lunatik_toruntime(L);
+	lunatik_getobject(xtable->runtime);
+	lunatik_registerobject(L, idx, object);
+}
+
+#define LUAXTABLE_NEWHOOK(hook, HOOK)					\
+static int luaxtable_new##hook(lua_State *L) 				\
+{									\
+	luaL_checktype(L, 1, LUA_TTABLE);				\
```
lneto, line 129: let's move it to luaxtable_new, so everything that doesn't depend on macro concat is out of the template..

lneto (approved):

PR #114, opened 2024-05-15
add attach/detach APIs to luaxdp

lneto:
* add offset parameter on bpf_luaxdp_run

### DIFF
--- a/lib/luaxdp.c
+++ b/lib/luaxdp.c
@@ -48,43 +48,31 @@ __diag_ignore_all("-Wmissing-prototypes",
                   "Global kfuncs as their definitions will be in BTF");
 #endif
 
-static inline int luaxdp_callback(lua_State *L)
-{
-	if (lunatik_getregistry(L, luaxdp_callback) == LUA_TNIL) {
-		lua_pop(L, 1); /* pop nil */
-		if (lua_type(L, -1) != LUA_TFUNCTION) {
-			pr_err("couldn't find callback");
-			return -1;
-		}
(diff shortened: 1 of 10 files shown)


lib/luaxdp.c
```diff
@@ -48,43 +48,31 @@ __diag_ignore_all("-Wmissing-prototypes",
...
 
 	action = lua_tointeger(L, -1);
-pop:
-	lua_pop(L, 1);
 out:
 	return action;
 }
 
-__bpf_kfunc int bpf_luaxdp_run(struct xdp_md *xdp_ctx, char *key, size_t key__sz)
+__bpf_kfunc int bpf_luaxdp_run(struct xdp_md *xdp_ctx, char *key, size_t key__sz, size_t offset)
```
sheharyaar, line 73: does `offset` refer to the offset of packet data inside the buffer ?

sheharyaar:
I have added my queries in the code comments

lib/luaxdp.c
lneto: Yes, a offset within the packet, passed from the eBPF program.. I'm using currently for passing the payload offset, but it could be used by eBPF to point to any specific position of the packet.. my plan is to evolve this parm to pass a `void *` with a packed data from the eBPF program.. but, baby steps ;-)

lneto: I can share a sample eBPF code to make it clearer if you think it's worth

sheharyaar: I understood the concept.

PR #116, opened 2024-05-30
require luadata on xdp.attach()

lneto:
* add LUNATIK_LIB and lunatik_require APIs

### DIFF
--- a/examples/filter/sni.lua
+++ b/examples/filter/sni.lua
@@ -22,7 +22,6 @@
 --
 
 local xdp  = require("xdp")
-local data = require("data")
 
 local action = xdp.action
 
--- a/lib/luadata.h
+++ b/lib/luadata.h
@@ -24,6 +24,8 @@
 #ifndef luadata_h
 #define luadata_h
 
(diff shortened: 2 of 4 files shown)


glk0:
If I may, what is the motivation for this change of approach ?

lneto:
> If I may, what is the motivation for this change of approach ?

Some folks found it confusing having this _apparently_ unused `local data = require("data")` in the `sni.lua` script. It was necessary to access the methods of the callback arguments. For instance, if you haven't such `require` and  called `xdp.attach()` it would raise a `"metatable not found (data)"` error trying to create the argument data objects. I would like to improve this to use a string with the `libname` on `lunatik_require`, so we could load the metatables of the parameters on creating new threads; perhaps using `getglobal(L, "require")`, but I've mixed feelings about this.

BTW, working on this [article](https://medium.com/@lourival.neto/is-ebpf-driving-you-crazy-let-it-run-lunatik-instead-4aca7d63e6fd); you might wanna take a look ;-).

PR #119, opened 2024-06-03
add lunatik_checkfield method

sheharyaar:


lunatik.h
```diff
@@ -138,6 +138,15 @@ static inline void *lunatik_checknull(lua_State *L, void *ptr)
 
 #define lunatik_checkalloc(L, s)	(lunatik_checknull((L), lunatik_malloc((L), (s))))
 
+static inline void lunatik_checkfield(lua_State *L, int idx, const char *opt, int type)
```
lneto, line 141:
```suggestion
static inline void lunatik_checkfield(lua_State *L, int idx, const char *field, int type)
```

lunatik.h
```diff
@@ -138,6 +138,15 @@ static inline void *lunatik_checknull(lua_State *L, void *ptr)
...
+{
+	int tname;
+	lua_getfield(L, idx, opt);
+	if ((tname = lua_type(L, -1)) != type)
+		luaL_error(L, "bad field '%s' (%s expected, got %s)", opt, \
+			lua_typename(L, type), lua_typename(L, tname));
```
lneto, line 147:
```suggestion
                luaL_error(L, "bad field '%s' (%s expected, got %s)", field,
                        lua_typename(L, type), lua_typename(L, tname));
```

lneto (changes requested):
please, rename this commit as well and apply this API in the places we are using this pattern in the code base..

sheharyaar:
Done

lib/luadevice.c
```diff
@@ -95,11 +95,7 @@ static int luadevice_fop(lua_State *L, luadevice_t *luadev, const char *fop, int
 		goto err;
 	}
 
-	if (lua_getfield(L, -1, fop) != LUA_TFUNCTION) {
-		lua_getfield(L, -2, "name");
-		pr_err("%s: operation isn't defined for /dev/%s\n", fop, lua_tostring(L, -1));
-		goto err;
-	}
+	lunatik_checkfield(L, -1, fop, LUA_TFUNCTION);
```
lneto, line 98: we cannot raise an error here.. we are not running in a pcall.. I would simply revert this particular change..

lunatik.h
lneto: @sheharyaar please notice that you don't need this `\`.. it's not a macro..

sheharyaar:
Done

lunatik.h
```diff
@@ -138,6 +138,15 @@ static inline void *lunatik_checknull(lua_State *L, void *ptr)
...
 
+static inline void lunatik_checkfield(lua_State *L, int idx, const char *field, int type)
+{
+	int tname;
+	lua_getfield(L, idx, field);
+	if ((tname = lua_type(L, -1)) != type)
+		luaL_error(L, "bad field '%s' (%s expected, got %s)", field,
+			lua_typename(L, type), lua_typename(L, tname));
+}
+
```
lneto, line 149: perhaps, though I'm not very found of, it's better to use [GNU extensions](https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html) here.. it might cleanup LUAXTABLE_NEWHOOK a bit..

```suggestion
#define lunatik_checkfield(L, idx, field, type, TYPE)                                           \
do {                                                                                            \
        int _type = lua_getfield(L, idx, field);                                                \
        if (_type != LUA_T#TYPE) {                                                              \
                const char *err = "bad field '%s' (%s expected, got %s)";                       \
                luaL_error(L, err, field, lua_typename(L, ##type), lua_typename(L, _type));       \
        }                                                                                       \
        lua_to#type(L, -1);                                                                     \
} while(0);

```

so we can do..
```
hook->foo = lunatik_checkfield(L, 1, "foo", number, NUMBER);
hook->bar = lunatik_checkfield(L, 1, "bar", string, STRING);
```

what do you think?

lunatik.h
```diff
@@ -138,6 +138,15 @@ static inline void *lunatik_checknull(lua_State *L, void *ptr)
 
 #define lunatik_checkalloc(L, s)	(lunatik_checknull((L), lunatik_malloc((L), (s))))
 
+static inline void lunatik_checkfield(lua_State *L, int idx, const char *field, int type)
+{
+	int tname;
+	lua_getfield(L, idx, field);
+	if ((tname = lua_type(L, -1)) != type)
```
lneto, line 145: you don't need to call `lua_type` here.. check the `lua_getfield` manual ;-)

lunatik.h
```diff
@@ -138,6 +138,15 @@ static inline void *lunatik_checknull(lua_State *L, void *ptr)
 
 #define lunatik_checkalloc(L, s)	(lunatik_checknull((L), lunatik_malloc((L), (s))))
 
+static inline void lunatik_checkfield(lua_State *L, int idx, const char *field, int type)
+{
+	int tname;
```
lneto, line 143: `tname` is a bit misnomer.. it's not the type name that you are storing but the type enum itself

lunatik.h
sheharyaar: Initially I had this idea too, but it won't accomodate `LUA_TFUNCTION` in the later PRs.

lneto: > Is this still needed after the macro addition in the other PR ?

I don't think so.. I think the changes there are good enough regarding this.. we can get merge as is and get back to it later if we think it will be a improvement..

lneto (approved):

PR #122, opened 2024-06-10
[GSoC 2024] add xtable rcu and lua callback execution

sheharyaar:


lib/luaxtable.c
```diff
@@ -31,6 +31,11 @@
 #include <lua.h>
 #include <lauxlib.h>
 #include <lunatik.h>
+#include "luadata.h"
+#include "luarcu.h"
+#include "luaxtable.h"
+
+#define XTABLE_DEFAULT_SIZE 1024
```
lneto, line 38: please, use `LUAXTABLE` prefix for this module definitions to not get mixed up with the actual `XTABLE` ones.. also, if we are going to use LUARCU default value, let's just use its define instead of creating another one..

lib/luaxtable.c
```diff
@@ -31,6 +31,11 @@
 #include <lua.h>
 #include <lauxlib.h>
 #include <lunatik.h>
+#include "luadata.h"
```
lneto, line 34:
```suggestion

#include "luadata.h"
```

lib/luaxtable.h
```diff
@@ -0,0 +1,34 @@
...
+*/
+
+#ifndef luaxtable_h
+#define luaxtable_h
+
+struct luaxtable_s;
+
+typedef struct luaxtable_info_s {
+	struct luaxtable_s *data __attribute__((aligned(8)));
+} luaxtable_info_t;
```
lneto, line 31: why do we need to export this struct and, btw, why do we need such struct in the first place?

lib/luaxtable.c
```diff
@@ -46,34 +51,154 @@ typedef struct luaxtable_s {
...
-{
-	return NF_ACCEPT;
+#define LUAXTABLE_HOOK_CB(T, U, V, hook, huk) 				\
+static T luaxtable_##hook(U skb, V par)					\
+{									\
+	int ret = NF_DROP;						\
+	const luaxtable_info_t *info = (const luaxtable_info_t *)par->huk##info;	\
+	luaxtable_t *xtable = info->data;				\
+									\
+	lunatik_run(xtable->runtime, luaxtable_invoke, ret, xtable, (void *)skb, sizeof(struct sk_buff), (void *)par, sizeof(struct xt_action_param), #hook, 2, 1);					\
```
lneto, line 159: why do you need the sizeofs here? and to use `void *`? if you have different parameters for target and match, let's just have to separate "invoke" functions..

sheharyaar: The following are the struct parameters that are used by match and target functions : 

- `match` : struct sk_buff, **const** struct **xt_action_param**
- `target` : **const** struct sk_buff, struct **xt_action_param**

- `match_checkentry` : const struct xt_**mtchk**_param
- `target_checkentry` : const struct xt_**tgchk**_param

- `match_destroy` : const struct xt_**mtdtor**_param
- `target_destory` : const struct xt_**tgdtor**_param

There are 5 different structs for us to handle.
These have a common pattern (just match is replaced by target, for the fields).

I used `void *` and `sizeof` for luadata object allocation.

lib/luaxtable.h
sheharyaar: This will be shared with the userspace program

lib/luaxtable.c
```diff
@@ -46,34 +50,123 @@ typedef struct luaxtable_s {
...
+		ret = -ECANCELED;
+		goto err;
+	}
+
+	if (nret > 0)
+		ret = luaL_optinteger(L, -1, 0);
+	else
+		ret = 0;
+
+	return ret;
```
lneto, line 114:
```suggestion
	return nret > 0 ? luaL_optinteger(L, -1, 0) : 0;
```

lib/luaxtable.c
```diff
@@ -43,37 +47,123 @@ typedef struct luaxtable_s {
...
+		return -1;						\
+	}								\
+	xtable = (luaxtable_t *)obj->private;				\
+} while(0)
+
+#define luaxtable_setrcu(object, hook, key, keylen)			\
+do {									\
+	if (luarcu_settable(xtable_##hook##hooks, key, keylen, object) != 0) {	\
+		luaL_error(L, "unable to set table");			\
+		return -1;						\
```
lneto, line 78: it's never reached.. do you intend to raise an error or return an error condition?

usr/lib/libxt_luaxtabletest.c
```diff
@@ -0,0 +1,81 @@
...
+* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+#include <xtables.h>
+#include <linux/netfilter.h>
+
+#include "../../lib/luaxtable.h"
```
lneto, line 27: adjust your CFLAGS instead of using this sort of relative path..

lib/luaxtable.c
```diff
@@ -43,37 +47,123 @@ typedef struct luaxtable_s {
...
+{									\
+	const luaxtable_info_t *info = (const luaxtable_info_t *)par->huk##info;	\
+	luaxtable_t *xtable = info->data;				\
+	int ret = xtable->fallback;					\
+									\
+	/* TODO: handle const type in luadata */			\
+	lunatik_run(xtable->runtime, luaxtable_invoke, ret, xtable, (struct sk_buff *)skb, #hook, 1);	\
+	if (ret < 0)							\
+		pr_err("error in running " #hook " lua handler : %d\n", ret);	\
+	return ret < 0 ? xtable->fallback : ret;					\
```
lneto, line 127:
```suggestion
	return ret;					\
```

lib/luaxtable.c
```diff
@@ -43,37 +47,123 @@ typedef struct luaxtable_s {
...
+		nargs = 1;
+	}
+
+	if (lua_pcall(L, nargs, nret, 0) != LUA_OK) {
+		pr_err("%s error: %s\n", op, lua_tostring(L, -1));
+		ret = -ECANCELED;
+		goto err;
+	}
+
+	return nret > 0 ? luaL_optinteger(L, -1, 0) : 0;
```
lneto, line 110:
```suggestion
	return luaL_optinteger(L, -1, 0);
```

I think it should be enough, doesn't it? take a look at the manual..

lib/luaxtable.c
```diff
@@ -43,37 +47,123 @@ typedef struct luaxtable_s {
...
+		goto err;
+	}
+
+	if (lua_getfield(L, -1, op) != LUA_TFUNCTION) {
+		pr_err("%s isn't defined\n", op);
+		goto err;
+	}
+
+	if (skb != NULL ){
+		luaxtable_newdata(L, skb, sizeof(struct sk_buff), false);
```
lneto, line 100: I don't think it's right.. we should push `skb->header` and `skb->data`, not the actual skb structure, to Lua, right? take a look at nflua..

lneto: I would start with `->data`..

lib/luaxtable.c
```diff
@@ -43,37 +47,123 @@ typedef struct luaxtable_s {
...
+	int ret = -EINVAL;							\
+										\
+	luaxtable_t *xtable;							\
+	luaxtable_getrcu(xtable, hook, par->hook->name, XT_EXTENSION_MAXNAMELEN);	\
+	luaxtable_info_t *info = (luaxtable_info_t *)par->huk##info; 		\
+	info->data = xtable;							\
+										\
+	lunatik_run(xtable->runtime, luaxtable_invoke, ret, xtable, NULL, "checkentry", 1);	\
+	if (ret < 0)								\
+		pr_err("error in running 'checkentry' lua handler : %d\n", ret);	\
```
lneto, line 142: we can move these prints to `invoke`, right? we have the `op` string there.. it will clean up a bit your macros..

lib/luaxtable.c
```diff
@@ -43,37 +47,123 @@ typedef struct luaxtable_s {
...
+
+	if (lua_pcall(L, nargs, nret, 0) != LUA_OK) {
+		pr_err("%s error: %s\n", op, lua_tostring(L, -1));
+		ret = -ECANCELED;
+		goto err;
+	}
+
+	return nret > 0 ? luaL_optinteger(L, -1, 0) : 0;
+err:
+	lua_settop(L, base); /* pop everything, including args */
```
lneto, line 112: I think `lunatik_run` will do this for you.. please, check..

lib/luaxtable.h
lneto, line 31: can't we simply use only one struct for both? why do we need the nested one?

lib/luaxtable.c
```diff
@@ -43,37 +47,123 @@ typedef struct luaxtable_s {
...
+	int fallback;
 	luaxtable_type_t type;
 } luaxtable_t;
 
-static bool luaxtable_match(const struct sk_buff *skb, struct xt_action_param *par)
-{
-	return true;
-}
+static lunatik_object_t *xtable_targethooks = NULL;
+static lunatik_object_t *xtable_matchhooks = NULL;
```
lneto, line 55: please, prefix your code with `luaxtable`, not `xtable`.. it makes it harder to track what is yours and what's original kernel.. btw, perhaps it's a good idea to structure it a bit.. but not sure.. I'll leave to you to think a bit more on this =)..

```
static struct {
	lunatik_object_t *target;
	lunatik_object_t *match;
} luaxtable_hooks = {NULL, NULL};
```

lib/luaxtable.c
sheharyaar: Here, `ret` is being changed by lunatik_run

lib/luaxtable.h
sheharyaar: > The following are the struct parameters that are used by match and target functions :
> 
> * `match` : struct sk_buff, **const** struct **xt_action_param**
> * `target` : **const** struct sk_buff, struct **xt_action_param**
> * `match_checkentry` : const struct xt_**mtchk**_param
> * `target_checkentry` : const struct xt_**tgchk**_param
> * `match_destroy` : const struct xt_**mtdtor**_param
> * `target_destory` : const struct xt_**tgdtor**_param

The struct `luaxtable_info_s` is set under `params->(target/match)info`. Now since the struct param is const in (match/target) checkentry, we cannot edit the info field. That's why we need another level of indirection (const param -> match/target info -> data), here data can be modified

sheharyaar: This is the pattern across xtable kernel and userspace addon source codes.
For example : 
`xt_quota.h` : https://elixir.bootlin.com/linux/v6.9/source/include/uapi/linux/netfilter/xt_quota.h#L14
`xt_quota.c` : https://elixir.bootlin.com/linux/v6.9/source/net/netfilter/xt_quota.c#L46

lib/luaxtable.c
sheharyaar: Yeah, I am working on this

lib/luaxtable.c
```diff
@@ -88,6 +165,15 @@ static const lunatik_class_t luaxtable_class = {
 	.sleep = false,
 };
 
+static void inline luaxtable_setbuf(lua_State *L, int idx, lunatik_object_t *buf)
```
lneto, line 168: `inline void`

lib/luaxtable.h
lneto: got it.. I just don't see the `__attribute__((aligned(8))` in the example you shared.. but agreed, let's use the same pattern..

lib/luaxtable.c
```diff
@@ -32,48 +32,125 @@
...
+		goto err;
+	}
+	ret = luaL_optinteger(L, -1, 0);
+
+err:
+	if (data)
+		luadata_clear(data);
+
+	if (ret < 0)
+		pr_err("error in running %s lua handler : %d\n", op, ret);
```
lneto, line 100: you don't need this extra print here, right? you can just add the `op` info to every error condition.. you can even have a macro for this..

lib/luaxtable.c
lneto: which is set by your `invoke` function, doesn't it? why you need to check it again?

lib/luaxtable.h
sheharyaar: That is given under  `Section 4.5 :  Attaching kernel-specific data` in https://inai.de/documents/Netfilter_Modules.pdf

Other examples use this : 
1. xt_rateest.h : https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/netfilter/xt_rateest.h#L21
2. xt_TEE.h : https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/netfilter/xt_TEE.h#L12
3. xt_cgroup.h : https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/netfilter/xt_cgroup.h#L22

and many more under `include/uapi/linux/netfilter/*.h`

lib/luaxtable.c
sheharyaar: I need to differentiate between negative returns from pcall and an error condition during or before pcall in `invoke`. If the return is negative from the pcall result, it can be returned, but for errors we need to have a fallback value.

sheharyaar: Ah, I just remembered, `match` should return bool and `target` should return `unsigned int`, so we can't return negative values, So only for these callbacks we need to send that fallback value. For checkentry, a negative return is fine (a success is 0) and for destroy no value is to be returned.

examples/xt.lua
```diff
@@ -0,0 +1,51 @@
+local xt = require("xtable")
```
lneto, line 1: add your copyright and use a more descriptive name for this script.. what is it doing?

examples/xt.lua
```diff
@@ -0,0 +1,51 @@
...
+local action = xt.action
+local family = xt.family
+local match_packet
+local check_packet
+local destroy_packet
```
lneto, line 6: you don't need this, right/

examples/xt.lua
```diff
@@ -0,0 +1,51 @@
...
+
+	return action.DROP
+end
+
+function check_packet()
+	print("check_packet called\n")
+	return 0
+end
+
+function destroy_packet()
```
lneto, line 36: `destroy_packet` is a bit misleading.. you are not "destroying" packet but the xtable object, right..

examples/xt.lua
```diff
@@ -0,0 +1,51 @@
...
+			return action.ACCEPT
+		end
+	end
+
+	return action.DROP
+end
+
+function check_packet()
+	print("check_packet called\n")
+	return 0
```
lneto, line 33: is this return really necessary? I would simplify your example and do something like..

```
local function nop() end

...

local dnsdrop = {
	name= "dnsdrop",
	...
	checkentry = nop,
	destroy = nop,
	...
}
```

examples/xt.lua
```diff
@@ -0,0 +1,51 @@
...
+local destroy_packet
+
+local udp = 0x11
+local dns = 0x35
+
+function match_packet(skb)
+	print("match_packet called\n")
+	-- ip header
+	local ipb = skb:getbyte(0)
+	local ihlen = (( ipb << 4 ) & 0xFF ) >> 4
```
lneto, line 15: please, mind the code style.. `s/(( /((/`

examples/xt.lua
```diff
@@ -0,0 +1,51 @@
...
+	local iphdrlen = ihlen * 4
+	local proto = skb:getbyte(9)
+
+	if proto == udp then
+		local srcport = skb:getbyte(iphdrlen) << 8 | skb:getbyte(iphdrlen + 1)
+		local dstport = skb:getbyte(iphdrlen + 2) << 8 | skb:getbyte(iphdrlen + 3)
+		if srcport == dns or dstport == dns then
```
lneto, line 22: I didn't verify it, but just a reminder that you should take endianess in consideration.. probably it's something that should be added to luadata (as the `const` thing.. probably it would be the case of having `->flags`)

lib/luaxtable.c
```diff
@@ -32,48 +32,127 @@
...
+	if (data)
+		luadata_clear(data);
+	return ret;
 }
 
-static unsigned int luaxtable_target(struct sk_buff *skb, const struct xt_action_param *par)
-{
-	return NF_ACCEPT;
+static bool match_fallback = false;
+static unsigned int target_fallback = XT_CONTINUE;
```
lneto, line 104: as already pointed out, you should prefix every global with `luaxtable_`.. in this case, I think it's a good idea to follow the same pattern you have done for the RCU tables, using a struct.. perhaps we can have a single struct to hold table pointe and the default return.. it's up to you

examples/xt.lua
sheharyaar: check packet will be needed for now, destroy can be a nop

examples/xt.lua
sheharyaar: We should also have luadata `short` get/set too, there we can handle the endianness.

lneto: I'm not sure what would be better.. to have separate APIs per byte length (such as {get,set}short) or having the byte length as argument on {get,set}number.. mixed feelings.. what do you think?

examples/xt.lua
lneto: why? aren't you doing `optint(..., 0)`? it will evaluate to 0 anyway if return none, doesn't it?

sheharyaar: Ah, yes

examples/xt.lua
sheharyaar: Both approaches are fine.

examples/xt.lua
```diff
@@ -0,0 +1,67 @@
+--
+-- Copyright (c) 2024 Mohammad Shehar Yaar Tausif <sheharyaar48@gmail.com>
```
lneto, line 2: we've moved to SPDX tags.. please, adjust your copyright notices accordingly

lib/luaxtable.c
```diff
@@ -32,48 +32,161 @@
...
 } luaxtable_t;
 
-static bool luaxtable_match(const struct sk_buff *skb, struct xt_action_param *par)
+static struct {
+	lunatik_object_t *target;
+	lunatik_object_t *match;
+	bool match_fallback;
+	unsigned int target_fallback;
+} luaxtable_hooks = {NULL, NULL, false, XT_CONTINUE};
+
```
lneto, line 60:
```suggestion
```

examples/xt.lua
lneto: *all* your copyright notices, I meant ;-)

lib/luaxtable.c
```diff
@@ -32,48 +32,160 @@
...
+		return -1;
+	}
+
+	if (skb_is_nonlinear(skb)) {
+		pr_err("skb is non_linear!\n");
+		return -1;
+	}
+
+	luadata_reset(data, skb->data, skb_headlen(skb), LUADATA_OPT_KEEP);
+	return luaxtable_invoke(L, xtable, "match", 1, 1);
```
lneto, line 137:
```suggestion
	return luaxtable_invoke(L, xtable, "target", 1, 1);
```
?

lib/luaxtable.c
```diff
@@ -32,48 +32,160 @@
...
+	lunatik_object_t *data = luaxtable_getskb(L, xtable);
+	if (data == NULL) {
+		pr_err("could not find skb\n");
+		return -1;
+	}
+
+	if (skb_is_nonlinear(skb)) {
+		pr_err("skb is non_linear!\n");
+		return -1;
+	}
```
lneto, line 134: you can move this to an aux function.. e.g., luaxtable_checkskb()

lib/luaxtable.c
```diff
@@ -32,48 +32,160 @@
...
+
+	luadata_reset(data, skb->data, skb_headlen(skb), LUADATA_OPT_READONLY);
+	return luaxtable_invoke(L, xtable, "match", 1, 1);
 }
 
-static unsigned int luaxtable_target(struct sk_buff *skb, const struct xt_action_param *par)
+static int luaxtable_dotarget(lua_State *L, luaxtable_t *xtable, struct sk_buff *skb, const struct xt_action_param *par)
 {
-	return NF_ACCEPT;
+	if (skb == NULL || par == NULL) {
```
lneto, line 120: is it supposed to happen? if so, in which condition? is it being checked in another xtable extension?

lib/luaxtable.c
sheharyaar: I have done something similar to this in the next PR, because apart from skb, we will have params here as well.

lib/luaxtable.c
sheharyaar: It's not being used by any extension, we can remove it

lib/luaxtable.c
sheharyaar: Ah, I forgot this when copying the function

lib/luaxtable.c
```diff
@@ -32,48 +14,150 @@
...
+	return ret < 0 ? luaxtable_hooks.hook##_fallback : ret;		\
 }
 
-static void luaxtable_target_destroy(const struct xt_tgdtor_param *par)
-{
+#define LUAXTABLE_CHECKER_CB(hook, hk, huk, HOOK)				\
+static int luaxtable_##hook##_check(const struct xt_##hk##chk_param *par)	\
+{										\
+	int ret = -EINVAL;							\
+										\
```
lneto, line 127:
```suggestion
```

examples/xt.lua
```diff
@@ -0,0 +1,49 @@
...
+			print("DNS packet\n")
+			return action.ACCEPT
+		end
+	end
+
+	return action.DROP
+end
+
+local match_ops = {
+	name= "luaxtabletest",
```
lneto, line 39:
```suggestion
	name = "luaxtabletest",
```

please, mind the code style.. `"%s/\([a-z]\)=/\1 =/g"`

lneto: I wouldn't name this "luaxtabletest".. I think you should use a more descriptive name indicating what your extension really does..

examples/xt.lua
```diff
@@ -0,0 +1,49 @@
+--
+-- SPDX-FileCopyrightText: (c) 2024 Mohammad Shehar Yaar Tausif <sheharyaar48@gmail.com>
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+-- Match DNS packets using netfilter match
```
lneto, line 6: I would rename the file accordingly.. instead of calling this `xt.lua`.. let's avoid such generic names.. also, I would expect this example to be more "useful"..  instead of simply matching proto and ports, do something that you can't do simply using iptables rules.. for instance, parse something on the DNS request and block it accordingly to a blocklist based on such parameter..

usr/lib/libxt_luaxtabletest.c
```diff
@@ -0,0 +1,63 @@
...
+}
+
+static void ipaddr_mt_save(const void *entry,
+    const struct xt_entry_match *match)
+{
+}
+
+static struct xtables_match ipaddr_mt_reg = {
+	.version       = XTABLES_VERSION,
+	.name          = "luaxtabletest",
```
lneto, line 41: if it's tied up to your example, this should live just besides it on `examples/`.. similarly to bpf program of `examples/filter`

lib/luaxtable.c
```diff
@@ -32,48 +14,150 @@
...
+	lunatik_object_t *data = luaxtable_getskb(L, xtable);
+	if (data == NULL) {
+		pr_err("could not find skb\n");
+		return -1;
+	}
+
+	if (skb_is_nonlinear(skb)) {
+		pr_err("skb is non_linear!\n");
+		return -1;
+	}
```
lneto, line 106: do we really need this here? shouldn't we let the Lua script to decide what to do?

lib/luaxtable.c
```diff
@@ -32,48 +14,150 @@
...
+		goto err;
+	}
+
+	if (lua_getfield(L, -1, op) != LUA_TFUNCTION) {
+		pr_err("%s isn't defined\n", op);
+		goto err;
+	}
+
+	lua_insert(L, base + 1); /* op */
+	lua_pop(L, 1); /* table */
```
lneto, line 59: you can just use https://www.lua.org/manual/5.4/manual.html#lua_remove here, right?

lib/luaxtable.c
```diff
@@ -32,48 +14,150 @@
...
+
+	if (lua_pcall(L, nargs, nret, 0) != LUA_OK) {
+		pr_err("%s error: %s\n", op, lua_tostring(L, -1));
+		ret = -ECANCELED;
```
lneto, line 63: who's expecting this? does anyone check for `ECANCELED`?

lib/luaxtable.c
```diff
@@ -32,48 +14,150 @@
...
+		goto err;
+	}
+	if (nret != 0)
+		ret = luaL_optinteger(L, -1, 0);
```
lneto, line 67: wouldn't be simple to just  unconditionally `return luaL_optinteger(L, -1, 0);` here?

lib/luaxtable.c
sheharyaar: The stack contains : | base | arg1 |... | argn | table | function |
We need : | base | function |arg1 |... | argn |

We need to shift the function below args and remove the table from the stack. So, if we use remove, it would just remove the table, we still need to move the function below args.

lib/luaxtable.c
```diff
@@ -32,48 +14,150 @@
...
-	return 0;
+#define LUAXTABLE_HOOK_CB(T, U, V, hook, huk) 				\
+static T luaxtable_##hook(U skb, V par)					\
+{									\
+	const luaxtable_info_t *info = (const luaxtable_info_t *)par->huk##info;	\
+	luaxtable_t *xtable = info->data;				\
+	int ret;							\
+									\
+	lunatik_run(xtable->runtime, luaxtable_do##hook, ret, xtable, skb, par);	\
+	return ret < 0 ? luaxtable_hooks.hook##_fallback : ret;		\
```
lneto, line 120: I think we can pass `fallback` as an argument to `lunatik_run`, then avoiding this `ret < 0 ?` and unifying `LUAXTABLE_HOOK_CB` and `LUAXTABLE_DESTROYER_CB` in a single macro..

lib/luaxtable.c
sheharyaar: No, we can remove this.

lib/luaxtable.c
```diff
@@ -32,48 +14,150 @@
...
+
+	lua_insert(L, base + 1); /* op */
+	lua_pop(L, 1); /* table */
+
+	if (lua_pcall(L, nargs, nret, 0) != LUA_OK) {
+		pr_err("%s error: %s\n", op, lua_tostring(L, -1));
+		ret = -ECANCELED;
+		goto err;
+	}
+	if (nret != 0)
```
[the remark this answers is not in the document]
sheharyaar: Yeah..  0 for void {mt/tg}_destroy callbacks and 1 for others

lib/luaxtable.c
lneto: ah.. I see! ack

lib/luaxtable.c
sheharyaar: I think it will be better to keep logically different sections to be seperate (destroy and hook), what do you suggest ?

lib/luaxtable.c
lneto: perhaps..
```suggestion
	return ret ? luaL_optinteger(L, -1, 0) : fallback;
```

lib/luaxtable.c
lneto: but why repeating code? you can have a `LUAXTABLE_CB` and define `LUAXTABLE_HOOK_CB` and `LUAXTABLE_DESTROYER_CB` using it..

```suggestion
	lunatik_run(xtable->runtime, cb, ret, xtable, skb, par, fallback);	\
	return ret;		\
```

lib/luaxtable.c
lneto: yeah

lib/luaxtable.c
```diff
@@ -32,48 +14,150 @@
...
+	lunatik_run(xtable->runtime, luaxtable_do##hook, ret, xtable, skb, par);	\
+	return ret < 0 ? luaxtable_hooks.hook##_fallback : ret;		\
 }
 
-static void luaxtable_target_destroy(const struct xt_tgdtor_param *par)
-{
+#define LUAXTABLE_CHECKER_CB(hook, hk, huk, HOOK)				\
+static int luaxtable_##hook##_check(const struct xt_##hk##chk_param *par)	\
+{										\
+	int ret = -EINVAL;							\
```
lneto, line 126: it looks like this value is not being used anywhere, that's right?

README.md
```diff
@@ -1103,6 +1103,43 @@ sudo lunatik run examples/filter/sni false   # runs the Lua kernel script
 sudo xdp-loader load -m skb <ifname> https.o # loads the XDP/eBPF program
 ```
 
+### dnsblock
+
+This is a lua program to block DNS from a blacklist. To run this program follow the instructions below.
+
+#### Usage
+
+1. Build the userspace program
```
lneto, line 1112: Please mind the other examples.. we are using one single shell listing and using comments (e.g., `# builds the userspace module`) to give instructions like this

README.md
```diff
@@ -1103,6 +1103,43 @@ sudo lunatik run examples/filter/sni false   # runs the Lua kernel script
 sudo xdp-loader load -m skb <ifname> https.o # loads the XDP/eBPF program
 ```
 
+### dnsblock
+
+This is a lua program to block DNS from a blacklist. To run this program follow the instructions below.
```
lneto, line 1108: Please mind the other examples, we are linking the actual example code on every doc.. see filter above, for instance..

examples/dnsblock/libxt_dnsblock.c
```diff
@@ -0,0 +1,63 @@
...
+}
+
+static struct xtables_match ipaddr_mt_reg = {
+	.version       = XTABLES_VERSION,
+	.name          = "dnsblock",
+	.revision      = 1,
+	.family        = NFPROTO_UNSPEC,
+	.size          = XT_ALIGN(sizeof(luaxtable_info_t)),
+	.userspacesize = 0,
+	/* called when user execs "iptables -m ipaddr -h" */
```
lneto, line 46: I think you used `ipaddr` module as a base and forgot to rename your namespace, right? Please, adjust accordingly..

lib/luaxtable.c
```diff
@@ -88,6 +191,15 @@ static const lunatik_class_t luaxtable_class = {
 	.sleep = false,
 };
 
+static inline void luaxtable_setbuf(lua_State *L, int idx, luaxtable_t *xtable)
```
lneto, line 194:
```suggestion
static inline void luaxtable_setbuffer(lua_State *L, int idx, luaxtable_t *xtable)
```

README.md
```diff
@@ -1103,6 +1103,24 @@ sudo lunatik run examples/filter/sni false   # runs the Lua kernel script
...
+This kernel extension drops any outbound DNS packet with questions matching the blacklist
+provided by the user.
+
+#### Usage
+
+```
+sudo make examples_install						# installs examples
+cd examples/dnsblock
+make											# builds the userspace extension for netfilter
+sudo make install								# installs the extension to Xtables directory
```
lneto, line 1119: it's unaligned.. please, check how it's rendered on [GitHub](https://github.com/sheharyaar/lunatik/tree/xtables-rcu?tab=readme-ov-file#usage-8)..

README.md
```diff
@@ -1103,6 +1103,24 @@ sudo lunatik run examples/filter/sni false   # runs the Lua kernel script
 sudo xdp-loader load -m skb <ifname> https.o # loads the XDP/eBPF program
 ```
 
+### dnsblock
+
+[dnsblock](examples/dnsblock) is a kernel extension composed by
```
lneto, line 1108: ?

examples/dnsblock/libxt_dnsblock.c
```diff
@@ -0,0 +1,57 @@
...
+	.parse         = ipaddr_mt_parse,
+	.final_check   = ipaddr_mt_check,
+	.print         = ipaddr_mt_print,
+	.save          = ipaddr_mt_save,
+};
+
+static void __attribute__((constructor)) _init(void)
+{
+	xtables_register_match(&ipaddr_mt_reg);
+}
```
lneto, line 57:
```suggestion
}

```

examples/dnsblock/libxt_dnsblock.c
lneto: I believe it's not resolved.. you should adjust your namespace as well, right?

sheharyaar: Ah, yes

lib/luaxtable.c
sheharyaar: In the latest code, the number of parameters have changed, only 2 lines are similar, should I move them to a new macro ?

examples/dnsblock/dnsblock.lua
```diff
@@ -0,0 +1,89 @@
...
+end
+
+local function get_domainname (skb, off)
+	local name = ""
+	local i = 0
+	while true do
+		local d = skb:getbyte(off + i)
+		if d == 0 then
+			break
+		end
```
lneto, line 40: I'm not very fond of loops without proper ending conditions.. you could change this to `repeat ... until d == 0` right?

examples/dnsblock/dnsblock.lua
```diff
@@ -0,0 +1,89 @@
...
+
+local blacklist = {
+	"github.com",
+	"gitlab.com",
+}
+
+local function nop() end
+
+local function check_blacklist (name)
+	for _,v in ipairs(blacklist) do
```
lneto, line 25:
```suggestion
local function check_blacklist(name)
	for _, v in ipairs(blacklist) do
```

please review the code style in this file as a whole.. this pattern is scattered along the file..

examples/dnsblock/dnsblock.lua
```diff
@@ -0,0 +1,89 @@
...
+		name = name .. "."
+	end
+	return name
+end
+
+local function match_packet(skb, par)
+	print("match_packet called")
+
+	-- ip header
+	local thoff = par["thoff"]
```
lneto, line 60:
```suggestion
	local thoff = par.thoff
```

examples/dnsblock/dnsblock.lua
```diff
@@ -0,0 +1,89 @@
...
+		local d = skb:getbyte(off + i)
+		if d == 0 then
+			break
+		end
+
+		i = i + 1
+		local len = d
+
+		while len > 0 do
+			name = name .. string.char(skb:getbyte(off + i))
```
lneto, line 46:
```suggestion
			name = name .. skb:getstring(off + i, 1)
```

examples/dnsblock/dnsblock.lua
lneto: btw, perhaps it's a better idea to just find the last position (e.g., `d == 0`) and then convert the string as a whole, using `skb:getstring(off, length)`

examples/dnsblock/dnsblock.lua
```diff
@@ -0,0 +1,89 @@
...
+local blacklist = {
+	"github.com",
+	"gitlab.com",
+}
+
+local function nop() end
+
+local function check_blacklist (name)
+	for _,v in ipairs(blacklist) do
+		if string.find(name, v) ~= nil then
```
lneto, line 26: are you using `find` to handle subdomains? is that the case?

sheharyaar: Yes

examples/dnsblock/dnsblock.lua
```diff
@@ -0,0 +1,89 @@
...
+				print("DNS query for " .. name .. " blocked\n")
+				return action.ACCEPT
+			end
+		end
+	end
+
+	return action.DROP
+end
+
+local match_ops = {
```
lneto, line 78: this is quite a minor comment and totally up to you.. but it would look better to me to have use the netfilter module name on this table.. (e.g., `local dnsblock_mt = { ... }`) at the beginning of this module and then `function dnsblock_mt.match(...)` or suppressing the local variable here and doing:

```
xt.match{
	name = "dnsblock",
	...
}
```

both ways look better to me.. but again.. up to you ;-)

lib/luaxtable.c
```diff
@@ -32,48 +14,169 @@
...
+static inline void luaxtable_pushparams(lua_State *L, const struct xt_action_param *par)
 {
+	lua_newtable(L);
+	lua_pushboolean(L, par->hotdrop);
+	lua_setfield(L, -2, "hotdrop");
+	lua_pushinteger(L, par->thoff);
+	lua_setfield(L, -2, "thoff");
+	lua_pushinteger(L, par->fragoff);
+	lua_setfield(L, -2, "fragoff");
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(4, 10, 0))
```
lneto, line 90: mixed feelings about supporting such old kernels.. the code base isn't compatible with this already, but perhaps it might be useful if one wants to use only luaxtable.. what do you think? @sav any thoughts?

lib/luaxtable.c
```diff
@@ -32,48 +14,169 @@
...
 }
 
-static int luaxtable_match_check(const struct xt_mtchk_param *par)
+static inline lunatik_object_t *luaxtable_getskb(lua_State *L, luaxtable_t *xtable)
 {
-	return 0;
+	if (lunatik_getregistry(L, xtable->skb) != LUA_TUSERDATA)
+		return NULL;
+
+	return (lunatik_object_t *)lunatik_toobject(L, -1);
```
lneto, line 78:
```suggestion
	return lunatik_getregistry(L, xtable->skb) == LUA_TUSERDATA ? lunatik_toobject(L, -1) : NULL;
```

lneto: btw, shouldn't we pop userdatum from stack here? not really sure.. please check if needed..

examples/dnsblock/dnsblock.lua
sheharyaar: The issue here is that, DNS uses question name of the following format :
`<length1><prefix1><length2><prefix2><length3><prefix3>....<lengthN><prefixN><00>`

Example : 
`\x03www\x06google\x03com\x00`

I think `repeat..until d == 0` will be a better choice

sheharyaar: `repeat..until d == 0`, will also have an issue, 

```lua
	repeat
		d = skb:getbyte(off + i)
		if d == 0 then
			break
		end
		...
```

I still need this condition `d == 0` check to prevent the loop reading further bytes

sheharyaar: `repeat..until d == 0`, will also have an issue, 

```lua
	repeat
		d = skb:getbyte(off + i)
		if d == 0 then
			break
		end
		...
```

I still need this condition `d == 0` check to prevent the loop reading further bytes

sheharyaar:
One of my comments on this suggestion is not showing up (though I posted it multiple times, some issue from Github side) 

Suggestion : https://github.com/luainkernel/lunatik/pull/122#discussion_r1677131899

Comment : The space between function name and `(` for the rest of the file, right ?

examples/dnsblock/dnsblock.lua
lneto: > The issue here is that, DNS uses question name of the following format : `<length1><prefix1><length2><prefix2><length3><prefix3>....<lengthN><prefixN><00>`
> 
> Example : `\x03www\x06google\x03com\x00`

you can just use gsub later on, right?

lneto: btw, instead of looking for `\0`, checking byte per byte, you could just transform the rest of the payload into a string and apply pattern matching using string facilities (e.g., `match`, `gsub`), right? wouldn't that be a simpler approach?

lib/luaxtable.c
```diff
@@ -32,48 +14,169 @@
...
+
+	luadata_reset(data, skb->data, skb_headlen(skb), LUADATA_OPT_READONLY);
+	luaxtable_pushparams(L, par);
+	/* take reference of param table */
+	lua_pushvalue(L, -1);
+	lua_insert(L, lua_gettop(L) - 2);
+
+	ret = luaxtable_invoke(L, xtable, "match", 2, 1, luaxtable_hooks.match_fallback);
+	lua_getfield(L, -2, "hotdrop");
+	par->hotdrop = lua_toboolean(L, -1);
```
lneto, line 115: you should pop the the par table here, right?

lib/luaxtable.c
```diff
@@ -32,48 +14,169 @@
...
 {
-	return NF_ACCEPT;
+	int ret;
+	lunatik_object_t *data = luaxtable_getskb(L, xtable);
+	if (data == NULL) {
+		pr_err("could not find skb\n");
+		return luaxtable_hooks.match_fallback;
+	}
+
+	luadata_reset(data, skb->data, skb_headlen(skb), LUADATA_OPT_READONLY);
```
lneto, line 107: this, at least, can be moved to an auxiliary function.. perhaps an inline.. (e.g., `pushskb`)

lib/luaxtable.c
```diff
@@ -32,48 +14,169 @@
...
+	luaxtable_pushparams(L, par);
+	/* take reference of param table */
+	lua_pushvalue(L, -1);
+	lua_insert(L, lua_gettop(L) - 2);
```
lneto, line 111:
```suggestion
	lua_pushvalue(L, -1); /* param table */
	lua_insert(L, lua_gettop(L) - 2); /* stack: param, skb, param */
```

lib/luaxtable.c
```diff
@@ -32,48 +14,169 @@
...
+	if (lua_pcall(L, nargs, nret, 0) != LUA_OK) {
+		pr_err("%s error: %s\n", op, lua_tostring(L, -1));
+		goto err;
+	}
+	if (nret == 1) {
+		ret = luaL_optinteger(L, -1, 0);
+		ret = ret < 0 ? fallback : ret;
+	}
+err:
+	return ret;
```
lneto, line 70: here we are not returning fallback in case of error.. is it expected?

sheharyaar: I should probable initialise `ret = fallback` instead of `ret = -1`, It should handle all the edge cases.

lneto: in that case, I think you don't need `ret` you can return fallback directly on `err: `, it might be clearer..

sheharyaar: Ok

sheharyaar: Ah but, I will have to add another return statement above `err:`

sheharyaar: So it would be : 
```c
if (nret == 1) {
		ret = luaL_optinteger(L, -1, 0);
		ret = ret < 0 ? fallback : ret;
	}
	return ret;
err:
	return fallback;
```

examples/dnsblock/dnsblock.lua
```diff
@@ -0,0 +1,74 @@
...
+end
+
+local function get_domainname(skb, off)
+	local name = ""
+	local i = 0
+
+	repeat
+		local d = skb:getbyte(off + i)
+		i = i + 1
+	until (d == 0)
```
lneto, line 40:
```suggestion
	until d == 0
```

examples/dnsblock/dnsblock.lua
```diff
@@ -0,0 +1,74 @@
...
+	local proto = skb:getuint8(9)
+
+	if proto == udp then
+		local dstport = linux.ntoh16(skb:getuint16(thoff + 2))
+		if dstport == dns then
+			local qoff = thoff + 20
+			local name = get_domainname(skb, qoff)
+			if check_blacklist(name) then
+				print("DNS query for " .. name .. " blocked\n")
+				return action.ACCEPT
```
lneto, line 57: shouldn't it be `DROP` here? actually, `match` shouldn't return a bool?

examples/dnsblock/dnsblock.lua
```diff
@@ -0,0 +1,74 @@
...
+	repeat
+		local d = skb:getbyte(off + i)
+		i = i + 1
+	until (d == 0)
+
+	return name .. skb:getstring(off, i)
+end
+
+local function dnsblock_mt(skb, par)
+	-- ip header
```
lneto, line 46: what you mean here?

lib/luaxtable.c
```diff
@@ -88,6 +192,15 @@ static const lunatik_class_t luaxtable_class = {
 	.sleep = false,
 };
 
+static inline void luaxtable_setbuffer(lua_State *L, int idx, luaxtable_t *xtable)
+{
+	lunatik_require(L, data);
+	xtable->skb = lunatik_checknull(L, luadata_new(NULL, 0, false, LUADATA_OPT_NONE));
+	lunatik_cloneobject(L, xtable->skb);
```
lneto, line 199: why cloning the object here?

lib/luaxtable.c
```diff
@@ -88,6 +192,15 @@ static const lunatik_class_t luaxtable_class = {
 	.sleep = false,
 };
 
+static inline void luaxtable_setbuffer(lua_State *L, int idx, luaxtable_t *xtable)
```
lneto, line 195: it's actually `newbuffer`, right? as you are creating new one bellow on every call..

lib/luaxtable.c
```diff
@@ -32,48 +14,170 @@
...
+}
+
+#define LUAXTABLE_CHECKER_CB(hook, hk, huk, HOOK)				\
+static int luaxtable_##hook##_check(const struct xt_##hk##chk_param *par)	\
+{										\
+	int ret;								\
+	luaxtable_t *xtable;							\
+	lunatik_object_t *obj = luarcu_gettable(luaxtable_hooks.hook, par->hook->name, XT_EXTENSION_MAXNAMELEN);	\
+	if (obj == NULL) {							\
+		pr_err("could not find table for %s\n", par->hook->name);	\
```
lneto, line 151: which table do you refer here? `xtable`? `rcu`?

lib/luaxtable.c
```diff
@@ -32,48 +14,170 @@
...
+	return 0;
 }
 
-static int luaxtable_target_check(const struct xt_tgchk_param *par)
+static int luaxtable_domatch(lua_State *L, luaxtable_t *xtable, const struct sk_buff *skb, struct xt_action_param *par)
 {
-	return 0;
+	int ret;
+	if (luaxtable_pushskb(L, xtable, skb, LUADATA_OPT_READONLY) == -1)
+		return luaxtable_hooks.match_fallback;
```
lneto, line 116: I would move this to `pushparams` as well..

lib/luaxtable.c
```diff
@@ -32,48 +14,170 @@
...
+	lua_pushinteger(L, par->fragoff);
+	lua_setfield(L, -2, "fragoff");
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(4, 10, 0))
+	lua_pushinteger(L, xt_hooknum(par));
+#else
+	lua_pushinteger(L, par->hooknum);
+#endif
+	lua_setfield(L, -2, "hooknum");
+	lua_pushvalue(L, -1); /* param table */
+	lua_insert(L, lua_gettop(L) - 2); /* stack: param, skb, param */
```
lneto, line 97: you can't assume `skb` is here at this point, unless you move `pushskb` to this function.. in this case, I would update the comment (because) you can't assume what is below them in the stack..
```suggestion
	lua_insert(L, lua_gettop(L) - 2); /* stack: (...), param, skb, param */
```

lib/luaxtable.c
```diff
@@ -137,13 +251,22 @@ static int luaxtable_new##hook(lua_State *L) 				\
...
+	lunatik_checkfield(L, 1, #hook, LUA_TFUNCTION);			\
 									\
+	hook->usersize = 0;						\
+	hook->hook##size = sizeof(luaxtable_info_t);			\
 	hook->hook = luaxtable_##hook;					\
 	hook->checkentry = luaxtable_##hook##_check;			\
 	hook->destroy = luaxtable_##hook##_destroy;			\
 									\
+	if (luarcu_settable(luaxtable_hooks.hook, hook->name, XT_EXTENSION_MAXNAMELEN, object) != 0)	\
+		luaL_error(L, "unable to set key : %s\n", hook->name);	\
```
lneto, line 265: I think we might have a better error message to the API user.. `unable to set key` is a bit cryptic.. which key? key for what? perhaps we just need to signalize that we were `unable to create the hook` (or something like that)

lib/luaxtable.c
```diff
@@ -32,48 +14,170 @@
...
+	lua_pop(L, 1); /* table */
+
+	if (lua_pcall(L, nargs, nret, 0) != LUA_OK) {
+		pr_err("%s error: %s\n", op, lua_tostring(L, -1));
+		goto err;
+	}
+	if (nret == 1) {
+		int ret = luaL_optinteger(L, -1, 0);
+		ret = ret < 0 ? fallback : ret;
+		return ret;
```
lneto, line 67:
```suggestion
		int ret = luaL_optinteger(L, -1, 0);
		return ret < 0 ? fallback : ret;
```

what happens if it's returned a positive invalid value? are we handling it somewhere.. what would happen if we simply
```suggestion
		return luaL_optinteger(L, -1, fallback);
```?

what netfilter will do?

File: lib/luaxtable.h [hunk 68]
```diff
@@ -0,0 +1,16 @@
...
+* SPDX-License-Identifier: MIT OR GPL-2.0-only
+*/
+
+#ifndef luaxtable_h
+#define luaxtable_h
+
+struct luaxtable_s;
+
+typedef struct luaxtable_info_s {
+	struct luaxtable_s *data __attribute__((aligned(8)));
```
lneto, line 12: can we use this to pass "generic" parameters from the userspace? similarly to what we are doing on eBPF -> luaxdp? I would add an parm in your example.. perhap the domain to be blocked.. or a the name of a lua table containing the blacklist.. (or anything else ;-)

lib/luaxtable.c
```diff
@@ -32,48 +14,170 @@
...
 {
-	return 0;
+	int ret;
+	if (luaxtable_pushskb(L, xtable, skb, LUADATA_OPT_READONLY) == -1)
+		return luaxtable_hooks.match_fallback;
+	luaxtable_pushparams(L, par);
+
+	ret = luaxtable_invoke(L, xtable, "match", 2, 1, luaxtable_hooks.match_fallback);
+	lua_getfield(L, -2, "hotdrop");
+	par->hotdrop = lua_toboolean(L, -1);
```
lneto, line 121: btw, why can't we set `hotdrop` on target?

examples/dnsblock/dnsblock.lua
sheharyaar: Actually `ACCEPT` evaluates to true -> the packet matches correctly. Should I  just return true ?

`ACCEPT`, `DROP` and others are generally used by `nfhook`.

examples/dnsblock/dnsblock.lua
sheharyaar: Ah, that should not be here, it was there earlier, when I was manually extracting the udp offset. Now we use `thoff`

lib/luaxtable.c
sheharyaar: we are reusing the skb luadata object, hence I name it setbuffer.

lib/luaxtable.c
sheharyaar: `rcu` table

lib/luaxtable.c
sheharyaar: Ah, yes, got your point, it should be a `newbuffer`, but since I was using `set{integer,string}` for setting xtable values, I used it.

lib/luaxtable.c
lneto: the API client has no clue of the RCU table, right? it's transparent to them.. I would change this error message..

lib/luaxtable.c
lneto: I meant, we are always creating a new luadata object (therefore, a buffer).. I think `set` is misleading here.. `new` would be clearer..

sheharyaar: Yes, I will change it in the next push

examples/dnsblock/dnsblock.lua
sheharyaar: Yes

lib/luaxtable.c
sheharyaar: What can be a better message here ?

lib/luaxtable.h
sheharyaar: We can pass parameters to this module, that would be covered in the user-library generation.

lib/luaxtable.c
sheharyaar: Because target callback sends it's arguments as `const xt_action_param`, so we are not allowed to set `hotdrop` here, though we can force it, but that is not intended. `hotdrop` is generally used when the packet qualifies for that particular module, but is not in the correct format or somehow corrupt. For example, `conntrack` uses hotdrop to drop packets if it was not able to get memory using `kmalloc`.

lib/luaxtable.c
sheharyaar: The return value from target or match is used at many places in the kernel. For each type of table or module (ipv4, ipv6, arp, netlink). For `target` verdicts, it is either compared to a `NF_*` value or first `AND` with `NF_VERDICT_MASK (0x000000ff)` and then compared using `== NF_*`. The behaviour is neither documented nor uniform. It's better we sanitise the positive value also.

Regarding the second example, we already discussed that `fallback` will be set if ret is either `nil` or not of the correct type, but if the module itself returns an illegal value (negative in case `match` which expects bool), it would pass this check.

lneto: > It's better we sanitise the positive value also.

agreed

lib/luaxtable.c
lneto: perhaps..
```suggestion
		pr_err("could not find hook (%s)\n", par->hook->name);	\
```

didn't like that much, but I could think on something much better.. =)

lib/luaxtable.c
sheharyaar: sanitising using the following condition : 

```c
	if (nret == 1) {
		int ret = luaL_optinteger(L, -1, 0);
		ret = (ret < 0 || ret > NF_MAX_VERDICT) ? fallback : ret;
		return ret;
	}

```

lib/luaxtable.c
sheharyaar: `luadata_new` allocates memorym sets the class and returns a pointer to luadata object, right ? So it's not on the stack yet. So we need to clone it, then `lunatik_setregistry` in the next line will call : 
```c
	lua_pushvalue(L, ix);
	lua_rawsetp(L, LUA_REGISTRYINDEX, key); /* pop value */
```

Otherwise, it panics. I used `lunatik_cloneobject` since it calls lunatik_setclass, which pushes it onto the stack : 
```c
	if (luaL_getmetatable(L, class->name) == LUA_TNIL)
		luaL_error(L, "metatable not found (%s)", class->name);
	lua_setmetatable(L, -2);
	lua_pushlightuserdata(L, (void *)class);
	lua_setiuservalue(L, -2, 1); /* pop class */
```

lib/luaxtable.c
```diff
@@ -88,6 +192,15 @@ static const lunatik_class_t luaxtable_class = {
 	.sleep = false,
 };
 
+static inline void luaxtable_setbuffer(lua_State *L, int idx, luaxtable_t *xtable)
+{
+	lunatik_require(L, data);
+	xtable->skb = lunatik_checknull(L, luadata_new(NULL, 0, false, LUADATA_OPT_NONE));
+	lunatik_cloneobject(L, xtable->skb);
+	lunatik_setregistry(L, -1, xtable->skb);
+	lua_pop(L, 1);
```
lneto, line 201:
```suggestion
	lua_pop(L, 1); /* skb */
```

lib/luaxtable.c
```diff
@@ -32,48 +14,190 @@
...
 
-static unsigned int luaxtable_target(struct sk_buff *skb, const struct xt_action_param *par)
+static int luaxtable_domatch(lua_State *L, luaxtable_t *xtable, const struct sk_buff *skb, struct xt_action_param *par)
 {
-	return NF_ACCEPT;
+	int ret;
+	if (luaxtable_pushparams(L, par, xtable, skb, LUADATA_OPT_READONLY) == -1)
+		goto err;
+
+	if (luaxtable_invoke(L, xtable, "match", 2, 1) == -1)
```
lneto, line 118: this can be an `or` and moved to a macro

lib/luaxtable.c
```diff
@@ -32,48 +14,190 @@
...
+	if (obj == NULL) {							\
+		pr_err("could not find hook (%s)\n", par->hook->name);		\
+		return -EINVAL;							\
+	}									\
+	xtable = (luaxtable_t *)obj->private;					\
+	luaxtable_info_t *info = (luaxtable_info_t *)par->huk##info; 		\
+	info->data = xtable;							\
+										\
+	lunatik_run(xtable->runtime, luaxtable_docheckentry, ret, xtable);	\
+	return ret;								\
```
lneto, line 179:
```suggestion
	return ret != 0 ? -EINVAL : 0;								\
```

lneto, line 179:
```suggestion
	lunatik_run(xtable->runtime, luaxtable_invoke, ret, xtable, "checkentry", 0, 1);	\
	return ret != 0 ? -EINVAL : 0;								\
```

lib/luaxtable.c
```diff
@@ -32,48 +14,190 @@
...
+	return luaxtable_hooks.target_fallback;
 }
 
-static void luaxtable_target_destroy(const struct xt_tgdtor_param *par)
+static int luaxtable_docheckentry(lua_State *L, luaxtable_t *xtable)
 {
+	if (luaxtable_invoke(L, xtable,"checkentry", 0, 1) == -1)
+		return -EINVAL;
+	return 0;
+}
```
lneto, line 151:
```suggestion
```

lib/luaxtable.c
```diff
@@ -32,48 +14,190 @@
...
+
+	if (luaxtable_invoke(L, xtable, "target", 2, 1) == -1)
+		goto err;
+
+	ret = lua_tointeger(L, -1);
+	if (ret < 0 || ret > NF_MAX_VERDICT)
+		goto err;
+	return ret;
+err:
+	return luaxtable_hooks.target_fallback;
```
lneto, line 143:
```suggestion
	int fallback = luaxtable_hooks.target_fallback;
	
	if (luaxtable_call(L, xtable, par, skb, "target", LUADATA_OPT_NONE) != 0)
		return fallback;

	int ret = lua_tointeger(L, -1);
	return ret >= 0 && ret <= NF_MAX_VERDICT ? ret : fallback;
```

lib/luaxtable.c
```diff
@@ -32,48 +14,190 @@
...
+}
+
+#define LUAXTABLE_HOOK_CB(T, U, V, hook, huk) 				\
+static T luaxtable_##hook(U skb, V par)					\
+{									\
+	int ret;							\
+	const luaxtable_info_t *info = (const luaxtable_info_t *)par->huk##info;	\
+	luaxtable_t *xtable = info->data;				\
+									\
+	lunatik_run(xtable->runtime, luaxtable_do##hook, ret, xtable, skb, par);	\
```
lneto, line 160: you can actually pass `fallback` here, so you can avoid repeat code on `do{match,target}`

lneto, line 160:
```suggestion
	luaxtable_t *xtable = info->data;				\
	T fallback = luaxtable_hooks.##hook##_fallback;	\
									\
	lunatik_run(xtable->runtime, luaxtable_do##hook, ret, xtable, skb, par, fallback);	\
```

lib/luaxtable.c
lneto:
```
#define luaxtable_call(L, xtable, skb, par, op, opt)	\
	(luaxtable_pushparams((L), (par), (xtable), (skb), (opt)) || luaxtable_invoke((L), (xtable), op, 2, 1))
```

then

```
if (luaxtable_call(L, xtable, par, skb, "match", LUADATA_OPT_READONLY))
	goto err;
```

lib/luaxtable.c
```diff
@@ -32,48 +14,190 @@
...
+		pr_err("could not find skb\n");
+		return -1;
+	}
+
+	luadata_reset(data, skb->data, skb_headlen(skb), opt);
 	return 0;
 }
 
-static void luaxtable_match_destroy(const struct xt_mtdtor_param *par)
+static inline int luaxtable_pushparams(lua_State *L, const struct xt_action_param *par, luaxtable_t *xtable, const struct sk_buff *skb, uint8_t opt)
```
lneto, line 89: this function has grow a bit.. I think there's no point in being an `inline` anymore.. btw, perhaps it's a good idea to move the body of `pushskb` inhere because it's sufficient small and it's not being used anywhere else.. it might clean up the code a bit.. what do you think?

lib/luaxtable.c
```diff
@@ -32,48 +14,165 @@
...
+{										\
+	int ret;								\
+	luaxtable_info_t *info = (luaxtable_info_t *)par->huk##info; 		\
+	luaxtable_t *xtable = (luaxtable_t *)info->data;			\
+										\
+	lunatik_run(xtable->runtime, luaxtable_invoke, ret, xtable, "destroy", 0, 0);	\
+}
+
+LUAXTABLE_HOOK_CB(bool, const struct  sk_buff *, struct xt_action_param *, match, match);
+LUAXTABLE_HOOK_CB(unsigned int, struct sk_buff *, const struct xt_action_param *, target, targ);
```
lneto, line 168: I would change the args ordering.. the hook itself sounds way more relevant while reading this.. and return type sounds like the latest thing
```suggestion
LUAXTABLE_HOOK_CB(match, match, const struct  sk_buff *, struct xt_action_param *, bool);
LUAXTABLE_HOOK_CB(target, targ, struct sk_buff *, const struct xt_action_param *, unsigned int);
```

lib/luaxtable.c
```diff
@@ -32,48 +14,165 @@
...
+	lua_pushinteger(L, par->hooknum);
+#endif
+	lua_setfield(L, -2, "hooknum");
+	lua_pushvalue(L, -1); /* param table */
+	lua_insert(L, lua_gettop(L) - 2); /* stack: (...), param, skb, param */
+	return 0;
 }
 
-static int luaxtable_target_check(const struct xt_tgchk_param *par)
+#define luaxtable_call(L, xtable, par, skb, op, opt) \
```
lneto, line 104: `op` seems the most relevant here as well.. and let's preserve the ordering `skb, par` across the internal API
```suggestion
#define luaxtable_call(L, op, xtable, skb, par, opt) \
```

lib/luaxtable.c
```diff
@@ -32,48 +14,165 @@
...
+	int ret = lua_toboolean(L, -1);
+	lua_getfield(L, -2, "hotdrop");
+	par->hotdrop = lua_toboolean(L, -1);
+	return ret;
 }
 
-static void luaxtable_target_destroy(const struct xt_tgdtor_param *par)
+static int luaxtable_dotarget(lua_State *L, luaxtable_t *xtable, struct sk_buff *skb, const struct xt_action_param *par, int fallback)
 {
+	if (luaxtable_call(L, xtable, par, skb, "match", LUADATA_OPT_READONLY))
```
lneto, line 121: I think `!= 0` or `< 0` would be a bit clearer..
```suggestion
	if (luaxtable_call(L, xtable, par, skb, "match", LUADATA_OPT_READONLY) != 0)
```

lib/luaxtable.c
```diff
@@ -32,48 +14,165 @@
...
-}
+static struct {
+	lunatik_object_t *target;
+	lunatik_object_t *match;
+	bool match_fallback;
+	unsigned int target_fallback;
+} luaxtable_hooks = {NULL, NULL, false, XT_CONTINUE};
 
-static int luaxtable_match_check(const struct xt_mtchk_param *par)
+static int luaxtable_invoke(lua_State *L, luaxtable_t *xtable, const char *op, int nargs, int nret)
```
lneto, line 43: as we have `call` now, perhaps it would be clearer to rename this to `docall`?

sheharyaar: Yes

lib/luaxtable.c
```diff
@@ -209,11 +327,15 @@ LUNATIK_NEWLIB(xtable, luaxtable_lib, &luaxtable_class, luanetfilter_flags);
 
 static int __init luaxtable_init(void)
 {
+	luaxtable_hooks.match = luarcu_newtable(LUARCU_DEFAULT_SIZE, false);
```
lneto, line 330: Actually, this can fail (e.g., return NULL), in this case you should return an error condition.. probably `-ENOMEM`

lib/luaxtable.c
```diff
@@ -137,13 +246,22 @@ static int luaxtable_new##hook(lua_State *L) 				\
...
+	lunatik_checkfield(L, 1, #hook, LUA_TFUNCTION);			\
 									\
+	hook->usersize = 0;						\
+	hook->hook##size = sizeof(luaxtable_info_t);			\
 	hook->hook = luaxtable_##hook;					\
 	hook->checkentry = luaxtable_##hook##_check;			\
 	hook->destroy = luaxtable_##hook##_destroy;			\
 									\
+	if (luarcu_settable(luaxtable_hooks.hook, hook->name, XT_EXTENSION_MAXNAMELEN, object) != 0)	\
+		luaL_error(L, "unable to create the hook: %s\n", hook->name);	\
```
lneto, line 260:
```suggestion
		luaL_error(L, "unable to hook %s\n", hook->name);	\
```

README.md
```diff
@@ -1127,6 +1127,53 @@ flags to Lua.
...
+* `match_opts`: a table containing the following fields:
+  * `name`: string representing the xtable extension name.
+  * `revision`: integer representing the xtable extension revision.
+  * `family`: address family, one of [xtable.family](https://github.com/luainkernel/lunatik#xtablefamily).
+  * `proto`: protocol number.
+  * `match` : function to be called for matching packets. It receives the following arguments:
+	* `skb`: a `data` object representing the socket buffer.
+	* `par`: a table containing `hotdrop`, `thoff` (transport header offset) and `fragoff` (fragment offset) fields.
+    * The function must return `true` if the packet matches the extension; otherwise, it must return `false`.
+  * `checkentry` : function to be called for checking the entry.
```
lneto, line 1147:
```suggestion
  * `checkentry`: function to be called for checking the entry.
```

lib/luaxtable.c
```diff
@@ -209,11 +327,19 @@ LUNATIK_NEWLIB(xtable, luaxtable_lib, &luaxtable_class, luanetfilter_flags);
 
 static int __init luaxtable_init(void)
 {
+	if ((luaxtable_hooks.match = luarcu_newtable(LUARCU_DEFAULT_SIZE, false)) == NULL)
```
lneto, line 330: you can move this to a macro and use an or here to avoid repeating this pattern..

README.md
```diff
@@ -1127,6 +1127,53 @@ flags to Lua.
 * `"TX"`: n/a.
 * `"REDIRECT"`: n/a.
 
+### xtable
+
+The `xtable` library provides support for developing netfilter xtable extensions.
+
+#### `xtable.match(match_opts)`
```
lneto, line 1134:
```suggestion
#### `xtable.match(opts)`
```
or
```suggestion
#### `xtable.match(desc)`
```
didn't like much both.. but could think on a better name.. anyway, there's no point on repeating `match` prefix

README.md
```diff
@@ -1127,6 +1127,53 @@ flags to Lua.
...
+
+_xtable.match()_ returns a new `xtable` object for matching packets.
+This function receives the following arguments:
+* `match_opts`: a table containing the following fields:
+  * `name`: string representing the xtable extension name.
+  * `revision`: integer representing the xtable extension revision.
+  * `family`: address family, one of [xtable.family](https://github.com/luainkernel/lunatik#xtablefamily).
+  * `proto`: protocol number.
+  * `match` : function to be called for matching packets. It receives the following arguments:
```
lneto, line 1143: If we are going to name this field `match`, we could have a single function for adding hooks (e.g., `xtable.new()`) and take the decision based on such field.. otherwise, I would change this field name (e.g., `callback`)

README.md
```diff
@@ -1127,6 +1127,53 @@ flags to Lua.
 * `"TX"`: n/a.
 * `"REDIRECT"`: n/a.
 
+### xtable
+
+The `xtable` library provides support for developing netfilter xtable extensions.
```
lneto, line 1132: It would be good to add a link to a xtable reference

lib/luaxtable.c
lneto: yeah.. you're right, I missed that.. sorry.. but I would move this to a macro anyway.. something like..
```
#define luaxtable_inithook(hook)	(luaxtable_hooks.##hook = luarcu_newtable(LUARCU_DEFAULT_SIZE, false))
```

lib/luaxtable.c
```diff
@@ -209,11 +327,19 @@ LUNATIK_NEWLIB(xtable, luaxtable_lib, &luaxtable_class, luanetfilter_flags);
...
+		return -ENOMEM;
+	if ((luaxtable_hooks.target = luarcu_newtable(LUARCU_DEFAULT_SIZE, false)) == NULL) {
+		lunatik_putobject(luaxtable_hooks.match);
+		return -ENOMEM;
+	}
```
lneto, line 335:
```suggestion
	if (luaxtable_inithook(match) == NULL)
		goto err;
	if (luaxtable_inithook(target) == NULL)
		goto put;
	return 0;
put:
	luaxtable_puthook(match);
err:
	return -ENOMEM;
}
```

lib/luaxtable.c
```diff
@@ -209,11 +327,19 @@ LUNATIK_NEWLIB(xtable, luaxtable_lib, &luaxtable_class, luanetfilter_flags);
...
 	return 0;
 }
 
 static void __exit luaxtable_exit(void)
 {
+	lunatik_putobject(luaxtable_hooks.match);
+	lunatik_putobject(luaxtable_hooks.target);
```
lneto, line 342:
```suggestion
	luaxtable_puthook(match);
	luaxtable_puthook(target);
```

README.md
lneto: this doesn't look resolved.. please, see xdp section as an example.. it links to XDP documentation.. it's a good thing especially for who is not that familiar with kernel internals..

README.md
```diff
@@ -1259,6 +1322,22 @@ sudo lunatik run examples/filter/sni false   # runs the Lua kernel script
...
+This script drops any outbound DNS packet with question matching the blacklist provided by the user.
+
+#### Usage
+
+```
+sudo make examples_install  			# installs examples
+cd examples/dnsblock
+make    					# builds the userspace extension for netfilter
+sudo make install   				# installs the extension to Xtables directory
+sudo lunatik run examples/dnsblock/dnsblock     # runs the Lua kernel script
```
lneto, line 1337:
```suggestion
sudo lunatik run examples/dnsblock/dnsblock false    # runs the Lua kernel script
```

lib/luaxtable.c
```diff
@@ -174,9 +293,19 @@ static const lunatik_reg_t luanetfilter_action[] = {
...
+	{"FORWARD", NF_INET_FORWARD},
+	{"LOCAL_OUT", NF_INET_LOCAL_OUT},
+	{"POST_ROUTING", NF_INET_POST_ROUTING},
+	{NULL, 0}
+};
+
 static const lunatik_namespace_t luanetfilter_flags[] = {
 	{"action", luanetfilter_action},
 	{"family", luanetfilter_family},
+	{"hooks", luanetfilter_hooks},
```
lneto, line 308:
```suggestion
	{"inet", luanetfilter_inet},
```

PR #134, opened 2024-06-25
add set,get (u)int8,16,24,32,64 support to luadata

sheharyaar:


lib/luadata.c
```diff
@@ -24,6 +24,7 @@
 #define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
 #include <linux/kernel.h>
 #include <linux/module.h>
+#include <linux/types.h>
```
lneto, line 27: I don't think it's necessary as we include this on [include/stdint.h](https://github.com/luainkernel/lunatik/blob/master/include/stdint.h) header.

lib/luadata.c
```diff
@@ -50,41 +51,76 @@ static inline luadata_t *luadata_checkdata(lua_State *L, lua_Integer *offset, lu
...
 
-	*(lua_Integer *)(data->ptr + offset) = luaL_checkinteger(L, 3);
-	return 0;
+static inline int luadata_setbyte(lua_State *L) {
+	return luadata_setuint8(L);
 }
 
-static int luadata_getbyte(lua_State *L)
+#ifdef __LP64__
+static inline int luadata_getnumber(lua_State *L)
```
lneto, line 104: I think we can deprecate it.. as it can be a simply alias in lua (e.g., `local getnumber = data.getint64`).. what do you think?

lib/luadata.c
```diff
@@ -50,41 +51,76 @@ static inline luadata_t *luadata_checkdata(lua_State *L, lua_Integer *offset, lu
...
-{
-	lua_Integer offset;
-	luadata_t *data = luadata_checkdata(L, &offset, LUADATA_NUMBER_SZ);
+LUADATA_NEWSETTER(int8);
+LUADATA_NEWSETTER(uint8);
+LUADATA_NEWSETTER(int16);
+LUADATA_NEWSETTER(uint16);
+LUADATA_NEWSETTER(int32);
+LUADATA_NEWSETTER(uint32);
+
```
lneto, line 87:
```suggestion
```

lib/luadata.c
```diff
@@ -50,41 +51,76 @@ static inline luadata_t *luadata_checkdata(lua_State *L, lua_Integer *offset, lu
...
+}
 
-	lua_pushinteger(L, *(lua_Integer *)(data->ptr + offset));
-	return 1;
+LUADATA_NEWGETTER(int8);
+LUADATA_NEWGETTER(uint8);
+LUADATA_NEWGETTER(int16);
+LUADATA_NEWGETTER(uint16);
+LUADATA_NEWGETTER(int32);
+LUADATA_NEWGETTER(uint32);
```
lneto, line 69: let's move this to below `LUADATA_NEWSETTER` definition and declare getters and setters alternately.. the ifdef 64 will be cleaner in this way I guess.. e.g.,
```
LUADATA_NEWGETTER(int8);
LUADATA_NEWSETTER(int8);
...
```

actually, we can also have a macro that bundle both.. it will clean up the code a bit more I think.. e.g.,

```
#define LUADATA_NEWINT(T)	(LUADATA_NEWGETTER(T); LUADATA_NEWSETTER(T))
```

lib/luadata.c
```diff
@@ -197,4 +223,3 @@ module_init(luadata_init);
 module_exit(luadata_exit);
 MODULE_LICENSE("Dual MIT/GPL");
 MODULE_AUTHOR("Lourival Vieira Neto <lourival.neto@ring-0.io>");
```
lneto, line 225:
```suggestion
MODULE_AUTHOR("Lourival Vieira Neto <lourival.neto@ring-0.io>");

```

lib/luadata.c
```diff
@@ -50,41 +50,41 @@ static inline luadata_t *luadata_checkdata(lua_State *L, lua_Integer *offset, lu
...
-	luadata_t *data = luadata_checkdata(L, &offset, LUADATA_NUMBER_SZ);
-
-	lua_pushinteger(L, *(lua_Integer *)(data->ptr + offset));
-	return 1;
+#define LUADATA_NEWINT_GETTER(T) 	\
+static int luadata_get##T(lua_State *L) \
+{					\
+	lua_Integer offset;		\
+	luadata_t *data = luadata_checkdata(L, &offset, sizeof(T##_t));	\
+					\
```
lneto, line 58:
```suggestion
```

lib/luadata.c
```diff
@@ -50,41 +50,41 @@ static inline luadata_t *luadata_checkdata(lua_State *L, lua_Integer *offset, lu
...
-	luadata_t *data = luadata_checkdata(L, &offset, LUADATA_NUMBER_SZ);
-
-	*(lua_Integer *)(data->ptr + offset) = luaL_checkinteger(L, 3);
-	return 0;
+#define LUADATA_NEWINT_SETTER(T)		\
+static int luadata_set##T(lua_State *L)		\
+{						\
+	lua_Integer offset;			\
+	luadata_t *data = luadata_checkdata(L, &offset, sizeof(T##_t));	\
+						\
```
lneto, line 68:
```suggestion
```

lib/luadata.c
```diff
@@ -50,41 +50,41 @@ static inline luadata_t *luadata_checkdata(lua_State *L, lua_Integer *offset, lu
...
-	lua_Integer offset;
-	luadata_t *data = luadata_checkdata(L, &offset, LUADATA_NUMBER_SZ);
-
-	lua_pushinteger(L, *(lua_Integer *)(data->ptr + offset));
-	return 1;
+#define LUADATA_NEWINT_GETTER(T) 	\
+static int luadata_get##T(lua_State *L) \
+{					\
+	lua_Integer offset;		\
+	luadata_t *data = luadata_checkdata(L, &offset, sizeof(T##_t));	\
```
lneto, line 57:
```
#define luadata_checkint(L, offset, T)  	luadata_checkdata((L), &(offset), sizeof(T##_t))
```

just to avoid repeat this on getter and setter..

lib/luadata.c
```diff
@@ -122,10 +122,31 @@ static const luaL_Reg luadata_lib[] = {
 static const luaL_Reg luadata_mt[] = {
 	{"__index", lunatik_monitorobject},
 	{"__gc", lunatik_deleteobject},
-	{"getnumber", luadata_getnumber},
-	{"setnumber", luadata_setnumber},
-	{"getbyte", luadata_getbyte},
-	{"setbyte", luadata_setbyte},
+	{"getbyte", luadata_getuint8},
```
lneto, line 125: wise.. in that way we don't break filter, right?

lneto (changes requested):
just few comments more.. everything looks to me besides that.. I can merge it promptly after this changes.. thanks!

sav (approved):
looks good to me. good job.

lib/luadata.c
```diff
@@ -50,41 +50,41 @@ static inline luadata_t *luadata_checkdata(lua_State *L, lua_Integer *offset, lu
...
-{
-	lua_Integer offset;
-	luadata_t *data = luadata_checkdata(L, &offset, 1);
-
-	lua_pushinteger(L, (lua_Integer)*(data->ptr + offset));
-	return 1;
-}
+#define LUADATA_NEWINT(T)	\
+	LUADATA_NEWINT_GETTER(T); 	\
+	LUADATA_NEWINT_SETTER(T);
```
lneto, line 75:
```suggestion
#define LUADATA_NEWINT(T)	(LUADATA_NEWINT_GETTER(T);  LUADATA_NEWINT_SETTER(T))
```

lib/luadata.c
lneto: originally we were using a signed integer for byte, right?

lib/luadata.c
lneto: my bad.. you can ignore it =)

lneto (approved):

Issue #128, opened 2024-06-21
luaxdp: failure to load on linux 6.9.4


glk0:
Using insmod to load luaxdp returns the following error:
```
insmod: ERROR: could not insert module lib/luaxdp.ko: Invalid parameters
```
I found out that the error wasn't related to the params of the module (insmod output is misleading) and that the failure to load the module was due to the init function of the module returning `-EINVAL` from:

```c
return register_btf_kfunc_id_set(BPF_PROG_TYPE_XDP, &bpf_luaxdp_kfunc_set);
```

Turns out that [here](https://elixir.bootlin.com/linux/latest/source/kernel/bpf/btf.c#L8209),  the `register_btf_kfunc_id_set` function in my kernel have this line:

```c
if (!(kset->set->flags & BTF_SET8_KFUNCS)) {
    WARN_ON(!kset->owner);
    return -EINVAL;
}
```
And that the BTF_SET8_START (name) always set the flag to 0 [here](https://elixir.bootlin.com/linux/latest/source/include/linux/btf_ids.h#L201) so the check always fails. 
I'm not familiar with XDP, BPF nor luaxdp, just followed the error. What does the call to register_btf_kfunc_id_set do anyway ?

lneto:
@glk0 can you try this on luaxdp.c?
```
s/BTF_SET8_START/BTF_KFUNCS_START/g
s/BTF_SET8_END/BTF_KFUNCS_END/g
s/__diag_pop/__bpf_kfunc_end_defs/g
```

lneto:
> What does the call to register_btf_kfunc_id_set do anyway ?

It's necessary for us to register our kfunc "bpf_luaxdp_run" to be called from an eBPF program (such as examples/filter/https.c)..

glk0:
OK thanks. Made the substitutions, and the new module loads without issue.

lneto:
> @lneto OK thanks. Made the substitutions, and the new module loads without issue.

awesome, thanks! can you please open a PR with that change? we'll need ifdefs for the kernel version that introduce it.. I believe it was 6.9

lneto:
@glk0 btw, which distro are you using with 6.9.x?

glk0:
@lneto Arch

PR #137, opened 2024-06-30
fix luaxdp BTF flags for kernel >= 6.9.0

sheharyaar:
Fixes : #128

### DIFF
--- a/lib/luaxdp.c
+++ b/lib/luaxdp.c
@@ -104,11 +104,21 @@ out:
 	return action;
 }
 
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 7, 0))
+__bpf_kfunc_end_defs();
+#else
 __diag_pop();
+#endif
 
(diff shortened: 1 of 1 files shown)


lib/luaxdp.c
```diff
@@ -106,9 +106,15 @@ __bpf_kfunc int bpf_luaxdp_run(char *key, size_t key__sz, struct xdp_md *xdp_ctx
 
 __diag_pop();
```
lneto, line 107: I think we should also add this..
```suggestion
#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 9, 0))
__bpf_kfunc_end_defs();
#else
__diag_pop();
#endif
```

sheharyaar: This should be for >= 6.4.0, right?

lneto: Yup, you're right!

lneto (approved):

PR #139, opened 2024-07-01
add editable field to luadata

sheharyaar:
This PR allows us to handle const data using luadata.

### DIFF
--- a/lib/luadata.c
+++ b/lib/luadata.c
@@ -13,10 +13,12 @@
 
 #include <lunatik.h>
 
+#include "luadata.h"
+
 typedef struct luadata_s {
 	char *ptr;
 	size_t size;
-	bool free;
+	uint8_t opt;
 } luadata_t;
 
 #define LUADATA_NUMBER_SZ	(sizeof(lua_Integer))
@@ -35,6 +37,8 @@ static inline luadata_t *luadata_checkdata(lua_State *L, lua_Integer *offset, lu
 
 #define luadata_checkint(L, offset, T)	luadata_checkdata((L), &(offset), sizeof(T##_t))
 
+#define luadata_checkwritable(L, data)	luaL_argcheck((L), ((data)->opt & LUADATA_OPT_READONLY), 1, "read only")
+
 #define LUADATA_NEWINT_GETTER(T) 	\
 static int luadata_get##T(lua_State *L) \
 {					\
@@ -49,6 +53,7 @@ static int luadata_set##T(lua_State *L)		\
 {						\
 	lua_Integer offset;			\
 	luadata_t *data = luadata_checkint(L, offset, T);			\
+	luadata_checkwritable(L, data);		\
 	*(T##_t *)(data->ptr + offset) = (T##_t)luaL_checkinteger(L, 3);	\
 	return 0;				\
 }
@@ -85,6 +90,7 @@ static int luadata_setstring(lua_State *L)
 	size_t length;
 	const char *str = luaL_checklstring(L, 3, &length);
 	luadata_t *data = luadata_checkdata(L, &offset, (lua_Integer)length);
+	luadata_checkwritable(L, data);
 
 	memcpy(data->ptr + offset, str, length);
 	return 0;
@@ -100,7 +106,7 @@ static int luadata_length(lua_State *L)
 static void luadata_release(void *private)
 {
 	luadata_t *data = (luadata_t *)private;
-	if (data->free)
+	if (data->opt & LUADATA_OPT_FREE)
 		lunatik_free(data->ptr);
 }
 
@@ -157,13 +163,13 @@ static int luadata_lnew(lua_State *L)
 
 	data->ptr = lunatik_checkalloc(L, size);
 	data->size = size;
-	data->free = true;
+	data->opt = LUADATA_OPT_FREE;
 	return 1; /* object */
 }
 
 LUNATIK_NEWLIB(data, luadata_lib, &luadata_class, NULL);
 
-lunatik_object_t *luadata_new(void *ptr, size_t size, bool sleep)
+lunatik_object_t *luadata_new(void *ptr, size_t size, bool sleep, uint32_t opt)
 {
 	lunatik_object_t *object = lunatik_createobject(&luadata_class, sizeof(luadata_t), sleep);
 
@@ -171,26 +177,28 @@ lunatik_object_t *luadata_new(void *ptr, size_t size, bool sleep)
 		luadata_t *data = (luadata_t *)object->private;
 		data->ptr = ptr;
 		data->size = size;
-		data->free = false;
+		data->opt = opt;
 	}
 	return object;
 }
 EXPORT_SYMBOL(luadata_new);
 
-int luadata_reset(lunatik_object_t *object, void *ptr, size_t size)
+int luadata_reset(lunatik_object_t *object, void *ptr, size_t size, uint8_t opt)
 {
 	luadata_t *data;
 
 	lunatik_lock(object);
 	data = (luadata_t *)object->private;
 
-	if (data->free) {
+	if (data->opt & LUADATA_OPT_FREE) {
 		lunatik_unlock(object);
 		return -1;
 	}
 
 	data->ptr = ptr;
 	data->size = size;
+	data->opt = opt & LUADATA_OPT_KEEP ? data->opt : opt;
+
 	lunatik_unlock(object);
 	return 0;
 }
--- a/lib/luadata.h
+++ b/lib/luadata.h
@@ -6,16 +6,23 @@
 #ifndef luadata_h
 #define luadata_h
 
+#include <lunatik.h>
+
 LUNATIK_LIB(data);
 
-#define luadata_clear(o)	(luadata_reset((o), NULL, 0))
+#define	LUADATA_OPT_NONE	0x00
+#define	LUADATA_OPT_READONLY	0x01
+#define	LUADATA_OPT_FREE	0x02
+#define	LUADATA_OPT_KEEP  	0x80
 
-lunatik_object_t *luadata_new(void *ptr, size_t size, bool sleep);
-int luadata_reset(lunatik_object_t *object, void *ptr, size_t size);
+#define luadata_clear(o)	(luadata_reset((o), NULL, 0, LUADATA_OPT_KEEP))
+
+lunatik_object_t *luadata_new(void *ptr, size_t size, bool sleep, uint32_t opt);
+int luadata_reset(lunatik_object_t *object, void *ptr, size_t size, uint8_t opt);
 
 static inline void luadata_close(lunatik_object_t *object)
 {
-	luadata_reset(object, NULL, 0);
+	luadata_clear(object);
 	lunatik_putobject(object);
 }
 
--- a/lib/luaxdp.c
+++ b/lib/luaxdp.c
@@ -36,7 +36,7 @@ static inline lunatik_object_t *luaxdp_pushdata(lua_State *L, int upvalue, void 
 
 	lua_pushvalue(L, lua_upvalueindex(upvalue));
 	data = (lunatik_object_t *)lunatik_toobject(L, -1);
-	luadata_reset(data, ptr, size);
+	luadata_reset(data, ptr, size, LUADATA_OPT_KEEP);
 	return data;
 }
 
@@ -129,7 +129,7 @@ static const struct btf_kfunc_id_set bpf_luaxdp_kfunc_set = {
 
 static inline void luaxdp_newdata(lua_State *L)
 {
-	lunatik_object_t *data = lunatik_checknull(L, luadata_new(NULL, 0, false));
+	lunatik_object_t *data = lunatik_checknull(L, luadata_new(NULL, 0, false, LUADATA_OPT_NONE));
 	lunatik_cloneobject(L, data);
 }
 


lib/luadata.c
```diff
@@ -140,21 +150,25 @@ static const lunatik_class_t luadata_class = {
 	.release = luadata_release,
 };
 
-static int luadata_lnew(lua_State *L)
-{
+#define LUADATA_LNEW(L, edit)		\
+do {					\
+} while(0)
```
lneto, line 155: ?

lib/luadata.c
```diff
@@ -47,7 +50,11 @@ static int luadata_get##T(lua_State *L) \
 static int luadata_set##T(lua_State *L)		\
 {						\
 	lua_Integer offset;			\
-	luadata_t *data = luadata_checkint(L, offset, T);			\
+	luadata_t *data = luadata_checkint(L, offset, T);	\
```
lneto, line 53: what did change here?

lib/luadata.c
```diff
@@ -47,7 +50,11 @@ static int luadata_get##T(lua_State *L) \
...
+						\
```
lneto, line 54:
```suggestion
```

lib/luadata.c
```diff
@@ -47,7 +50,11 @@ static int luadata_get##T(lua_State *L) \
...
+	if (!data->editable)			\
+		luaL_error(L, "data is not editable");		\
+						\
```
lneto, line 57:
```suggestion
```

lib/luadata.c
```diff
@@ -163,30 +177,46 @@ lunatik_object_t *luadata_new(void *ptr, size_t size, bool sleep)
...
+	}					\
+						\
+	data->ptr = p;				\
+	data->size = sz;			\
+} while(0)
+
+int luadata_reset(lunatik_object_t *object, void *ptr, size_t size, bool editable)
 {
 	luadata_t *data;
-
```
lneto, line 174: ?

lib/luadata.c
```diff
@@ -163,30 +177,46 @@ lunatik_object_t *luadata_new(void *ptr, size_t size, bool sleep)
...
 		data->size = size;
 		data->free = false;
+		data->editable = editable;
 	}
 	return object;
 }
 EXPORT_SYMBOL(luadata_new);
 
-int luadata_reset(lunatik_object_t *object, void *ptr, size_t size)
+#define luadata_resetter(L, p, sz)		\
```
lneto, line 186: why can't it be an inline?

lib/luadata.c
```diff
@@ -85,6 +92,9 @@ static int luadata_setstring(lua_State *L)
 	const char *str = luaL_checklstring(L, 3, &length);
 	luadata_t *data = luadata_checkdata(L, &offset, (lua_Integer)length);
 
+	if (!data->editable)
+		luaL_error(L, "data is not editable");
```
lneto, line 96: you can move this to an inline, right? see luasocket for this sort of checker example..

lib/luadata.h
```diff
@@ -6,16 +6,17 @@
 #ifndef luadata_h
 #define luadata_h
 
-LUNATIK_LIB(data);
+#include <lunatik.h>
```
lneto, line 9: why doing this?

lib/luadata.c
```diff
@@ -163,30 +177,46 @@ lunatik_object_t *luadata_new(void *ptr, size_t size, bool sleep)
...
+	if (data->free) {			\
+		lunatik_unlock(object);		\
+		return -1;			\
+	}					\
+						\
+	data->ptr = p;				\
+	data->size = sz;			\
+} while(0)
+
+int luadata_reset(lunatik_object_t *object, void *ptr, size_t size, bool editable)
```
lneto, line 199: what's the case for changing the `editable` property of a data object?

lib/luadata.c
sheharyaar: Ah, I forgot to clean it up

lib/luadata.c
sheharyaar: Nothing, just the `\` indentation

lib/luadata.c
sheharyaar: I need to have `data->editable` after the macro to set it here : https://github.com/luainkernel/lunatik/pull/139/files#diff-7af6ccd3b938d3efe6965a52e59f84c7f34aae6f825864536661cf83f544a277R199-R200

lneto: I still miss the point.. AFAIK, you can do the same in a inline function.. perhaps I'm missing something.. the same accounts for the checker, for me

lib/luadata.c
```diff
@@ -13,9 +13,12 @@
 
 #include <lunatik.h>
 
+#include "luadata.h"
+
 typedef struct luadata_s {
 	char *ptr;
 	size_t size;
+	uint32_t flags;
```
lneto, line 21: if you are using `typedef enum ... luadata_flags_t`, you should use this type here, right?

lib/luadata.h
```diff
@@ -6,16 +6,23 @@
 #ifndef luadata_h
 #define luadata_h
 
+#include <lunatik.h>
+
 LUNATIK_LIB(data);
 
+typedef enum {
+	LUADATA_CONST = 1,
```
lneto, line 14: my point on using a flag was to handle every "bool" with it.. e.g., `LUADATA_SLEEP` instead of having a separate field in the struct for it.. (and anything else we need in the future)

lib/luadata.c
```diff
@@ -84,7 +94,7 @@ static int luadata_setstring(lua_State *L)
 	size_t length;
 	const char *str = luaL_checklstring(L, 3, &length);
 	luadata_t *data = luadata_checkdata(L, &offset, (lua_Integer)length);
-
```
lneto, line 87: let's not remove these `\n`'s.. I've asked to remove them only from templates.. to "force" us to keep such macros small and coarse ;-).. avoiding doing too much on them and structure our code a bit better..

lib/luadata.c
```diff
@@ -174,19 +186,32 @@ int luadata_reset(lunatik_object_t *object, void *ptr, size_t size)
 
 	lunatik_lock(object);
 	data = (luadata_t *)object->private;
-
 	if (data->free) {
 		lunatik_unlock(object);
 		return -1;
 	}
-
```
lneto, line 182: and here...

lib/luadata.c
```diff
@@ -174,19 +186,32 @@ int luadata_reset(lunatik_object_t *object, void *ptr, size_t size)
...
 	data->ptr = ptr;
 	data->size = size;
 	lunatik_unlock(object);
 	return 0;
 }
 EXPORT_SYMBOL(luadata_reset);
 
+int luadata_setflags(lunatik_object_t *object, uint32_t flags)
```
lneto, line 200: I don't think we need this API, I would use `luadata_reset` for that..

lib/luadata.c
```diff
@@ -34,6 +37,12 @@ static inline luadata_t *luadata_checkdata(lua_State *L, lua_Integer *offset, lu
 
 #define luadata_checkint(L, offset, T)	luadata_checkdata((L), &(offset), sizeof(T##_t))
 
+static inline void luadata_editablechecker(lua_State *L, luadata_t *data)
```
lneto, line 40: I would use `*CHECKER` for the template and `check*` for the function name..
```suggestion
static inline void luadata_checkeditable(lua_State *L, luadata_t *data)
```

lib/luadata.c
```diff
@@ -34,6 +37,12 @@ static inline luadata_t *luadata_checkdata(lua_State *L, lua_Integer *offset, lu
...
+{
+	if (data->flags & LUADATA_CONST)
+		luaL_error(L, "data is not editable");
```
lneto, line 43:
```suggestion
		luaL_error(L, "not editable");
```
I would follow the pattern of the other error messages.. e.g., "out of bounds".. perhaps "read only" would be a better message in this case.. I'm not very fond of `editable`, btw

lib/luadata.c
```diff
@@ -157,45 +167,69 @@ static int luadata_lnew(lua_State *L)
...
+		data->flags = *flags;
+
 	lunatik_unlock(object);
 	return 0;
 }
 EXPORT_SYMBOL(luadata_reset);
 
+int luadata_getflags(lunatik_object_t *object, luadata_flags_t *flags)
+{
+	if (flags == NULL)
```
sav, line 214: Do we expect this function to be called with `flags=NULL`? If not, I suggest removing the validation and assuming the argument to be non-NULL by contract. Feel free to ignore this suggestion if you like.

sav (approved):

lib/luadata.c
lneto: we moved this discussion to [matrix](https://matrix.to/#/!RHdYAjlWKfZRHOoOoB:matrix.org/$Oz5M8Aks4B1mffk6L881yyc90KYBCqRuRY0YC5eoln4?via=matrix.org&via=envs.net&via=nixserve.net).. please hop into there ;-)

lib/luadata.c
```diff
@@ -157,40 +167,43 @@ static int luadata_lnew(lua_State *L)
...
-	if (data->free) {
+	if (data->flags & LUADATA_FREE) {
 		lunatik_unlock(object);
 		return -1;
 	}
 
 	data->ptr = ptr;
 	data->size = size;
+	if (flags < 0)
+		data->flags = flags;
```
lneto, line 205:
```suggestion
	data->opt = opt & LUADATA_OPT_KEEP ? data->opt : opt;
```

lib/luaxdp.c
```diff
@@ -129,7 +129,7 @@ static const struct btf_kfunc_id_set bpf_luaxdp_kfunc_set = {
 
 static inline void luaxdp_newdata(lua_State *L)
 {
-	lunatik_object_t *data = lunatik_checknull(L, luadata_new(NULL, 0, false));
+	lunatik_object_t *data = lunatik_checknull(L, luadata_new(NULL, 0, false, 0));
```
lneto, line 132:
```suggestion
	lunatik_object_t *data = lunatik_checknull(L, luadata_new(NULL, 0, LUADATA_OPT_NONE));
```

lneto (commented):
@sav I would appreciate if you can give your opinion on my suggestions 🙏

lib/luadata.c
```diff
@@ -35,6 +37,12 @@ static inline luadata_t *luadata_checkdata(lua_State *L, lua_Integer *offset, lu
 
 #define luadata_checkint(L, offset, T)	luadata_checkdata((L), &(offset), sizeof(T##_t))
 
+static inline void luadata_checkeditable(lua_State *L, luadata_t *data)
+{
+	if (data->flags & LUADATA_READONLY)
+		luaL_error(L, "read only");
```
sav, line 43:
```suggestion
		luaL_error(L, "read only data");
```

lib/luadata.h
```diff
@@ -6,16 +6,23 @@
...
 
+#include <lunatik.h>
+
 LUNATIK_LIB(data);
 
-#define luadata_clear(o)	(luadata_reset((o), NULL, 0))
+typedef enum {
+	LUADATA_READONLY = 1,
+	LUADATA_FREE = 2,
+} luadata_flags_t;
```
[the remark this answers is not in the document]
sav: If I got this right, if `luadata_flags_t` really changes to `uint8_t`, these defines must be at max `0xff`. In that case I'd go with:
```C
#define LUADATA_OPT_KEEP	0x80
```

sav:
> @sav I would appreciate if you can give your opinion on my suggestions 🙏

sav:
Sorry, I closed the issue accidentally.

lib/luadata.c
```diff
@@ -157,40 +167,43 @@ static int luadata_lnew(lua_State *L)
...
 	data->size = size;
-	data->free = true;
+	data->flags = LUADATA_FREE;
 	return 1; /* object */
 }
 
 LUNATIK_NEWLIB(data, luadata_lib, &luadata_class, NULL);
 
-lunatik_object_t *luadata_new(void *ptr, size_t size, bool sleep)
+lunatik_object_t *luadata_new(void *ptr, size_t size, bool sleep, uint32_t flags)
```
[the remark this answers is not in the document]
sheharyaar: For sleep, we can't change it later, right ?? it is used in `lunatik_createobject`, I think we should keep it the same.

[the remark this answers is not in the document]
sheharyaar: @sav @lneto

lib/luadata.c
```diff
@@ -3,6 +3,7 @@
 * SPDX-License-Identifier: MIT OR GPL-2.0-only
 */
 
+#include <stdint.h>
```
lneto, line 6: we don't need it will be included by `lunatik.h`

lib/luadata.c
```diff
@@ -35,6 +38,12 @@ static inline luadata_t *luadata_checkdata(lua_State *L, lua_Integer *offset, lu
 
 #define luadata_checkint(L, offset, T)	luadata_checkdata((L), &(offset), sizeof(T##_t))
 
+static inline void luadata_checkwritable(lua_State *L, luadata_t *data)
+{
+	if (data->opt & LUADATA_OPT_READONLY)
+		luaL_error(L, "read only data");
+}
+
```
lneto, line 46: in this case, of adding "data" to the error message, I would use argcheck instead, so it will print the `data` object as well giving as a better error handling.. 
```suggestion
#define luadata_checkwritable(L, data, ix)	luaL_argcheck((L), (data)->opt & LUADATA_OPT_READONLY), 2, "read only")
```

PR #141, opened 2024-07-01
add endian conversion support to lualinux

sheharyaar:


lneto (commented):
as we are creating higher level APIs, I believe we don't need to be tied up to the Linux's namespace.. I think [NetBSD's API names](https://man.netbsd.org/le16dec.9) are more sane.. e.g., `be32toh` instead of `be32_to_cpu`. What do you think? Also, aren't `ntoh*` and `hton*` just aliases to `be*_to_cpu` and `cpu_to_be*` respectively?

@sav would be good to hear your thoughts about this as well ;-)

lib/lualinux.c
```diff
@@ -104,6 +105,34 @@ static int lualinux_lookup(lua_State *L)
...
+LUALINUX_NEW_ENDIAN_CONVERTER(htonl, uint32_t);
+LUALINUX_NEW_ENDIAN_CONVERTER(cpu_to_be16, uint32_t);
+LUALINUX_NEW_ENDIAN_CONVERTER(cpu_to_be32, uint32_t);
+LUALINUX_NEW_ENDIAN_CONVERTER(cpu_to_le16, uint32_t);
+LUALINUX_NEW_ENDIAN_CONVERTER(cpu_to_le32, uint32_t);
+LUALINUX_NEW_ENDIAN_CONVERTER(be16_to_cpu, uint32_t);
+LUALINUX_NEW_ENDIAN_CONVERTER(be32_to_cpu, uint32_t);
+LUALINUX_NEW_ENDIAN_CONVERTER(le16_to_cpu, uint32_t);
+LUALINUX_NEW_ENDIAN_CONVERTER(le32_to_cpu, uint32_t);
+
```
lneto, line 128:
```suggestion
```

lib/lualinux.c
```diff
@@ -104,6 +105,34 @@ static int lualinux_lookup(lua_State *L)
 	return 1;
 }
 
+#define LUALINUX_NEW_ENDIAN_CONVERTER(method, T) \
```
lneto, line 108:
```suggestion
#define LUALINUX_NEW_BYTESWAPPER(func, T) \
```

"byte swapper" is a bit shorter than "endian converter".. and `method` is a bit misleading as it is not operating on an object..

lneto: good work on this template, btw 💪

sheharyaar:
> as we are creating higher level APIs, I believe we don't need to be tied up to the Linux's namespace.. I think [NetBSD's API names](https://man.netbsd.org/le16dec.9) are more sane.. e.g., `be32toh` instead of `be32_to_cpu`. What do you think? Also, aren't `ntoh*` and `hton*` just aliases to `be*_to_cpu` and `cpu_to_be*` respectively?
> 
> @sav would be good to hear your thoughts about this as well ;-)

Yes, `hton*` and `ntoh*` are aliases to `be*_to_cpu`, we can move to it, I used the native names for easy macro.

lneto (approved):

PR #144, opened 2024-07-08
support an specific device when using AF_PACKET

sav:
Currently, the `AF_PACKET` socket does not allow modifying the `sockaddr_ll` to specify a particular device. This changes allow the application to pass the interface as the address argument of `send`.

### DIFF
--- a/lib/lualinux.c
+++ b/lib/lualinux.c
@@ -9,6 +9,7 @@
 #include <linux/sched.h>
 #include <linux/jiffies.h>
 #include <linux/ktime.h>
+#include <linux/netdevice.h>
 #include <linux/byteorder/generic.h>
 
 #include <lua.h>
@@ -102,6 +103,17 @@ static int lualinux_lookup(lua_State *L)
 	const char *symbol = luaL_checkstring(L, 1);
 
 	lua_pushlightuserdata(L, lunatik_lookup(symbol));
+	return 1;
+}
+
+static int lualinux_ifindex(lua_State *L)
+{
+	const char *ifname = luaL_checkstring(L, 1);
+	struct net_device *dev = dev_get_by_name(&init_net, ifname);
+
+	luaL_argcheck(L, dev != NULL, 1, "device not found");
+	lua_pushinteger(L, dev->ifindex);
+	dev_put(dev);
 	return 1;
 }
 
@@ -213,6 +225,7 @@ static const luaL_Reg lualinux_lib[] = {
 	{"time", lualinux_time},
 	{"difftime", lualinux_difftime},
 	{"lookup", lualinux_lookup},
+	{"ifindex", lualinux_ifindex},
 	{"ntoh16", lualinux_be16_to_cpu},
 	{"ntoh32", lualinux_be32_to_cpu},
 	{"hton16", lualinux_cpu_to_be16},
--- a/lib/luasocket.c
+++ b/lib/luasocket.c
@@ -64,6 +64,10 @@ static void luasocket_checkaddr(lua_State *L, struct socket *socket, luasocket_a
 		addr_in->sin_addr.s_addr = htonl((u32)luaL_checkinteger(L, ix));
 		addr_in->sin_port = htons((u16)luaL_checkinteger(L, ix + 1));
 	}
+	else if (addr->family == AF_PACKET) {
+		struct sockaddr_ll *addr_ll = (struct sockaddr_ll *)addr;
+		addr_ll->sll_ifindex = (int)luaL_checkinteger(L, ix);
+	}
 	else {
 		size_t len;
 		const char *addr_data = luaL_checklstring(L, ix, &len);


lib/luasocket.c
```diff
@@ -56,15 +56,32 @@ typedef struct luasocket_addr_s {
...
+	const char *ifname = luaL_checklstring(L, ix, NULL);
+	struct net_device *dev = dev_get_by_name(&init_net, ifname);
+	if (!dev) {
+		lua_pushinteger(L, -ENODEV);
+		lua_error(L);
+	}
+	ifindex = dev->ifindex;
+	dev_put(dev);
+	return ifindex;
+}
```
lneto, line 71: perhaps this should be added as a Lua binding.. ideally in a `net` library, but could be put temporarily on `linux` lib.. in this way it could be used to other purposes as well.. and in this case we could do something like..
```lua
local ifindex = linux.ifindex("eth0")
sock:send("hello", ifindex)
```

what do you think? btw, I didn't like much the name `ifindex()` for the API, but I couldn't think on something better..

@sheharyaar would be good to hear your thoughts here as well..

lib/luasocket.c
```diff
@@ -56,15 +56,32 @@ typedef struct luasocket_addr_s {
 static int luasocket_new(lua_State *L);
 static int luasocket_accept(lua_State *L);
 
+static int luasocket_checkiface(lua_State *L, int ix)
+{
+	int ifindex;
+	const char *ifname = luaL_checklstring(L, ix, NULL);
+	struct net_device *dev = dev_get_by_name(&init_net, ifname);
```
lneto, line 63: in the case of creating such APIs, we could handle namespaces as well..

lib/luasocket.c
```diff
@@ -56,15 +56,32 @@ typedef struct luasocket_addr_s {
...
 	addr->family = socket->sk->sk_family;
 	if (addr->family == AF_INET) {
 		struct sockaddr_in *addr_in = (struct sockaddr_in *)addr;
 		addr_in->sin_addr.s_addr = htonl((u32)luaL_checkinteger(L, ix));
 		addr_in->sin_port = htons((u16)luaL_checkinteger(L, ix + 1));
-	}
-	else {
+	} else if (addr->family == AF_PACKET) {
+		struct sockaddr_ll *addr_ll = (struct sockaddr_ll *)addr;
+		addr_ll->sll_protocol = htons(ETH_P_ALL);
```
lneto, line 82: shouldn't we allow users to specify the ETH proto? or AF_PACKET can only be used with `ETH_P_ALL`?

lib/luasocket.c
sav: you mean something like `dev_get_by_name(current->nsproxy->net_ns)`?
I can try to add support to it right away, just need to understand what you have in mind.

lneto: I meant having namespace as an optional argument.. if it's not passed, falling back to `init_net`, but I would prefer to do this as a next step.. however, it's totally up to you ;-)

lib/luasocket.c
sav: sure, probably. good thing you spotted it, I forgot to mention. I tried to get the `protocol` from `socket->sk`, so that would be enough to specify, like:
```
local socket = socket.new(PACKET, RAW, <PROTO>)
socket:send(frame, 'eth0')
```
and it would re-use the `protocol` passed while creating the socket, but well, I failed badly. 

by the looks of it `socket->sk` does not store the `protocol` directly. it converts to all sorts of things, and it was kind of tough to unravel the code to a meaningful end, so I guessed it wouldn't hurt to pass `ETH_P_ALL` here, but it's still just a guess.

I'll look deeper and try to provide more information on this matter.

sav: just in case, and to make sure we're on the same page: I'm not really fond of the idea of passing the (iface, protocol) as distinct arguments to `send`. point being, it would sound odd to anyone familiar with the "classic" socket API.

The way I see, the best approach would be to re-use the value that should be already there, but as a last resort I'd vouch for something similar to what you suggested above:
```
local addr = socket:addr('eth0', ETH_P_ALL)
socket:send(frame, addr)
```

sav: Sorry, I believe it's safe to omit the protocol. According to the documentation:
```
         sll_protocol
                is the standard ethernet protocol type in network byte order as defined in the <linux/if_ether.h> include file.  It defaults to the socket's protocol.

```

lib/lualinux.c
```diff
@@ -105,6 +106,22 @@ static int lualinux_lookup(lua_State *L)
...
+	if (!dev) {
+		lua_pushinteger(L, -ENODEV);
+		lua_error(L);
+	}
+	ifindex = dev->ifindex;
+	dev_put(dev);
+
+	lua_pushinteger(L, ifindex);
+	return 1;
+}
```
lneto, line 123:
```suggestion
static int lualinux_ifindex(lua_State *L)
{
	const char *ifname = luaL_checkstring(L, 1);
	struct net_device *dev = dev_get_by_name(&init_net, ifname);

	luaL_argcheck(L, dev != NULL, 1, "device not found");
	lua_pushinteger(L, dev->ifindex);
	dev_put(dev);
	return 1;
}
```

as we are not in the realm of luasocket anymore, I would raise a proper error message instead of throwing the `errno`.. also, I would use argcheck to both simplify the code and indicate that it's an invalid argument.. also, though it's not written code style, I would mind that we are using a `\n` after the declarations..

lib/luasocket.c
```diff
@@ -63,8 +63,10 @@ static void luasocket_checkaddr(lua_State *L, struct socket *socket, luasocket_a
 		struct sockaddr_in *addr_in = (struct sockaddr_in *)addr;
 		addr_in->sin_addr.s_addr = htonl((u32)luaL_checkinteger(L, ix));
 		addr_in->sin_port = htons((u16)luaL_checkinteger(L, ix + 1));
-	}
-	else {
+	} else if (addr->family == AF_PACKET) {
+		struct sockaddr_ll *addr_ll = (struct sockaddr_ll *)addr;
+		addr_ll->sll_ifindex = (int)luaL_checkinteger(L, ix);
+	} else {
```
lneto, line 69:
```suggestion
	}
	else if (addr->family == AF_PACKET) {
		struct sockaddr_ll *addr_ll = (struct sockaddr_ll *)addr;
		addr_ll->sll_ifindex = (int)luaL_checkinteger(L, ix);
	}
	else {
```
I would do that for both preserving the code style and to avoid change unnecessary lines in the patch.. I don't mind to change the code style, but I would do this as a separate commit and change the code base as a whole.. at least the whole module..

lneto (commented):
quite minor comments..

lib/luasocket.c
sav: that was my bad. I didn't notice the style. I always go with whatever is already there.

lneto: It's actually my bad it's not documented 😉

lneto (approved):

PR #145, opened 2024-07-08
allow default argument in luadata:getstring

sav:


lib/luadata.c
```diff
@@ -27,11 +27,17 @@ static int luadata_lnew(lua_State *L);
 
 LUNATIK_PRIVATECHECKER(luadata_check, luadata_t *);
 
+#define luadata_checkarg(L, arg, size, offset, length)					\
```
lneto, line 30: I believe this could be an inline, right? I'm trying to use macros just in the case it can't be implemented as inline or it's a simple expression..

lib/luadata.c
```diff
@@ -27,11 +27,17 @@ static int luadata_lnew(lua_State *L);
 
 LUNATIK_PRIVATECHECKER(luadata_check, luadata_t *);
 
+static inline void luadata_checkarg(lua_State *L, int arg, size_t size, lua_Integer offset, lua_Integer length)
+{
+	int cond = offset >= 0 && length > 0 && offset + length <= size;
+	luaL_argcheck(L, cond, arg, "out of bounds");
+}
```
lneto, line 34:
```suggestion
static inline void luadata_checkbounds(lua_State *L, int ix, size_t size, lua_Integer offset, lua_Integer length)
{
	int bounds = offset >= 0 && length > 0 && offset + length <= size;
	luaL_argcheck(L, bounds, ix, "out of bounds");
}
```

lib/luadata.c
```diff
@@ -27,11 +27,17 @@ static int luadata_lnew(lua_State *L);
...
+
 static inline luadata_t *luadata_checkdata(lua_State *L, lua_Integer *offset, lua_Integer length)
 {
 	luadata_t* data = luadata_check(L, 1);
 	*offset = luaL_checkinteger(L, 2);
-	luaL_argcheck(L, *offset >= 0 && length > 0 && *offset + length <= data->size, 2, "out of bounds");
+	luadata_checkarg(L, 2, data->size, *offset, length);
```
lneto, line 40: I'm not sure we should maintain `luadata_checkdata()` only to get `offset`.. it sounds like a unnecessary indirection now we have `luadata_check()` and `luadata_checkbounds()`.. mixed feelings though.. what do you think?

@sheharyaar would like to hear your thoughts as well..

sav: I added it to a different commit in case we decide not to make this change.

lneto (approved):

PR #149, opened 2024-07-17
fix lunatik sleep check

sheharyaar:


lunatik.h
```diff
@@ -130,7 +130,7 @@ static inline void lunatik_checkfield(lua_State *L, int idx, const char *field,
 
 static inline void lunatik_checkclass(lua_State *L, const lunatik_class_t *class)
 {
-	if (lunatik_cannotsleep(L, class->sleep))
+	if (lunatik_checksleep(L, class->sleep))
 		luaL_error(L, "cannot use '%s' class on non-sleepable runtime", class->name);
```
lneto, line 134: I think would be clearer now to check the condition directly (as we are not using it elsewhere also).. we need also to update the error message..
```suggestion
	if (class->sleep != lunatik_toruntime(L)->sleep)
		luaL_error(L, "cannot use '%s' class on runtime (sleep mismatch)", class->name);
```

lunatik.h
```diff
@@ -31,7 +31,7 @@ do {						\
 
 #define lunatik_toruntime(L)	(*(lunatik_object_t **)lua_getextraspace(L))
 
-#define lunatik_cannotsleep(L, s)	((s) && !lunatik_toruntime(L)->sleep)
+#define lunatik_checksleep(L, s)	((s) != lunatik_toruntime(L)->sleep)
```
lneto, line 34:
```suggestion
```

PR #152, opened 2024-07-25
fix dnsblock example, add hooks field

sheharyaar:


examples/dnsblock/dnsblock.lua
```diff
@@ -68,5 +68,6 @@ xt.match{
 	match = dnsblock_mt,
 	checkentry = nop,
 	destroy = nop,
+	hooks = 0,
```
lneto, line 71: Shouldn't this be optional?

sheharyaar: It can be optional, but we have not added optional table field checking in luaxtable.

lneto: what `0` means in this case.. that is, how netfilter will handle that? also, what happens if we pass a invalid value here?

sheharyaar: 0 means, that it can be added to any hook point that is provided by a particular nat table

lneto: Isn't there an "any" flag for this? And what happens in case of invalid value?

sheharyaar: The value is checked against a mask, and if it is not valid, `EINVAL` is returned.

sheharyaar:
```c
	if (par->target->hooks && (par->hook_mask & ~par->target->hooks) != 0) {
		char used[64], allow[64];

		pr_info_ratelimited("%s_tables: %s target: used from hooks %s, but only usable from %s\n",
				    xt_prefix[par->family], par->target->name,
				    textify_hooks(used, sizeof(used),
						  par->hook_mask, par->family),
				    textify_hooks(allow, sizeof(allow),
						  par->target->hooks,
						  par->family));
		return -EINVAL;
	}
```

PR #153, opened 2024-07-29
add xtable userspace library

sheharyaar:


usr/lib/xtable/Makefile.tmpl
```diff
@@ -0,0 +1,21 @@
+CFLAGS = -O2 -Wall -I../../../lib
```
lneto, line 1: add copyright/license to all new files please

usr/lib/xtable/luaxtable.lua
```diff
@@ -0,0 +1,65 @@
...
+
+local function write_file(filename, content)
+    local file = io.open(filename, "w")
+    if file == nil then
+        print("Failed to open file: "..filename)
+        return
+    end
+
+    file:write(content)
+    file:close()
```
lneto, line 24: `<close>` ;-)

usr/lib/xtable/luaxtable.lua
```diff
@@ -0,0 +1,65 @@
...
+    cfile = cfile:gsub("<EXT_REV>", ext_rev or 1)
+    cfile = cfile:gsub("<EXT_FAMILY>", ext_family or "NFPROTO_UNSPEC")
+
+    write_file("libxt_"..ext_name..".c", cfile)
+    write_file("libxt_"..ext_name..".lua", luafile)
+    write_file("Makefile", makefile)
+end
+
+if #arg < 3 then
+    print("Usage: lua "..arg[0].." <module_name> <module_revision> <proto_family>")
```
lneto, line 56: this all can be defined on the script using the Lua library.. you don't need to pass this in the command line..

usr/lib/xtable/luaxtable.c.tmpl
```diff
@@ -0,0 +1,169 @@
...
+    .family = <EXT_FAMILY>,
+    .help = <EXT_NAME>_tg_help,
+    .init = <EXT_NAME>_tg_init,
+    .parse = <EXT_NAME>_tg_parse,
+    .final_check = <EXT_NAME>_tg_check,
+    .print = <EXT_NAME>_tg_print,
+    .save = <EXT_NAME>_tg_save,
+};
+
+static void __attribute__((constructor)) _init(void)
```
lneto, line 148: is there a way to pass some parameters in here?

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,218 @@
...
+		fprintf(stderr, "Function %s not found\n", func_name);
+		return 0;
+	}
+
+	lua_insert(L, base + 1); /* func */
+	if (lua_pcall(L, nargs, nresults, 0) != LUA_OK) {
+		fprintf(stderr, "Failed to call Lua function %s: %s\n", func_name,
+				lua_tostring(L, -1));
+		lua_settop(L, base - 1);
+		return 0;
```
lneto, line 64:
```suggestion
		goto restore;
```

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// Helper function to call Lua functions
+static int libxtable_invoke(const char *func_name, int nargs, int nresults)
+{
+	int ret = 0;
+	int base = lua_gettop(L) - nargs;
+	if (lua_getfield(L, -1, func_name) != LUA_TFUNCTION) {
+		fprintf(stderr, "Function %s not found\n", func_name);
+		return 0;
```
lneto, line 56: shouldn't you restore the stack here as well?

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,218 @@
...
+		ret = lua_toboolean(L, -1);
+
+	lua_settop(L, base - 1);
+	return ret;
+}
+
+static inline int libxtable_run(lua_State *L, const char *func_name, const char *key, int nargs, int nresults)
+{
+	lua_rawgetp(L, LUA_REGISTRYINDEX, key);
+	return libxtable_invoke(func_name, nargs, nresults);
```
lneto, line 77: I don't think we need this indirection here.. I would just move the body of this function inhere..

README.md
```diff
@@ -1203,6 +1203,18 @@ netfilter hooks to Lua.
 * `"LOCAL_OUT"`: `NF_INET_LOCAL_OUT`. The packet is generated by the local system.
 * `"POST_ROUTING"`: `NF_INET_POST_ROUTING`. The packet is about to be sent out.
 
+### xtable userspace library (libxtable)
+
+The `libxtable` [userspace library](usr/lib/xtable) provides support for generating userspace code for [xtable extensions](https://inai.de/projects/xtables-addons/). The user can modify the generated lua code to implement the userspace handlers for the corresponding xtable extension.
+
+To generate the library, the following steps are required:
```
lneto, line 1210: I would document the API as well.. in the same way you have done for the kernel library

usr/lib/xtable/Makefile
```diff
@@ -0,0 +1,26 @@
...
+	sudo rm -f ${XTABLES_SO_DIR}/libxt_${LUAXTABLE_MODULE}.so
+
+clean:
+	rm -f libxt_*.so libxt_*.o
+
+lib%.so: lib%.o
+	gcc -shared -fPIC -o libxt_${LUAXTABLE_MODULE}.so libxt_${LUAXTABLE_MODULE}.o ${LUA_FLAGS};
+
+lib%.o: libxtable.c
+	gcc ${CFLAGS} ${LUA_FLAGS} -D_INIT=lib$*_init -fPIC -c -o libxt_${LUAXTABLE_MODULE}.o $<;
```
lneto, line 26:
```suggestion
	gcc ${CFLAGS} ${LUA_FLAGS} -D_INIT=lib$*_init -fPIC -c -o libxt_${LUAXTABLE_MODULE}.o $<;

```

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,216 @@
...
+* SPDX-License-Identifier: MIT OR GPL-2.0-only
+*/
+
+#include <lauxlib.h>
+#include <lua.h>
+#include <lualib.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <xtables.h>
```
lneto, line 12:
```suggestion
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

#include <lua.h>
#include <lauxlib.h>
#include <lualib.h>

#include <xtables.h>
```

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,216 @@
...
+
+#include "luaxtable.h"
+
+#ifndef LUAXTABLE_MODULE
+#define LUAXTABLE_MODULE "test"
```
lneto, line 17:
```suggestion
#error LUAXTABLE_MODULE is not defined
```

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,216 @@
...
+const char *libxtable_target_key = "libxtable_target";
+
+int luaopen_libxtable(lua_State *L);
+
+// Function to initialize Lua state
+static int init_lua(void)
+{
+	L = luaL_newstate();
+	if (L == NULL) {
+		fprintf(stderr, "Failed to create Lua state\n");
```
lneto, line 31: does other modules use `fprintf` directly? I would define a macro.. (e.g., `pr_err`)..

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,216 @@
...
+const char *libxtable_match_key = "libxtable_match";
+const char *libxtable_target_key = "libxtable_target";
+
+int luaopen_libxtable(lua_State *L);
+
+// Function to initialize Lua state
+static int init_lua(void)
+{
+	L = luaL_newstate();
+	if (L == NULL) {
```
lneto, line 30:
```suggestion
	L = luaL_newstate();
	if ((L = luaL_newstate()) == NULL)
```

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,216 @@
...
+#define LUAXTABLE_MODULE "test"
+#endif
+
+static lua_State *L = NULL;
+const char *libxtable_match_key = "libxtable_match";
+const char *libxtable_target_key = "libxtable_target";
+
+int luaopen_libxtable(lua_State *L);
+
+// Function to initialize Lua state
```
lneto, line 26: I would get rid of this sort of comment.. you don't need to "teach" Lua through them..

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,216 @@
...
+
+int luaopen_libxtable(lua_State *L)
+{
+	luaL_newlib(L, libxtable_lib);
+	return 1;
+}
+
+static void __attribute__((constructor)) _init(void)
+{
+	if (!init_lua()) {
```
lneto, line 198: I don't think we need this auxiliary function here.. IMHO, the code would be cleaner moving the function body inhere.. the same accounts to `load_lua_script` below.. just use Lua C API directly.. that is, `luaL_dofile` sounds better than `load_lua_script` to the Lua programmers' eyes ;-)

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,216 @@
...
+		fprintf(stderr, "Failed to initialize Lua\n");
+		return;
```
lneto, line 200: shouldn't we `exit()` in case of failure? how does other modules handle that?

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,216 @@
...
+	}
+}
+
+static void __attribute__((destructor)) _fini(void)
+{
+	if (L) {
+		libxtable_putregistry(L, libxtable_match_key);
+		libxtable_putregistry(L, libxtable_target_key);
+		lua_close(L);
+	}
```
lneto, line 215:
```suggestion
	if (L)
		lua_close(L);
```

close will clean the register..

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,216 @@
...
+#define LIBXTABLE_LIB_CB(hook)				\
+static int libxtable_##hook##_lib(lua_State *L)		\
+{							\
+	luaL_checktype(L, 1, LUA_TTABLE);		\
+							\
+	libxtable_##hook##_reg.revision = libxtable_checkint(L, 1, "revision");		\
+	libxtable_##hook##_reg.family = libxtable_checkint(L, 1, "family");		\
+								\
+	xtables_register_##hook(&libxtable_##hook##_reg);	\
+	libxtable_setregistry(L, 1, libxtable_##hook##_key);	\
```
lneto, line 177: IMHO, we don't need this function as well.. it's just two lines used only here and clean enough..

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,216 @@
...
+
+static inline void libxtable_putregistry(lua_State *L, const char *key)
+{
+	lua_pushnil(L);
+	lua_rawsetp(L, LUA_REGISTRYINDEX, key);
+}
+
+static inline int libxtable_checkint(lua_State *L, int idx, const char *key)
+{
+	if(lua_getfield(L, idx, key) != LUA_TNUMBER) {
```
lneto, line 159:
```suggestion
	if (lua_getfield(L, idx, key) != LUA_TNUMBER) {
```

please mind the code style =p

usr/lib/xtable/libxtable.c
sheharyaar: Yes, they use a either `fprintf` or `printf`.

usr/lib/xtable/libxtable.c
sheharyaar: I have changed it to `static int`, a non-zero value will make it a failure, other modules do this.

README.md
sheharyaar: Ok

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,182 @@
...
+	if ((L = luaL_newstate()) == NULL) {
+		fprintf(stderr, "Failed to create Lua state\n");
+		return -1;
+	}
+	luaL_openlibs(L);
+	luaL_requiref(L, "libxtable", luaopen_libxtable, 1);
+
+	if (luaL_dofile(L, "libxt_"LUAXTABLE_MODULE".lua") != LUA_OK) {
+		fprintf(stderr, "Failed to load Lua script: %s\n", lua_tostring(L, -1));
+		return -1;
```
lneto, line 172: I would return some actual errno here..

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	return 0;
+}
+
+static void __attribute__((destructor)) _fini(void)
+{
+	if (L)
```
lneto, line 180: after changing the returning value of `_init` we will never reach this condition, right? we can safely just close the Lua state..

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,182 @@
...
+{
+	luaL_newlib(L, libxtable_lib);
+	return 1;
+}
+
+static int __attribute__((constructor)) _init(void)
+{
+	if ((L = luaL_newstate()) == NULL) {
+		fprintf(stderr, "Failed to create Lua state\n");
+		return -1;
```
lneto, line 165: ENOMEM?

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,182 @@
...
+LIBXTABLE_LIB_CB(match);
+LIBXTABLE_LIB_CB(target);
+
+static const luaL_Reg libxtable_lib[] = {
+	{"match", libxtable_match_lib},
+	{"target", libxtable_target_lib},
+	{NULL, NULL}
+};
+
+int luaopen_libxtable(lua_State *L)
```
lneto, line 155: there's no reason to this not being static, right? we are not gonna link the Lua library dynamically..

usr/lib/xtable/Makefile
```diff
@@ -0,0 +1,27 @@
+# SPDX-FileCopyrightText: (c) 2024 Mohammad Shehar Yaar Tausif <sheharyaar48@gmail.com>
+# SPDX-License-Identifier: MIT OR GPL-2.0-only
+
+LUAXTABLE_MODULE ?= test
```
lneto, line 4: I would let it raise an error if not defined.. perhaps you can just say how it should be used in the error message

usr/lib/xtable/Makefile
```diff
@@ -0,0 +1,27 @@
...
+
+CFLAGS = -DLUAXTABLE_MODULE=\"${LUAXTABLE_MODULE}\" -O2 -Wall -I../../../lib
+XTABLES_SO_DIR = $(shell pkg-config xtables --variable xtlibdir)
+LUA_FLAGS = -llua -lm
+
+all:
+	make libxt_${LUAXTABLE_MODULE}.so
+
+install:
+	sudo cp libxt_${LUAXTABLE_MODULE}.so  ${XTABLES_SO_DIR}
```
lneto, line 14: I wouldn't require LUAXTABLE_MODULE to be defined on `install`, I would just copy `libxt_*.so`

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	return 0;
+}
+
+static void __attribute__((destructor)) _fini(void)
+{
+	if (L)
+		lua_close(L);
+}
```
lneto, line 182: Please mind the `\n` at the end of every file

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,182 @@
...
+int luaopen_libxtable(lua_State *L);
+
+static inline int libxtable_run(lua_State *L, const char *func_name, const char *key, int nargs, int nresults)
+{
+	int ret = 0;
+	int base = lua_gettop(L) - nargs;
+	lua_rawgetp(L, LUA_REGISTRYINDEX, key);
+
+	if (lua_getfield(L, -1, func_name) != LUA_TFUNCTION) {
+		fprintf(stderr, "Function %s not found\n", func_name);
```
lneto, line 35: I would define a macro for `fprint(stderr, ...)` anyway.. to avoid repeating it all around the code..

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,182 @@
...
+	lua_rawsetp(L, LUA_REGISTRYINDEX, libxtable_##hook##_key);	\
+	return 0;							\
+}
+
+LIBXTABLE_LIB_CB(match);
+LIBXTABLE_LIB_CB(target);
+
+static const luaL_Reg libxtable_lib[] = {
+	{"match", libxtable_match_lib},
+	{"target", libxtable_target_lib},
```
lneto, line 151: It's confusing call them `*_lib` here.. they aren't libs, right? They are API functions, doesn't them? I would call them simply `libxtable_{match, target}`..

usr/lib/xtable/libxtable.c
lneto, line 172: I would close the Lua state here..

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,202 @@
...
+#endif
+
+typedef struct libxtable_flags_s {
+	const char *name;
+	lua_Integer value;
+} libxtable_flags_t;
+
+static lua_State *L = NULL;
+const char *libxtable_match_key = "libxtable_match";
+const char *libxtable_target_key = "libxtable_target";
```
lneto, line 30: We don't need this indirection here.. you can just use the actual strings or even better.. use the function pointer.. e.g., `libxtable_match`

usr/lib/xtable/libxtable.c
```diff
@@ -0,0 +1,202 @@
+/*
```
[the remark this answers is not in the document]
sheharyaar: this is user library, so i thought it would be better to keep lib, what do you think ?

usr/lib/xtable/libxtable.c
sheharyaar: It will be handled in `_fini`, I tested it.

usr/lib/xtable/libxtable.c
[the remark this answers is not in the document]
lneto: I think it's weird to prefix it with `lib` and also we should indicate somehow it's a Lua binding.. I understand you wanna differentiate it from the kernel binding, but I don't think using `lib` is a good way.. the kernel binding is also a "library"..  perhaps we can name the file `luaxt.c` and prefix `luaxt_`, what do you think?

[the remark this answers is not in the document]
sheharyaar: `luaxt` seems better

usr/lib/xtable/luaxt.c
```diff
@@ -0,0 +1,203 @@
...
+#define LIBXTABLE_LIB_CB(hook)				\
+static int luaxt_##hook(lua_State *L)		\
+{							\
+	luaL_checktype(L, 1, LUA_TTABLE);		\
+							\
+	luaxt_##hook##_reg.revision = luaxt_checkint(L, 1, "revision");		\
+	luaxt_##hook##_reg.family = luaxt_checkint(L, 1, "family");		\
+	xtables_register_##hook(&luaxt_##hook##_reg);		\
+	lua_pushvalue(L, 1);						\
+	lua_rawsetp(L, LUA_REGISTRYINDEX, "luaxt_"#hook);		\
```
lneto, line 147: you don't need to use a string here.. you can just use `luaxt_##hook` and use a `void *` as key.. ;-)

usr/lib/xtable/luaxt.c
```diff
@@ -0,0 +1,203 @@
...
+	if (luaL_dofile(L, "libxt_"LUAXTABLE_MODULE".lua") != LUA_OK) {
+		pr_err("Failed to load Lua script: %s\n", lua_tostring(L, -1));
+		return -ENOENT;
+	}
+	return 0;
+}
+
+static void __attribute__((destructor)) _fini(void)
+{
+	if (L) lua_close(L);
```
lneto, line 202:
```suggestion
	if (L != NULL)
		lua_close(L);
```

usr/lib/xtable/luaxt.c
```diff
@@ -0,0 +1,203 @@
...
+	LIBXTABLE_HELPER_CB(hook)   	\
+	LIBXTABLE_INITER_CB(hook)   	\
+	LIBXTABLE_PARSER_CB(hook)   	\
+	LIBXTABLE__FINALCHECKER_CB(hook)	\
+	LIBXTABLE_PRINTER_CB(hook)  	\
+	LIBXTABLE_SAVER_CB(hook)
+
+LIBXTABLE_NEWCB(match);
+LIBXTABLE_NEWCB(target);
+
```
lneto, line 108:
```suggestion
```

usr/lib/xtable/luaxt.c
```diff
@@ -0,0 +1,203 @@
...
+	.save = luaxt_##hook##_save			 	\
+}
+
+LIBXTABLE_NEWREG(match);
+LIBXTABLE_NEWREG(target);
+
+static inline int luaxt_checkint(lua_State *L, int idx, const char *key)
+{
+	if (lua_getfield(L, idx, key) != LUA_TNUMBER) {
+		pr_err("invalid \'%s\' in match\n", key);
```
lneto, line 130: does this only apply to `match`?

usr/lib/xtable/luaxt.c
```diff
@@ -0,0 +1,203 @@
...
+
+	if (nresults == 1)
+		ret = lua_toboolean(L, -1);
+
+restore:
+	lua_settop(L, base);
+	return ret;
+}
+
+#define LIBXTABLE_HELPER_CB(hook)			   	\
```
lneto, line 61: you should prefix everything with `LUAXT`

usr/lib/xtable/luaxt.c
```diff
@@ -0,0 +1,203 @@
...
+		pr_err("Failed to load Lua script: %s\n", lua_tostring(L, -1));
+		return -ENOENT;
+	}
+	return 0;
+}
+
+static void __attribute__((destructor)) _fini(void)
+{
+	if (L) lua_close(L);
+}
```
lneto, line 203:
```suggestion
}

```

please mind this on every file

examples/dnsdoctor/dnsdoctor.lua
```diff
@@ -0,0 +1,80 @@
+--
+-- SPDX-FileCopyrightText: (c) 2024 Mohammad Shehar Yaar Tausif <sheharyaar48@gmail.com>
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+-- Filter DNS packets based on a blocklist
```
lneto, line 6: ?

examples/dnsdoctor/dnsdoctor.lua
```diff
@@ -0,0 +1,80 @@
...
+	local dst_ip, target_ip = string.unpack("I4I4", userdata:sub(sep + 1))
+	local thoff = par.thoff
+	local proto = skb:getuint8(9)
+
+	local packetdst = skb:getuint32(16)
+	if packetdst ~= dst_ip then
+		return action.ACCEPT
+	end
+
+	if proto == udp then
```
lneto, line 43:
```suggestion
	if proto ~= udp then
		goto accept
```

wondering if we really need this condition.. couldn't it be simply handle by the iptables rule?

examples/dnsdoctor/setup.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+set -eux
+
+# add namespaces ns1 for dns server ns2 for server
```
lneto, line 5: why do we need namespaces for this example? btw, I would change `resolv.conf` as well.. I would just `dig` to a well known domain (e.g., `ebpf.io`) and "redirect" it later to a localhost..

sheharyaar: I wanted to create the DNS doctoring scenario using namespaces.

examples/dnsdoctor/dnsdoctor.lua
sheharyaar: Yeah, the iptables rule part `-p udp --sport 53` should handle it.

examples/dnsdoctor/cleanup.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
lneto, line 1: Add your copyright

examples/dnsdoctor/setup.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
lneto, line 1: Add copyright

README.md
```diff
@@ -1148,9 +1148,10 @@ This function receives the following arguments:
   * `match` : function to be called for matching packets. It receives the following arguments:
 	* `skb` (readonly): a `data` object representing the socket buffer.
 	* `par`: a table containing `hotdrop`, `thoff` (transport header offset) and `fragoff` (fragment offset) fields.
+    * `userdata` : a lua string passed from the userspace xtable module.
     * The function must return `true` if the packet matches the extension; otherwise, it must return `false`.
-  * `checkentry`: function to be called for checking the entry.
-  * `destroy`: function to be called for destroying the xtable extension.
+  * `checkentry`: function to be called for checking the entry. This function receives `userdata` as it's argument.
+  * `destroy`: function to be called for destroying the xtable extension. This function receives `userdata` as it's argument.
```
lneto, line 1154: s/it's/its/

README.md
```diff
@@ -1165,9 +1166,10 @@ This function receives the following arguments:
   * `target` : function to be called for targeting packets. It receives the following arguments:
     * `skb`: a `data` object representing the socket buffer.
     * `par` (readonly): a table containing `hotdrop`, `thoff` (transport header offset) and `fragoff` (fragment offset) fields.
+    * `userdata` : a lua string passed from the userspace xtable module.
     * The function must return one of the values defined by the [xtable.action](https://github.com/luainkernel/lunatik#xtableaction) table.
-  * `checkentry`: function to be called for checking the entry.
-  * `destroy`: function to be called for destroying the xtable extension.
+  * `checkentry`: function to be called for checking the entry. This function receives `userdata` as it's argument.
+  * `destroy`: function to be called for destroying the xtable extension. This function receives `userdata` as it's argument.
```
lneto, line 1172: s/it's/its/

README.md
```diff
@@ -1355,6 +1410,46 @@ sudo lunatik run examples/dnsblock/dnsblock false	# runs the Lua kernel script
...
+
+# test the setup, a response with IP 192.168.10.1 should be returned
+dig lunatik.com
+
+# run the Lua kernel script
+sudo lunatik run examples/dnsdoctor/dnsdoctor false
+
+# copy the userspace extension to luaxt directory
+cp libxt_dnsdoctor.lua ../../usr/lib/xtable/
+cd ../../usr/lib/xtable
```
lneto, line 1434: instead of doing this, I would have the make file of this example including the make file from `../../usr/lib/xtable/` and setting the env var..

examples/dnsdoctor/dnsdoctor.lua
```diff
@@ -0,0 +1,77 @@
...
+local action = xt.action
+local family = xt.family
+
+local udp = 0x11
+local dns = 0x35
+
+local function nop() end
+
+local function get_domain(skb, off)
+	local name = ""
```
lneto, line 20: why do you need this? you are only using it to prepend `""` in the return value.. it doesn't make much sense.. perhaps I'm missing something..

examples/dnsdoctor/dnsdoctor.lua
```diff
@@ -0,0 +1,77 @@
...
+	local i = 0
+
+	repeat
+		local d = skb:getbyte(off + i)
+		i = i + 1
+	until d == 0
+
+	return name .. skb:getstring(off, i-1), i
```
lneto, line 28: mind the code style.. `i - 1`

examples/dnsdoctor/dnsdoctor.lua
```diff
@@ -0,0 +1,77 @@
...
+local function nop() end
+
+local function get_domain(skb, off)
+	local name = ""
+	local i = 0
+
+	repeat
+		local d = skb:getbyte(off + i)
+		i = i + 1
+	until d == 0
```
lneto, line 26: wouldn't it be better to `tostring(skb):find('\0')`?

examples/dnsdoctor/dnsdoctor.lua
```diff
@@ -0,0 +1,77 @@
...
+
+	return name .. skb:getstring(off, i-1), i
+end
+
+local function dnsdoctor_tg(skb, par, userdata)
```
lneto, line 31: I think userdata here is misleading.. because it corresponds to a Lua type name.. I would `arg` or something else..

examples/dnsdoctor/dnsdoctor.lua
```diff
@@ -0,0 +1,77 @@
...
+	local sep = userdata:find(" ")
```
lneto, line 32: you can encode strings prepended with its size.. take a look at https://www.lua.org/manual/5.4/manual.html#6.4.2:

```
s[n]: a string preceded by its length coded as an unsigned integer with n bytes (default is a size_t)
```

then, you just  need to unpack firstly

```
T: a size_t (native size)
```

examples/dnsdoctor/dnsdoctor.lua
```diff
@@ -0,0 +1,77 @@
...
+	local target_dns = userdata:sub(1, sep - 1)
+	local dst_ip, target_ip = string.unpack("I4I4", userdata:sub(sep + 1))
+	local thoff = par.thoff
+
+	local packetdst = skb:getuint32(16)
```
lneto, line 37: you need to handle endianness here, right?

examples/dnsdoctor/setup.sh
```diff
@@ -0,0 +1,64 @@
...
+
+[[zones]]
+host = 'lunatik.com'
+type = 'NS'
+answer = 'ns2.lunatik.com.'
+""" > zones.toml
+sudo ip netns exec ns1 .venv/bin/dnserver --no-upstream zones.toml &
+
+# run webserver in ns2
+sudo ip netns exec ns2 python3 -m http.server --bind 10.1.2.3 8000 &
```
lneto, line 63: python? really? heheh.. I'll let it go ;-).. but if it's easy to run another DNS server, especially a Lua one, I would make the change..

lib/luaxtable.c
```diff
@@ -56,8 +56,9 @@ static int luaxtable_docall(lua_State *L, luaxtable_t *xtable, const char *op, i
 
 	lua_insert(L, base + 1); /* op */
 	lua_pop(L, 1); /* table */
+	lua_pushstring(L, info->userdata); /* userdata */
```
lneto, line 59: I would rename it.. `userdata` clashes with Lua's userdada and make it a bit confusing

lib/luaxtable.h
```diff
@@ -6,9 +6,14 @@
 #ifndef luaxtable_h
 #define luaxtable_h
 
+#define LUAXTABLE_USERDATA_SIZE 256
+
 struct luaxtable_s;
 
 typedef struct luaxtable_info_s {
+	char userdata[LUAXTABLE_USERDATA_SIZE];
```
lneto, line 14: I would rename this.. `info`, `userinfo`, `arg`.. not sure what's the best option..

usr/lib/xtable/luaxt.c
```diff
@@ -0,0 +1,238 @@
...
+#include <xtables.h>
+
+#include "luaxtable.h"
+
+#ifndef LUAXTABLE_MODULE
+#error "LUAXTABLE_MODULE is not defined"
+#endif
+
+#define pr_err(fmt, ...) fprintf(stderr, fmt, ##__VA_ARGS__)
+#define MIN(a,b) (((a)<(b))?(a):(b))
```
lneto, line 24: mind code style, please

usr/lib/xtable/luaxt.c
```diff
@@ -0,0 +1,238 @@
...
+}
+
+static int luaxt_doparams(lua_State *L, const char *op, const char *key, unsigned int *flags, luaxtable_info_t *info)
+{
+	int ret;
+	lua_newtable(L);
+	lua_pushvalue(L, -1); /* stack : param param */
+
+	ret = luaxt_run(L, op, key, 1, 1);
+	if (ret == -1)
```
lneto, line 70:
```suggestion
	if ((ret =  luaxt_run(L, op, key, 1, 1)) == -1)
```

usr/lib/xtable/luaxt.c
```diff
@@ -0,0 +1,238 @@
...
+		ret = lua_toboolean(L, -1);
+restore:
+	lua_settop(L, base);
+	return ret;
+}
+
+static int luaxt_doparams(lua_State *L, const char *op, const char *key, unsigned int *flags, luaxtable_info_t *info)
+{
+	int ret;
+	lua_newtable(L);
```
lneto, line 66:
```suggestion

	lua_newtable(L);
```

usr/lib/xtable/luaxt.c
```diff
@@ -0,0 +1,238 @@
...
+	lua_pushvalue(L, -1); /* stack : param param */
```
lneto, line 67: "param param"? is it a typo?

examples/dnsdoctor/dnsdoctor.lua
sheharyaar: I am sending the IP address from the userspace in the network order using `string.pack(">I4>I4", ... )`

lib/luaxtable.c
sheharyaar: Can I rename it to `userargs` ?

lneto: yup.. just fine for me..

examples/dnsdoctor/setup.sh
sheharyaar: The HTTP server is not needed, however I could not find an easy to setup DNS server apart from python.

usr/lib/xtable/luaxt.c
sheharyaar: Now, one param is the one passed to the luadata, and the other one is later used to get the options (like we did in luaxtable for `par`).

usr/lib/xtable/luaxt.c
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	if (nresults == 1)
+		ret = lua_toboolean(L, -1);
+restore:
+	lua_settop(L, base);
+	return ret;
+}
+
+static int luaxt_doparams(lua_State *L, const char *op, const char *key, unsigned int *flags, luaxtable_info_t *info)
```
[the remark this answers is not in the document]
sheharyaar: moving with `pushuserargs`

examples/dnsdoctor/dnsdoctor.lua
sheharyaar: I used this because, `lunatik.com` is actually encoded as `\x07lunatik\x03com` in the lua program, this directly matches that of the DNS answer format that is `<len of string>string`, so I used a seperator to seperate this and the IP addresses.

sheharyaar: Unpacking this into `lunatik` and `com` will need additional step to search for each string in the DNS answer, whereas `x07lunatik\x03com` directly matches the string in DNS packet.

examples/dnsdoctor/setup.sh
sheharyaar: I have removed the HTTP server, just using the DNS server.

README.md
sheharyaar: done

examples/dnsdoctor/dnsdoctor.lua
lneto: you mean because `off`? wouldn't be it just the case of `skb:getstring(off, len):find('\0')` or even `tostring(skb):sub(off + 1):find('\0')`? I miss the point of verifying it byte per byte instead of using a higher-level string API.. however, it's up to you.. we can merge it as is..

examples/dnsdoctor/dnsdoctor.lua
lneto: sorry.. I think I'm missing something..

```
> domain = "\x07lunatik\x03com"
> userarg = string.pack("sI8", "\x07lunatik\x03com", 0xFF)
> string.unpack("sI8", userarg)
lunatikcom	255	29
```

wouldn't this be enough?

examples/dnsdoctor/dnsdoctor.lua
sheharyaar: Yeah, this works : `tostring(skb):sub(off + 1):find('\0')`

examples/dnsdoctor/dnsdoctor.lua
sheharyaar: Yes, this can be done, through your approach, I identified a bug in my code too. Earlier, if the binary string returned by pack had `\x00` in the beginning part, the userdata was not passed correctly to the kernel lua program, since I was using `lua_pushstring`, now I changed it to `lua_pushlstring(info->userdata, LUAXTABLE_USERDATA_SIZE)`

examples/dnsdoctor/dnsdoctor.lua
sheharyaar: Done in new push

examples/dnsdoctor/dnsdoctor.lua
sheharyaar: Reference : https://github.com/luainkernel/lunatik/pull/153/files#diff-292f6a11c6f52f91544b0b47343e04b5d6516316ca5a075844659348899709ffR59

examples/dnsdoctor/Makefile
```diff
@@ -0,0 +1,14 @@
...
+	LUAXTABLE_MODULE=dnsdoctor $(MAKE) -C ../../usr/lib/xtable
+
+install:
+	sudo LUAXTABLE_MODULE=dnsdoctor $(MAKE) -C ../../usr/lib/xtable install
+
+uninstall:
+	sudo rm -f ${XTABLES_SO_DIR}/libxt_${LUAXTABLE_MODULE}.so
+
+clean:
+	LUAXTABLE_MODULE=dnsdoctor $(MAKE) -C ../../usr/lib/xtable clean
```
lneto, line 14:
```suggestion
	LUAXTABLE_MODULE=dnsdoctor $(MAKE) -C ../../usr/lib/xtable clean

```

examples/dnsdoctor/dnsdoctor.lua
```diff
@@ -0,0 +1,68 @@
...
+local family = xt.family
+
+local udp = 0x11
+local dns = 0x35
+
+local function nop() end
+
+local function get_domain(skb, off)
+	local i = tostring(skb):sub(off+1):find('\0')
+	return skb:getstring(off, i - 1), i
```
lneto, line 21:
```suggestion
	local _, nameoff, name = skb:getstring(off):find("([^\0]*)")
	return name, nameoff
```

I think this version is better, because we convert `skb + off` to a string , instead of starting it from the very beginning.. also, we don't create another string afterwards.. we just let `find()` to capture our pattern.. moreover, it looks a bit clearer what we are really doing here.. what do you think?

examples/dnsdoctor/dnsdoctor.lua
```diff
@@ -0,0 +1,68 @@
...
+end
+
+local function dnsdoctor_tg(skb, par, userargs)
+	local target_dns, dst_ip, target_ip = string.unpack(">s4 I4 I4", userargs)
```
lneto, line 25:
```suggestion
	local target_dns, dst_ip, target_ip = string.unpack(">s4I4I4", userargs)
```

examples/dnsdoctor/dnsdoctor.lua
```diff
@@ -0,0 +1,68 @@
...
+		local nanswers = linux.ntoh16(skb:getuint16(dnsoff + 6))
+
+		-- check the domain name
+		dnsoff = dnsoff + 12
+		local domainname, nameoff = get_domain(skb, dnsoff)
+
+		if domainname == target_dns then
+			dnsoff = dnsoff + nameoff + 4 -- skip over type, label fields
+			-- iterate over answers
+			for i = 1,nanswers do
```
lneto, line 45:
```suggestion
			for i = 1, nanswers do
```

examples/dnsdoctor/libxt_dnsdoctor.lua
```diff
@@ -0,0 +1,28 @@
...
+local function dnsdoctor_init(par)
+	local target_ip = "10.1.2.3"
+	local target = 0
+	target_ip:gsub("%d+", function(s) target = target * 256 + tonumber(s) end)
+
+	local src_ip = "10.1.1.2"
+	local src = 0
+	src_ip:gsub("%d+", function(s) src = src * 256 + tonumber(s) end)
+
+	par.userargs = string.pack(">s4 I4 I4", "\x07lunatik\x03com", src, target)
```
lneto, line 15:
```suggestion
	par.userargs = string.pack(">s4I4I4", "\x07lunatik\x03com", src, target)
```

examples/dnsdoctor/libxt_dnsdoctor.lua
```diff
@@ -0,0 +1,28 @@
...
+local function noop() end
+
+local function dnsdoctor_init(par)
+	local target_ip = "10.1.2.3"
+	local target = 0
+	target_ip:gsub("%d+", function(s) target = target * 256 + tonumber(s) end)
+
+	local src_ip = "10.1.1.2"
+	local src = 0
+	src_ip:gsub("%d+", function(s) src = src * 256 + tonumber(s) end)
```
lneto, line 13: you can require `socket.inet` and just use `aton()`

examples/dnsdoctor/libxt_dnsdoctor.lua
```diff
@@ -0,0 +1,28 @@
+local luaxt = require("luaxt")
+local family = luaxt.family
+
+local function noop() end
```
lneto, line 4:
```suggestion
local function nop() end
```

usr/lib/xtable/luaxt.c
```diff
@@ -0,0 +1,239 @@
...
+}
+
+static int luaxt_runwithuserargs(lua_State *L, const char *op, const char *key, unsigned int *flags, luaxtable_info_t *info)
+{
+	int ret;
+
+	lua_newtable(L);
+	lua_pushvalue(L, -1); /* stack : param, param */
+
+	if((ret = luaxt_run(L, op, key, 1, 1)) == -1)
```
lneto, line 70:
```suggestion
	if ((ret = luaxt_run(L, op, key, 1, 1)) == -1)
```

usr/lib/xtable/luaxt.c
```diff
@@ -0,0 +1,239 @@
...
+	lua_settop(L, base);
+	return ret;
+}
+
+static int luaxt_runwithuserargs(lua_State *L, const char *op, const char *key, unsigned int *flags, luaxtable_info_t *info)
+{
+	int ret;
+
+	lua_newtable(L);
+	lua_pushvalue(L, -1); /* stack : param, param */
```
lneto, line 68:
```suggestion
	lua_pushvalue(L, -1); /* stack: param, param */
```

usr/lib/xtable/luaxt.c
```diff
@@ -0,0 +1,239 @@
...
+
+	if (flags && (lua_getfield(L, -1, "flags") == LUA_TNUMBER)) {
+		*flags = lua_tointeger(L, -1);
+		lua_pop(L, 1);
+	}
+
+	if (lua_getfield(L, -1, "userargs") == LUA_TSTRING) {
+		size_t len = 0;
+		const char *ldata = lua_tolstring(L, -1, &len);
+		memset(info->userargs, 0, LUAXTABLE_USERDATA_SIZE);
```
lneto, line 81: you can just `memset` the rest (that's it, `LUAXTABLE_USERDATA_SIZE - MIN(len, LUAXTABLE_USERDATA_SIZE)`), right? after copying ldata..

sheharyaar: It will still result in a memset call, right ?

examples/dnsdoctor/libxt_dnsdoctor.lua
sheharyaar: That library is available in the kernel space, right ?

lneto: yeah.. you're right.. I'd leave it as is.. but could you create an issue to move the utilities from `socket.inet` such as `aton()` to a separate script to be shared between user space and kernel? I think it would be a good first issue for new contributors =).

usr/lib/xtable/luaxt.c
lneto: yeah, but it would memset only the reminder of the "buffer".. perhaps it's a bit of premature optimization from my side =).. I just found it a bit weird to memset to 0 and immediately after memcpy to the very same portion of the memory.. but we can leave it as is.. it's up to you =)

lneto (approved):

examples/dnsdoctor/libxt_dnsdoctor.lua
sheharyaar: Done, created #169

usr/lib/xtable/luaxt.c
sheharyaar: We can optimise it later if needed.

PR #155, opened 2024-08-12
Fix skb len calculation

jperon:
Fix a bug that prevented TCP payload to be passed to lua scripts.

### DIFF
--- a/lib/luaxtable.c
+++ b/lib/luaxtable.c
@@ -77,11 +77,11 @@ static inline lunatik_object_t *luaxtable_getskb(lua_State *L, luaxtable_t *xtab
 static int luaxtable_pushparams(lua_State *L, const struct xt_action_param *par, luaxtable_t *xtable, const struct sk_buff *skb, uint8_t opt)
 {
 	lunatik_object_t *data = luaxtable_getskb(L, xtable);
-	if (data == NULL) {
-		pr_err("could not find skb\n");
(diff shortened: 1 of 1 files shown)


sheharyaar (approved):
LGTM.

lib/luaxtable.c
```diff
@@ -81,7 +81,9 @@ static int luaxtable_pushparams(lua_State *L, const struct xt_action_param *par,
 		pr_err("could not find skb\n");
 		return -1;
 	}
-	luadata_reset(data, skb->data, skb_headlen(skb), opt);
+	if (skb_linearize(skb) != 0)
+          return -1;
+        luadata_reset(data, skb->data, skb->len, opt);
```
lneto, line 86:
```suggestion
	if (unlikely(data == NULL || skb_linearize(skb) != 0)) {
		pr_err("could not get skb\n");
		return -1;
	}
        luadata_reset(data, skb->data, skb->len, opt);
```

I would simplify the patch in this way.. please notice that we would need to remove the [lines above](https://github.com/luainkernel/lunatik/pull/155/files#diff-292f6a11c6f52f91544b0b47343e04b5d6516316ca5a075844659348899709ffR80-R83) as well.. what do you think?

jperon:
I think it should be good now.

lneto:
@jperon it looks good to me.. can you just squash your changes into a single commit? So we have a more comprehensive history.. thanks!

jperon:
Here you are.

PR #156, opened 2024-08-14
SNI filter in MoonScript

jperon:


lneto:
Can you remove the transpiled Lua files from your patch? Also, can you add the steps to install moonscript in the readme? Thanks!

jperon:
Just done and force-pushed.

lneto:
now we have https://github.com/luainkernel/snihook, I think we can close this.. however, @jperon, it would be much appreciated if you just add an entry on the examples section linking snihook repo.. thanks!

PR #160, opened 2024-08-17
fail when there is no space available in the kfifo

sav:


sav:
sorry, it impacted significantly the benchmarks. I'll hold this change.

lneto (approved):

PR #164, opened 2024-08-19
fixes and improvements on sleep modes

lneto:


sav (approved):

lneto:
@jperon can you help on reviewing this.. also applying `runtime:resume(...)` to your module to replace `thread.run(runtime, script, ...)` to dispatch objects to your xtable hook?

jperon:
@lneto After first tries, it seems runtime:resume() isn’t equivalent to thread.run(), as I get this error message in dmesg:
```
[  331.269137] luathread: [00000000d4a6fbf2] bad argument #1 to '?' (invalid object)
```

With thread.run, I got this error message when it gets to running `hook.lua`:
```
[  113.172540] shelterbox kernel: luathread: [0000000024b11b39] /lib/modules/lua/snihook/main.lua:40: bad argument #1 to 'run' (cannot use non-sleepable
```
After changing the sleep parameter of snihook/hook [here](https://github.com/jperon/lunatik_snihook/blob/master/main.moon#L22) to `true`, it works. I must confess that for me, the effect of this parameter is a mystery, and that mutexes and spinlocks documentation looks like ancient chinese…

lneto:
> @lneto After first tries, it seems runtime:resume() isn’t equivalent to thread.run(), as I get this error message in dmesg:
> 
> ```
> [  331.269137] luathread: [00000000d4a6fbf2] bad argument #1 to '?' (invalid object)
> ```
> 
> With thread.run, I got this error message when it gets to running `hook.lua`:
> 
> ```
> [  113.172540] shelterbox kernel: luathread: [0000000024b11b39] /lib/modules/lua/snihook/main.lua:40: bad argument #1 to 'run' (cannot use non-sleepable
> ```

so, that's the thing.. you should replace this line

https://github.com/jperon/lunatik_snihook/blob/75665e10676c67cac1369852c4e035dc0aad76bf/main.moon#L41

with

```
rt\resume dev_hook.queue, log.queue
```

and get rid completely from `thread.run`

> After changing the sleep parameter of snihook/hook [here](https://github.com/jperon/lunatik_snihook/blob/master/main.moon#L22) to `true`, it works. I must confess that for me, the effect of this parameter is a mystery, and that mutexes and spinlocks documentation looks like ancient chinese…

hehe.. so, `sleep = true` uses mutexes, what basically can be only used by threads/processes as they might be scheduled.. `sleep = false` sues spinlocks, what can be used from softirqs (such as netfilter hooks), which doesn't have process context and cannot be scheduled.. so, it "spins" in a busy wait loop trying to acquire the lock.. we can discuss this in depth in Matrix if you like.. feel free to suggest improvements (or directly update the documentation =).. the plan is to make this clear as possible.. thank you!

jperon:
> so, that's the thing.. you should replace this line
> 
> [jperon/lunatik_snihook@`75665e1`/main.moon#L41](https://github.com/jperon/lunatik_snihook/blob/75665e10676c67cac1369852c4e035dc0aad76bf/main.moon#L41)
> 
> with
> 
> ```
> rt\resume dev_hook.queue, log.queue
> ```

This is what I did, and I think I found from where did the error come: MoonScript has implicit return, and runtime:resume didn’t like it. When I add "return" at the end of the main functions of `dev.moon` and `hook.moon`, it works.

lneto:
> > so, that's the thing.. you should replace this line
> > [jperon/lunatik_snihook@`75665e1`/main.moon#L41](https://github.com/jperon/lunatik_snihook/blob/75665e10676c67cac1369852c4e035dc0aad76bf/main.moon#L41)
> > with
> > ```
> > rt\resume dev_hook.queue, log.queue
> > ```
> 
> This is what I did, and I think I found from where did the error come: MoonScript has implicit return, and runtime:resume didn’t like it. When I add "return" at the end of the main functions of `dev.moon` and `hook.moon`, it works.

ah.. awesome! yes, the function executed by `resume` can return only lunatik objects.. in that you way you can pass back objects from the callee to the caller runtime.. I still need to document it.. thanks for testing it!

lneto:
@sav I've pushed -f, can you take a final look? for me, it's good enough to merge.. thanks!

sav (approved):

PR #170, opened 2024-08-30
add luacompletion

sav:


lib/luacompletion.c
```diff
@@ -0,0 +1,103 @@
...
+#include <linux/completion.h>
+
+#include <lua.h>
+#include <lualib.h>
+#include <lauxlib.h>
+
+#include <lunatik.h>
+
+typedef struct luacompletion_s {
+	struct completion event;
```
lneto, line 18: we don't need to nest this struct, right? We can just use it directly..

lib/luacompletion.c
```diff
@@ -0,0 +1,103 @@
...
+	ret = wait_for_completion_timeout(&completion->event, msecs_to_jiffies(timeout));
+	lua_pushboolean(L, ret > 0 ? true : false);
+	return 1;
+}
+
+
+static void luacompletion_release(void *private)
+{
+	luacompletion_t *completion = (luacompletion_t *)private;
+	(void)completion; /* do nothing */
```
lneto, line 54: If we have no destructor, you can just avoid this function completely.. e.g., `.release = NULL`

lib/luacompletion.c
```diff
@@ -0,0 +1,103 @@
...
+static const luaL_Reg luacompletion_lib[] = {
+	{"new", luacompletion_new},
+	{NULL, NULL}
+};
+
+static const luaL_Reg luacompletion_mt[] = {
+	{"__gc", lunatik_deleteobject},
+	{"complete", luacompletion_complete},
+	{"wait", luacompletion_wait},
+	{"wait_timeout", luacompletion_wait_timeout},
```
lneto, line 68: Instead of having two separate APIs, I would just let `timeout` arg to be optional on `wait()`

lunatik.h
```diff
@@ -239,10 +240,13 @@ static inline T checker(lua_State *L, int ix)			\
...
 
-#define LUNATIK_PRIVATECHECKER(checker, T)			\
+#define LUNATIK_PRIVATECHECKER(checker, T, monitor)		\
 static inline T checker(lua_State *L, int ix)			\
 {								\
-	T private = (T)lunatik_toobject(L, ix)->private;	\
+	lunatik_object_t *object = likely(monitor) ?		\
+		lunatik_toobject(L, ix) :			\
+		lunatik_checkobject(L, ix);			\
+	T private = (T)object->private;				\
```
lneto, line 249: I'm needing some similar right now and I realized it's not thread safe when you are not under a monitor.. because the object can be closed (that's, `->private = NULL` between `checkobject()` and `private =`.. I think we would need monitor on luacompletion anyways..

lneto: see https://github.com/luainkernel/lunatik/blob/master/lunatik_obj.c#L89

lneto: however, if not using `close`, I think there's no other flow that can set `private` to NULL.. it must be safe for this particular use case, I think.. I'm just not sure that we should provide this API.. it's kinda error prone IMO.. what do you think?

lneto: to give you more context.. that's what I'm doing right now..

```
        lunatik_object_t *runtime = lunatik_checkobject(L, 1);

        lunatik_lock(runtime); /* now it's safe to check private */
        lua_State *Lr = (lua_State *)runtime->private;
        if (Lr == NULL)
                goto err;
        ...
   ```

File: lib/mailbox.lua [hunk 5]
```diff
@@ -39,7 +47,8 @@ function MailBox:receive()
 end
 
 function MailBox:send(message)
-	return self.queue:push(string.pack("s", message))
+	self.queue:push(string.pack("s", message))
```
lneto, line 50: shouldn't we return `push` result no more?

lib/luacompletion.c
```diff
@@ -0,0 +1,88 @@
...
+	{"__gc", lunatik_deleteobject},
+	{"complete", luacompletion_complete},
+	{"wait", luacompletion_wait},
+	{NULL, NULL}
+};
+
+static const lunatik_class_t luacompletion_class = {
+	.name = "completion",
+	.methods = luacompletion_mt,
+	.sleep = false,
```
lneto, line 61: btw, is it safe to call `wait()` from a softirq? It shouldn't, right? If it isn't, I would add a guard to raise an error.. I've done similar guards in my PR (`lneto_sleep`)

README.md
```diff
@@ -1263,6 +1263,29 @@ address families to Lua.
 * `"NETDEV"`: Device ingress and egress path
 * `"BRIDGE"`: Ethernet Bridge.
 
+### `completion`
+
+The `completion` library provides functions to handle task completion in Lua.
```
lneto, line 1268: Can you add a link to Linux kernel documentation if any or to its implementation otherwise.. we are trying to have these links on every binding.. you can refer to luathread, for instance..

lib/luacompletion.c
```diff
@@ -0,0 +1,88 @@
...
+{
+	struct completion *completion = luacompletion_check(L, 1);
+	lua_Integer timeout = luaL_optinteger(L, 2, MAX_SCHEDULE_TIMEOUT);
+	unsigned long timeout_jiffies = msecs_to_jiffies((unsigned long)timeout);
+	unsigned long ret;
+
+	ret = wait_for_completion_interruptible_timeout(completion, timeout_jiffies);
+	if (ret == 0 || ret == -ERESTARTSYS) {
+		lua_pushnil(L);
+		lua_pushstring(L, ret == 0 ? "timeout" : "interrupted");
```
lneto, line 37: I think "interrupt" is more concise.. but I have no strong feelings about this...feel free to choose =)

lib/luacompletion.c
```diff
@@ -0,0 +1,88 @@
...
+
+static int luacompletion_wait(lua_State *L)
+{
+	struct completion *completion = luacompletion_check(L, 1);
+	lua_Integer timeout = luaL_optinteger(L, 2, MAX_SCHEDULE_TIMEOUT);
+	unsigned long timeout_jiffies = msecs_to_jiffies((unsigned long)timeout);
+	unsigned long ret;
+
+	ret = wait_for_completion_interruptible_timeout(completion, timeout_jiffies);
+	if (ret == 0 || ret == -ERESTARTSYS) {
```
lneto, line 35: I would handle other negative cases as "unknown".. just for "completion" (no pun intended 😂).. in this case, I would first check for `> 0` and pushbool/return 1, then I would switch-case with a default: "unknown".. just a suggestion though ;-)

lib/mailbox.lua
sav: it was wrong to begin with; `push` never returned a value.

lib/luacompletion.c
sav: it could be. I hadn't considered it, but it looks good. will change.

lib/luacompletion.c
sav: it could be, for the sake of completeness, in case the internal API changes. will change.

lib/mailbox.lua
sav: unless you want to foresee potential future changes, like in that other point you came up with. I find it misleading, but it's really up to you here.

lneto: No need.. the thing is `push()` used to return but now it raises an error.. all good 💪🏿

lib/luacompletion.c
sav: I missed something important. `wait_for_completion_interruptible_timeout` returns an `unsigned long`; if I were to check for anything other than the specific known return values, I'd have to cast it to signed. That might be too much of an assumption, don't you think?

sav: sorry, I made a confusion here with the previous function we were using, the prototypes are actually: 
- `unsigned long wait_for_completion_timeout(struct completion *done, unsigned long timeout)`
- `long wait_for_completion_interruptible_timeout(struct completion *done, unsigned long timeout)`

lneto:
> I would just hold it until we merge `lneto_sleep`

@sav it's merged =)..

lib/mailbox.lua
```diff
@@ -3,29 +3,37 @@
...
-local fifo = require("fifo")
+local fifo       = require("fifo")
+local completion = require("completion")
 
 local mailbox = {} 
 local MailBox = {}
 MailBox.__index = MailBox
 
-local function new(o, allowed, forbidden)
+local function new(o, e, allowed, forbidden)
```
lneto, line 13: Actually, I think I would keep just `o` here.. and then
```
local mbox = type(o) == 'userdata' and o or
  {queue = fifo.new(o.queue), event = completion.new(o.event)}
```
What do you think?

sav: it could be. then we'd have to call it like `local outbox = mailbox.outbox({queue = queue, event = event})` in dome (and, potentially, in other impls), right? it's a bit clumsy IMO, but it makes sense if we want to support native objects in the future; it could be reduced to `mailbox.outbox(inbox)`. these are two separate things albeit related, right? because impls will often have to pass them as `fifo, completion`.

lneto: you're absolutely right.. somewhat I thought would be possible to support `mailbox.outbox(inbox)` simply by doing this.. sorry for the noise.. let's keep it as is by now..

PR #190, opened 2024-09-15
protect lunatik_runtime() call

lneto:
* change ENINVAL to ENOEXEC
* replace lunatik.runtimes() with lunatik.runtimes
* require libs automatically on lunatik_pushobject()

### DIFF
--- a/lunatik.h
+++ b/lunatik.h
@@ -180,10 +180,18 @@ int lunatik_monitorobject(lua_State *L);
 #define lunatik_getobject(o)		kref_get(&(o)->kref)
 #define lunatik_putobject(o)		kref_put(&(o)->kref, lunatik_releaseobject)
 
+static inline void lunatik_require(lua_State *L, const char *libname)
+{
+	lua_getglobal(L, "require");
+	lua_pushstring(L, libname);
+	lua_call(L, 1, 0);
+}
+
 static inline void lunatik_pushobject(lua_State *L, lunatik_object_t *object)
 {
-	lunatik_getobject(object);
+	lunatik_require(L, object->class->name);
 	lunatik_cloneobject(L, object);
+	lunatik_getobject(object);
 }
 
 static inline bool lunatik_hasindex(lua_State *L, int index)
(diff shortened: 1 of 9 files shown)


lunatik_core.c
```diff
@@ -194,21 +196,15 @@ static int lunatik_newruntime(lunatik_object_t **pruntime, lua_State *parent, co
...
-	luaL_openlibs(L);
-	if (sleep) {
-		luaL_requiref(L, "lunatik", luaopen_lunatik, 0);
-		lua_pop(L, 1); /* lunatik library */
-	}
-
-	if (lunatik_dofile(L, script) != LUA_OK) {
-		lunatik_runerror(L, parent, lua_tostring(L, -1));
+	lua_pushcfunction(L, lunatik_runscript);
+	lua_pushlightuserdata(L, (void *)script);
```
sav, line 200: not a factual suggestion for this review, per se, but rather, just to mention: about all these type casts, I really like those `cast_*` macros Lua define in `lua/llimits.h`; maybe Lunatik could have their own.

sav (approved):

lunatik.h
```diff
@@ -180,10 +180,18 @@ int lunatik_monitorobject(lua_State *L);
 #define lunatik_getobject(o)		kref_get(&(o)->kref)
 #define lunatik_putobject(o)		kref_put(&(o)->kref, lunatik_releaseobject)
 
+static inline void lunatik_require(lua_State *L, const char *libname)
```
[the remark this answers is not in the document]
lneto: problem here is that it's now used by `pushobjects`.. will need to move both around "dirtying" the patch.. if you think it's worth, I can move..

lunatik_core.c
lneto: we could use it if you think it's worth.. I wouldn't replicate them though.. we could include this header on lunatik.h perhaps..

lneto: but I would apply this to the code as a whole, not only in this single place now.. what do you think?

lunatik.h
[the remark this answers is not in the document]
sav: no worries.

lunatik_core.c
sav: I just mentioned as something to consider for future changes.

lneto: btw, you can create a issue for that..

PR #196, opened 2024-09-22
Debian: install Lunatik on kernel upgrade

### LINKED ISSUE #195: Script to automatically install Lunatik on Debian update
[Here is](https://gist.github.com/jperon/46f17707425b87ec6ed1dc56724e3cd8) a script that, in `/etc/kernel/postinst.d/`, installs Lunatik for the new kernel. Would it be worth a PR? If so, where should I put the script?

jperon:
Fix #195

### DIFF
--- a/tools/debian_kernel_postinst_lunatik.sh
+++ b/tools/debian_kernel_postinst_lunatik.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: (c) 2024 jperon
+# SPDX-License-Identifier: MIT OR GPL-2.0-only
+
+KERNEL_RELEASE=$(uname -r)
+# This ugly hack is needed because Ubuntu has a weird kernel package versionning.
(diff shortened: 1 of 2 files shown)


lneto (approved):

tools/debian_kernel_postinst_lunatik.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
lneto, line 1: Can you please add license here? Dual MIT/GPL2 is preferable.. thanks!

jperon: Just done, with copyright assignment so that you may change it at your will.

jperon:
Just added a small change to use all detected cpu cores, and not previously hardcoded `4`.

tools/debian_kernel_postinst_lunatik.sh
```diff
@@ -0,0 +1,30 @@
...
+# SPDX-FileCopyrightText: (c) 2024 Ring Zero Desenvolvimento de Software LTDA
```
lneto, line 2: I think you should use your own copyright ;-).. btw, can you also squash your commits? It helps to sanitize the history.. thanks!

jperon:
Here you are!

Issue #193, opened 2024-09-19
SNI filter sni.lua can't filter chrome browser with reassembled TCP segments?


vincentmli:
when use chrome browser, the ssl client hello is sent in two TCP segments, and reassembled, it appears sni.lua can't handle TCP reassembled clienthello SNI?

![Screenshot 2024-09-19 at 7 38 38 AM](https://github.com/user-attachments/assets/30ba0867-93ef-4f4d-84a6-0ef0557afe86)

vincentmli:
the first TCP segments contains handshake and clienthello and up to extension `key_share` which has 1263 bytes. `server_name` extension is in the second TCP segments, remove the handshake and clienthello check in sni.lua seems not enough to get `server_name`, the offset could be incorrect still.

lneto:
wrapping up our discussion on Matrix.. it might be resolved on kernel 6.9 by this:

https://elixir.bootlin.com/linux/v6.9/source/net/core/dev.c#L4914

see also: https://lore.kernel.org/bpf/cover.1707729884.git.lorenzo@kernel.org/

vincentmli:
@lneto I tried upstream stable kernel 6.10.11 release, same problem

```
[root@bpfire-3 ~]# uname -a
Linux bpfire-3.localdomain 6.10.11-ipfire #1 SMP PREEMPT_DYNAMIC Fri Sep 20 15:26:24 GMT 2024 x86_64 Intel(R) Celeron(R) CPU G1820 @ 2.70GHz GenuineIntel GNU/Linux
```

lneto:
I'm not sure why it's not triggering skb_linearize(), I would need to have a better look at this.. thinking also on a simple hack on Lua side to keep context of the parser.. will try it out later..

vincentmli:
> I'm not sure why it's not triggering skb_linearize(), I would need to have a better look at this.. thinking also on a simple hack on Lua side to keep context of the parser.. will try it out later..

is there some issue with  `skb_linearize()` not triggering in kernel or in lunatik?

lneto:
> > I'm not sure why it's not triggering skb_linearize(), I would need to have a better look at this.. thinking also on a simple hack on Lua side to keep context of the parser.. will try it out later..
> 
> is there some issue with `skb_linearize()` not triggering in kernel or in lunatik?

the issue is probably with my understanding of what should actual happen ;-).. I was expecting the following flow..

[...]

PR #198, opened 2024-09-23
Add netfilter hooks on filter example

jperon:
Related to #193 

Could’nt test it, as I don’t have the necessary machines at hand: should be tested before merge.

### DIFF
--- a/examples/filter/sni.lua
+++ b/examples/filter/sni.lua
@@ -4,8 +4,7 @@
 --
 
 local xdp  = require("xdp")
-
-local action = xdp.action
+local netfilter = require("netfilter")
 
 local function set(t)
 	local s = {}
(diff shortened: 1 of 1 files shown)


lneto:
merging on lneto_filter integration branch.. so we can use to develop the multi layer filter before merging it into master..

vincentmli:
@lneto @jperon I simply replaced the sni.lua, did nothing else, also loaded https.o program, it stops working, even xdp stops working, here is the kernel error log:
```
Sep 26 00:13:11 bpfire-2 kernel: luanetfilter: luanetfilter hook: pcall error /lib/modules/lua/examples/filter/sni.lua:30: attempt to index a number value (upvalue 'packet')
```

PR #192, opened 2024-09-19
install ebpf https.o to /usr/local/lib/bpf/lunatik

vincentmli:
Makefile miss installing ebpf https.o program.

install https.o in /usr/local/lib/bpf/lunatik since xdp-tools by default install ebpf program in /usr/local/lib/bpf.

PR #202, opened 2024-09-26
fix bpf and bpf_install Makefile targets

lneto:
fix https://github.com/luainkernel/lunatik/pull/192#issuecomment-2374650775


vincentmli (commented):
`.PHONY: ebpf` should be removed

PR #210, opened 2024-11-04
Ipparse integration

jperon:
Makefile action to install / uninstall `ipparse` from Moontastik releases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ MODULES_INSTALL_PATH = /lib/modules/${KERNEL_RELEASE}
 SCRIPTS_INSTALL_PATH = /lib/modules/lua
 LUNATIK_INSTALL_PATH = /usr/local/sbin
 LUNATIK_EBPF_INSTALL_PATH = /usr/local/lib/bpf/lunatik
+MOONTASTIK_RELEASE ?= v0.1c
 LUA_API = lua/lua.h lua/lauxlib.h lua/lualib.h
 KDIR ?= ${MODULES_INSTALL_PATH}/build
 RM = rm -f
@@ -88,4 +89,15 @@ uninstall: scripts_uninstall modules_uninstall
 
 lunatik_sym.h: $(LUA_API)
 	${shell ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
+
+moontastik_install_%:
+	[ $* ] || (echo "usage: make moontastik_install_TARGET" ; exit 1)
+	wget https://github.com/luainkernel/moontastik/releases/download/${MOONTASTIK_RELEASE}/moontastik_lua.zip -O moontastik_lua.zip
+	[ -d moontastik_lua ] && ${RM} -r moontastik_lua || true
+	unzip moontastik_lua.zip
+	cd moontastik_lua/"$*" && ./install.sh ; cd -
+
+moontastik_uninstall_%:
+	[ $* ] || (echo "usage: make moontastik_uninstall_TARGET" ; exit 1)
+	${RM} -r ${SCRIPTS_INSTALL_PATH}/$*
 
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,4 @@ modules.builtin
 modules.order
 build/
 lunatik_sym.h
-
+moontastik*


Makefile
```diff
@@ -58,6 +58,13 @@ examples_install:
 	${MKDIR} ${SCRIPTS_INSTALL_PATH}/examples/dnsdoctor
 	${INSTALL} -m 0644 examples/dnsdoctor/*.lua ${SCRIPTS_INSTALL_PATH}/examples/dnsdoctor
 
+ipparse_install:
```
lneto, line 61: I would call this `moonstatik_install`
```suggestion
moonstatik_install:
```

tools/debian_kernel_postinst_lunatik.sh
```diff
@@ -42,6 +42,7 @@ else
   cd "${LUNATIK_DIR}"
 fi
 make -j"${CPU_CORES}" KERNEL_RELEASE="${KERNEL_RELEASE}" && make KERNEL_RELEASE="${KERNEL_RELEASE}" install || exit 1
+make ipparse_install
```
lneto, line 45: I would make it optional

Makefile
```diff
@@ -58,6 +58,13 @@ examples_install:
...
+	wget https://github.com/luainkernel/moontastik/releases/latest/download/moontastik_lua.zip -O moontastik_lua.zip
+	[ -d moontastik_lua ] && ${RM} -r moontastik_lua || true
+	unzip moontastik_lua.zip
+	${MKDIR} ${SCRIPTS_INSTALL_PATH}/ipparse
+	cd moontastik_lua/ipparse && find . -type f -exec install -vDm 644 {} ${SCRIPTS_INSTALL_PATH}/ipparse/{} \; ; cd -
```
lneto, line 66: I would move this to an "install" script in the `moonstatik` package

Makefile
```diff
@@ -58,6 +58,13 @@ examples_install:
 	${MKDIR} ${SCRIPTS_INSTALL_PATH}/examples/dnsdoctor
 	${INSTALL} -m 0644 examples/dnsdoctor/*.lua ${SCRIPTS_INSTALL_PATH}/examples/dnsdoctor
 
+ipparse_install:
+	wget https://github.com/luainkernel/moontastik/releases/latest/download/moontastik_lua.zip -O moontastik_lua.zip
```
lneto, line 62: I would define a specific version here.. for safety..

Makefile
jperon: The fact is it doesn’t install whole moontastik, but only ipparse, which is a lib, so intended to be useful for other scripts that could be part of lunatik's main repo. Moontastik contains also (for now) [hexdump](https://github.com/luainkernel/moontastik/tree/master/hexdump), which doesn’t get installed here.

tools/debian_kernel_postinst_lunatik.sh
jperon: OK so I’ll remove it from the script, as there is no option passing for it.

Makefile
jperon: I don’t understand this one: the whole point of this integration is to avoid MoonScript dependency. If I put this in `moontastik` (which doesn’t include the lua files), it will necessarily add steps to install `ipparse`.

Makefile
lneto: my point is to decouple lunatik from moonstatik.. that is, lunatik base doesn't need to have the visibility of moonstatik modules.. we can install "moonstatik" from here and let moonstatik base decide whatever will be present in the default package

Makefile
lneto: decoupling as well.. we can have `moonstatik/install.sh` for instance, and let the lunatik script just call it here.. lunatik doesn't need to know the inner details..

tools/debian_kernel_postinst_lunatik.sh
lneto: my suggestion is to have an option.. e.g., CONFIG_MOONSTATIK or something..

jperon:
@lneto What about the new approach I took? Installing ipparse would become:
```sh
make moontastik_install_ipparse
```

and uninstalling:
```sh
make moontastik_uninstall_ipparse
```

lneto:
> @lneto What about the new approach I took? Installing ipparse would become:
> 
> ```shell
> make moontastik_install_ipparse
> ```
> 
> and uninstalling:
> 
> ```shell
> make moontastik_uninstall_ipparse
> ```

my point is that Lunatik should have no visibility of the internal modules of Moonstatik.. IMO, we should just have `make moonstatik[un]install` here in Lunatik and let Moonstatik decide whatever is present in its release.. what do you think?

jperon:
The fact is Lunatik hasn't this visibility: the modules are hidden behind those targets: `moontastik_[un]install_%:`. That way, the user may choose which part he wants to install, and modules could be added to Moontastik without changing Lunatik’s makefile.

Makefile
```diff
@@ -89,3 +90,14 @@ uninstall: scripts_uninstall modules_uninstall
 lunatik_sym.h: $(LUA_API)
 	${shell ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
 
+moontastik_install_%:
+	[ $* ] || (echo "Which target would you like to install?" && exit 1)
+	wget https://github.com/luainkernel/moontastik/releases/download/v0.1c/moontastik_lua.zip -O moontastik_lua.zip
```
lneto, line 95: I would move the release version to a variable.. so it would be easier to adjust it later

Makefile
```diff
@@ -89,3 +90,14 @@ uninstall: scripts_uninstall modules_uninstall
 lunatik_sym.h: $(LUA_API)
 	${shell ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
 
+moontastik_install_%:
+	[ $* ] || (echo "Which target would you like to install?" && exit 1)
```
lneto, line 94: IMHO, a better idea is to print something like `usage: make moontastik_install <target>`

lneto (changes requested):
Can you change the commit message? s/ipparse/moontastik/

Makefile
```diff
@@ -89,3 +91,14 @@ uninstall: scripts_uninstall modules_uninstall
...
 
+moontastik_install_%:
+	[ $* ] || (echo "usage: make moontastik_install_TARGET" ; exit 1)
+	wget https://github.com/luainkernel/moontastik/releases/download/${MOONTASTIK_RELEASE}/moontastik_lua.zip -O moontastik_lua.zip
+	[ -d moontastik_lua ] && ${RM} -r moontastik_lua || true
+	unzip moontastik_lua.zip
+	cd moontastik_lua/"$*" && ./install.sh ; cd -
+
+moontastik_uninstall_%:
+	[ $* ] || (echo "Which target would you like to uninstall?" && exit 1)
```
lneto, line 102: Can you also change this to "usage ..."?

Makefile
```diff
@@ -58,6 +59,7 @@ examples_install:
 	${MKDIR} ${SCRIPTS_INSTALL_PATH}/examples/dnsdoctor
 	${INSTALL} -m 0644 examples/dnsdoctor/*.lua ${SCRIPTS_INSTALL_PATH}/examples/dnsdoctor
 
+
```
lneto, line 62: Can you remove this newline?

lneto (changes requested):
Besides these two comments.. I think we are good to merge.. thanks!

Makefile
jperon: Sorry, didn’t think about this one…

PR #212, opened 2024-12-03
pass ether header to the netfilter hook

sav:


lneto (changes requested):
we should update the documentation and examples as well and make sure they are still working properly..

lib/luanetfilter.c
```diff
@@ -45,7 +45,11 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
 		pr_err("could not get skb\n");
 		return -1;
 	}
-	luadata_reset(data, skb->data, skb_headlen(skb), LUADATA_OPT_NONE);
+
+	if (skb_mac_header_was_set(skb))
+		luadata_reset(data, skb_mac_header(skb), skb_headlen(skb) + skb_mac_header_len(skb), LUADATA_OPT_NONE);
+	else
+		luadata_reset(data, skb->data, skb_headlen(skb), LUADATA_OPT_NONE);
```
lneto, line 52:
```suggestion
	if (skb_mac_header_was_set(skb)) {
		ptr = skb_mac_header(skb);
		offset = skb_mac_header_len(skb);
	}
	luadata_reset(data, ptr, skb_headlen(skb) + offset, LUADATA_OPT_NONE);
```

I'm not sure which version would be cleaner.. feel free to decide =)

lneto (approved):

lib/luanetfilter.c
lneto: just let me know, then I can merge it promptly

sav: sorry for the delay. I'd go with the original version, as I find it more direct. what do you think?

lneto: no worries.. I've no strong feeling on that ;-).. let's merge as is then..

PR #213, opened 2024-12-03
add linux.percpu api

sav:


lib/lunatik/runner.lua
```diff
@@ -55,8 +79,11 @@ function runner.list()
 end
 
 function runner.shutdown()
-	rcu.map(env.runtimes, function (script)
-		runner.stop(script)
+	rcu.map(env.threads, function (script)
+		stop(env.threads, script)
+	end)
+	rcu.map(env.runtimes, function (name)
+		stop(env.runtimes, name)
```
lneto, line 86: why don't we need to stop percpu here as well?

lib/lunatik/runner.lua
```diff
@@ -15,16 +16,36 @@ local function trim(script) -- drop ".lua" file extension
...
+function runner.run(script, ...)
+	local script = trim(script)
+	local name = runtime_name(script, 0)
+	return run(name, script, ...)
+end
+
+function runner.percpu(script, ...)
+	local script = trim(script)
+	local runtimes = {}
+	local args = {...}
```
lneto, line 41: we don't need this, right? couldn't we just pass `...` along?

Issue #215, opened 2024-12-28
unable to use `socket.inet` in non-sleepable contexts


sav:
some functions in `socket.inet` could come in handy while scripting filters (i.e., in softirq), but currently, we are unable to load them, as the following error occurs:

`cannot use 'socket' class on non-sleepable runtime`

could we have a workaround allowing some of these functions to be used from softirq? for instance, I wanted to use `socket.inet.ntoa`.

rustedusted:
i'll take up this issue
:D

PR #217, opened 2025-02-04
adding a net library

rustedusted:
draft: solution for the issue #215

yet to change the docs 
wanted to confirm if this is correct way to implement or is there any other better way

### DIFF
--- a/lib/net.lua
+++ b/lib/net.lua
@@ -1,0 +1,32 @@
+--
+-- SPDX-FileCopyrightText: (c) 2023-2025 Ring Zero Desenvolvimento de Software LTDA
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+local net = {}
+
+function net.aton(addr)
+	local i = 1
+	local bits = { 24, 16, 8, 0 }
+	local ip = 0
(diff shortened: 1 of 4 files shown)


rustedusted:
if the changes are fine i can change the docs

lib/net.lua
```diff
@@ -0,0 +1,31 @@
...
+	local bytes = {}
+	for i = 1, n do
+		local shift = (n - i) * 8
+		local mask = 0xFF << shift
+		bytes[i] = (ip & mask) >> shift
+	end
+	return table.concat(bytes, '.')
+end
+
+return net
```
lneto, line 31: We use a newline in the end of files on the code base

lneto (changes requested):
Please squash your two first commits.. the code shouldn't be committed without the original copyright

lib/net.lua
rustedusted: oh i missed that will do that and also squash the commits in some time

lneto (changes requested):
You changed it wrongly from 2023 to 2024, the correct is 2023-2025

rustedusted:
oh hey sorry for that

PR #233, opened 2025-04-02
Mailbox improvements for non-sleepable runtimes

jperon:
This PR adds:
- options to make mailboxes non-blocking;
- an alternative implementation using `lunatik._ENV`, so that it’s not necessary to pass the mailbox as an argument to the runtime:
```lua
inbox = require("mailbox").rcu_inbox
local box = inbox("my_test_msgbox")
```
```lua
outbox = require("mailbox").rcu_outbox
local box = outbox("my_test_msgbox")
```

### DIFF
--- a/lib/mailbox.lua
+++ b/lib/mailbox.lua
@@ -1,6 +1,6 @@
 --
 -- SPDX-FileCopyrightText: (c) 2024 Ring Zero Desenvolvimento de Software LTDA
--- SPDX-License-Identifier: MIT OR GPL-2.0-only 
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
 --
 
 ---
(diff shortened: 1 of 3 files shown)


lneto (changes requested):
we have discussed the changes on this PR on the matrix room..

lib/mailbox.lua
```diff
@@ -1,21 +1,32 @@
 --
 -- SPDX-FileCopyrightText: (c) 2024 Ring Zero Desenvolvimento de Software LTDA
--- SPDX-License-Identifier: MIT OR GPL-2.0-only 
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
```
lneto, line 3: please, don't add trailing whitespaces on your patch =)

jperon: I believe I didn’t add, but remove one.

PR #234, opened 2025-04-05
separate kernel build directory and module build path

glk0:
The MODULES_BUILD_PATH variable was used to designate both the location of kernel headers and where to copy `/sys/kernel/btf/vmlinux` when
`btf_install` target. when building and installing a module for the same machine, they are indeed the same; however they differ if we pass
a MODULES_BUILD_PATH externally, for example the kernel source, with a config for a different target platform.
 
The default values remain the same as they were.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,9 +6,11 @@ MODULES_PATH := /lib/modules
 MODULES_RELEASE_PATH := ${MODULES_PATH}/${KERNEL_RELEASE}
 MODULES_ORDER_LIST := lunatik/lunatik.ko kernel/zfs/zfs.ko # needed for Ubuntu
 MODULES_ORDER_FILE := ${MODULES_RELEASE_PATH}/modules.order
-MODULES_BUILD_PATH ?= ${MODULES_RELEASE_PATH}/build
+BTF_INSTALL_PATH = ${MODULES_RELEASE_PATH}/build
+MODULES_BUILD_PATH ?= ${BTF_INSTALL_PATH}
 MODULES_INSTALL_PATH := ${MODULES_RELEASE_PATH}/kernel
 SCRIPTS_INSTALL_PATH := ${MODULES_PATH}/lua
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -6,7 +6,8 @@ MODULES_PATH := /lib/modules
 MODULES_RELEASE_PATH := ${MODULES_PATH}/${KERNEL_RELEASE}
 MODULES_ORDER_LIST := lunatik/lunatik.ko kernel/zfs/zfs.ko # needed for Ubuntu
 MODULES_ORDER_FILE := ${MODULES_RELEASE_PATH}/modules.order
-MODULES_BUILD_PATH ?= ${MODULES_RELEASE_PATH}/build
+KERNEL_BUILD_DIR = ${MODULES_RELEASE_PATH}/build
+MODULES_BUILD_PATH ?= ${KERNEL_BUILD_DIR}
```
lneto, line 10:
```suggestion
MODULES_BUILD_PATH = ${MODULES_RELEASE_PATH}/build
BTF_INSTALL_PATH ?= ${MODULES_BUILD_PATH}
```
I think it would be clearer to state this var is used to install "btf" here..

glk0: I think we dont want to override MODULES_BUILD_PATH if it's passed from the command line ... that's why I use `?=` instead of `=`.

lneto: ah.. got you, sorry.. so, I think this would comprehend both concerns:
```
BTF_INSTALL_PATH = ${MODULES_RELEASE_PATH}/build
MODULES_BUILD_PATH ?= ${BTF_INSTALL_PATH}
```
that is, just `s/KERNEL_BUILD_DIR/BTF_INSTALL_PATH/g` in you original patch..

what do you think?

glk0: It makes sense. i will apply it

lneto (approved):

PR #236, opened 2025-04-14
added vt support to luanotifier

glk0:
This enables registration of notifier blocks for virtual terminal events. Here's an example of use:

```lua
local notifier = require 'notifier'
local notify = notifier.notify
local vt_ev = notifier.vt

function vt_monitor (event, c, vc_num)
    print (string.format("event=%d, c=%d, vc_num=%d", event, c, vc_num))
return notify.OK end

notifier.vterm (vt_monitor)
```

### DIFF
--- a/lib/luanotifier.c
+++ b/lib/luanotifier.c
@@ -8,6 +8,8 @@
 #include <linux/version.h>
 #include <linux/keyboard.h>
 #include <linux/netdevice.h>
+#include <linux/vt_kern.h>
+#include <linux/vt.h>
 
 #include <lua.h>
 #include <lualib.h>
@@ -42,6 +44,15 @@ static int luanotifier_netdevice_handler(lua_State *L, void *data)
 
 	lua_pushstring(L, dev->name);
 	return 1;
+}
+
+static int luanotifier_vt_handler (lua_State* L, void* data)
+{
+	struct vt_notifier_param* param = data;
+
+	lua_pushinteger(L, param->c);
+	lua_pushinteger(L, param->vc->vc_num);
+	return 2;
 }
 
 static int luanotifier_handler(lua_State *L, luanotifier_t *notifier, unsigned long event, void *data)
@@ -93,6 +104,7 @@ static int luanotifier_##name(lua_State *L)					\
 
 LUANOTIFIER_NEWCHAIN(keyboard);
 LUANOTIFIER_NEWCHAIN(netdevice);
+LUANOTIFIER_NEWCHAIN(vt);
 
 static void luanotifier_release(void *private)
 {
@@ -147,6 +159,7 @@ static int luanotifier_delete(lua_State *L)
 static const luaL_Reg luanotifier_lib[] = {
 	{"keyboard", luanotifier_keyboard},
 	{"netdevice", luanotifier_netdevice},
+	{"vterm", luanotifier_vt},
 	{NULL, NULL}
 };
 
@@ -223,10 +236,19 @@ static const lunatik_reg_t luanotifier_netdev[] = {
 	{NULL, 0}
 };
 
+static const lunatik_reg_t luanotifier_vt_evs[] = {
+	{"VT_ALLOCATE", VT_ALLOCATE},
+	{"VT_DEALLOCATE", VT_DEALLOCATE},
+	{"VT_WRITE", VT_WRITE},
+	{"VT_UPDATE", VT_UPDATE},
+	{"VT_PREWRITE", VT_PREWRITE},
+	{NULL, 0}
+};
 static const lunatik_namespace_t luanotifier_flags[] = {
 	{"notify", luanotifier_notify},
 	{"kbd", luanotifier_kbd},
 	{"netdev", luanotifier_netdev},
+	{"vt", luanotifier_vt_evs},
 	{NULL, NULL}
 };
 
--- a/README.md
+++ b/README.md
@@ -590,6 +590,30 @@ _notifier.netdev_ is a table that exports
 [NETDEV](https://elixir.bootlin.com/linux/v6.3/source/include/linux/netdevice.h#L2812)
 flags to Lua.
 
+### `notifier.vterm(callback)`
+
+_notifier.vterm()_ returns a new virtual terminal `notifier` object and installs it in the system. The `callback` function is called whenever a virtual terminal event happens (e.g. new character written to the terminal, new terminal allocated, etc).
+The callback function receives the following arguments:
+* `event`: the available _events_ are defined by the
+[notifier.vt](https://github.com/luainkernel/lunatik#notifiervt) table.
+* `c`: character related to the event.
+* `vc_num`: virtual console number associated with the event.
+
+The `callback` function might return the values defined by the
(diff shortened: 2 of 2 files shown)


lib/luanotifier.c
```diff
@@ -147,6 +159,7 @@ static int luanotifier_delete(lua_State *L)
 static const luaL_Reg luanotifier_lib[] = {
 	{"keyboard", luanotifier_keyboard},
 	{"netdevice", luanotifier_netdevice},
+    {"vt", luanotifier_vt},
```
lneto, line 162: \t

lib/luanotifier.c
```diff
@@ -223,10 +236,19 @@ static const lunatik_reg_t luanotifier_netdev[] = {
 	{NULL, 0}
 };
 
+static const lunatik_reg_t luanotifier_vt_evs [] = {
+    {"VT_ALLOCATE", VT_ALLOCATE},
+    {"VT_DEALLOCATE", VT_DEALLOCATE},
+    {"VT_WRITE", VT_WRITE},
+    {"VT_UPDATE", VT_UPDATE},
+    {"VT_PREWRITE", VT_PREWRITE},
+    {NULL, 0}
```
lneto, line 245: \t

lib/luanotifier.c
```diff
@@ -44,6 +46,15 @@ static int luanotifier_netdevice_handler(lua_State *L, void *data)
 	return 1;
 }
 
+static int luanotifier_vt_handler (lua_State* L, void* data)
+{
+    struct vt_notifier_param* param = data;
+
+    lua_pushinteger(L, param->c);
+    lua_pushinteger(L, param->vc->vc_num);
+    return 2;
```
lneto, line 55: \t

lneto (commented):
This looks cool! I would merge this if you add a section in README and use tabs for indenting your code.. thanks!

lib/luanotifier.c
glk0: Addressed

lib/luanotifier.c
```diff
@@ -223,10 +236,19 @@ static const lunatik_reg_t luanotifier_netdev[] = {
...
+	{"VT_WRITE", VT_WRITE},
+	{"VT_UPDATE", VT_UPDATE},
+	{"VT_PREWRITE", VT_PREWRITE},
+	{NULL, 0}
+};
 static const lunatik_namespace_t luanotifier_flags[] = {
 	{"notify", luanotifier_notify},
 	{"kbd", luanotifier_kbd},
 	{"netdev", luanotifier_netdev},
+	{"vt_ev", luanotifier_vt_evs},
```
lneto, line 251: I couldn't think on a better name for `vt_ev`, but it doesn't look alike the other flags.. I would try to figure a better name.. perhaps calling the flag table `vt` and the chain `vterm` (or something like that).. it would be more in sync to `keyboard` and `kbd`, for instance..

glk0: Interesting. But the current pattern (short name for hte flag table, long name for the chain) cannot pass here ... I'd like to use LUANOTIFIER_NEWCHAIN macro, and the name for the register/unregister callbacks for vterm uses "vt", not vterm, unlike for keyboard and netdevice where it uses the full name.

lneto: we can name the Lua API "notifier.vt" and still call the function named `luanotifier_vt`, we would just need to s/vt/vterm/ on `luanotifier_lib`, right?

btw, wondering if a better API would be something like this..

```Lua
local notifier = require("notifier")
local vt = notifier.vt

vt.register(function (...) ... end)
```

however, it would be easier to implement something like this..

```Lua
local notifier = require("notifier")
local vt = notifier.vt

notifier.register_vt(function (...) ... end)
```

and then having a second layer in Lua (like socket/inet) to do the `vt.register` "magic".. food for thought ;-)

lneto: the latter looks alright too.. I could live with that.. ;-)

glk0: The last update changes the export name, but keeps the internal names, so now its `vt` for flags, and `vterm` for the notifier. For the proposed improvement of the API, for the sake of consistency, this requires changing the structure of the module I think ?

lneto: it looks good to me, I would just remove test.lua from this patch.. if you like, you could add a an actual example.. or changing an existent one (e.g., the keylogger, I've the impression that using `vterm` we can have the final resulting char instead separated keys, we could add an option to use one or another).. also, I would add a documentation section before merging this..

lneto: >  this requires changing the structure of the module I think ?

Not necessarily; if we follow the second approach, I think we would just need to prefix the callback registers to `register_<chain>`, then we can have `notifier/chains.lua` to do the magic.. however, we don't need to do it now.. it's completely up to you if you would like to explore further such ideas.. =)

glk0: Sorry, the test.lua file was intended for local testing only, commited by mistake.

glk0: For the `register_` edit, I think I will address it in a separate patch.

lib/luanotifier.c
```diff
@@ -223,10 +236,19 @@ static const lunatik_reg_t luanotifier_netdev[] = {
 	{NULL, 0}
 };
 
+static const lunatik_reg_t luanotifier_vt_evs [] = {
```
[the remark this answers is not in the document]
glk0: I think we need to add a Lua formatter spec to the repo. Which one do you think the community will be prefer ?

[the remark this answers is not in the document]
glk0: And one for C too ...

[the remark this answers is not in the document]
lneto: I would prefer one which adds no external dependencies ;-).. I found `.editorconfig` a good approach, for instance..

lib/luanotifier.c
```diff
@@ -282,4 +304,3 @@ module_init(luanotifier_init);
 module_exit(luanotifier_exit);
 MODULE_LICENSE("Dual MIT/GPL");
 MODULE_AUTHOR("Lourival Vieira Neto <lourival.neto@ring-0.io>");
-
```
lneto, line 285: you shouldn't do this.. we should keep \n at the end of the files.. btw, didn't you have added such a rule in .editorconfig?

glk0: I did. Did you try to run `tail -c1 <file`. I think the display in the editor somehow does not display the newline ...

glk0: Anyway, the pull request is updated. That is actually 2 \n at the end of the file now.

lneto: weird.. will check it later.. anyways, if you can just add a section in README and remove the draft status from this PR, I will merge it promptly.. thanks!

lib/luanotifier.c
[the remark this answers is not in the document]
glk0: Or we could add some Github workflow actions to format PR code before merge ...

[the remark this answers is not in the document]
lneto: I think this is a good approach =)

PR #239, opened 2025-04-20
add support for netfilter mark

lneto:


lib/luanetfilter.c
```diff
@@ -119,17 +131,20 @@ static int luanetfilter_register(lua_State *L)
...
-	lunatik_setinteger(L, 1, nfops, pf);
-	lunatik_setinteger(L, 1, nfops, hooknum);
-	lunatik_setinteger(L, 1, nfops, priority);
+	luanetfilter_setinteger(L, 1, nfops, pf);
+	luanetfilter_setinteger(L, 1, nfops, hooknum);
+	luanetfilter_setinteger(L, 1, nfops, priority);
+
+	lua_getfield(L, 1, "mark");
+	nf->mark = lua_tointeger(L, -1);
+	lua_pop(L, 1); /* mark */
```
sheharyaar, line 140: can we replace this with `luanetfilter_setinteger(L, 1, nf, mark)` ??

sheharyaar (changes requested):
We also need to add the changes to the example scripts and documentation.

lib/luanetfilter.c
lneto: I thought the same, but `luanetfilter_setinteger` uses `lunatik_checkfield`, which will fail if `mark = nil`, the plan is to only care about `mark` if it's set and different than 0.. perhaps we can create another macro, but I found it a bit overkill for just this case.. if you think it's worth it, I can add..

lneto:
> We also need to add the changes to the example scripts and documentation.

agreed with documentation.. I was just a bit lazy to do that.. but you are completely right =).. however, it shouldn't impact on the examples.. is this breaking anything?

sheharyaar (approved):

sheharyaar:
Yeah, it should not break the examples.

lib/luanetfilter.c
sheharyaar: I don't think we have more use cases for it as of now.

sheharyaar: We can add it later

marcelstanley (approved):

PR #244, opened 2025-05-07
fix luarcu_map synchronization

lneto:


jperon (approved):

jperon:
Just compiled and relaunch with last force-push, and all seems OK.

Issue #250, opened 2025-05-16
[GSoC 2025] submodule of luahid, luahid.input


qrsikno2:
`luahid` has a submodule named `luahid.inputs` for processing inputs and provides some sys infos, contains:

- `KEY`, `KEYID`: table of datas to mapping & unmaping of marco-defined codes in linux.
- `REL`, `RELID`: like above, contains about relative codes.
- `ABS`, `ABSID`: like above, contains about absolute codes.
- `EV`: a table contains several event types.
- `input_event(type, code, value)`: create an input event with specified parameters for user to process custom event.
there are some tool funcs for mapping:
1. `map_key(key_name, clear)`: Map a key named key_name to a key code.  The `clear` is a boolean value, which indicates whether to clear the key mapping or not. 
2. `map_rel(rel_name)`: Map a relative named rel_name to a relative code.
3. `map_abs(abs_name)`: Map an absolute named abs_name to an absolute code.
4. map_other(event_id, event_name): Map an event named event_name to an event id. 
event_id is a number, beside key, rel, and abs event, it’s given in the evtable.

Issue #249, opened 2025-05-16
[GSoC 2025] Basic APIs for luahid & its submodule


qrsikno2:
For `luahid`, as i mentioned in my design, I list the following APIs for my "class":

- `name` for driver's name
- `match_list` for matching the fixed device in bus(es). it is a table and must cantains the `vendor_id` and `product_id`.
- `match`: a function for dynamic matching devices, like
```lua
-- other codes
match = function(device)  
    return device.vendor == 0x0111
end
```
- report_descriptor, a tables contains several byted number, like 
```lua
report_descriptor = {
0x05, 0x01, 0x09, 0x01, 0xa1, 0x01,
0x05, 0x09, 0x19, 0x01, 0x29, 0x08,
-- ... etc
}` 
```
- init(device): a function  to initialize the driver.
- on_event: a table to process diffierent types of event, like:
```lua
on_event = {
  key = {
    [hid.input.KEY.A] = function(value)
      print("unsupported A, filtered", value)
      return true
    end,
    [hid.input.KEY.B] = function(value)
      print("Key B occured:", value)
      return false
    end,
    default = function(key, value)
      print("unknown key, filtered", key.name, value)
      return true
    end
  },
-- there is no ABS.Z in the example,
-- so the default event handler of it will be used.
-- you can also redefine the default event handler.
  abs = {
    [hid.input.ABS.X] = function(value)
      print("position X:", value)
      return true
    end,
    [hid.input.ABS.Y] = function(value)
      print("position Y:", value)
      return true
    end
  },
  raw = function(data)
    if data[1] == 0x01 then
      print("find a special data:", data[2])
      return true
    end
    return false
  end
}
```
This is my inital design(given in examples) of on_event, but it is a little bit difficult to implement & process the hole `on_event` in C. 
But names like`on_event_raw`, `on_event_key` are lack of extensibility.

- `input_mapping`: a function to process data "before" submit to Linux input subsys.
- `input_mapped`: a function to process data "after" submit to Linux input subsys.
the two above may use the submodule `luahid.input`, but it is optional to use or not.

and the class is loaded by an API `register` in `luahid` and the other API funcs `send_report` is used to operate the device if you want.

feel free to discuss, thanks. =)

lneto:
I would draft a simple driver based on an existing one (e.g., generic mouse/keyboard) using this API to make it clear and help us to brush it up..

qrsikno2:
[...]

PR #252, opened 2025-05-17
Luahid api draft

qrsikno2:
There is the background of this api:
[#249 ](https://github.com/luainkernel/lunatik/issues/249)
[#250 ](https://github.com/luainkernel/lunatik/issues/250)


examples/slow_mouse.lua
```diff
@@ -0,0 +1,96 @@
+local hid = require("luahid")
```
lneto, line 1: always add license and copyright

examples/slow_mouse.lua
```diff
@@ -0,0 +1,96 @@
+local hid = require("luahid") 
+local dummy_submitx = function(value)
+	-- pass
+end
+local dummy_submity = function(value)
```
lneto, line 5: mind our code style, please

test_main.lua
```diff
@@ -0,0 +1,29 @@
+local linux = require("linux")
```
lneto, line 1: what does this test do? it's quite confusing for me.. how does it relate to your hid draft?

qrsikno2: wrong mixed. removed.

examples/slow_mouse.lua
qrsikno2: fixed

examples/slow_mouse.lua
qrsikno2: fixed

examples/slow_mouse.lua
```diff
@@ -0,0 +1,102 @@
...
+-- SPDX-FileCopyrightText: (c) 2025 Jieming Zhou <qrsikno@gmail.com>
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+-- example of a slow mouse driver
+local hid = require("luahid") 
+local function dummy_submitx(value)
+	-- pass
+end
+local function dummy_submity(value)
```
lneto, line 11: please, mind \n's.. observer the style of other examples..

lneto, line 11: what's the purpose of these two funcs?

examples/slow_mouse.lua
```diff
@@ -0,0 +1,102 @@
...
+	-- pass
+end
+local slow_mouse_driver = {
```
lneto, line 14: please, add readme entry for describing this API and every field in this table..

examples/slow_mouse.lua
```diff
@@ -0,0 +1,102 @@
...
+	name = "Luahid Half-Speed Mouse Example",
+	match_list = {
+		{ vendor_id = 0x1234, product_id = 0x5678 }, 
+	},
+	match = function(device) 
+		return device.vendor_id == 0x1233
+	end,
+
+	report_descriptor = {
```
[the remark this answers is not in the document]
qrsikno2: I write it to run a imaginary driver. That's because a set of devices may have their own quirks(maybe) which make it very difficult to cope with diffierent situation. But The draft needs a full example for diffierent apis i think.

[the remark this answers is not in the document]
lneto: I would start by "porting" a driver using the proposed API.. so we can have a better sense..

examples/xiaomi_mouse.lua
```diff
@@ -0,0 +1,65 @@
+--
+-- SPDX-FileCopyrightText: (c) 2025 Jieming Zhou <qrsikno@gmail.com>
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+-- example of a slow mouse driver
+local hid = require("luahid")
+
+local fixed_xiaomi_report_descriptor = {
```
lneto, line 9: please, add a comment with a link to reference kernel code

examples/xiaomi_mouse.lua
```diff
@@ -0,0 +1,65 @@
...
+	0xC0                --  End Collection                      --
+}
+
+local xiaomi_mouse_driver = {
+	name = "luahid_xiaomi_mouse_driver",
+	match_list = {
+		{ vendor_id = 0x2717, product_id = 0x5014 },  -- copied from kernel src 
+	},
+
+	report_descriptor = fixed_xiaomi_report_descriptor
```
lneto, line 62: that's it? don't you need any callback? just this descriptor?

lneto: you probably should support callbacks on your API, instead of just assuming such table.. devices drivers should use such callbacks to do their own work.. e.g., https://elixir.bootlin.com/linux/v6.14.7/source/drivers/hid/hid-xiaomi.c#L64

lneto: btw, that's one reason this example (xiaome mouse) is not that good.. because it's very limited.. if you see, https://elixir.bootlin.com/linux/v6.14.7/source/include/linux/hid.h#L811, we have many other callbacks to be defined that you *must* give support in your API..

examples/simple_generic_mouse.lua
```diff
@@ -0,0 +1,13 @@
...
+--
+
+-- example of a simple generic mouse driver
+local hid = require("luahid")
+
+local mouse_driver = {
+	name = "luahid_simple_generic_mouse_driver",
+}
+
+hid.register(mouse_driver)
```
lneto, line 13: I don't see the point of this script.. can you explain it or remove it from the PR?

examples/xiaomi_mouse.lua
sheharyaar: @qrsikno2 , I agree with Lourival here. Using callbacks will make the luahid library "generic". By generic we mean:

1. You cannot directly call the kernel C functions from lua scripts (E.g: xiaomi_mouse.lua)
2. So you create a library `luahid` that will provide an interface (API) to the lua scripts.
3. The `luahid` must allow the scripts to register their own callbacks, in this sense, this library becomes generic. The use of callbacks will allow you to register your own functions for the xiaomi mouse or apple keyboard. This is how even the kernel works. The `hid_driver` is a struct with function pointers so that any driver can use this struct and define their own drivers, hence the `hid` API is generic.

So `xiaomi_mouse.lua` calls the `luahid` functions, which call the Kernel C functions.

qrsikno2: I think there is no need to have callbacks in this scene. that is because the hid-xiaomi is going to fix special device's report descriptor, so you just need to provide the vendor_id+product_id+descriptor. and this makes the example very clear and simple

examples/simple_generic_mouse.lua
qrsikno2: done

examples/xiaomi_mouse.lua
```diff
@@ -0,0 +1,65 @@
...
+	0x75, 0x08,         --      Report Size (8),                --
+	0x95, 0x04,         --      Report Count (4),               --
+	0xB1, 0x02,         --      Feature (Variable),             --
+	0xC0                --  End Collection                      --
+}
+
+local xiaomi_mouse_driver = {
+	name = "luahid_xiaomi_mouse_driver",
+	match_list = {
+		{ vendor_id = 0x2717, product_id = 0x5014 },  -- copied from kernel src
```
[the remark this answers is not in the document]
qrsikno2: done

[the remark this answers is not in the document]
qrsikno2: I think i would maybe change the `match_list` just because some devices need to given a `bus_type` and `device_type` is that ok for my apis?

examples/xiaomi_mouse.lua
qrsikno2: > please, add a comment with a link to reference kernel code of the original driver请添加一条评论，并附上引用原始驱动内核代码的链接

done

examples/xiaomi_mouse.lua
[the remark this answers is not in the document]
sheharyaar: As my previous comment, the match_list is of type `struct hid_device_id` in the kernel right ?

```c
struct hid_device_id {
	__u16 bus;
	__u16 group;
	__u32 vendor;
	__u32 product;
	kernel_ulong_t driver_data;
};
```

So this can just be a device id table, 

```lua
local xiaomi_id = {
    bus = 0x123,
    group = 0x2345, -- sample values
    vendor = 0x11,
    procuct = 0x01
}

-- under the xiaomi driver, you can use this
local xiaomi_mouse_driver = {
	name = "luahid_xiaomi_mouse_driver",
        ....
        id_table = xiaomi_id
}
...
```

examples/xiaomi_mouse.lua
sheharyaar: So your `luahid` will use this table to then use your functions and tables to pass on to the kernel in a general way for every program.

examples/xiaomi_mouse.lua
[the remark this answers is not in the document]
qrsikno2: yes quite right. i think the table can be write into 'plural' form:
```lua
local xiaomi_id1 = {
    bus = 0x123,
    group = 0x2345, -- sample values
    vendor = 0x11,
    procuct = 0x01
}
local xiaomi_id2 = {
    bus = 0x123,
    group = 0x2345, -- sample values
    vendor = 0x11,
    procuct = 0x02
}

-- under the xiaomi driver, you can use this
local xiaomi_mouse_driver = {
	name = "luahid_xiaomi_mouse_driver",
        ....
        id_table = {xiaomi_id1, xiaomi_id2},
        -- or id_table = {xiaomi_id1}, embraced with '{}' maybe better.
}
...
```

PR #253, opened 2025-05-20
add support for ct->mark on luanetfilter

lneto:


sav (approved):

lneto:
skb->mark seems to be enough right now and will simplify our code base.. if needed, we can set ctmark based on a skb->mark using ntf

PR #255, opened 2025-05-26
Skel2hid

qrsikno2:
copy from `hid-generic` to lunatik.


lib/luahid.c
```diff
@@ -0,0 +1,123 @@
...
+
+#include <linux/hid.h>
+
+#include <lua.h>
+#include <lualib.h>
+#include <lauxlib.h>
+
+#include <lunatik.h>
+
+#include "luanetfilter.h"
```
lneto, line 23: ?

lneto: so we should move this macro to lunatik.h

lunatik.h
```diff
@@ -283,6 +283,17 @@ static inline T checker(lua_State *L, int ix)			\
 
 #define lunatik_getregistry(L, key)	lua_rawgetp((L), LUA_REGISTRYINDEX, (key))
 
+#define lunatik_setstring(L, idx, hook, field, maxlen)        \
```
lneto, line 286: You should remove it from the original place as well, instead of duplicating it..

lneto: I think you missed the point.. you should do both the addition and removal into a single commit.. if you don't know how to fixup a commit, please ask..

qrsikno2: but before your msg i have added in a commit. do you mean i need to change commit history using things like rebase and cherry-pick & so on

Issue #262, opened 2025-06-06
Build failure on kernels < 6.4 due to undeclared luaxp_runtimes variable


ShivamVashisth28:
### Steps to reproduce the issue : 
1. Clone the lunatik repo
2. Run the make command
3. You will be encounterd with the following error message

![Image](https://github.com/user-attachments/assets/2bcabc34-c6cc-4e53-bca2-bcf5ad0643a2)

PR #263, opened 2025-06-06
Fix :  add kernel version guard to luaxdp_exit() for compatibility with < 6.4

ShivamVashisth28:
fixes : #262 

### Summary

This pull request fixes a build failure when compiling Lunatik on Linux kernels older than 6.4.0 (e.g., Debian 12 with 6.1.x).

### Problem

The `luaxdp_exit()` function references `luaxdp_runtimes`  which is only available in Linux >= 6.4. This causes compilation error.

### Solution 

Added an version guard in before using the `luaxdp_runtimes` variable.

### DIFF
--- a/lib/luaxdp.c
+++ b/lib/luaxdp.c
@@ -205,8 +205,10 @@ static int __init luaxdp_init(void)
 
 static void __exit luaxdp_exit(void)
 {
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 4, 0))
 	if (luaxdp_runtimes != NULL)
 		lunatik_putobject(luaxdp_runtimes);
+#endif
 }
 
 module_init(luaxdp_init);


lib/luaxdp.c
```diff
@@ -205,8 +205,10 @@ static int __init luaxdp_init(void)
 
 static void __exit luaxdp_exit(void)
 {
-	if (luaxdp_runtimes != NULL)
-		lunatik_putobject(luaxdp_runtimes);
+	#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 4, 0))
+		if (luaxdp_runtimes != NULL)
+			lunatik_putobject(luaxdp_runtimes);
+	#endif
 }
```
lneto:
```suggestion
#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 4, 0))
	if (luaxdp_runtimes != NULL)
		lunatik_putobject(luaxdp_runtimes);
#endif
}
```

lneto (changes requested):
please, mind the indentation and code style.. you just need to add the ifdef (without spaces) and not changing the indentation of the original code..

ShivamVashisth28:
> please, mind the indentation and code style.. you just need to add the ifdef (without spaces) and not changing the indentation of the original code..

okay 👍

lneto (changes requested):
@ShivamVashisth28 please squash your commit.. let's keep the history as clean as possible and blame working

ShivamVashisth28:
> @ShivamVashisth28 please squash your commit.. let's keep the history as clean as possible and blame working

Done

PR #267, opened 2025-06-12
hotfix: correct function name to fix build failure

qrsikno2:
re-added the netfilter's marco and kept the `lunatik.h`'s marco as well.
consider that can be saved and changed in the `lunatik.h`.
the one in the `luanetfilter.h` can be specialized for netfilter

### DIFF
--- a/lib/luaxtable.c
+++ b/lib/luaxtable.c
@@ -215,7 +215,7 @@ static int luaxtable_new##hook(lua_State *L) 				\
 	struct xt_##hook *hook = &xtable->hook;				\
 	hook->me = THIS_MODULE;						\
 									\
-	luanetfilter_setstring(L, 1, hook, name, XT_EXTENSION_MAXNAMELEN - 1);	\
+	lunatik_setstring(L, 1, hook, name, XT_EXTENSION_MAXNAMELEN - 1);	\
 	luanetfilter_setinteger(L, 1, hook, revision);			\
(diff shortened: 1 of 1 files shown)


lneto:
Why not just changing the calls on luanetfilter to lunatik_setstring?

qrsikno2:
That's a great point. You're absolutely right, refactoring all the call sites would be a much cleaner solution.

As it's nearly 2 AM in my timezone, I'd prefer to tackle this with a fresh mind to ensure I don't miss anything.

I'll make this my first priority when I'm back online in a few hours. Thanks for the catch!

qrsikno2:
> Why not just changing the calls on luanetfilter to lunatik_setstring?

fixed

lib/luaxtable.c
```diff
@@ -253,7 +253,7 @@ static const luaL_Reg luaxtable_lib[] = {
 static void luaxtable_release(void *private)
 {
 	luaxtable_t *xtable = (luaxtable_t *)private;
-	if (!xtable->runtime) 
+	if (!xtable->runtime)
```
lneto: Please don't mix this in the same commit

lneto (changes requested):
Please rephrase your commit message to something more descriptive re what you are really doing, e.g., "fix luanetfilter_setstring call"

qrsikno2:
> Please rephrase your commit message to something more descriptive re what you are really doing, e.g., "fix luanetfilter_setstring call"

fixed

PR #270, opened 2025-06-14
construct a skeleton for luahid

qrsikno2:
Construct a skeleton for `luahid`
Add Makefile and Add entry in `lunatik`.
The `hid.register()` can received a table with attr `name`, like `{name = "hid_ok_driver"}`.


lib/luahid.c
```diff
@@ -0,0 +1,102 @@
...
+
+#include <lunatik.h>
+
+typedef struct luahid_s {
+	lunatik_object_t *runtime;
+	struct hid_driver driver;
+} luahid_t;
+
+static const struct hid_device_id hid_table[] = {
+	{HID_DEVICE(HID_BUS_ANY, HID_GROUP_ANY, HID_ANY_ID, HID_ANY_ID)},
```
lneto, line 18: wouldn't this need to be dynamically defined per driver? or we will always used this definition?

lib/luahid.c
```diff
@@ -0,0 +1,102 @@
...
+#include <lauxlib.h>
+
+#include <lunatik.h>
+
+typedef struct luahid_s {
+	lunatik_object_t *runtime;
+	struct hid_driver driver;
+} luahid_t;
+
+static const struct hid_device_id hid_table[] = {
```
lneto: please, prefix *everything" with `luahid_`; let's keep your namespace clear, so we can spot easily what was defined by your module, what's lunatik, what's lua, what's kernel..
```suggestion
static const struct hid_device_id luahid_table[] = {
```

lib/luahid.c
```diff
@@ -0,0 +1,102 @@
...
+}
+
+static void __exit luahid_exit(void)
+{
+}
+
+module_init(luahid_init);
+module_exit(luahid_exit);
+MODULE_LICENSE("Dual MIT/GPL");
+MODULE_AUTHOR("Jieming Zhou <qrsikno@gmail.com>");
```
lneto, line 91: always add `\n` at the end of files

lib/luahid.c
```diff
@@ -0,0 +1,102 @@
...
+* SPDX-License-Identifier: MIT OR GPL-2.0-only
+*/
+
+#define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+#include <linux/kernel.h>
+#include <linux/module.h>
+#include <linux/version.h>
+#include <linux/slab.h>
+#include <linux/limits.h>
+#include <asm/byteorder.h>
```
lneto: please, review your headers.. it looks like you have many unused here.. always clean up your code

lib/luahid.c
```diff
@@ -0,0 +1,102 @@
...
+
+	/*
+	 * configure the driver's properties & callbacks
+	 */
+	struct hid_driver *user_driver = &(hid->driver);
+	user_driver->name = lunatik_checkalloc(L, NAME_MAX);
+	lunatik_setstring(L, 1, user_driver, name, NAME_MAX);
+	user_driver->id_table = hid_table;
+
+	lunatik_registerobject(L, 1, object);
```
lneto: why not registering the object only if you have successfully registered the hid driver?

lib/luahid.c
```diff
@@ -0,0 +1,102 @@
...
+	luaL_checktype(L, 1, LUA_TTABLE); /* assure that is a driver */
+
+	lunatik_object_t *object = lunatik_newobject(L, &luahid_class, sizeof(luahid_t));
+	luahid_t *hid = (luahid_t *)object->private;
+
+	/*
+	 * configure the driver's properties & callbacks
+	 */
+	struct hid_driver *user_driver = &(hid->driver);
+	user_driver->name = lunatik_checkalloc(L, NAME_MAX);
```
lneto, line 62: you're leaking this

lneto: btw, where is NAM_MAX defined?

lib/luahid.c
```diff
@@ -0,0 +1,102 @@
...
+
+static int luahid_register(lua_State *L)
+{
+	luaL_checktype(L, 1, LUA_TTABLE); /* assure that is a driver */
+
+	lunatik_object_t *object = lunatik_newobject(L, &luahid_class, sizeof(luahid_t));
+	luahid_t *hid = (luahid_t *)object->private;
+
+	/*
+	 * configure the driver's properties & callbacks
```
lneto: it looks like unnecessary comment.. just add explanations where is necessary

lib/luahid.c
```diff
@@ -0,0 +1,102 @@
...
+static const lunatik_class_t luahid_class = {
+	.name = "hid",
+	.methods = luahid_mt,
+	.release = luahid_release,
+	.sleep = true,
+};
+
+static int luahid_register(lua_State *L)
+{
+	luaL_checktype(L, 1, LUA_TTABLE); /* assure that is a driver */
```
lneto: it looks like unnecessary comment.. just add explanations where is necessary

lib/luahid.c
```diff
@@ -0,0 +1,102 @@
...
+	 * configure the driver's properties & callbacks
+	 */
+	struct hid_driver *user_driver = &(hid->driver);
+	user_driver->name = lunatik_checkalloc(L, NAME_MAX);
+	lunatik_setstring(L, 1, user_driver, name, NAME_MAX);
+	user_driver->id_table = hid_table;
+
+	lunatik_registerobject(L, 1, object);
+	lunatik_setruntime(L, hid, hid);
+	lunatik_getobject(hid->runtime);
```
lneto, line 67: I believe the pattern is to getobject(runtime) before setruntime().. btw, you didn't assign `hid->runtime`.. did you try your code? I think it is broken..

lib/luahid.c
qrsikno2: To be honest, I don't know whether that can be removed. I used the table to accept any hid device. and i will use user's own `id_table` or maybe `match` to judge whether received it. like generals pattern+characteristic lua script, but i don't know how will effect after removing it.

I find some chinese posts says it can be removed, and the others says it cannot be removed. so i don't know how to process it.

lib/luahid.c
qrsikno2: Good idea!

lib/luahid.c
qrsikno2: if the member `name` is removed, emmm.... I did an experiment shows whether you say `user_driver->name=NULL;` or delete the two lines will cause kernel panic. so i think we should really have a name. if you don't like it, can i write a static string for the value. 
what is a good name?

The NAME_MAX is defined in linux/limits.h. for eg, https://elixir.bootlin.com/linux/v6.13.7/source/include/uapi/linux/limits.h#L12

lib/luahid.c
qrsikno2: I copied from the `notifier`:
https://github.com/luainkernel/lunatik/blob/c58ec0b301f10821ec4e15123d9aff3950d9854f/lib/luanotifier.c#L272C1-L275C1.
it also write like this: 
1. setruntime; 2.getobject.

qrsikno2: I tried my code. it runs well in my VM and baremetal. 
I have no idea what can i do for `hid->runtime=abc;` do you mean `hid->runtime = NULL;'?

lib/luahid.c

lib/luahid.c

lib/luahid.c
qrsikno2: Oh, no. let me think about it.
your idea can reduce unnecessary trouble, that's right.
but when i'm debugging, i find once i reload the module, the kernel will call my callbacks. 
so if you don't register firstly, i think it is a strange way for me.

but let me confirm that firstly.

qrsikno2: Oho, I cant reproduce it. sorry, i will change it.

lib/luahid.c
qrsikno2: done.

lib/luahid.c
qrsikno2: done

lib/luahid.c

lib/luahid.c

lib/luahid.c
lneto: it's not that I don't like it, but you are memory leaking this.. you should release this memory when the object is released..

lib/luahid.c
lneto: my bad on that.. i misread the code.. sorry..

lib/luahid.c
lneto: my point is not to get rid of it.. but whether we will need different parameters on this table or not.. for instance, if we should let the API client defining this table.. (what looks to be the case).. we can start in this way, but we should have an answer (and a ticket) for that

qrsikno2: If you say so, i would to say this table is fixed with such values.
I have a success trying, just put user's own `id_table` in the `hid_driver`'s member `id_table`, and this table always stay the same. It runs as expected.

I have thought about this problem. you know it's a `static const` global var, which seems to be impossible to change dynamic. that's why i have done the experiment before.

sheharyaar: Yeah, we should let the Lua script (API client) send the id table. One important thing to remember for kernel panics is that, you should remember that the `struct hid_driver` and the `struct hid_device_id` are accessed by the kernel when a callback comes. So it needs to be in memory for the duration. Hence, in normal modules they are declared as `static`, but for us, since we have many API clients, we need to allocate these structures so that they are in the heap and not on the stack.

sheharyaar: I had to spent some time debugging it, so giving you the heads up.

lib/luahid.c
qrsikno2: done

lib/luahid.c
```diff
@@ -0,0 +1,90 @@
...
+	user_driver->id_table = luahid_table;
+
+	lunatik_setruntime(L, hid, hid);
+	lunatik_getobject(hid->runtime);
+
+	int ret = __hid_register_driver(user_driver, THIS_MODULE, KBUILD_MODNAME);
+	if (ret) {
+		luaL_error(L, "failed to register hid driver: %s", user_driver->name);
+		lunatik_putobject(hid->runtime);
+	}
```
```suggestion
	if (__hid_register_driver(user_driver, THIS_MODULE, KBUILD_MODNAME) != 0)
		luaL_error(L, "failed to register hid driver: %s", user_driver->name);
```
* you don't use ret anywhere else;
* you're already putting the object on release(), what will be called when the object is collected after you raise the error;
* what happens if you register more than one hid driver with the same KBUILD_MODNAME? Is it safe?

qrsikno2: changed. but about the third one, i have no idea about that.

qrsikno2: About the third problem. I have dome some easy experiment about it.
i performed:
1. prepare 2 lua driver with different name(if same, it will occur error).
2. load the first one(by doas lunatik run xxxx)
3. load the second one(by doas lunatik run xxxx1)
4. inspect the `dmesg` and find no error log as follow:
```text
[  177.566350] input: VirtualBox USB Tablet as /devices/pci0000:00/0000:00:06.0/usb1/1-1/1-1:1.0/0003:80EE:0021.0001/input/input12
[  177.567197] hid_axis_swapper 0003:80EE:0021.0001: input,hidraw0: USB HID v1.10 Mouse [VirtualBox USB Tablet] on usb-0000:00:06.0-1/input0
[  177.592623] input: Logitech Wireless Receiver Mouse as /devices/pci0000:00/0000:00:06.0/usb1/1-2/1-2:1.0/0003:046D:C542.0002/input/input13
[  177.592702] hid_axis_swapper 0003:046D:C542.0002: input,hidraw1: USB HID v1.11 Mouse [Logitech Wireless Receiver] on usb-0000:00:06.0-2/input0
```
swap the step3 and step2, it will show as:
```lua
[   87.109638] input: VirtualBox USB Tablet as /devices/pci0000:00/0000:00:06.0/usb2/2-1/2-1:1.0/0003:80EE:0021.0001/input/input15
[   87.111082] hid_axis_swapper1 0003:80EE:0021.0001: input,hidraw0: USB HID v1.10 Mouse [VirtualBox USB Tablet] on usb-0000:00:06.0-1/input0
[   87.132924] input: Logitech Wireless Receiver Mouse as /devices/pci0000:00/0000:00:06.0/usb2/2-2/2-2:1.0/0003:046D:C542.0002/input/input16
[   87.134059] hid_axis_swapper1 0003:046D:C542.0002: input,hidraw1: USB HID v1.11 Mouse [Logitech Wireless Receiver] on usb-0000:00:06.0-2/input0
```

I think it's no error if you have two driver, but it will only load the first driver, and the second driver is blocked without error.

lneto: @qrsikno2 it's good to run this kind of experiment.. but you also need go through the kernel API implementation and make sure you understand why it works in this way.. if you cannot proper understand, please, discuss it on matrix.

lib/luahid.c
```diff
@@ -0,0 +1,87 @@
...
+};
+
+static int luahid_register(lua_State *L)
+{
+	luaL_checktype(L, 1, LUA_TTABLE);
+
+	lunatik_object_t *object = lunatik_newobject(L, &luahid_class, sizeof(luahid_t));
+	luahid_t *hid = (luahid_t *)object->private;
+
+	struct hid_driver *user_driver = &(hid->driver);
```
lneto, line 61: I would call this simply `driver`.. I think it's clear enough.. but I don't have any strong feeling on this.. please, feel free to decide..
```suggestion
	struct hid_driver *driver = &(hid->driver);
```

lib/luahid.c
```diff
@@ -0,0 +1,87 @@
...
+static const struct hid_device_id luahid_table[] = {
+	{HID_DEVICE(HID_BUS_ANY, HID_GROUP_ANY, HID_ANY_ID, HID_ANY_ID)},
+	{ }
+};
+MODULE_DEVICE_TABLE(hid, luahid_table);
+
+static void luahid_release(void *private)
+{
+	luahid_t *hid = (luahid_t *)private;
+	if (hid) {
```
lneto: what exactly are you verifying in here? it's to supposed to always be set for a given hid object.. it's static memory, right? the thing you allocate dynamically is `name`, right? and if it fails to alloc with `lunatik_checkalloc`, you will ending up freeing a "dirty" pointer.. If I got it correctly, you would need to initialize name with NULL, then checkalloc().. and in here you should check if `name != NULL`, right?

lib/luahid.c
```diff
@@ -0,0 +1,87 @@
...
+		hid_unregister_driver(&hid->driver);
```
lneto, line 27: also, what happens when you fail on `__hid_register_driver`? release will be called for this object and `hid_unregister_driver()` will be called from here.. is it safe? shouldn't you have a guard for this situation?

lib/luahid.c
```diff
@@ -0,0 +1,96 @@
...
+	user_driver->name = lunatik_checkalloc(L, NAME_MAX);
+	lunatik_setstring(L, 1, user_driver, name, NAME_MAX);
+	user_driver->id_table = luahid_table;
+
+	lunatik_setruntime(L, hid, hid);
+	lunatik_getobject(hid->runtime);
+
+	if (__hid_register_driver(user_driver, THIS_MODULE, KBUILD_MODNAME) != 0)
+		luaL_error(L, "failed to register hid driver: %s", user_driver->name);
+	else
```
lneto: unnecessary.. luaL_error does not return

lib/luahid.c

lib/luahid.c
qrsikno2: done. just because i prefer the `user_driver` :-)

lib/luahid.c
```diff
@@ -0,0 +1,96 @@
...
+		hid->registered = true;
```
lneto: I don't think we need an extra flag for this.. you can just move lunatik_setstring() to here, right?

lib/luahid.c
qrsikno2: i would like to ask, i was confused about the process

lib/luahid.c
qrsikno2: done

lib/luahid.c
```diff
@@ -0,0 +1,96 @@
...
+
```
lneto: please, mind code style..

lib/luahid.c
```diff
@@ -0,0 +1,96 @@
...
+};
+
+static int luahid_register(lua_State *L)
+{
+	luaL_checktype(L, 1, LUA_TTABLE);
+
+	lunatik_object_t *object = lunatik_newobject(L, &luahid_class, sizeof(luahid_t));
+	luahid_t *hid = (luahid_t *)object->private;
+	hid->registered = false;
+	hid->runtime = NULL;
```
lneto: perhaps a better idea is to memset to 0 the the whole private structure.. take a look at the code base.. I think we do this in other modules

lib/luahid.c
lneto: just ask ;-).. we can discuss this on matrix as well.. just raise the topic..

lib/luahid.c
qrsikno2: is this a good idea? Do u think so?
```c
	object = lunatik_newobject(L, &luadevice_class, sizeof(luadevice_t));
	luadev = (luadevice_t *)object->private;

	memset(luadev, 0, sizeof(luadevice_t));

	lunatik_setruntime(L, device, luadev);
	lunatik_getobject(luadev->runtime);

```

lib/luahid.c

lib/luahid.c
qrsikno2: i think this is talked in the group

lib/luahid.c

lib/luahid.c
```diff
@@ -0,0 +1,94 @@
...
+static int luahid_register(lua_State *L)
+{
+	luaL_checktype(L, 1, LUA_TTABLE);
+
+	lunatik_object_t *object = lunatik_newobject(L, &luahid_class, sizeof(luahid_t));
+	luahid_t *hid = (luahid_t *)object->private;
+	memset(hid, 0, sizeof(luahid_t));
+
+	struct hid_driver *user_driver = &(hid->driver);
+	user_driver->name = NULL;
```
sheharyaar: Why the double assignment ?

```
user_driver->name = NULL;
user_driver->name = lunatik_checkalloc(L, NAME_MAX);
```

Why not just : `user_driver->name = lunatik_checkalloc(L, NAME_MAX);` ??
Is this intended ? Sorry if I missed something from the discussion.

lib/luahid.c
```diff
@@ -0,0 +1,94 @@
...
+
+	user_driver->name = lunatik_checkalloc(L, NAME_MAX);
+	lunatik_setstring(L, 1, user_driver, name, NAME_MAX);
+	user_driver->id_table = luahid_table;
+
+	lunatik_setruntime(L, hid, hid);
+	lunatik_getobject(hid->runtime);
+
+	if (__hid_register_driver(user_driver, THIS_MODULE, KBUILD_MODNAME) != 0)
```
sheharyaar, line 69: Why not just set `user_driver = NULL` if this fails. And then in the release part, you can just check:
CC: @lneto 

```
if (hid->driver)
		hid_unregister_driver(&hid->driver);
```

lib/luahid.c
qrsikno2: That's because if allocation is failed, the action `lunatik_free` which aims to free user_driver->name will free a dirty pointer.

lib/luahid.c
lneto: I knew there was a way.. hehe.. thanks, @sheharyaar!

qrsikno2: If you set `user_driver = NULL`, how can we free the data structure or i would like to call it memory? These are some of my simple thoughts

lib/luahid.c
lneto: You are already memset'ing to zero.. you don't need this here, right.. if the allocation fails, it will be already set to NULL

sheharyaar: `lunatik_checkalloc` calls `lunatik_checknull` : https://github.com/luainkernel/lunatik/blob/master/lunatik.h#L115

Which has luaL_error guard. So it won't return here anyway, right?

lib/luahid.c
lneto: Actually, it will already set to NULL due to memset.. and if it fails to register, you will never set it with the mem address.. so, when releasing you just need to check if not NULL.. this case will happen if and only if you've registered the hid driver

qrsikno2: Oh i was misunderstanding your meaning..

lib/luahid.c
qrsikno2: Oh, my mistake.
changed.

lneto (approved):

PR #277, opened 2025-06-18
allow Lua to send signals to kernel threads 

Physic69:
The title is what it was supposed to do.
The pid needs to be conditional and have to improved
the send_sig is better than what i used

### DIFF
--- a/lib/luathread.c
+++ b/lib/luathread.c
@@ -187,7 +187,6 @@ static int luathread_task(lua_State *L)
 * @usage
 * t:allow(15)  -- Allow SIGTERM for this thread
 */
-
 static int luathread_allow(lua_State *L)
 {
    	luathread_t *thread = luathread_check(L, 1);
@@ -332,7 +331,6 @@ static const lunatik_class_t luathread_class = {
 	.sleep = true,
 };
 
-
 #define luathread_new(L)	(lunatik_newobject((L), &luathread_class, sizeof(luathread_t)))
 
 /***
@@ -386,7 +384,6 @@ static int luathread_run(lua_State *L)
 * local current_task_as_thread = thread.current()
 * @within thread
 */
-
 static int luathread_current(lua_State *L)
 {
 	lunatik_object_t *object = luathread_new(L);


lib/luathread.c
```diff
@@ -37,6 +39,9 @@ typedef struct luathread_s {
 
 static int luathread_run(lua_State *L);
 static int luathread_current(lua_State *L);
+static int luathread_allow_signal(lua_State *L);
+static int luathread_signal_pending(lua_State *L);
+static int luathread_send_signal(lua_State *L);
```
lneto: you don't need this.. just move your static functions to before the metatable definition..

lib/luathread.c
```diff
@@ -259,6 +269,59 @@ static int luathread_current(lua_State *L)
...
+ 	int signum = luaL_checkinteger(L, 2);
+ 	allow_signal(signum);
+	
+ 	return 0;
+}
+
+/***
+*
+*
+*/
```
lneto: do or do not -- Yoda. Don't add incomplete things if you are asking for merging.. it's ok for a draft only

lib/luathread.c
```diff
@@ -259,6 +269,59 @@ static int luathread_current(lua_State *L)
...
+
+static int luathread_send_signal(lua_State *L)
+{
+ 	lunatik_object_t *object = lunatik_toobject(L, 1);
+ 	int signum = luaL_checkinteger(L, 2);
+	
+ 	luathread_t *thread = (luathread_t *)object->private;
+ 	struct task_struct *task = thread->task;
+	if (!task)
+ 	 	return luaL_error(L, "thread task is NULL");
```
lneto: you can use LUNATIK_CHECKER for this.. if you really need..

lib/luathread.c
```diff
@@ -259,6 +269,59 @@ static int luathread_current(lua_State *L)
...
+	 int ret = send_sig(signum, task, 0);
```
lneto: if you don't use ret no where else, just check for the function return on the if condition..

lib/luathread.c
```diff
@@ -187,6 +192,10 @@ static const luaL_Reg luathread_mt[] = {
 	{"__gc", lunatik_deleteobject},
 	{"stop", luathread_stop},
 	{"task", luathread_task},
+        {"allow_signal", luathread_allow_signal},
+        {"send_signal", luathread_send_signal},
+        {"signal_pending", luathread_signal_pending},
```
lneto: I think you can sanitize these APIs names.. at least prefixing it with `signal`, e.g., `signal_{allow, send, pending}`.. although I'm not very fond of such long names.. perhaps @jperon have good insights on this.. not sure if simply `allow`, `signal` and `pending` looks better as there's no naming clash.. `t:allow(SIGALRM)`, `t:signal(SIGALRM)`, `t:pending(SIGALRM)` look clear to me.. btw, you might want to add SIG enums in this PR.. it would be quite a easy stretch

lib/luathread.c
```diff
@@ -259,6 +313,7 @@ static int luathread_current(lua_State *L)
 	return 1; /* object */
 }
 
+
```
lneto: Please don't change anything that isn't related to what you're are implementing in this PR.. it includes spaces and style changing in general..

lib/luathread.c
```diff
@@ -274,4 +329,3 @@ module_init(luathread_init);
 module_exit(luathread_exit);
 MODULE_LICENSE("Dual MIT/GPL");
 MODULE_AUTHOR("Lourival Vieira Neto <lourival.neto@ring-0.io>");
-
```
lneto: ^

lib/luathread.c
```diff
@@ -187,6 +236,10 @@ static const luaL_Reg luathread_mt[] = {
 	{"__gc", lunatik_deleteobject},
 	{"stop", luathread_stop},
 	{"task", luathread_task},
+        {"allow", luathread_allow_signal},
+        {"send", luathread_send_signal},
+        {"sigpending", luathread_signal_pending},
+
```
lneto: I would rather prefix everything or nothing to be consistent.. also, I would name the C function the same

lib/luathread.c
```diff
@@ -175,6 +177,53 @@ static int luathread_task(lua_State *L)
...
+ 	int signum = luaL_checkinteger(L, 2);
+ 	allow_signal(signum);
+	
+ 	return 0;
+}
+
+static int luathread_send_signal(lua_State *L)
+{
+ 	int signum = luaL_checkinteger(L, 2);
+   	lunatik_object_t *object = lunatik_toobject(L, 1);
```
lneto: I think you miss my point on the checker function.. take a look on the code base, you will find that we have a macro ready for this pattern

Physic69: lunatik_object_t *object = lunatik_checkobject(L, 1);
 you are referring to this one right?.

lneto: No.. I'm referring to this https://github.com/luainkernel/lunatik/blob/270622c10aa238a5405f6f1122cd66597dcbe288/lib/luasocket.c#L122

lib/luathread.c
Physic69: Here, do you want me to do smtg similar to https://github.com/Physic69/lunatik/blob/NEWPATCH/lib/lualinux.c#L465
or do you want me to do with enums where I declare enum like
```
 enum signa
{
    SIGHANGUP, SIGINT, SIGTERM, SIGALRM //can add more
}; 
```
And then mapping these to actual signal values with something like:

```
 for (int i = 0; i < sizeof(signal_map) / sizeof(signal_map[0]); i++) {
        if (strcmp(str, signal_map[i].name) == 0) {
            return signal_map[i].value;
        }
    } 
```
signal map is map which maps our enums with original signals,?
Or do you think there is  a better way to implement enums?

lneto: > Here, do you want me to do smtg similar to https://github.com/Physic69/lunatik/blob/NEWPATCH/lib/lualinux.c#L465
or do you want me to do with enums where I declare enum like

we should replicate this.. we can just require("linux") and use that definition on Lua side.. no need to define the same thing multiple times..

> And then mapping these to actual signal values with something like:

sorry, why do we need to map this?

Physic69: ah thanks for pointing out, i forgot to remove that enums.. 
When you said "enum", I  thought of doing an enum {} declaration in C and mapping it, which I later realized wasn’t needed at all, my bad for complicating it.
Later this made me realize that all signals are allowed by default, then i thought of adding block(), as well, does it sound good?

Physic69:
I have added printallowed() just for debugging, but I’m thinking of switching to isallowed(signal) instead to check if a particular signal is allowed. Does that sound good?

lib/luathread.c
lneto: Yup.. sure, go ahead and add block()

lib/luathread.c
```diff
@@ -175,6 +179,111 @@ static int luathread_task(lua_State *L)
...
+   	struct task_struct *task = thread->task;
+	
+	if (!task)
+ 		return luaL_error(L, "thread task is NULL");
+ 	lua_pushboolean(L, signal_pending(task));
+	 
+	return 1;
+}
+
+static int luathread_allowedsignals(lua_State *L)
```
Physic69: Shall I modify this function to isallowed? might be better?

lib/luathread.c
```diff
@@ -175,6 +179,132 @@ static int luathread_task(lua_State *L)
...
+static int luathread_allow(lua_State *L)
+{
+   	luathread_t *thread = luathread_check(L, 1);
+    	struct task_struct *task = thread->task;
+    	if (!task)
+        	return luaL_error(L, "thread task is NULL");
+    
+    	int signum = luaL_checkinteger(L, 2);
+    
+	spin_lock_irq(&task->sighand->siglock);
```
lneto, line 199: is this lock really necessary? can you point to from where you get this pattern?

lib/luathread.c
```diff
@@ -175,6 +179,132 @@ static int luathread_task(lua_State *L)
...
+*/
+static int luathread_send(lua_State *L)
+{
+    	luathread_t *thread = luathread_check(L, 1);
+    	struct task_struct *task = thread->task;
+    	if (!task)
+        	return luaL_error(L, "thread task is NULL");
+    
+    	int signum = luaL_checkinteger(L, 2);
+	if (sigismember(&task->blocked, signum))
```
lneto, line 226: why do we need this? can you provide some documentation?

lib/luathread.c
```diff
@@ -175,6 +179,132 @@ static int luathread_task(lua_State *L)
...
+* @treturn nil
+* @usage
+* t:allow(15)  -- Allow SIGTERM for this thread
+*/
+static int luathread_allow(lua_State *L)
+{
+   	luathread_t *thread = luathread_check(L, 1);
+    	struct task_struct *task = thread->task;
+    	if (!task)
+        	return luaL_error(L, "thread task is NULL");
```
lneto, line 195: see this https://github.com/luainkernel/lunatik/pull/289#discussion_r2178772646

lib/luathread.c
```diff
@@ -175,6 +179,132 @@ static int luathread_task(lua_State *L)
...
+* @usage
+* local a = t:isallowed(15) -- Returns 1 (true) if allowed else 0 (false)
+*/
+static int luathread_isallowed(lua_State *L)
+{
+	luathread_t *thread = luathread_check(L, 1);
+	struct task_struct *task = thread->task;
+	
+	if (!task)
+		return luaL_error(L, "thread task is NULL");
```
lneto, line 295: we should decide if you gonna throw errors or return bool, errno

lib/luathread.c
Physic69: [Linux/signal.c](https://docs.huihoo.com/doxygen/linux/kernel/3.7/kernel_2signal_8c_source.html#l02996)

This lock prevents race condition between concurrent signal senders (we can say signal masks) {safety :)}

lib/luathread.c
Physic69: If we dont check before sending signal, the blocked signal wont be handled properly,  it will be queued. This is what I remember. I will add the source as well (have to look where I read this )

lib/luathread.c
Physic69: Oh,Ill check it again and let you know , but we can change the errors to bool

lib/luathread.c
Physic69: actually I wanted your opnion on it 
        lua_pushinteger(L, -EACCES);  // permission denied (signal blocked) 
        lua_pushinteger(L, -ESRCH);  // no such process (send)
        if (ret) {
        lua_pushboolean(L, false);
        lua_pushinteger(L, ret);
        return 2;  // similar to one in lualinux
    }

we have to write sepeartly for every case like above...
we can use goto like we did lin lulinux

lib/luathread.c
lneto: my impression is that it should be done internally by signal API, not by its clients.. can you point to some API client doing this?

lib/luathread.c
lneto: wasn't it expected? why shouldn't it be queued?

lib/luathread.c
lneto: if goto will simplify the code, let's just use it ;-)

Physic69: Ill update code and let you know

lib/luathread.c
Physic69: Yes , i found some warnings in KBUILD, Ill try to reduce the usage of that. Thanks for letting me know that

lib/luathread.c
Physic69: my bad, we don't need to check this here. Thanks for pointing this out. I mixed two different concepts here 😅,
I just checked the official docs. This is required in synchronous signals, which we are not doing. I'll update it soon

lib/luathread.c
Physic69: [flush_signals](https://docs.huihoo.com/doxygen/linux/kernel/3.7/kernel_2signal_8c_source.html#l00421)
The above function is a API client used by us to flush and pending signals and takes lock manually before manipulating them
Sorry for taking too much of time to respond, 
I was searching for other examples...

lneto: it's the internal API implementation.. I meant a client from this API.. I think you are mixing the internal implementation with the API usage..

lneto: I think you should just use this.. https://elixir.bootlin.com/linux/v6.14.7/source/include/linux/signal.h#L301

lneto: hmm.. I see the problem, you want to allow/block a signal on a specific task.. I think we shouldn't go to this path.. the common pattern as I see is to allow/block signal in the current task.. so such functions would probably belong to lualinux functions and not to luathread methods

Physic69: that makes sense now.
You are right, I was trying to manipulate signals on a specific task, 
ok ill move these into Lualinux . instead and stick to using the standard allow_signal() API for the current task.
Ill make a new updated PR for that, Shall I close this one?

lneto: I think you still need send() and pending(), right?

Physic69: Yes, you're right. I’ll check once send() and pending() in threads to make sure everything is good.
For now, I’ll go ahead  
and move allow(), isallowed() and block() to lualinux
I’ll update everything properly and move on from there. Sounds good?

lneto: Yup.. thanks!

Physic69:
Will open a new PR regarding this, with some example

PR #278, opened 2025-06-18
Fix fib api change in kernel >= 6.15

jperon:
Fib api has changed in kernel 6.15 :

```diff
-int fib_nl_newrule(struct sk_buff *skb, struct nlmsghdr *nlh,
-		  struct netlink_ext_ack *extack);
-int fib_nl_delrule(struct sk_buff *skb, struct nlmsghdr *nlh,
-		  struct netlink_ext_ack *extack);
+int fib_newrule(struct net *net, struct sk_buff *skb, struct nlmsghdr *nlh,
+		struct netlink_ext_ack *extack, bool rtnl_held);
+int fib_delrule(struct net *net, struct sk_buff *skb, struct nlmsghdr *nlh,
+		struct netlink_ext_ack *extack, bool rtnl_held);
```

This PR intends to fix it.

@lneto I’ve no real idea how to test it, so may you please have a glance ?


lib/luafib.c
```diff
@@ -30,7 +31,11 @@
 
 typedef struct luafib_rule_s {
 	struct net *net;
+	#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0)
```
lneto:
```suggestion
#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0)
```

lib/luafib.c
```diff
@@ -30,7 +31,11 @@
...
+	int (*command)(struct net *, struct sk_buff *, struct nlmsghdr *, struct netlink_ext_ack *, bool);
+	#else
```
lneto:
```suggestion
#else
```

lib/luafib.c
```diff
@@ -30,7 +31,11 @@
...
 	int (*command)(struct sk_buff *, struct nlmsghdr *, struct netlink_ext_ack *);
+	#endif
```
lneto:
```suggestion
#endif
```

lib/luafib.c
```diff
@@ -60,20 +65,30 @@ static int luafib_nl_rule(luafib_rule_t *rule)
 	nlmsg_end(skb, nlh);
 
 	skb->sk = rule->net->rtnl;
+	#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0)
```
lneto:
```suggestion
#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0)
```

lib/luafib.c
```diff
@@ -60,20 +65,30 @@ static int luafib_nl_rule(luafib_rule_t *rule)
...
+	ret = rule->command(rule->net, skb, nlh, NULL, true);
+	#else
```
lneto:
```suggestion
#else
```

lib/luafib.c
```diff
@@ -60,20 +65,30 @@ static int luafib_nl_rule(luafib_rule_t *rule)
...
 	ret = rule->command(skb, nlh, NULL);
+	#endif
```
lneto:
```suggestion
#endif
```

lib/luafib.c
```diff
@@ -60,20 +65,30 @@ static int luafib_nl_rule(luafib_rule_t *rule)
...
+#endif
+
 #define LUAFIB_OPRULE(op) 				\
 static int luafib_##op(lua_State *L)			\
 {							\
 	luafib_rule_t rule;				\
 							\
 	rule.net = &init_net;				\
-	rule.command = fib_nl_##op;			\
+	rule.command = LUAFIB_RULECOMMAND(op);			\
```
lneto:
```suggestion
	rule.command = LUAFIB_RULECOMMAND(op);		\
```

lib/luafib.c
```diff
@@ -60,20 +65,30 @@ static int luafib_nl_rule(luafib_rule_t *rule)
...
 	nlmsg_free(skb);
 out:
 	return ret;
 }
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0)
+#define LUAFIB_RULECOMMAND(op) fib_##op
+#else
+#define LUAFIB_RULECOMMAND(op) fib_nl_##op
+#endif
```
lneto, line 83:
```suggestion
#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0)
#define LUAFIB_PREFIX		fib
#else
#define LUAFIB_PREFIX		fib_nl
#endif
```

it's completely up to you.. but I think this version is simpler ;-)

jperon: I tried, but then I can’t get to find the right syntax to concatenate it with `op` within `#define LUAFIB_OPRULE(op)`.

lneto: you're absolutely right.. to do this macro concat we would need some dirty cpp tricks..  it would end up being unnecessarily complex.. thanks

lneto (approved):

PR #280, opened 2025-06-22
Add lualinux_kill() 

Physic69:
Add new function kill, takes pid as input and terminates the process for given pid

### DIFF
--- a/lib/lualinux.c
+++ b/lib/lualinux.c
@@ -21,6 +21,10 @@
 #include <linux/ktime.h>
 #include <linux/netdevice.h>
 #include <linux/byteorder/generic.h>
+#include <linux/sched/signal.h>
+#include <linux/pid.h> 
+#include <linux/signal.h>
+#include <linux/errno.h>
 
 #include <lua.h>
 #include <lauxlib.h>
@@ -111,6 +115,47 @@ static int lualinux_schedule(lua_State *L)
 
 	lua_pushinteger(L, jiffies_to_msecs(schedule_timeout(timeout)));
 	return 1;
+}
+
+/***
+* Kills a process by sending a signal.
+* By default, sends SIGKILL.
+* An optional second argument can specify a different signal (either by number or by using the constants from `linux.signal`).
+*
+* @function kill
+* @tparam integer pid Process ID to kill.
+* @tparam[opt] integer sig Signal number to send (default: `linux.signal.KILL`).
+* @treturn boolean `true` if the signal was sent successfully.
+* @treturn[error] boolean `false` followed by an error number if the operation fails.
+* @raise Errors:
+*   - (3): The specified PID doesn't exist
+*   - other errno values depending on the failure cause (e.g., `EPERM`, `EINVAL`, etc.)
+* @usage
+*   linux.kill(1234)  -- Kill process 1234 with SIGKILL (default)
+*   linux.kill(1234, linux.signal.TERM)  -- Kill process 1234 with SIGTERM
+*/
+static int lualinux_kill(lua_State *L)
+{
+	pid_t nr = (pid_t)luaL_checkinteger(L, 1);
+  	int sig = luaL_optinteger(L, 2, SIGKILL);
+ 	struct pid *pid = find_get_pid(nr);
+
+	int ret = ESRCH;
+	if (pid == NULL) 
+        	goto err;    
+	
+    	ret = kill_pid(pid, sig, 1);
+    	put_pid(pid);
+    
+    	if (ret) 
+		goto err;
+    
+    	lua_pushboolean(L, true);
+    	return 1;
+err:
+	lua_pushboolean(L, false);
+	lua_pushinteger(L, ret);
+	return 2;
 }
 
 /***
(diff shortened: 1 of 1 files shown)


lib/lualinux.c
```diff
@@ -564,4 +594,3 @@ module_init(lualinux_init);
 module_exit(lualinux_exit);
 MODULE_LICENSE("Dual MIT/GPL");
 MODULE_AUTHOR("Lourival Vieira Neto <lourival.neto@ring-0.io>");
-
```
lneto: Please, keep \n at the end

lib/lualinux.c
```diff
@@ -113,6 +115,33 @@ static int lualinux_schedule(lua_State *L)
...
+* @tparam integer pid Process ID to kill.
+* @treturn boolean `true` if successful.
+* @raise Error if process not found or kill fails.
+* @usage 
+* linux.kill(1234) 
+*/
+static int lualinux_kill(lua_State *L)
+{
+    pid_t pid = (pid_t)luaL_checkinteger(L, 1);
+    struct pid *pid_struct = find_get_pid(pid);
```
```suggestion
	pid_t nr = (pid_t)luaL_checkinteger(L, 1);
	struct pid *pid = find_get_pid(nr);
```

lib/lualinux.c
```diff
@@ -113,6 +115,33 @@ static int lualinux_schedule(lua_State *L)
 	return 1;
 }
 
+/***
+* Kills a process by sending SIGKILL to pid.
```
lneto: I would use SIGKILL as default and use a optional argument for overriding it.. what do you think?

lib/lualinux.c
```diff
@@ -113,6 +115,33 @@ static int lualinux_schedule(lua_State *L)
...
+* @function kill
+* @tparam integer pid Process ID to kill.
+* @treturn boolean `true` if successful.
```
lneto: if you raise an error, it doesn't make much sense to return true on success.. we should decide if it must raise an error or return false on failure.. what do you think?

Physic69: I thought this approach might be better since it can help the user understand exactly which error occurred. there were 3-4 values of ret other than zero.
We can change the error part if you feel that its not that of use

lib/lualinux.c
Physic69: Sounds good, but a different function for it might be better
if someone wants to pass some signal, they can just use different function mentioned above eg: lualinux_sig(pid, sig).
Just feels better than overloading a function. Simplicity :)

lib/lualinux.c
lneto: perhaps it's better to return true on success or false, err on failure.. what do you think?

Physic69: yup i agree. This seems better

lib/lualinux.c
lneto: hm.. I disagree.. IMO, it would be way simpler to just have a optinteger call (such as on linux.schedule).. also the kill API in linux is not meant for SIGKILL only.. for me, the best approach would be to have the signals exported as a lunatik_flag.. something like..

```
local thread = require("thread")

local sig = linux.signal

local function psycho_killer()
  linux.kill(linux.random()) -- send SIGKILL to random PID
end

local function suicidal_tendencies()
  linux.kill(thread.current(), sig.TERM) -- send SIGTERM to the current thread
end
```

Physic69: Okay,
Ill try to implement it and reach out to you if I find any difficulties.

lib/lualinux.c
```diff
@@ -113,6 +116,46 @@ static int lualinux_schedule(lua_State *L)
...
+        	lua_pushboolean(L, false);
+        	lua_pushstring(L, "process not found");
+        	return 2;
+    	}	    
+    	int ret = kill_pid(pid, sig, 0);
+    	put_pid(pid);
+    
+    	if (ret) {
+		lua_pushboolean(L, false);
+        	return luaL_error(L, "kill failed with error %d", ret);
```
lneto: You should just `pushinteger(L, ret)` here..

lib/lualinux.c
```diff
@@ -113,6 +116,46 @@ static int lualinux_schedule(lua_State *L)
...
+	if (pid == NULL) {
+        	lua_pushboolean(L, false);
+        	lua_pushstring(L, "process not found");
+        	return 2;
+    	}	    
+    	int ret = kill_pid(pid, sig, 0);
+    	put_pid(pid);
+    
+    	if (ret) {
+		lua_pushinteger(L, ret);
```
lneto: I think you missed my point here.. the idea is to `return false, <integer>` instead of raising an error.. not to `return <integer>`, right?

Physic69: Ah, got it now, thanks for clarifying. I misunderstood earlier.
makes sense.

lib/lualinux.c
```diff
@@ -381,10 +425,84 @@ static const lunatik_reg_t lualinux_errno[] = {
...
+*   @tfield integer VTALRM SIGVTALRM (26) - Virtual alarm clock
+*   @tfield integer PROF SIGPROF (27) - Profiling alarm clock
+*   @tfield integer WINCH SIGWINCH (28) - Window size change
+*   @tfield integer IO   SIGIO (29) - I/O now possible
+*   @tfield integer PWR  SIGPWR (30) - Power failure
+*   @tfield integer SYS  SIGSYS (31) - Bad system call
+*/
+static const lunatik_reg_t lualinux_signal[] = {
+    	{ "HUP",    SIGHUP  },
+    	{ "INT",    SIGINT  },
```
lneto: Please mind our code style.. always replicate the style that is already there

lib/lualinux.c
```diff
@@ -113,6 +116,47 @@ static int lualinux_schedule(lua_State *L)
...
+*/
+static int lualinux_kill(lua_State *L)
+{
+	pid_t nr = (pid_t)luaL_checkinteger(L, 1);
+  	int sig = luaL_optinteger(L, 2, SIGKILL);
+ 	struct pid *pid = find_get_pid(nr);
+
+	if (pid == NULL) {
+        	lua_pushboolean(L, false);
+        	lua_pushstring(L, "process not found");
```
lneto: I'm not sure about mixing integer and string in the error message.. I think it would be confusing for the API's client.. the way I see we should use it would be something like..

```
local ok, err = linux.kill(pid)
if not ok then
  If err == errno.X then
   ...
  end
end
```

I'd like to hear @jperon's input on this as well.. thanks!

lib/lualinux.c
Physic69: my bad, sorry about that! I’ll follow the existing code style carefully and update it accordingly. thanks for pointing it out

lib/lualinux.c
Physic69: That makes sense. I added the string mainly to make it easier for us during testing, but I agree it might be inconsistent for actual use in scripts.
Maybe we can push integer like we did with ret, which is thrown by kill_pid normally i.e., -3 (-ESRCH) when a process is not found

lneto: In this case, we push ESRCH then.. we are not using negative for this

lneto: You can move the error case to the end of the function and `goto err`, so we don't need to replicate the code across your func

Physic69: Got it.
Thanks for the suggestion. Let me know if anything else looks off.

lib/lualinux.c
```diff
@@ -113,6 +116,47 @@ static int lualinux_schedule(lua_State *L)
...
+*   linux.kill(1234)  -- Kill process 1234 with SIGKILL (default)
+*   linux.kill(1234, linux.signal.TERM)  -- Kill process 1234 with SIGTERM
+*/
+static int lualinux_kill(lua_State *L)
+{
+	pid_t nr = (pid_t)luaL_checkinteger(L, 1);
+  	int sig = luaL_optinteger(L, 2, SIGKILL);
+ 	struct pid *pid = find_get_pid(nr);
+
+	int ret = 3;
```
lneto: please, use the actual define/enum instead of "magic" numbers

lib/lualinux.c
```diff
@@ -113,6 +117,47 @@ static int lualinux_schedule(lua_State *L)
...
+{
+	pid_t nr = (pid_t)luaL_checkinteger(L, 1);
+  	int sig = luaL_optinteger(L, 2, SIGKILL);
+ 	struct pid *pid = find_get_pid(nr);
+
+	int ret = ESRCH;
+	if (pid == NULL) 
+        	goto err;    
+	
+    	ret = kill_pid(pid, sig, 0);
```
lneto: Why using 0 here? I found that many kernel modules use 1..

Physic69: Ah, yes. 
using 1 skips permission checks (similar to using sudo), which can be dangerous. It would allow the script to send signals to any process, including critical kernel tasks, potentially destabilizing the system. (danger)
sticking with 0 ensures that standard permission checks are enforced, making it safer

Physic69: Would it make sense to support an optional third argument to bypass permission checks (1 as input)?

lneto: Thing is.. we are running at the kernel level.. we already have the permissions at this point.. if we want to restrict this call I think we should use sandboxing techniques.. I would use the same pattern as other kernel modules and call it with `1` here

Physic69: Maybe we can rely on custom policies in lua for sandboxing for now, like defining can_kill rules per use case.
```
local policy = {
  can_kill = function(pid)
    return pid ~= 1 and pid ~= 2 
  end
}
```
Here is a basic technique which skips init and system processes 
 And I agree, using 1 as you said totally makes sense here. Thanks, that clears it up.

lneto (approved):
I think we are good to go.. can you squash your commits? Then I can merge it

lneto:
@Physic69 sorry, quite a minor in your commit message.. it should be `lualinux_kill` or `linux.kill`, not `lualinux.kill` ;-)

PR #281, opened 2025-06-24
add dynamic input support for id_table

qrsikno2:
This pull request introduces enhancements to the `lib/luahid.c` file, focusing on memory management, Lua table parsing, and improving code clarity. The most significant changes include adding a helper function to parse integer fields from Lua tables, implementing a function to parse `id_table` entries, and updating memory cleanup logic in the `luahid_release` function.

### Lua table parsing improvements:
* Added a helper function `get_int_field` to safely retrieve integer fields from Lua tables with a default value fallback.
* Implemented the `luahid_parse_id_table` function to parse the `id_table` field from Lua tables, dynamically allocate memory for HID device IDs, and validate table entries.

### Memory management updates:
* Updated the `luahid_release` function to free the `id_table` field in the HID driver if it is not `NULL`, ensuring proper cleanup of dynamically allocated memory.

### DIFF
--- a/config.ld
+++ b/config.ld
@@ -19,6 +19,7 @@ file = {
     './lib/luadevice.c',
     './lib/luafib.c',
     './lib/luafifo.c',
+    './lib/luahid.c',
     './lunatik_core.c',
     './lib/lunatik/runner.lua',
     './lib/lualinux.c',
--- a/lib/luahid.c
+++ b/lib/luahid.c
@@ -3,22 +3,34 @@
 * SPDX-License-Identifier: MIT OR GPL-2.0-only
 */
 
-#include <string.h>
+/***
+* High-level Lua interface to the Linux HID subsystem.
+* This module allows registering Lua table with functions as a HID driver,
+* support for id_table is provided, which allows matching HID devices
+*
+* @module hid
+*/
+
 #define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+#include <string.h>
 #include <linux/hid.h>
 #include <lunatik.h>
+#include <lauxlib.h>
 
+/***
+* Represents a registered HID driver.
+* This is a userdata object returned by `hid.register()`. It encapsulates
+* the kernel `struct hid_driver` and associated Lunatik runtime information
+* necessary to invoke the Lua callback when a HID device is matched.
+* The `registered` field indicates whether the driver is currently registered
+* @type hid_driver
+*/
 typedef struct luahid_s {
 	lunatik_object_t *runtime;
 	struct hid_driver driver;
 	bool registered;
 } luahid_t;
 
-static const struct hid_device_id luahid_table[] = {
-	{HID_DEVICE(HID_BUS_ANY, HID_GROUP_ANY, HID_ANY_ID, HID_ANY_ID)},
-	{ }
-};
-MODULE_DEVICE_TABLE(hid, luahid_table);
 
 static void luahid_release(void *private)
 {
@@ -27,6 +39,8 @@ static void luahid_release(void *private)
 		hid_unregister_driver(&hid->driver);
 	if (hid->runtime != NULL)
 		lunatik_putobject(hid->runtime);
+	if (hid->driver.id_table != NULL)
+		lunatik_free(hid->driver.id_table);
 	if (hid->driver.name != NULL)
 		lunatik_free(hid->driver.name);
 }
@@ -50,6 +64,64 @@ static const lunatik_class_t luahid_class = {
 	.sleep = true,
 };
 
+static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
+{
+	size_t len = luaL_len(L, idx);
+	struct hid_device_id *user_table = lunatik_checkalloc(L, sizeof(struct hid_device_id) * (len + 1));
+
+	struct hid_device_id *cur_id = user_table;
+	for (size_t i = 0; i < len; i++, cur_id++) {
+		if (lua_geti(L, idx, i + 1) != LUA_TTABLE) { /* table entry */
+			lua_pop(L, 1); /* table entry */
+			lunatik_free(user_table);
+			user_table = NULL;
+			goto out;
+		}
+
+		lunatik_optinteger(L, -1, cur_id, bus, HID_BUS_ANY);
+		lunatik_optinteger(L, -1, cur_id, group, HID_GROUP_ANY);
+		lunatik_optinteger(L, -1, cur_id, vendor, HID_ANY_ID);
+		lunatik_optinteger(L, -1, cur_id, product, HID_ANY_ID);
+		lunatik_optinteger(L, -1, cur_id, driver_data, 0);
+
+		lua_pop(L, 1); /* table entry */
+	}
+
+	memset(cur_id, 0, sizeof(struct hid_device_id));
+out:
+	lua_pop(L, 1); /* id_table */
+	return user_table;
+}
+
+/***
+* Registers a new HID driver.
+* This function creates a new HID driver object from a Lua table and registers it with the kernel.
+* The Lua table must contain the following fields:
+*
+* - `name`: The name of the driver (string).
+* - `id_table`: A table of device IDs that this driver can match against (lua_table).
+* 	- each id is a table consisting of fields:
+*
+* 		- `bus`: The bus type (integer, default: `HID_BUS_ANY`).
+* 		- `group`: The group type (integer, default: `HID_GROUP_ANY`).
+* 		- `vendor`: The vendor ID (integer, default: `HID_ANY_ID`).
+* 		- `product`: The product ID (integer, default: `HID_ANY_ID`).
+* 		- `driver_data`: Additional driver data (integer, default: `0`).
+*
+* @function hid.register
+* @tparam table table The Lua table containing driver information.
+* @treturn hid_driver The registered HID driver object.
+* @usage
+*	 local hid = require("hid")
+*	 local hid_driver = {
+*	 	name = "my_hid_driver",
+*	 	id_table = {
+*	 		{ bus = 1, vendor = 0x1234, product = 0x5678 },
+*	 		{ bus = 2, vendor = 0x4321, product = 0x8765 },
+*	 	}
+*	 }
+*	 hid.register(hid_driver)
+*/
 static int luahid_register(lua_State *L)
 {
 	luaL_checktype(L, 1, LUA_TTABLE);
@@ -61,7 +133,10 @@ static int luahid_register(lua_State *L)
 	struct hid_driver *user_driver = &(hid->driver);
 	user_driver->name = lunatik_checkalloc(L, NAME_MAX);
 	lunatik_setstring(L, 1, user_driver, name, NAME_MAX);
-	user_driver->id_table = luahid_table;
+
+	lunatik_checkfield(L, 1, "id_table", LUA_TTABLE);
+	luaL_argcheck(L, (user_driver->id_table = luahid_setidtable(L, -1)) != NULL,
+			   2, "invaild id_table");
 
 	lunatik_setruntime(L, hid, hid);
 	lunatik_getobject(hid->runtime);


lib/luahid.c
```diff
@@ -50,6 +52,53 @@ static const lunatik_class_t luahid_class = {
 	.sleep = true,
 };
 
+/*
+ * Helper function to safely get an integer field from a Lua table.
+ */
+static lua_Integer luahid_get_int_field(lua_State *L, int table_idx, const char *field_name, lua_Integer default_val)
```
lneto: For the sake of consistency, I would move this getter to lunatik.h in the same way we do the setter

lib/luahid.c
```diff
@@ -50,6 +52,53 @@ static const lunatik_class_t luahid_class = {
...
+		user_table[i].bus = luahid_get_int_field(L, -1, "bus", HID_BUS_ANY);
+		user_table[i].group = luahid_get_int_field(L, -1, "group", HID_GROUP_ANY);
+		user_table[i].vendor = luahid_get_int_field(L, -1, "vendor", HID_ANY_ID);
+		user_table[i].product = luahid_get_int_field(L, -1, "product", HID_ANY_ID);
+		user_table[i].driver_data = luahid_get_int_field(L, -1, "driver_data", 0);
+
+		lua_pop(L, 1);
+	}
+
+	memset(&user_table[len], 0, sizeof(struct hid_device_id));
```
lneto: Why setting it to zero here?

lib/luahid.c
```diff
@@ -50,6 +52,53 @@ static const lunatik_class_t luahid_class = {
...
+	lua_pop(L, 1);
```
lneto: What are you poping here?

lib/luahid.c
qrsikno2: Is this what you mean? I thought we were going to use `optinteger`
 and those macros.

lib/luahid.c
qrsikno2: most of the driver will set it to zero by not to define it and use the default `0`.

lib/luahid.c
qrsikno2: i guess it was the `id_table` pushed at first.
```c
	if (lua_getfield(L, idx, "id_table") != LUA_TTABLE)
		goto err;
```

lib/luahid.c
lneto: ah.. yeah.. I misread this.. so, you need to rebase, right?

lib/luahid.c
lneto: you mean the last element of this table is expected to be zero? can you point to an example on the linux code base?

lib/luahid.c
lneto: please, don't guess.. be sure of what you are removing from the Lua stack and add a comment.. look form lua_pop comments in our code base.. there are plenty of examples..

lib/luahid.c
```diff
@@ -50,6 +52,53 @@ static const lunatik_class_t luahid_class = {
...
+static lua_Integer luahid_get_int_field(lua_State *L, int table_idx, const char *field_name, lua_Integer default_val)
+{
+	lua_Integer result = default_val;
+	if (lua_getfield(L, table_idx, field_name) == LUA_TNUMBER)
+		result = lua_tointeger(L, -1);
+	lua_pop(L, 1);
+	return result;
+}
+
+static const struct hid_device_id *luahid_parse_id_table(lua_State *L, int idx)
```
lneto: I'm not fond of long names like this.. I think `luahid_setidtable` would be better.. it would be more harmonic to lunatik API

lib/luahid.c

lib/luahid.c

lib/luahid.c
qrsikno2: deleted the funcs

lib/luahid.c
qrsikno2: About the first one need to pay attn is the table is usually `static`. and a not initialized struct member with attr `static` will be given the value '0'

qrsikno2: by the way you see how a normal driver use the `driver_data`, it defines like:
https://elixir.bootlin.com/linux/v6.13.7/source/drivers/hid/hid-microsoft.c#L459
and used as:
https://elixir.bootlin.com/linux/v6.13.7/source/drivers/hid/hid-microsoft.c#L382
the `quirks` is the driver_data. which allows the driver to produce the quirks. 
it use the operator `&` to identify the quirks just like checking some bits.
so the default value is 0 which means it has no quirks

lib/luahid.c
lneto: which funcs?

lib/luahid.c
```diff
@@ -50,6 +52,42 @@ static const lunatik_class_t luahid_class = {
...
+		}
+
+		struct hid_device_id *cur_id = &user_table[i];
+		lunatik_optinteger(L, 1, cur_id, bus, HID_BUS_ANY);
+		lunatik_optinteger(L, 1, cur_id, group, HID_GROUP_ANY);
+		lunatik_optinteger(L, 1, cur_id, vendor, HID_ANY_ID);
+		lunatik_optinteger(L, 1, cur_id, product, HID_ANY_ID);
+		lunatik_optinteger(L, 1, cur_id, driver_data, 0);
+
+		lua_pop(L, 1); /* pop the table entry */
```
```suggestion
		lua_pop(L, 1); /* table entry */
```

lneto: let's keep code and comments less verbose..

lib/luahid.c
```diff
@@ -50,6 +52,42 @@ static const lunatik_class_t luahid_class = {
...
+	}
+
+	memset(&user_table[len], 0, sizeof(struct hid_device_id));
+	lua_pop(L, 1); /* pop the id_table table */
+
+	return user_table;
+err:
+	lua_pop(L, 1);
+	return luahid_table;
```
lneto: why doing this?

lib/luahid.c
qrsikno2: changed name. misunderstanding for your comment, sorry.

lib/luahid.c
qrsikno2: return a table with infos and pop the idtable to make the stack balance.

lib/luahid.c

lib/luahid.c
lneto: let's get rid of the static table.. if you can't load the user defined one return NULL

lib/luahid.c
```diff
@@ -50,6 +52,42 @@ static const lunatik_class_t luahid_class = {
...
+	size_t len = luaL_len(L, -1);
+	if (len == 0)
+		goto err;
+
+	struct hid_device_id *user_table = lunatik_checkalloc(L, sizeof(struct hid_device_id) * (len + 1));
+
+	for (size_t i = 0; i < len; i++) {
+		if (lua_geti(L, -1, i + 1) != LUA_TTABLE) { /* i-th entry of the table */
+			lunatik_free(user_table);
+			luaL_error(L, "id_table entry #%zu is not a table", i + 1);
```
lneto: I'm not sure about raising error in here.. I think it's make harder to read the register function.. perhaps it's better to return NULL and raise an error on register() instead.. I would like also to hear @sheharyaar's inputs on this..

lib/luahid.c

lib/luahid.c
qrsikno2: done. but i comment original codes for @sheharyaar to read & comment.

if the review is done, i would delete it. Plz don't be angry about that.

lib/luahid.c
```diff
@@ -50,6 +46,45 @@ static const lunatik_class_t luahid_class = {
...
+		goto err;
+
+	struct hid_device_id *user_table = lunatik_checkalloc(L, sizeof(struct hid_device_id) * (len + 1));
+
+	for (size_t i = 0; i < len; i++) {
+		if (lua_geti(L, -1, i + 1) != LUA_TTABLE) { /* i-th entry of the table */
+			/*
+			 * lunatik_free(user_table);
+			 * luaL_error(L, "id_table entry #%zu is not a table", i + 1);
+			 */
```
lneto: Please, don't do this if it's not a draft.. you shouldn't be asking to merge commented out code

lib/luahid.c
```diff
@@ -50,6 +46,45 @@ static const lunatik_class_t luahid_class = {
...
+		lunatik_optinteger(L, 1, cur_id, product, HID_ANY_ID);
+		lunatik_optinteger(L, 1, cur_id, driver_data, 0);
+
+		lua_pop(L, 1); /* table entry */
+	}
+
+	memset(&user_table[len], 0, sizeof(struct hid_device_id));
+	lua_pop(L, 1); /* id_table */
+
+	return user_table;
```
lneto, line 93: Why not initializing it with NULL, so we can avoid the last two lines, avoiding unnecessarily replicating code

lib/luahid.c
```diff
@@ -50,6 +46,45 @@ static const lunatik_class_t luahid_class = {
...
+err:
```
lneto: you are also leaking memory in this case.. please, be more careful..

lib/luahid.c

lib/luahid.c
qrsikno2: done

lib/luahid.c

lib/luahid.c
```diff
@@ -50,6 +46,42 @@ static const lunatik_class_t luahid_class = {
...
+		lunatik_optinteger(L, 1, cur_id, driver_data, 0);
+
+		lua_pop(L, 1); /* table entry */
+	}
+
+	memset(&user_table[len], 0, sizeof(struct hid_device_id));
+err:
+	lua_pop(L, 1); /* id_table */
+
+	return user_table;
```
```suggestion
out:
	lua_pop(L, 1); /* id_table */
	return user_table;
```

lib/luahid.c
```diff
@@ -50,6 +46,42 @@ static const lunatik_class_t luahid_class = {
...
+	if (lua_getfield(L, idx, "id_table") != LUA_TTABLE) /* id_table */
+		goto err;
+
+	size_t len = luaL_len(L, -1);
+	if (len == 0)
+		goto err;
+
+	user_table = lunatik_checkalloc(L, sizeof(struct hid_device_id) * (len + 1));
+	for (size_t i = 0; i < len; i++) {
+		if (lua_geti(L, -1, i + 1) != LUA_TTABLE) { /* i-th entry of the table */
```
```suggestion
		if (lua_geti(L, -1, i + 1) != LUA_TTABLE) { /* table entry */
```

lib/luahid.c
```diff
@@ -50,6 +46,42 @@ static const lunatik_class_t luahid_class = {
...
 };
 
+static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
+{
+	struct hid_device_id *user_table = NULL;
+	if (lua_getfield(L, idx, "id_table") != LUA_TTABLE) /* id_table */
+		goto err;
+
+	size_t len = luaL_len(L, -1);
+	if (len == 0)
```
lneto: why do you need that?

lib/luahid.c
```diff
@@ -50,6 +46,42 @@ static const lunatik_class_t luahid_class = {
...
+	user_table = lunatik_checkalloc(L, sizeof(struct hid_device_id) * (len + 1));
+	for (size_t i = 0; i < len; i++) {
+		if (lua_geti(L, -1, i + 1) != LUA_TTABLE) { /* i-th entry of the table */
+			lua_pop(L, 1); /* table entry */
+			lunatik_free(user_table);
+			user_table = NULL;
+			goto err;
+		}
+
+		struct hid_device_id *cur_id = &user_table[i];
```
lneto: you can update cur_id in the for incremental step.. I think it will be clearer..

lib/luahid.c
```diff
@@ -61,7 +93,10 @@ static int luahid_register(lua_State *L)
 	struct hid_driver *user_driver = &(hid->driver);
 	user_driver->name = lunatik_checkalloc(L, NAME_MAX);
 	lunatik_setstring(L, 1, user_driver, name, NAME_MAX);
-	user_driver->id_table = luahid_table;
+
+	user_driver->id_table = luahid_setidtable(L, 1);
+	if (user_driver->id_table == NULL)
```
```suggestion
	if ((user_driver->id_table = luahid_setidtable(L, 1)) == NULL)
```

lib/luahid.c
```diff
@@ -61,7 +93,10 @@ static int luahid_register(lua_State *L)
...
+		luaL_error(L, "id_table is not specified or invalid");
```
lneto: perhaps we should use checkarg for that to indicate it's an invalid argument

lib/luahid.c
qrsikno2: what doyou mean about the "incremental step"?

lib/luahid.c
qrsikno2: what doyou mean? what's checkarg

lib/luahid.c

lib/luahid.c

lib/luahid.c

lib/luahid.c

lib/luahid.c
qrsikno2: such as luaL_checkinteger?

lneto: Grep our code base and the Lua manual.. if you don't find it, let me know..

lib/luahid.c
```diff
@@ -50,6 +46,39 @@ static const lunatik_class_t luahid_class = {
 	.sleep = true,
 };
 
+static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
+{
+	struct hid_device_id *user_table = NULL;
+	if (lua_getfield(L, idx, "id_table") != LUA_TTABLE) /* id_table */
```
lneto: I think it will simplify your code if you check for idx type before calling this function

lib/luahid.c
qrsikno2: do you mean this?
```c
	/* random integer in the interval [low, up] */
	luaL_argcheck(L, low <= up, 1, "interval is empty");
	luaL_argcheck(L, low >= 0 || up <= LUA_MAXINTEGER + low, 1, "interval too large");
```

lib/luahid.c

lib/luahid.c
lneto: yeah

lib/luahid.c
```diff
@@ -50,6 +46,39 @@ static const lunatik_class_t luahid_class = {
...
+		lunatik_optinteger(L, 1, cur_id, bus, HID_BUS_ANY);
+		lunatik_optinteger(L, 1, cur_id, group, HID_GROUP_ANY);
+		lunatik_optinteger(L, 1, cur_id, vendor, HID_ANY_ID);
+		lunatik_optinteger(L, 1, cur_id, product, HID_ANY_ID);
+		lunatik_optinteger(L, 1, cur_id, driver_data, 0);
+
+		lua_pop(L, 1); /* table entry */
+	}
+
+	memset(&user_table[len], 0, sizeof(struct hid_device_id));
```
lneto: you can just use `cur_id` here, right?

qrsikno2: i disagree with it . i think the len is better

qrsikno2: as a loop var, the cur_id may changed due to the method of increase. it would be dangerous.

lneto: how? how would it reach out to this line without ending the loop normally? can you give an actual case?

lib/luahid.c

lib/luahid.c
qrsikno2: if you write `for (...; cur_id++ < start_ptr + len; ...)`
vs.
`for (...; ++cur_id < start_ptr + len; ...)`

lneto: I mean in your actual code..

lneto: `i` will have the same effect.. I don't follow your point.. my point is that you have a well defined loop.. just use the control variable for that..

qrsikno2: it's normal to put the loop var into the condition, and you can also simplify the cond with ++.
i think's danger for use the loop var outside the loop.
on one hand, it will be recycled in many programming language.
on the other hand, you may not use the var as it's meaning is ambiguous because of outside of the loop.

and if you want simplify it, i can also remove the loop var i, and use the ptr as the loop var. 
Why not? we should simplify it completely rather than give up halfway.
and if you're not going to simplify it, i prefer to continue to use the len. static is reliable than dynamic.

lib/luahid.c
lneto: It's not, right? I meant it for the pattern above if-then-error

lib/luahid.c
lneto: @sheharyaar can you share your thoughts on this discussion?

lib/luahid.c
```diff
@@ -61,7 +87,10 @@ static int luahid_register(lua_State *L)
 	struct hid_driver *user_driver = &(hid->driver);
 	user_driver->name = lunatik_checkalloc(L, NAME_MAX);
 	lunatik_setstring(L, 1, user_driver, name, NAME_MAX);
-	user_driver->id_table = luahid_table;
+
+	luaL_argcheck(L, lua_getfield(L, 1, "id_table") == LUA_TTABLE, 1, "invalid 'id_table'");
```
sheharyaar: Why not simply use `lunatik_checkfield()` ?

```
```
- `lunatik_checkfield` calls `lua_getfield` and if the type does not match, calls `luaL_error`
- `lua_getfield` documentation: https://www.lua.org/manual/5.4/manual.html#lua_getfield

And since `LUA_TTABLE` is a lua defined type under (`lunatik/lua/lua.h`) it should work
```
#define LUA_TNONE		(-1)
#define LUA_TNIL		0
#define LUA_TBOOLEAN		1
#define LUA_TLIGHTUSERDATA	2
#define LUA_TNUMBER		3
#define LUA_TSTRING		4
#define LUA_TTABLE		5
#define LUA_TFUNCTION		6
#define LUA_TUSERDATA		7
#define LUA_TTHREAD		8
#define LUA_NUMTYPES		9
```

lib/luahid.c
```diff
@@ -50,6 +47,35 @@ static const lunatik_class_t luahid_class = {
...
+	struct hid_device_id *cur_id = user_table;
+	for (size_t i = 0; i < len; i++, cur_id++) {
+		if (lua_geti(L, -1, i + 1) != LUA_TTABLE) { /* table entry */
+			lua_pop(L, 1); /* table entry */
+			lunatik_free(user_table);
+			user_table = NULL;
+			goto out;
+		}
+
+		lunatik_optinteger(L, 1, cur_id, bus, HID_BUS_ANY);
```
sheharyaar: Refer to the discussion on `lunatik` channel. This should be relative indexes.

lib/luahid.c
sheharyaar: I don't think Lourival is angry. Usually we do not see the person behind the screen, but rest assured, he is a very cheerful person. Don't worry. Since, this is a kernel module we are working on, and it can get complex. It gets better to follow a correct style.

sheharyaar: Yes, its better to return NULL and check for error. Makes it easier to read the code.

sheharyaar:
Before merging this, it needs to be fixed. The code is still buggy, I got errors related to `attempting to index nil value`, due to some stack operation error. I have attached the test lua script in the lunatik chat.

CC: @lneto @qrsikno2

lib/luahid.c
sheharyaar: Why are we `memset` -ing this to `0` ? Why not just set this to NULL ?

`id_table` is used here, which runs in a loop until NULL or `0` in the memset case occurs.
Source: https://elixir.bootlin.com/linux/v6.11/source/drivers/hid/hid-core.c#L2122-L2131

Just a simple, `user_table->cur_id = NULL`, should be fine.

sheharyaar: I can understand the scepticism behind not using `cur_id` outside the loop. But since we are aware of the `cur_id++` behaviour, we can use it. I have seen code written in this manner. Usually when we are dealing with an array of pointers (when we want to fill in values and keep the last pointer NULL). For example when building the argument list for passing to the `exec` family of code. Or creating a `\0` terminated string of given size.

sheharyaar: Sorry, I misread the code. We need the bus value as 0 to terminate the array. `memset(&user_id[cur_id], 0, ...)` should be fine. This pattern is used in many places, such as to terminate strings with given lengths, to form argument array for passing to the `exec` family, etc.

Since we know the behaviour of `curr_id++`, it should be fine using that.
I am fine with both approaches.

sheharyaar: >  it will be recycled in many programming language. on the other hand

In other programming languages that happens if you declare and use a variable inside the loop with the _same_ name as the one outside it. Am I right ? If yes, then here that is not the case.

sheharyaar: Tho, I am fine with both approaches.

lib/luahid.c
I've just been trying hard to find these functions but failed. Thank you for helping me locate them.

lib/luahid.c
qrsikno2: done.
added new marcos

qrsikno2:
> Before merging this, it needs to be fixed. The code is still buggy, I got errors related to `attempting to index nil value`, due to some stack operation error. I have attached the test lua script in the lunatik chat.
> 
> CC: @lneto @qrsikno2

this is done due to add a new marco

lib/luahid.c
```diff
@@ -50,6 +47,45 @@ static const lunatik_class_t luahid_class = {
 	.sleep = true,
 };
 
+#define luahid_optinteger(L, idx, nf, field, opt)	\
+do {							\
+	lua_getfield(L, idx, #field);			\
+	if (!lua_isnil(L, -1)) 			\
```
sheharyaar: we can simplify this too,
```
nf->field = lua_isnil(L, -1) ? opt : lua_tointeger(L, -1);  \
```

But it's upto you, it's your code.

qrsikno2: what's @lneto 's idea of this?

lib/luahid.c

lib/luahid.c

lneto: I agree with @sheharyaar.. I don't see any reason to not follow his suggestion..

lib/luahid.c

lib/luahid.c
```diff
@@ -50,6 +47,42 @@ static const lunatik_class_t luahid_class = {
 	.sleep = true,
 };
 
+#define luahid_optinteger(L, idx, nf, field, opt)			\
```
lneto: You should remove it, right?

lib/luahid.c
```diff
@@ -61,7 +87,10 @@ static int luahid_register(lua_State *L)
 	struct hid_driver *user_driver = &(hid->driver);
 	user_driver->name = lunatik_checkalloc(L, NAME_MAX);
 	lunatik_setstring(L, 1, user_driver, name, NAME_MAX);
-	user_driver->id_table = luahid_table;
+
+	lunatik_checkfield(L, 1, "id_table", LUA_TTABLE);
+	luaL_argcheck(L, (user_driver->id_table = luahid_setidtable(L, -1)) == NULL,
+			   2, "id_table is not specified or invalid");
```
lneto: "invalid id_table" should be enough for the error message

lib/luahid.c
```diff
@@ -50,6 +47,35 @@ static const lunatik_class_t luahid_class = {
...
+	struct hid_device_id *cur_id = user_table;
+	for (size_t i = 0; i < len; i++, cur_id++) {
+		if (lua_geti(L, idx, i + 1) != LUA_TTABLE) { /* table entry */
+			lua_pop(L, 1); /* table entry */
+			lunatik_free(user_table);
+			user_table = NULL;
+			goto out;
+		}
+
+		lunatik_optinteger(L, -1, cur_id, bus, HID_BUS_ANY);
```
lneto, line 81: You should export this enum to Lua as well, right? If you don't plan to do it in this PR, please, create an issue

lib/luahid.c
```diff
@@ -3,22 +3,34 @@
...
 #include <lunatik.h>
-
+#include <lauxlib.h>
+
+/**
+ * Represents a registered HID driver.
+ * This is a userdata object returned by `hid.register()`. It encapsulates
+ * the kernel `struct hid_driver` and associated Lunatik runtime information
+ * necessary to invoke the Lua callback when a HID device is matched.
+* The `registered` field indicates whether the driver is currently registered
```
lneto, line 25: not aligned

lib/luahid.c
```diff
@@ -3,22 +3,34 @@
 * SPDX-License-Identifier: MIT OR GPL-2.0-only
 */
 
-#include <string.h>
```
lneto, line 6: it's weird

qrsikno2: it's just the lsp notes me that the marco `pr_fmt` defines dupilicated, and i reviewed all the lib code in our repo. 
only to find the defined the marco at top. so i do as the same way.

lib/luahid.c
qrsikno2: done

lneto (approved):
@qrsikno2 can you just confirm what happens when `id_table = {}`?

@sheharyaar this PR is good to me, if you approve, we can merge.

@jperon please, can you take a look at the doc? 🙏🏿

sheharyaar (approved):

lib/luahid.c
```diff
@@ -3,22 +3,34 @@
...
+/***
+* This is a high-level Lua interface to the Linux HID subsystem.
```
jperon:
```suggestion
* High-level Lua interface to the Linux HID subsystem.
```

lib/luahid.c
```diff
@@ -50,6 +64,52 @@ static const lunatik_class_t luahid_class = {
...
+* - `id_table`: A table of device IDs that this driver can match against (lua_table).
+* 	- each id is a table consist of fields:
+* 		- `bus`: The bus type (integer, default: `HID_BUS_ANY`).
+* 		- `group`: The group type (integer, default: `HID_GROUP_ANY`).
+* 		- `vendor`: The vendor ID (integer, default: `HID_ANY_ID`).
+* 		- `product`: The product ID (integer, default: `HID_ANY_ID`).
+* 		- `driver_data`: Additional driver data (integer, default: `0`).
+*
+* @function hid.register
+* @param {table} table The Lua table containing driver information.
```
jperon:
```suggestion
* @tparam table table The Lua table containing driver information.
```

`@tparam` and `@treturn` are better-suited for typed params / returns.

lib/luahid.c
```diff
@@ -50,6 +64,52 @@ static const lunatik_class_t luahid_class = {
...
+* @return {hid_driver} The registered HID driver object.
```
jperon:
```suggestion
* @treturn hid_driver The registered HID driver object.
```

lib/luahid.c
```diff
@@ -50,6 +64,52 @@ static const lunatik_class_t luahid_class = {
...
+out:
+	lua_pop(L, 1); /* id_table */
+	return user_table;
+}
+
+/***
+* Registers a new HID driver.
+* This function creates a new HID driver object from a Lua table and registers it with the kernel.
+* The Lua table must contain the following fields:
+* - `name`: The name of the driver (string).
```
jperon, line 101:
```suggestion
*
* - `name`: The name of the driver (string).
```

LDoc is a bit fussy about newilnes before lists. Cf. luadevice.c for example.

lib/luahid.c
```diff
@@ -50,6 +64,52 @@ static const lunatik_class_t luahid_class = {
...
+* - `id_table`: A table of device IDs that this driver can match against (lua_table).
+* 	- each id is a table consist of fields:
+* 		- `bus`: The bus type (integer, default: `HID_BUS_ANY`).
```
jperon:
```suggestion
* - `id_table`: A table of device IDs that this driver can match against (lua_table).
*   each id is a table consist of fields:
*
* 		- `bus`: The bus type (integer, default: `HID_BUS_ANY`).
```

jperon:
@qrsikno2 You must add `lib/luahid.c` to the `file` list of `config.ld` to get it rendered by `make doc-site`.

jperon:
Another thing you _should_ do (though not necessary, it makes it much easier to understand the api): adding a `@usage` section at the bottom of your LDoc comment. See [luadevice.c](https://github.com/luainkernel/lunatik/blob/13b01197368aa492899831a50b94102bd2e44681/lib/luadevice.c#L322) for an example.

config.ld
```diff
@@ -35,6 +35,7 @@ file = {
     './lib/luathread.c',
     './lib/luaxdp.c',
     './lib/luaxtable.c',
+    './lib/luahid.c',
```
jperon: Please put it just after `luafifo.c`, in order to keep alphabetical order in the left menu.

lib/luahid.c
jperon: @qrsikno2 It seems you forgot this one?

jperon: Never mind, I overlooked it.

lib/luahid.c
```diff
@@ -50,6 +64,64 @@ static const lunatik_class_t luahid_class = {
...
+}
+
+/***
+* Registers a new HID driver.
+* This function creates a new HID driver object from a Lua table and registers it with the kernel.
+* The Lua table must contain the following fields:
+*
+* - `name`: The name of the driver (string).
+* - `id_table`: A table of device IDs that this driver can match against (lua_table).
+* 	- each id is a table consist of fields:
```
jperon:
```suggestion
* 	- each id is a table consisting of fields:
```

jperon:
After the small change in `config.ld` and the small typo, will be OK for me.

lib/luahid.c

lib/luahid.c
qrsikno2: done

lib/luahid.c

lib/luahid.c

lib/luahid.c

qrsikno2:
> @qrsikno2 You must add `lib/luahid.c` to the `file` list of `config.ld` to get it rendered by `make doc-site`.

done

qrsikno2:
> Another thing you _should_ do (though not necessary, it makes it much easier to understand the api): adding a `@usage` section at the bottom of your LDoc comment. See [luadevice.c](https://github.com/luainkernel/lunatik/blob/13b01197368aa492899831a50b94102bd2e44681/lib/luadevice.c#L322) for an example.

done

config.ld
qrsikno2: done

jperon (approved):

PR #248, opened 2025-05-13
Add bindings to kernel crypto api

jperon:
hmac, shash, skcipher, aead, rng and comp

PR #283, opened 2025-06-24
Linux crypto api: crypto_shash

jperon:
First step to integrate Linux crypto api: replaces #248.


config.ld
```diff
@@ -13,28 +13,29 @@ title = 'Lunatik API Documentation'
...
+	'./lib/luanotifier.c',
+	'./lib/luaprobe.c',
+	'./lib/luarcu.c',
+	'./lib/luasocket.c',
+	'./lib/socket/inet.lua',
+	'./lib/luasyscall.c',
+	'./lib/syscall/table.lua',
+	'./lib/luathread.c',
+	'./lib/luaxdp.c',
+	'./lib/luaxtable.c',
```
lneto: Please, let's make formatting changes in separate PRs.. it's make it harder to review

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,346 @@
...
+static int luacrypto_shash_new(lua_State *L) {
+	const char *algname = luaL_checkstring(L, 1);
+	luacrypto_shash_tfm_t *tfm_ud;
+	gfp_t gfp = lunatik_gfp(lunatik_toruntime(L));
+	size_t desc_size;
+	lunatik_object_t *object = lunatik_newobject(
+		L, &luacrypto_shash_tfm_class,
+		sizeof(luacrypto_shash_tfm_t)
+	);
+	if (!object) {
```
lneto: Actually, newobject throws an error if it's not enough mem.. so, we don't need this verification here

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,343 @@
...
+* @treturn string The computed digest (hash output).
+* @raise Error on failure (e.g., allocation error, crypto API error).
+*/
+static int luacrypto_shash_tfm_digest(lua_State *L) {
+	luacrypto_shash_tfm_t *tfm_ud = luacrypto_check_shash_tfm(L, 1);
+	size_t datalen;
+	const char *data = luaL_checklstring(L, 2, &datalen);
+	gfp_t gfp = lunatik_gfp(lunatik_toruntime(L));
+	unsigned int digestsize = crypto_shash_digestsize(tfm_ud->tfm);
+	u8 *digest_buf = kmalloc(digestsize, gfp);
```
lneto: We could use checkalloc here, right?

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,343 @@
...
+					algname);
+	}
+	tfm_ud->kernel_desc->tfm = tfm_ud->tfm;
+	tfm_ud->desc_alloc_len = desc_size;
+	return 1;
+}
+
+
+static const luaL_Reg luacrypto_shash_lib_funcs[] = {
+	{"new", luacrypto_shash_new},
```
lneto, line 271: I was thinking if it wasn't the case of having a luacrypto_lib with all constructors and have sub bindings only for the classes.. in this case, we could have:
```
local crypto = require("crypto")
local shash = crypto.shash()

shash:init(...)
...
```
What do you think?

jperon: Why not; IMHO, the easiest (and cleanest?) way would be to have a `lib/crypto/init.lua` file, which would contain following:

```lua
return {
  shash = require("crypto_shash").new,
}
```

and then other constructors when following PRs will come.

lneto: I'm not sure it would be easier, but one advantage of your approach is that we can load submodules dynamically.. so we don't need to have everything linked on a luacrypto.ko.. up to you =)

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,343 @@
...
+*/
+static int luacrypto_shash_new(lua_State *L) {
+	const char *algname = luaL_checkstring(L, 1);
+	luacrypto_shash_tfm_t *tfm_ud;
+	gfp_t gfp = lunatik_gfp(lunatik_toruntime(L));
+	size_t desc_size;
+	lunatik_object_t *object = lunatik_newobject(
+		L, &luacrypto_shash_tfm_class,
+		sizeof(luacrypto_shash_tfm_t)
+	);
```
lneto: please, let's keep our code style..
```suggestion
	lunatik_object_t *object = lunatik_newobject(L, &luacrypto_shash_tfm_class,
		sizeof(luacrypto_shash_tfm_t));
```

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,343 @@
...
+	gfp_t gfp = lunatik_gfp(lunatik_toruntime(L));
+	unsigned int digestsize = crypto_shash_digestsize(tfm_ud->tfm);
+	u8 *digest_buf = kmalloc(digestsize, gfp);
+	if (!digest_buf) return luaL_error(L, "shash:digest: failed to allocate digest buffer");
+	int ret = crypto_shash_digest(tfm_ud->kernel_desc, data, datalen, digest_buf);
+	if (ret) {
+		kfree(digest_buf);
+		return luaL_error(L, "shash:digest: crypto_shash_digest failed (%d)", ret);
+	}
+	lua_pushlstring(
```
lneto: also here.. please review your patch regarding our code style.. it will make it easier for reviewing

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,343 @@
...
+* This must be called before using `update()` or `final()`.
+* @function crypto_shash:init
+* @raise Error on failure.
+*/
+static int luacrypto_shash_tfm_init(lua_State *L) {
+	luacrypto_shash_tfm_t *tfm_ud = luacrypto_check_shash_tfm(L, 1);
+	int ret = crypto_shash_init(tfm_ud->kernel_desc);
+	if (ret) {
+		return luaL_error(L, "shash:init: failed (%d)", ret);
+	}
```
lneto: I see this patter quite a lot in your code, perhaps we can leverage the try_ret approach from luasocket.. we can even move it to lunatik.h.. please take a look

jperon: OK. Should I factor out `luasocket_try` into lunatik.h in this PR, or submit a preliminary one?

jperon: I did it in a65fe5c. I’ll revert it if you want to.

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,295 @@
...
+
+	tfm_ud->tfm = crypto_alloc_shash(algname, 0, 0);
+	if (IS_ERR(tfm_ud->tfm)) {
+		long err = PTR_ERR(tfm_ud->tfm);
+		return luaL_error(L, "failed to allocate SHASH transform for %s (err %ld)", algname, err);
+	}
+
+	desc_size = sizeof(struct shash_desc) + crypto_shash_descsize(tfm_ud->tfm);
+	tfm_ud->kernel_desc = lunatik_checkalloc(L, desc_size);
+	tfm_ud->kernel_desc->tfm = tfm_ud->tfm;
```
lneto: Why do you need both? Couldn't you keep this on kernel_desc->tfm only? Btw, why naming this "kernel_"? I found this a bit odd.. how does the clients of this kernel API usually name desc?

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,298 @@
...
+LUNATIK_PRIVATECHECKER(luacrypto_check_shash_tfm, luacrypto_shash_tfm_t *);
+static void luacrypto_shash_tfm_release(void *private)
+{
+	struct crypto_shash *tfm;
+	luacrypto_shash_tfm_t *tfm_ud = (luacrypto_shash_tfm_t *)private;
+
+	if (tfm_ud->sdesc) {
+		tfm = tfm_ud->sdesc->tfm;
+		lunatik_free(tfm_ud->sdesc);
+		if (tfm && !IS_ERR(tfm))
```
lneto: if `IS_ERR(tfm)`, you are not assigning it to `tfm_ud->sdesc->tfm`, right? it looks like you don't need to verify it here..

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,298 @@
...
+	struct shash_desc *sdesc; /* Points to kmalloc'ed shash_desc + space */
+} luacrypto_shash_tfm_t;
+
+LUNATIK_PRIVATECHECKER(luacrypto_check_shash_tfm, luacrypto_shash_tfm_t *);
+static void luacrypto_shash_tfm_release(void *private)
+{
+	struct crypto_shash *tfm;
+	luacrypto_shash_tfm_t *tfm_ud = (luacrypto_shash_tfm_t *)private;
+
+	if (tfm_ud->sdesc) {
```
lneto: I would put this in a local variable.. I think it would clean up a bit the code..

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,298 @@
...
+#include <lunatik.h>
+
+static const lunatik_class_t luacrypto_shash_tfm_class;
+
+typedef struct {
+	struct shash_desc *sdesc; /* Points to kmalloc'ed shash_desc + space */
+} luacrypto_shash_tfm_t;
+
+LUNATIK_PRIVATECHECKER(luacrypto_check_shash_tfm, luacrypto_shash_tfm_t *);
+static void luacrypto_shash_tfm_release(void *private)
```
lneto:
```suggestion
LUNATIK_PRIVATECHECKER(luacrypto_check_shash_tfm, luacrypto_shash_tfm_t *);

static void luacrypto_shash_tfm_release(void *private)
```

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,298 @@
...
+
+	if (tfm_ud->sdesc) {
+		tfm = tfm_ud->sdesc->tfm;
+		lunatik_free(tfm_ud->sdesc);
+		if (tfm && !IS_ERR(tfm))
+			crypto_free_shash(tfm);
+	}
+}
+
+
```
lneto:
```suggestion

```

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,298 @@
...
+* Gets the digest size (output length) of the hash algorithm.
+* @function crypto_shash:digestsize
+* @treturn integer The digest size in bytes.
+*/
+static int luacrypto_shash_tfm_digestsize(lua_State *L) {
+	luacrypto_shash_tfm_t *tfm_ud = luacrypto_check_shash_tfm(L, 1);
+	lua_pushinteger(L, crypto_shash_digestsize(tfm_ud->sdesc->tfm));
+	return 1;
+}
+/***
```
lneto:
```suggestion
}

/***
```

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,298 @@
...
+#include <lauxlib.h>
+#include <lunatik.h>
+
+static const lunatik_class_t luacrypto_shash_tfm_class;
+
+typedef struct {
+	struct shash_desc *sdesc; /* Points to kmalloc'ed shash_desc + space */
+} luacrypto_shash_tfm_t;
+
+LUNATIK_PRIVATECHECKER(luacrypto_check_shash_tfm, luacrypto_shash_tfm_t *);
```
lneto: let's prefix everything with `luacrypto_shash`, so `sluacrypto_check_shash_tfm/luacrypto_shash_check_tfm/g`..

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,298 @@
...
+
+#include <lua.h>
+#include <lualib.h>
+#include <lauxlib.h>
+#include <lunatik.h>
+
+static const lunatik_class_t luacrypto_shash_tfm_class;
+
+typedef struct {
+	struct shash_desc *sdesc; /* Points to kmalloc'ed shash_desc + space */
```
lneto: if only a nested pointer, you don't need this structure.. you can just use the inner struct as you private and set object as a pointer in the class descriptor.. see luasocket

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,298 @@
...
+#include <crypto/hash.h>
+#include <linux/err.h>
+#include <linux/slab.h>
+
+#include <lua.h>
+#include <lualib.h>
+#include <lauxlib.h>
+#include <lunatik.h>
+
+static const lunatik_class_t luacrypto_shash_tfm_class;
```
lneto: you don't need this, right/

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,298 @@
...
+
+typedef struct {
+	struct shash_desc *sdesc; /* Points to kmalloc'ed shash_desc + space */
+} luacrypto_shash_tfm_t;
```
lneto: I would call your type just `luacrypto_shash_t` for the sake of simplicity.. you have no name clashing with that and I think it's clear enough..

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,298 @@
...
+static const luaL_Reg luacrypto_shash_tfm_mt[] = {
+	{"digestsize", luacrypto_shash_tfm_digestsize},
+	{"setkey", luacrypto_shash_tfm_setkey},
+	{"digest", luacrypto_shash_tfm_digest},
+	{"init", luacrypto_shash_tfm_init},
+	{"update", luacrypto_shash_tfm_update},
+	{"final", luacrypto_shash_tfm_final},
+	{"finup", luacrypto_shash_tfm_finup},
+	{"export", luacrypto_shash_tfm_export},
+	{"import", luacrypto_shash_tfm_import},
```
lneto: I would suppress `_tfm`, it would make the code easier for reading, IMO

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,292 @@
...
+#include <crypto/hash.h>
+#include <linux/err.h>
+#include <linux/slab.h>
+
+#include <lua.h>
+#include <lualib.h>
+#include <lauxlib.h>
+#include <lunatik.h>
+
+LUNATIK_PRIVATECHECKER(luacrypto_shash_check_tfm, struct shash_desc *);
```
lneto:
```suggestion
LUNATIK_PRIVATECHECKER(luacrypto_shash_check, struct shash_desc *);
```

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,292 @@
...
+	return 1;
+}
+
+/***
+* Initializes a multi-part hash operation.
+* This must be called before using `update()` or `final()`.
+* @function crypto_shash:init
+* @raise Error on failure.
+*/
+static int luacrypto_shash_init_method(lua_State *L) {
```
lneto, line 105: Is there a reason to not doing it automatically on instantiation? I don't see the use case for calling new() and not calling init()

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,292 @@
...
+*/
+static int luacrypto_shash_import(lua_State *L) {
+	struct shash_desc *sdesc = luacrypto_shash_check_tfm(L, 1);
+	size_t statelen;
+	const char *state = luaL_checklstring(L, 2, &statelen);
+	unsigned int expected_statesize;
+	expected_statesize = crypto_shash_statesize(sdesc->tfm);
+	luaL_argcheck(L, statelen == expected_statesize,
+		2, "incorrect state length for import"
+	);
```
lneto:
```suggestion
	luaL_argcheck(L, statelen == expected_statesize, 2,
		"incorrect state length for import");
```

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,292 @@
...
+* @function crypto_shash:import
+* @tparam string state The previously exported hash state (binary string).
+* @raise Error on failure or if the provided state length is incorrect for the algorithm.
+*/
+static int luacrypto_shash_import(lua_State *L) {
+	struct shash_desc *sdesc = luacrypto_shash_check_tfm(L, 1);
+	size_t statelen;
+	const char *state = luaL_checklstring(L, 2, &statelen);
+	unsigned int expected_statesize;
+	expected_statesize = crypto_shash_statesize(sdesc->tfm);
```
lneto:
```suggestion
	unsigned int expected_statesize = crypto_shash_statesize(sdesc->tfm);
```

lib/crypto/init.lua
```diff
@@ -0,0 +1,32 @@
...
+
+--- Wrapper for the Linux Kernel Crypto API.
+--
+-- This module provides a simplified interface to the underlying `crypto_*`
+-- C modules. It allows for a more natural and convenient API, such as
+-- using `require("crypto").shash()` instead of `require("crypto_shash").new()`.
+--
+-- @module crypto
+
+local crypto_shash = require("crypto_shash")
```
lneto, line 14: I might be wrong, but can't we `s/_/./` here?

```suggestion
local crypto_shash = require("crypto.shash")
```
can you confirm that `.` won't work in this case? if it works, IMO, it looks much more aligned to the rest of the code base ;-)

lneto:
@jperon we are almost there on shash, if it's not difficult to you, can we leave hkdf and util for the next PR? so we get shash merged fastly =)

jperon:
> @jperon we are almost there on shash, if it's not difficult to you, can we leave hkdf and util for the next PR? so we get shash merged fastly =)

Not a problem _in se_ for me, but hkdf serves as a very good test for shash. I was "saved" several times by the fact I just have to run `lunatik unload; find tests/crypto/ | grep lua | while read test; do lunatik run "$test"; done` to detect any regression while enhancing code.

lib/crypto/init.lua
jperon: As is, we can’t:
```c
LUNATIK_NEWLIB(crypto_shash, luacrypto_shash_lib_funcs, &luacrypto_shash_class, NULL);
```
We could add a `lib/crypto/shash.lua` file, which would just contain:
```lua
return require"crypto_shash"
```

That said:

- IMHO, the whole purpose of `init.lua` is to make it useless by "hiding" `crypto_hash` behind `require("crypto").shash()`;
- while less aligned with the rest of the code base, the `crypto_hash` syntax makes it clear that each module from the `crypto` api is independent, and could be compiled/used without any other one.

lib/luacrypto_shash.c
jperon: `init` is required for multi-part operation; but when using single operations like `digest`, initialization is handled internally. I tried adding `lunatik_try(L, crypto_shash_init, sdesc);` within `new`, and running hkdf test caused system hang without any log in dmesg.

lib/crypto/init.lua
lneto: Yeah.. let's keep it as is.. I just had the impression that '.' would match '_' in the finder..

lib/luacrypto_shash.c
lneto: Ack

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,291 @@
...
+* This allows suspending and later resuming a hash calculation via `import()`.
+* Must be called after `init()` and any `update()` calls if part of a multi-step operation.
+* @function crypto_shash:export
+* @treturn string The internal hash state as a binary string.
+* @raise Error on failure (e.g., allocation error).
+*/
+static int luacrypto_shash_export(lua_State *L) {
+	struct shash_desc *sdesc = luacrypto_shash_check(L, 1);
+	unsigned int statesize = crypto_shash_statesize(sdesc->tfm);
+	void *state_buf = lunatik_checkalloc(L, statesize); // Allocate on Lua stack for safety
```
lneto: You can use Lua Buffer for that, right? Also, please avoid cpp style comments..

jperon: Is my last change what you expect?

lib/luacrypto_shash.c
```diff
@@ -0,0 +1,289 @@
...
+* This allows suspending and later resuming a hash calculation via `import()`.
+* Must be called after `init()` and any `update()` calls if part of a multi-step operation.
+* @function crypto_shash:export
+* @treturn string The internal hash state as a binary string.
+* @raise Error on failure (e.g., allocation error).
+*/
+static int luacrypto_shash_export(lua_State *L) {
+	struct shash_desc *sdesc = luacrypto_shash_check(L, 1);
+	unsigned int statesize = crypto_shash_statesize(sdesc->tfm);
+	void *state_udata = lua_newuserdatauv(L, statesize, 0);
```
lneto: You are returning a userdata not a string.. it's not clear to me how it should be used from Lua

lneto (approved):

PR #285, opened 2025-06-28
update the optinteger with removed nil version

qrsikno2:
This pull request makes a single change to the macro definition in `lunatik.h`, renaming it and simplifying its logic.

* Renamed the macro `lunatik_optinteger` to `luahid_optinteger` and simplified its logic by removing conditional checks for `lua_isnil`. The new implementation directly assigns `opt` if the field is `nil`, reducing redundancy.

### DIFF
--- a/lunatik.h
+++ b/lunatik.h
@@ -301,15 +301,11 @@ do {								\
 	lua_pop(L, 1);						\
 } while (0)
 
-#define lunatik_optinteger(L, idx, nf, field, opt)	\
-do {							\
-	lua_getfield(L, idx, #field);			\
-	if (!lua_isnil(L, -1)) {			\
-		nf->field = lua_tointeger(L, -1);	\
-		lua_pop(L, 1);				\
-	}						\
-	else						\
-		nf->field = opt;			\
+#define lunatik_optinteger(L, idx, priv, field, opt)			\
+do {									\
+	lua_getfield(L, idx, #field);					\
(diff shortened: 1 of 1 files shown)


lunatik.h
```diff
@@ -301,15 +301,11 @@ do {								\
...
-#define lunatik_optinteger(L, idx, nf, field, opt)	\
-do {							\
-	lua_getfield(L, idx, #field);			\
-	if (!lua_isnil(L, -1)) {			\
-		nf->field = lua_tointeger(L, -1);	\
-		lua_pop(L, 1);				\
-	}						\
-	else						\
-		nf->field = opt;			\
+#define luahid_optinteger(L, idx, nf, field, opt)			\
```
lneto: you should not do this.. please be careful otherwise you are breaking the build

lneto: btw, while here, let's generalize `nf` name.. perhaps `priv`

lneto (commented):
please, use a more accurate commit message.. e.g., "pop field on lunatik_optinteger"

lunatik.h
qrsikno2: done

lneto (approved):

qrsikno2:
> please, use a more accurate commit message.. e.g., "pop field on lunatik_optinteger"

done

lunatik.h
qrsikno2: done

PR #286, opened 2025-06-29
crypto_shash tests

jperon:


lib/crypto/hkdf.lua
```diff
@@ -0,0 +1,113 @@
...
+	if self.tfm and self.tfm.__close then
+		self.tfm:__close()
+		self.tfm = nil
+	end
+end
+
+HKDF.__gc = HKDF.close
+HKDF.__index = HKDF
+
+return setmetatable(HKDF, HKDF)
```
lneto: why do we need this? btw, we usually use modules as plain libraries not objects on lunatik.. e.g., we don't do this..
```
local lib = require("lib")
lib:meth()
```
instead, we do..
```
local lib = require("lib")
lib.func()
```

lib/crypto/hkdf.lua
```diff
@@ -0,0 +1,113 @@
...
+function HKDF.new(alg)
+	local hmac = shash.new("hmac(" .. alg .. ")")
+	local hash_len = hmac:digestsize()
+	local default_salt = rep("\0", hash_len)
+
+	local instance = {
+		tfm = hmac,
+		hash_len = hash_len,
+		default_salt = default_salt,
+	}
```
lneto:
```suggestion
	local hash = {}
	hash.tfm = shash.new("hmac(" .. alg .. ")")
	hash.len = hmac:digestsize()
	hash.salt = rep("\0", hash_len)
```

I think it's unnecessary to put everything in a local just to store them in a table afterwords..

lneto: btw, do we really need to "cash" its digestsize? can't we just use its metatable to fallback to userdata (shash) methods?

tests/init.lua
```diff
@@ -0,0 +1,18 @@
+--
```
lneto: we should add this in a separate commit

tests/crypto/shash.lua
```diff
@@ -0,0 +1,109 @@
+--
```
lneto, line 1: we should add this in a separate commit as well..

lib/util.lua
```diff
@@ -0,0 +1,54 @@
...
+-- @tparam table tbl The table to iterate over.
+-- @tparam[opt] function fn A comparison function to sort the keys.
+-- @treturn function An iterator function that, when called, returns the next key (`any`) and value (`any`) in sorted order, or `nil` once iteration is complete.
+local function opairs(tbl, fn)
+	local i, keys = 1, {}
+	for k in pairs(tbl) do
+		keys[i] = k
+		i = i + 1
+	end
+	sort(keys, fn)
```
lneto: it feels awkward to me.. we are calling this on every iteration step.. why not sorting before iterating?

lib/crypto/hkdf.lua
```diff
@@ -0,0 +1,113 @@
...
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+--- HMAC-based Extract-and-Expand Key Derivation Function (HKDF) based on RFC 5869.
+-- This module provides functions to perform HKDF operations, utilizing the
+-- underlying `crypto_shash` C module for HMAC calculations.
+-- @classmod crypto.hkdf
+
+local shash = require("crypto_shash")
+local char, sp, rep, sub = string.char, string.pack, string.rep, string.sub
```
lneto:
```suggestion
local char, pack, rep, sub = string.char, string.pack, string.rep, string.sub
```

lib/crypto/hkdf.lua
```diff
@@ -0,0 +1,113 @@
...
+function HKDF:tls13_expand_label(prk, label, context, length)
+	local hkdf_label_info = sp(">Hs1s1", length, "tls13 " .. label, context)
+	return self:expand(prk, hkdf_label_info, length)
+end
+
+--- Closes the HKDF instance and releases the underlying HMAC transform.
+-- This method is also called by the garbage collector.
+-- @function HKDF:close
+function HKDF:close()
+	if self.tfm and self.tfm.__close then
```
lneto: we don't need to verify if `tfm.__close` is defined, right? it's always defined for shash

lib/crypto/hkdf.lua
```diff
@@ -0,0 +1,113 @@
...
+		self.tfm:__close()
+		self.tfm = nil
+	end
+end
+
+HKDF.__gc = HKDF.close
```
lneto: why not defining __close as well?

lib/util.lua
jperon: It seems to me we’re not sorting on every step: what gets called on every step is the closure that gets returned l. 42. Before returning this closure, we build an "array" with keys, then sort this array.

lneto: IMHO, it's unnecessarily complex.. I'd rather keep it simpler in the code base.. and explicit where we are ordering something (outside the iterator).. however, we could use a third opinion.. I don't want to impose my own view.. =).. @sheharyaar could you share your thoughts on this?

lib/crypto/hkdf.lua
```diff
@@ -0,0 +1,113 @@
...
+-- @tparam string alg The base hash algorithm name (e.g., "sha256", "sha512").
+-- The "hmac(" prefix will be added automatically.
+-- @treturn HKDF An HKDF instance table with methods for key derivation.
+-- @usage local hkdf_sha256 = require("crypto.hkdf").new("sha256")
+function HKDF.new(alg)
+	local hmac = setmetatable({},	HKDF)
+	hmac.tfm = shash.new("hmac(" .. alg .. ")")
+	hmac.salt = rep("\0", #hmac)
+
+	return hmac
```
lneto: generally we use this pattern:
```suggestion
	local hmac = {}
	hmac.tfm = shash.new("hmac(" .. alg .. ")")
	hmac.salt = rep("\0", #hmac)

	return setmetatable(hmac, HKDF)
```

lib/crypto/hkdf.lua
```diff
@@ -0,0 +1,113 @@
...
+--- Creates a new HKDF instance for a given hash algorithm.
+-- @function HKDF.new
+-- @tparam string alg The base hash algorithm name (e.g., "sha256", "sha512").
+-- The "hmac(" prefix will be added automatically.
+-- @treturn HKDF An HKDF instance table with methods for key derivation.
+-- @usage local hkdf_sha256 = require("crypto.hkdf").new("sha256")
+function HKDF.new(alg)
+	local hmac = setmetatable({},	HKDF)
+	hmac.tfm = shash.new("hmac(" .. alg .. ")")
+	hmac.salt = rep("\0", #hmac)
```
lneto, line 46: `#hmac` is zero, right? ah.. you are using the metamethod here? if yes, please disregard my comment below..

lib/crypto/hkdf.lua
```diff
@@ -0,0 +1,113 @@
...
+end
+
+--- Creates a new HKDF instance for a given hash algorithm.
+-- @function HKDF.new
+-- @tparam string alg The base hash algorithm name (e.g., "sha256", "sha512").
+-- The "hmac(" prefix will be added automatically.
+-- @treturn HKDF An HKDF instance table with methods for key derivation.
+-- @usage local hkdf_sha256 = require("crypto.hkdf").new("sha256")
+function HKDF.new(alg)
+	local hmac = setmetatable({},	HKDF)
```
lneto: `s/,\tHKDF/, HKDF/`

lib/crypto/hkdf.lua
```diff
@@ -0,0 +1,113 @@
...
+	hmac.tfm = shash.new("hmac(" .. alg .. ")")
```
lneto, line 45: is it safe to pass a invalid `alg` to the C binding here? what should happen?

lib/crypto/hkdf.lua
```diff
@@ -0,0 +1,113 @@
...
+	local hash_len = #self
+	local n = length / hash_len
+	n = (n * hash_len == length) and n or n + 1
+
+	local okm, t = "", ""
+	for i = 1, n do
+		t = self:hmac(prk, t .. info .. char(i))
+		okm = okm .. t
+	end
+	return sub(okm, 1, length)
```
lneto, line 86: I'm not sure I follow, but I'm assuming you know what you are doing and I see no potential harm on this as well.. =)

lib/util.lua
```diff
@@ -0,0 +1,32 @@
+--
+-- SPDX-FileCopyrightText: (c) 2025 jperon <cataclop@hotmail.com>
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+--- Utility functions.
+-- @module util
```
lneto, line 7: as it's restrict to bin2hex and hex2bin I would name this module something like: encode.{bin2hex, hex2bin} or hexbin.{to,from}.. of something else if you have a better idea ;-)

lib/util.lua
```diff
@@ -0,0 +1,32 @@
...
+
+local char, format, gsub, rep, su = string.char, string.format, string.gsub, string.rep, string.unpack
+local sort = table.sort
```
lneto: let's keep only the current used ones

lib/util.lua
```diff
@@ -0,0 +1,32 @@
...
+-- @tparam string hex The hexadecimal string to convert.
+-- @treturn string A string containing the binary representation of the input.
+local function hex2bin(hex)
+	return gsub(hex, "..", function(cc) return char(tonumber(cc, 16)) end)
+end
+
+return {
+	bin2hex = bin2hex,
+	hex2bin = hex2bin,
+}
```
lneto: the common patter here is to..
```
local mylib = {}
...
function mylib.func()
   ...
end
...
return mylib
```

tests/init.lua
```diff
@@ -0,0 +1,18 @@
...
+-- SPDX-FileCopyrightText: (c) 2025 jperon <cataclop@hotmail.com>
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+local function run_test(test_name, func)
+	local status, err = pcall(func)
+	if status then
+		print(string.format("[PASS] %s", test_name))
```
lneto: I would create an aux here like..
```
local function log(what, name, ...)
	print(string.format(what:upper .. "%s", name, ...)
end
...
 	log("pass", name)
```

just an idea

tests/init.lua
```diff
@@ -0,0 +1,18 @@
...
+	else
+		print(string.format("[FAIL] %s: %s\n%s", test_name, err, debug.traceback()))
+	end
+end
+
+return {
+	run_test = run_test,
+}
```
lneto: I would follow the library pattern I have mentioned in another comment..

lib/crypto/hkdf.lua
```diff
@@ -0,0 +1,115 @@
...
+-- @classmod crypto.hkdf
+
+local shash = require("crypto_shash")
+local char, pack, rep, sub = string.char, string.pack, string.rep, string.sub
+
+--- HKDF operations.
+-- This table provides the `new` method to create HKDF instances and also
+-- serves as the prototype for these instances.
+local HKDF = {}
+
```
lneto, line 18: I would move __gc, __close and __index to here

lib/crypto/hkdf.lua
```diff
@@ -0,0 +1,115 @@
...
+--- Creates a new HKDF instance for a given hash algorithm.
+-- @function HKDF.new
+-- @tparam string alg The base hash algorithm name (e.g., "sha256", "sha512").
+-- The "hmac(" prefix will be added automatically.
+-- @treturn HKDF An HKDF instance table with methods for key derivation.
+-- @usage local hkdf_sha256 = require("crypto.hkdf").new("sha256")
+function HKDF.new(alg)
+	local hmac = setmetatable({}, HKDF)
+	local ok, tfm = pcall(shash.new, "hmac(" .. alg .. ")")
+	assert(ok, "Invalid hash algorithm: " .. alg)
```
lneto: sorry, I think you misunderstood my point.. I was just asking if it would be safe to pass an invalid alg to shash.new.. it seems so, thus you don't need this assert, just let shash.new to raise this error..

lib/crypto/hkdf.lua
```diff
@@ -0,0 +1,115 @@
...
+end
+
+--- Derives a key using HKDF-Expand-Label as defined in RFC 8446 for TLS 1.3.
+-- @function HKDF:tls13_expand_label
+-- @tparam string prk Pseudorandom Key.
+-- @tparam string label The label for the derived secret.
+-- @tparam string context A string containing the hash of the transcript of the handshake messages.
+-- @tparam number length The desired length in bytes for the derived secret.
+-- @treturn string The derived secret of the specified `length`.
+function HKDF:tls13_expand_label(prk, label, context, length)
```
lneto: it looks to me that this function is more like a client of HKDF API than a feature from it.. I'd move it to a tls3 module and scope it out from this PR..

lib/util.lua
```diff
@@ -0,0 +1,54 @@
...
+-- @function log
+-- @tparam string what The prefix for the log message (e.g., "info", "error").
+-- @tparam ... Additional arguments to log, which will be concatenated with tabs.
+-- @usage util.log("info", "This is a message")
+-- @usage util.log("error", "An error occurred", "Error message")
+local function log(what, ...)
+	print(table.concat({what:upper(), ...}, "\t"))
+end
+
+util.log = log
```
lneto:
```suggestion
function util.log(what, ...)
	print(table.concat({what:upper(), ...}, "\t"))
end
```

lib/util.lua
```diff
@@ -0,0 +1,54 @@
...
+--- Converts a hexadecimal string to its binary representation.
+-- @function hex2bin
+-- @tparam string hex The hexadecimal string to convert.
+-- @treturn string A string containing the binary representation of the input.
+function util.hex2bin(hex)
+	return gsub(hex, "..", function(cc) return char(tonumber(cc, 16)) end)
+end
+
+--- Logs a message with a specific prefix.
+-- @function log
```
lneto, line 29: I would create a ticket for implement a proper binding for the kernel log levels.. perhaps adding some calls to lualinux.. not sure.. but it would be good to have that.. for now, this lua version is good enough ;-)

lib/crypto/hkdf.lua
```diff
@@ -0,0 +1,101 @@
...
+	return self.tfm:digest(data)
+end
+
+--- Performs the HKDF Extract step.
+-- @function HKDF:extract
+-- @tparam[opt] string salt Optional salt value. If nil or not provided, a salt of `hash_len` zeros is used.
+-- @tparam string ikm Input Keying Material.
+-- @treturn string The Pseudorandom Key (PRK).
+function HKDF:extract(salt, ikm)
+	return self:hmac((salt or self.default_salt), ikm)
```
lneto: it should be `self.salt` here, right?

PR #289, opened 2025-07-01
crypto_skcipher

jperon:
Second part of divided #248.


lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,277 @@
...
+	struct skcipher_request *req;
+
+	u8 *iv_data;
+	size_t iv_len;
+
+	u8 *work_buffer;
+	size_t work_buffer_len;
+
+	struct scatterlist sg_work;
+} luacrypto_skcipher_priv;
```
lneto:
```suggestion
} luacrypto_skcipher_t;
```
let's keep our style ;-)

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,277 @@
...
+	if (tfm_ud->req) {
+		skcipher_request_free(tfm_ud->req);
+		tfm_ud->req = NULL;
+	}
+
+	if (tfm_ud->tfm && !IS_ERR(tfm_ud->tfm))
+		crypto_free_skcipher(tfm_ud->tfm);
+}
+
+
```
lneto:
```suggestion
```

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,277 @@
...
+				   data_s_len, tfm_ud->iv_data);
+	skcipher_request_set_callback(tfm_ud->req, 0, NULL, NULL);
+
+	if (encrypt) {
+		lunatik_try(L, crypto_skcipher_encrypt, tfm_ud->req);
+	} else {
+		lunatik_try(L, crypto_skcipher_decrypt, tfm_ud->req);
+	}
+
+	lua_pushlstring(L, (const char *)tfm_ud->work_buffer, data_s_len);
```
lneto: wouldn't be the case of luaL_Buffer again?

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,277 @@
...
+* The IV (nonce) must be unique for each encryption operation with the same key for most modes.
+* Plaintext length should be appropriate for the cipher mode (e.g., multiple of blocksize).
+* @function crypto_skcipher:encrypt
+* @tparam string iv The Initialization Vector. Its length must match `ivsize()`.
+* @tparam string plaintext The data to encrypt.
+* @treturn string The ciphertext.
+* @raise Error on encryption failure, incorrect IV length, or allocation issues.
+*/
+static int luacrypto_skcipher_encrypt(lua_State *L) {
+	return luacrypto_skcipher_crypt_common(L, true);
```
lneto: we are using "template" macros in this case.. please, refer to luanetfilter

lneto: also, luanotifier

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,277 @@
...
+*/
+
+static const lunatik_class_t luacrypto_skcipher_class = {
+	.name = "crypto_skcipher",
+	.methods = luacrypto_skcipher_mt,
+	.release = luacrypto_skcipher_release,
+	.sleep = true,
+};
+
+
```
lneto:
```suggestion

```

let's please avoid multiple `\n`.. I just noticed we let it pass through review on luacrypto_shash

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,261 @@
...
+* @treturn crypto_skcipher The new SKCIPHER TFM object.
+* @raise Error if the TFM object or kernel request cannot be allocated/initialized.
+* @usage
+*   local skcipher_mod = require("crypto_skcipher")
+*   local cipher = skcipher_mod.new("cbc(aes)")
+*/
+static int luacrypto_skcipher_new(lua_State *L) {
+	const char *algname = luaL_checkstring(L, 1);
+	luacrypto_skcipher_t *tfm_ud;
+	gfp_t gfp = lunatik_gfp(lunatik_toruntime(L));
```
lneto: Is this being used?

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,261 @@
...
+	u8 *buf = luaL_buffinitsize(L, &work_buffer, data_s_len);
+	memcpy(buf, data_s, data_s_len);
+
+	sg_init_one(&sg_work, buf, data_s_len);
+
+	skcipher_request_set_crypt(tfm_ud->req, &sg_work, &sg_work,
+				   data_s_len, iv_data);
+	skcipher_request_set_callback(tfm_ud->req, 0, NULL, NULL);
+
+	if (encrypt) {
```
lneto: My point is to move this function inline to your template and completely avoid this if here

lneto: You can have all the common code that doesn't depend on anything in a inline function

lib/luacrypto_skcipher.c
jperon: Yes, `tfm_ud->req = skcipher_request_alloc(tfm_ud->tfm, gfp);`

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,248 @@
...
+	struct scatterlist sg_work; \
+\
+	u8 *buf = luaL_buffinitsize(L, &work_buffer, data_s_len); \
+	memcpy(buf, data_s, data_s_len); \
+\
+	sg_init_one(&sg_work, buf, data_s_len); \
+\
+	skcipher_request_set_crypt(tfm_ud->req, &sg_work, &sg_work, \
+				   data_s_len, iv_data); \
+	skcipher_request_set_callback(tfm_ud->req, 0, NULL, NULL); \
```
lneto: I think you misunderstood my point.. it's better to keep anything doesn't depend directly on `##name` in a static inline function.. and call it from the template..

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,248 @@
...
+	unsigned int expected_ivlen = crypto_skcipher_ivsize(tfm_ud->tfm); \
+	luaL_argcheck(L, iv_s_len == expected_ivlen, 2, "incorrect IV length"); \
+\
+	u8 *iv_data = lunatik_checkalloc(L, iv_s_len); \
+	memcpy(iv_data, iv_s, iv_s_len); \
+\
+	luaL_Buffer work_buffer; \
+	struct scatterlist sg_work; \
+\
+	u8 *buf = luaL_buffinitsize(L, &work_buffer, data_s_len); \
```
lneto: it might raise an error due to mem allocation.. in this case, you will leak iv_data..
https://www.lua.org/manual/5.4/manual.html#luaL_buffinitsize

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,248 @@
...
+* @function crypto_skcipher:blocksize
+* @treturn integer The block size in bytes.
+*/
+static int luacrypto_skcipher_blocksize(lua_State *L) {
+	luacrypto_skcipher_t *tfm_ud = luacrypto_skcipher_check(L, 1);
+	lua_pushinteger(L, crypto_skcipher_blocksize(tfm_ud->tfm));
+	return 1;
+}
+
+#define LUACRYPTO_SKCIPHER_CRYPT_FN(name) \
```
lneto: please mind our code style.. `\` should be aligned with tabs in the macro..

lib/luacrypto_skcipher.c
lneto: my suggestion is to create an aux inline function to prep request.. you can use try_ret to catch the error and then free iv_data, if it's the case

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,248 @@
...
+	const char *algname = luaL_checkstring(L, 1);
+	luacrypto_skcipher_t *tfm_ud;
+	gfp_t gfp = lunatik_gfp(lunatik_toruntime(L));
+	lunatik_object_t *object = lunatik_newobject(L, &luacrypto_skcipher_class, sizeof(luacrypto_skcipher_t));
+	tfm_ud = (luacrypto_skcipher_t *)object->private;
+
+	tfm_ud->tfm = crypto_alloc_skcipher(algname, 0, 0);
+	if (IS_ERR(tfm_ud->tfm)) {
+		long err = PTR_ERR(tfm_ud->tfm);
+		return luaL_error(L, "failed to allocate SKCIPHER transform for %s (err %ld)", algname, err);
```
lneto: we are not using this pattern `return luaL_error()` in lunatik because it raises warnings in kbuild if I'm not mistaken.. can you check if it's the case.. I'm not against using it in lunatik if it doesn't raise unnecessary warnings, but I would apply it globally not in a single module.. one of the things I think we should stick to is the uniformity of the code base.. I'm not against on changes as long as we keep the framework concise =)

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,248 @@
...
+	luacrypto_skcipher_t *tfm_ud = (luacrypto_skcipher_t *)private;
+	if (!tfm_ud)
+		return;
+
+	if (tfm_ud->req) {
+		skcipher_request_free(tfm_ud->req);
+		tfm_ud->req = NULL;
+	}
+
+	if (tfm_ud->tfm && !IS_ERR(tfm_ud->tfm))
```
lneto: is this condition, where tfm_ud->tfm and IS_ERR(tfm_ud->tfm) are both true, possible?

jperon: I think so, as the return of IS_ERR is a boolean: https://github.com/torvalds/linux/blob/66701750d5565c574af42bef0b789ce0203e3071/include/linux/err.h#L63-L71

Correct me if I understand it wrong.

lib/luacrypto_skcipher.c
jperon: I ended taking this approach: allocate iv_data after buf, just when needed, and freeing it if `crypt_func` fails.

lib/luacrypto_skcipher.c
lneto: my doubt was actually about in your own execution flow.. but now it's clear it's possible.. you raise an error on new after creating the object, so you'll end up here.. all good..

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,254 @@
...
+#include <lua.h>
+#include <lualib.h>
+#include <lauxlib.h>
+#include <lunatik.h>
+
+static const lunatik_class_t luacrypto_skcipher_class;
+
+typedef struct {
+	struct crypto_skcipher *tfm;
+	struct skcipher_request *req;
```
lneto: is the idea to have a single request per object? wondering if it wouldn't be the case to allocate it per request.. what's the common pattern in the kernel for using this API? can you describe pros and cons of allocating it per object instead of per request call? is it safe to reuse this structure on multiple requests?

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,254 @@
...
+	u8 *iv_data = lunatik_checkalloc(L, iv_s_len);
+	memcpy(iv_data, iv_s, iv_s_len);
+
+	skcipher_request_set_crypt(tfm_ud->req, &sg_work, &sg_work, data_s_len, iv_data);
+	skcipher_request_set_callback(tfm_ud->req, 0, NULL, NULL);
+
+	int ret = crypt_func(tfm_ud->req);
+	if (ret < 0) {
+		lunatik_free(iv_data);
+		lua_pushinteger(L, -ret);
```
lneto: I wouldn't `error(errno)`, I would rather use `error()` or `return false, errno`.. I'm not sure what's suite better in the usage of this API..

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,254 @@
...
+
+static inline int luacrypto_skcipher_crypt_common(lua_State *L, int (*crypt_func)(struct skcipher_request *req))
+{
+	luacrypto_skcipher_t *tfm_ud = luacrypto_skcipher_check(L, 1);
+	size_t iv_s_len, data_s_len;
+	const char *iv_s = luaL_checklstring(L, 2, &iv_s_len);
+	const char *data_s = luaL_checklstring(L, 3, &data_s_len); /* Data (plaintext or ciphertext) */
+
+	if (!tfm_ud->req)
+		luaL_error(L, "crypto_skcipher: TFM request not initialized (internal error)");
```
lneto: we don't need to prefix, right?

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,254 @@
...
+	if (IS_ERR(tfm_ud->tfm)) {
+		long err = PTR_ERR(tfm_ud->tfm);
+		luaL_error(L, "failed to allocate SKCIPHER transform for %s (err %ld)", algname, err);
+	}
+
+	tfm_ud->req = skcipher_request_alloc(tfm_ud->tfm, gfp);
+	if (!tfm_ud->req) {
+		/* tfm_ud->tfm is guaranteed not IS_ERR here. */
+		crypto_free_skcipher(tfm_ud->tfm);
+		luaL_error(L, "crypto_skcipher.new: failed to allocate kernel request for %s", algname);
```
lneto: we don't need to prefix

lib/luacrypto_skcipher.c
jperon: Thank you for pointing me to that: while checking, I saw that it could make it not thread-safe.

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,240 @@
...
+	const char *algname = luaL_checkstring(L, 1);
+	struct crypto_skcipher *tfm;
+	lunatik_object_t *object = lunatik_newobject(L, &luacrypto_skcipher_class, 0);
+
+	tfm = crypto_alloc_skcipher(algname, 0, 0);
+	if (IS_ERR(tfm)) {
+		long err = PTR_ERR(tfm);
+		luaL_error(L, "Failed to allocate SKCIPHER transform for %s (err %ld)", algname, err);
+	}
+	object->private = tfm;
```
lneto, line 212: Actually you don't need to check for PTR_ERR on release, right? As you are only updating priv if it's ok..

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,239 @@
...
+#include <crypto/skcipher.h>
+#include <linux/err.h>
+#include <linux/scatterlist.h>
+#include <linux/slab.h>
+
+#include <lua.h>
+#include <lualib.h>
+#include <lauxlib.h>
+#include <lunatik.h>
+
```
lneto, line 31: Please mind multiples `\n`

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,239 @@
...
+{
+	struct crypto_skcipher *tfm = luacrypto_skcipher_check(L, 1);
+	size_t iv_s_len, data_s_len;
+	const char *iv_s = luaL_checklstring(L, 2, &iv_s_len);
+	const char *data_s = luaL_checklstring(L, 3, &data_s_len); /* Data (plaintext or ciphertext) */
+
+	struct skcipher_request *req;
+	gfp_t gfp = lunatik_gfp(lunatik_toruntime(L));
+
+	req = skcipher_request_alloc(tfm, gfp);
```
lneto: I would try to move request alloc/prep to an aux function, if possible.. I think it would modularize better your code..

jperon: As said on Matrix, IMHO it’s not worth it _here_; I’ll think about it for aead.

lib/luacrypto_skcipher.c
```diff
@@ -0,0 +1,240 @@
...
+	buf = luaL_buffinitsize(L, &work_buffer, data_s_len);
+	memcpy(buf, data_s, data_s_len);
+	sg_init_one(&sg_work, buf, data_s_len);
+
+	expected_ivlen = crypto_skcipher_ivsize(tfm);
+	luaL_argcheck(L, iv_s_len == expected_ivlen, 2, "incorrect IV length");
+	iv_data = lunatik_checkalloc(L, iv_s_len);
+	memcpy(iv_data, iv_s, iv_s_len);
+
+	req = skcipher_request_alloc(tfm, gfp);
```
lneto:
```suggestion
	struct skcipher_request *req = skcipher_request_alloc(tfm, gfp);
```

we are using C99 declarations in the other modules.. I think it might reduce a little this function and make it more readable.. could you apply for all other vars as well? thanks

PR #290, opened 2025-07-01
crypto_aead

jperon:
Third part of divided #248


lib/crypto/aead.lua
```diff
@@ -0,0 +1,94 @@
...
+local sub = string.sub
+
+--- Prototype for AEAD instances.
+-- Objects of this type are created by `AeadModule.new()`.
+local Aead = {}
+
+--- Closes the cipher instance and releases underlying C resources.
+-- If not called explicitly, this will be called by the garbage collector.
+function Aead:close()
+	if self.c_tfm and self.c_tfm.__close then self.c_tfm:__close() end
```
lneto:
```suggestion
	if self.c_tfm then self.c_tfm:__close() end
```
I don't think we need to check for `self.c_tfm.__close`, `elf.c_tfm` if it is set, it comes from the C binding.. btw, I wouldn't prefix anything with `c_`.. `tfm` is good enough IMO

lib/crypto/aead.lua
```diff
@@ -0,0 +1,94 @@
...
+-- using the underlying 'crypto_aead' C module.
+--
+-- @classmod crypto.aead
+
+local new = require("crypto_aead").new
+local sub = string.sub
+
+--- Prototype for AEAD instances.
+-- Objects of this type are created by `AeadModule.new()`.
+local Aead = {}
```
lneto: on [HKDF](https://github.com/luainkernel/lunatik/blob/master/lib/crypto/hkdf.lua) we used all upper case.. let's agree on a standard..

lib/luacrypto_aead.c
```diff
@@ -0,0 +1,325 @@
...
+* @raise Error if setting the key fails (e.g., invalid key length for the algorithm).
+*/
+static int luacrypto_aead_setkey(lua_State *L) {
+	struct crypto_aead *tfm = luacrypto_aead_check(L, 1);
+	size_t keylen;
+	const char *key = luaL_checklstring(L, 2, &keylen);
+	int ret = crypto_aead_setkey(tfm, key, keylen);
+	if (ret < 0) {
+		luaL_error(L, "Failed to set AEAD key (err %d)", ret);
+	}
```
lneto:
```suggestion
	if (ret < 0)
		luaL_error(L, "Failed to set AEAD key (err %d)", ret);
```

please mind this style across this patch

lib/luacrypto_aead.c
```diff
@@ -0,0 +1,325 @@
...
+	struct aead_request *req = aead_request_alloc(tfm, gfp);
+	if (!req)
+		luaL_error(L, "Failed to allocate request");
+
+	aead_request_set_ad(req, aad_len);
+	aead_request_set_crypt(req, sg_in, sg_in, crypt_len, iv_data);
+	aead_request_set_callback(req, 0, NULL, NULL);
+
+	return req;
+}
```
lneto: I thought you have dropped this approach.. what changed? why using it here but not in the other module.. I feel like we could make a "generic" using macros

lib/luacrypto_aead.c
```diff
@@ -0,0 +1,325 @@
...
+	*/
+	luaL_argcheck(L, ciphertext_with_tag_len >= authsize_val, 3,
+		"input data (ciphertext+tag) too short for tag"
+	);
+	/* The length of this combined data is (aad_len + plaintext_len) */
+	luaL_addsize(&b, -(int)authsize_val); /* Remove the tag length from the buffer size */
+	luaL_pushresult(&b);
+	lunatik_free(iv_data);
+	return 1;
+}
```
lneto: why not using the macro template for encrypt/decrypt which we have discussed in the other PR.. my impression is that we could apply previous discussions to make this module more concise..

lib/luacrypto_aead.c
```diff
@@ -0,0 +1,325 @@
...
+
+	struct crypto_aead *tfm = crypto_alloc_aead(algname, 0, 0);
+	if (IS_ERR(tfm)) {
+		long err = PTR_ERR(tfm);
+		luaL_error(L, "Failed to allocate AEAD transform for %s (err %ld)", algname, err);
+	}
+	object->private = tfm;
+
+	return 1;
+}
```
lneto: I would take a look on the constructors of the other crypto modules.. perhaps we can make a generic as well..

lib/luacrypto_aead.c
jperon: Without the macro, the benefit was greater, avoiding code duplication. Now that I went for the macro solution, I came back to inlining it.

lib/luacrypto_aead.c
jperon: OK I ended going for this solution, though it was more complicated than for skcipher (as there were more differences between encrypt and decrypt).

lib/luacrypto.h
```diff
@@ -0,0 +1,67 @@
...
+ * @lunatik_class: The lunatik_class_t instance for this object type.
+ * @private_assign_expr: The expression to assign to object->private (e.g., tfm or sdesc).
+ * @extra_alloc_and_assign: Optional extra allocation and assignment logic (e.g., for shash_desc).
+ *
+ * This macro generates the `luacrypto_<_name>_new` function, which acts as the
+ * constructor for Lua crypto objects. It handles the common steps of checking
+ * the algorithm name, allocating the Lunatik object, allocating the kernel
+ * crypto transform, handling errors, and assigning the private data.
+ */
+#define LUACRYPTO_NEW(name, crypto_type, alloc_func, lunatik_class, private_assign_expr, extra_alloc_and_assign)	\
```
lneto:
```suggestion
#define LUACRYPTO_NEW(name, T, alloc, class, private, extra)	\
```

lib/luacrypto.h
```diff
@@ -0,0 +1,67 @@
...
+{															\
+	const char *algname = luaL_checkstring(L, 1);									\
+	lunatik_object_t *object = lunatik_newobject(L, &lunatik_class, 0);						\
+															\
+	crypto_type *tfm = alloc_func(algname, 0, 0);									\
+	if (IS_ERR(tfm)) {												\
+		long err = PTR_ERR(tfm);										\
+		luaL_error(L, "Failed to allocate " #name " transform for %s (err %ld)", algname, err);			\
+	}														\
+	extra_alloc_and_assign												\
```
lneto:
```suggestion
	extra()												\
```

lib/luacrypto.h
```diff
@@ -0,0 +1,67 @@
...
+ * @name: The base name of the crypto module (e.g., aead, skcipher, shash).
+ * @private_struct_type: The C struct type of the private data stored in the Lunatik object (e.g., struct crypto_aead, struct shash_desc).
+ * @free_private_func: The function to free the `private_struct_type` object itself (e.g., crypto_free_aead, lunatik_free).
+ * @extra_release_logic: Optional extra release logic, which can access `obj` (the casted `private_struct_type` pointer).
+ *
+ * This macro generates the `luacrypto_<_name>_release` function, which is used
+ * as the `release` callback for Lunatik objects. It handles freeing the primary
+ * private data and allows for additional custom cleanup.
+ */
+#define LUACRYPTO_RELEASE(name, private_struct_type, free_private_func, extra_release_logic)	\
```
lneto:
```suggestion
#define LUACRYPTO_RELEASE(name, T, release, extra)	\
```

lib/luacrypto.h
```diff
@@ -0,0 +1,67 @@
...
+static void luacrypto_##name##_release(void *private_ptr)					\
+{												\
+ private_struct_type *obj = (private_struct_type *)private_ptr;					\
+ if (obj) {											\
+ 	extra_release_logic									\
```
lneto:
```suggestion
 	extra()									\
```

lib/luacrypto_aead.c
```diff
@@ -0,0 +1,258 @@
...
+/* Auxiliary functions to define luacrypto_aead_encrypt and luacrypto_aead_decrypt */
+
+static inline size_t luacrypto_aead_get_encrypt_buffer_needed(lua_State *L, size_t combined_len, unsigned int authsize_val, size_t crypt_len) {
+	return combined_len + authsize_val;
+}
+
+static inline size_t luacrypto_aead_get_decrypt_buffer_needed(lua_State *L, size_t combined_len, unsigned int authsize_val, size_t crypt_len) {
+	luaL_argcheck(L, crypt_len >= authsize_val, 3, "input data (ciphertext+tag) too short for tag");
+	return combined_len;
+}
```
lneto: it seems to me that we don't need these aux funcs

jperon: I use them to avoid `if (is_encrypt)` within the main function. If there is a way to do otherwise with the main macro, I don’t know it.

lib/luacrypto_aead.c
```diff
@@ -0,0 +1,258 @@
...
+	luaL_pushresult(&b);
+	lunatik_free(iv_data);
+	return 1;
+}
+
+#define LUACRYPTO_AEAD_CRYPT_FN(name, crypt_func_ptr, get_buffer_needed_func_ptr, result_adjust_factor_val)		\
+static int luacrypto_aead_##name(lua_State *L) {									\
+	return luacrypto_aead_common_crypt(L, crypt_func_ptr, get_buffer_needed_func_ptr, result_adjust_factor_val);	\
+}
+
```
lneto: as you aren't using `name` anywhere, you don't actual need the "template".. I would rather have two separate functions here for encrypt/decrypt and just call common() passing whatever should be get_buffer_needed_func() result for each, as the actual value..

btw, I think `luacrypto_aead_request` is a better name for `common`.. as a general rule, I would a void very large names..

lib/luacrypto_aead.c
```diff
@@ -0,0 +1,258 @@
...
+	const char *iv = luaL_checklstring(L, 2, &iv_len);
+	const char *combined = luaL_checklstring(L, 3, &combined_len);
+	lua_Integer aad_l = luaL_checkinteger(L, 4);
+
+	unsigned int expected_ivlen = crypto_aead_ivsize(tfm);
+	luaL_argcheck(L, iv_len == expected_ivlen, 2, "incorrect IV length");
+	u8 *iv_data = lunatik_checkalloc(L, iv_len);
+	memcpy(iv_data, iv, iv_len);
+
+	luaL_argcheck(L, aad_l >= 0 && (size_t)aad_l <= combined_len, 4, "AAD length out of bounds");
```
lneto, line 118: please notice, that there are a bunch of potential error cases like this one where you are going to leak `iv_data`.. I've pointed a very similar situation in your previous PR on the usage of luaL_Buffer.. I think you should be more careful about this.. can you also take the opportunity and review the already merged crypto libs to make sure we don't have similar conditions there? thanks!

lib/luacrypto_aead.c
lneto: I've made a suggestion here https://github.com/luainkernel/lunatik/pull/290/files#r2185710812

lib/crypto/aead.lua
```diff
@@ -0,0 +1,94 @@
...
+-- @usage
+--  local aead = require("crypto.aead")
+--  local gcm_aes = aead.new("gcm(aes)")
+function AEAD.new(algname)
+	return setmetatable({tfm = new(algname)}, AEAD)
+end
+
+--- Sets the encryption key.
+-- @tparam string key The encryption key.
+-- @raise Error if C operation fails.
```
lneto: I think it's too vague.. not a blocker for merging, of course, but if we can be more specific here it would be better.. (quite minor comment)

lib/luacrypto_aead.c
```diff
@@ -0,0 +1,258 @@
...
+
+#include <lua.h>
+#include <lualib.h>
+#include <lauxlib.h>
+#include <lunatik.h>
+#include "luacrypto.h"
+
+LUNATIK_PRIVATECHECKER(luacrypto_aead_check, struct crypto_aead *);
+
+LUACRYPTO_RELEASE(aead, struct crypto_aead, crypto_free_aead, );
```
lneto: do we need this comma at the end? it looks odd to me.. if we need, probably it's better to use vararg.. there's some examples on lunatik.h

lib/luacrypto_aead.c
```diff
@@ -0,0 +1,258 @@
...
+	sg_init_one(&sg_work, buf, total_buffer_needed);
+
+	gfp_t gfp = lunatik_gfp(lunatik_toruntime(L));
+	struct aead_request *req = aead_request_alloc(tfm, gfp);
+	if (!req) {
+		lunatik_free(iv_data);
+		luaL_error(L, "Failed to allocate request");
+	}
+	aead_request_set_ad(req, aad_len);
+	aead_request_set_crypt(req, &sg_work, &sg_work, crypt_len, iv_data);
```
lneto, line 144: you only use iv_data here, right? you can allocate it just before.. btw, why do you need to memcpy this? why can't you just use iv directly here?

lib/luacrypto_aead.c
jperon: I’ve begun implementing your suggestion, but I can’t pass the total_buffer_needed as the actual value, as it gets calculated from combined_len and others that are known only within `request`. I could deport this from `request` into each function, but this would lead to code duplication. IMHO, those auxiliary functions (which get inlined at compilation, if I understand it well?) are the most elegant way to avoid code duplication and conditional blocks.

PR #291, opened 2025-07-02
Add crypto_skcipher to documentation

jperon:
Two commits, one for skcipher doc, another one to fix code style.


config.ld
```diff
@@ -13,29 +13,30 @@ title = 'Lunatik API Documentation'
...
-    './lib/luasocket.c',
-    './lib/socket/inet.lua',
-    './lib/luasyscall.c',
-    './lib/syscall/table.lua',
-    './lib/luathread.c',
-    './lib/luaxdp.c',
-    './lib/luaxtable.c',
+	'./lib/luacompletion.c',
+	'./lib/luacrypto_shash.c',
+	'./lib/luacrypto_skcipher.c',
```
lneto, line 18: This isn't right, right?

lneto: (perhaps that's because I viewing it on my phone.. will take a better look later)

PR #294, opened 2025-07-02
crypto_comp

jperon:
Last part of divided #248


lib/luacrypto_comp.c
```diff
@@ -0,0 +1,135 @@
...
+* These methods are available on COMP objects created by `crypto_comp.new()`.
+* @type crypto_comp
+*/
+
+#define LUACRYPTO_COMP_OPERATION(name)									\
+static int luacrypto_comp_##name(lua_State *L) {							\
+	struct crypto_comp *tfm = luacrypto_comp_check(L, 1);						\
+	size_t datalen;											\
+	const u8 *data = (const u8 *)luaL_checklstring(L, 2, &datalen);					\
+	luaL_argcheck(L, datalen > 0 && datalen <= UINT_MAX, 2, "out of bounds");			\
```
lneto: checkuint?

lneto: ah.. you don't check for int here.. I think it would be worth to have something as.. `lunatik_checkbounds()` and use both here and on checkuint.. perhaps on luadata as well..

lneto (approved):

lunatik.h
```diff
@@ -322,10 +322,15 @@ do {									\
 	lua_pop(L, 1);							\
 } while (0)
 
+static inline void lunatik_checkbounds(lua_State *L, int idx, lua_Integer val, lua_Integer min, lua_Integer max)
+{
+	luaL_argcheck(L, val > min && val <= max, idx, "out of bounds");
+}
```
lneto: I think in that case, the inline is a bit too much.. I would just use a macro here..
```suggestion
#lunatik_checkbounds(L, idx, val, min, max)	luaL_argcheck(L, val >= min && val <= max, idx, "out of bounds")
```
I would use `>=` to be consistent with `<=`, so, both boundaries are inclusive.. then, we can pass `1` instead of `0` on the call..

lib/luacrypto_comp.c
```diff
@@ -0,0 +1,135 @@
...
+* These methods are available on COMP objects created by `crypto_comp.new()`.
+* @type crypto_comp
+*/
+
+#define LUACRYPTO_COMP_OPERATION(name)									\
+static int luacrypto_comp_##name(lua_State *L) {							\
+	struct crypto_comp *tfm = luacrypto_comp_check(L, 1);						\
+	size_t datalen;											\
+	const u8 *data = (const u8 *)luaL_checklstring(L, 2, &datalen);					\
+	lunatik_checkbounds(L, 2, datalen, 0, UINT_MAX);							\
```
lneto:
```suggestion
	lunatik_checkbounds(L, 2, datalen, 1, UINT_MAX);						\
```

please, mind the `\` alignment

lunatik.h
```diff
@@ -322,10 +322,15 @@ do {									\
...
+
 static inline unsigned int lunatik_checkuint(lua_State *L, int idx)
 {
 	lua_Integer val = luaL_checkinteger(L, idx);
-	luaL_argcheck(L, val > 0 && val <= UINT_MAX, idx, "out of bounds");
+	lunatik_checkbounds(L, idx, val, 0, UINT_MAX);
```
```suggestion
	lunatik_checkbounds(L, idx, val, 1, UINT_MAX);
```

lneto (approved):

PR #295, opened 2025-07-04
add debian packaging submodule

ShivamVashisth28:
This PR adds the lunatik_packages repo as the submodule of lunatik .

@marcelstanley

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "klibc"]
 	path = klibc
 	url = https://github.com/luainkernel/klibc.git
+[submodule "lunatik_packages"]
+	path = lunatik_packages
+	url = https://github.com/luainkernel/lunatik_packages.git


lneto:
Why do we need this submodule?

ShivamVashisth28:
> Why do we need this submodule?

1. If someone wants to build the package from the source it self then they would require it.
2. @marcelstanley suggested for going with this approach

lneto:
I think you probably misunderstood him.. I don't see why in this case lunatik should have a submodule for the package repo.. it should be the other way around.. is the package that needs lunatik.. not the opposite

lneto:
See our OpenWRT feed

marcelstanley:
> I think you probably misunderstood him.. I don't see why in this case lunatik should have a submodule for the package repo.. it should be the other way around.. is the package that needs lunatik.. not the opposite

That's on me, @lneto. Let's close this issue and do the other way around, @ShivamVashisth28

PR #296, opened 2025-07-05
crypto_aead

jperon:
Supersedes #290


lib/luacrypto.h
```diff
@@ -0,0 +1,67 @@
...
+ * This macro generates the `luacrypto_<_name>_release` function, which is used
+ * as the `release` callback for Lunatik objects. It handles freeing the primary
+ * private data and allows for additional custom cleanup.
+ */
+#define LUACRYPTO_RELEASE(name, T, release, ...)		\
+static void luacrypto_##name##_release(void *private_ptr)	\
+{								\
+ T *obj = (T *)private_ptr;					\
+ if (obj) {							\
+	__VA_ARGS__						\
```
lneto: This looks weird to me.. I would instead, pass varargs to a function passed as another argument for this macro

lib/luacrypto_aead.c
```diff
@@ -0,0 +1,260 @@
...
+															\
+	LUACRYPTO_AEAD_CHECK_##NAME(L, 3, crypt_len, authsize);								\
+	size_t buffer_len = LUACRYPTO_AEAD_LEN_##NAME(combined_len, authsize);						\
+															\
+	luaL_Buffer B;													\
+	char *buffer = luaL_buffinitsize(L, &B, buffer_len);								\
+	luacrypto_aead_setrequest(L, request, combined, iv, combined_len, aad_len, crypt_len, buffer, buffer_len);	\
+	int ret = crypto_aead_##name(request);										\
+	luacrypto_aead_freerequest(request, iv);									\
+	if (ret < 0)													\
```
lneto, line 158: Wouldn't it be the case for lunatik_tryret?

lib/luacrypto_skcipher.c
```diff
@@ -82,52 +80,60 @@ static int luacrypto_skcipher_blocksize(lua_State *L) {
...
-	}
+	skcipher_request_set_crypt(request, &sg_work, &sg_work, data_len, iv);
+	skcipher_request_set_callback(request, 0, NULL, NULL);
+}
 
-	luaL_pushresultsize(&work_buffer, data_len);
-	skcipher_request_free(req);
-	lunatik_free(iv_data);
-	return 1;
+static inline void luacrypto_skcipher_freerequest(struct skcipher_request *request, u8 *iv)
```
lneto: Perhaps we could e move these to luacrypto.h as a template

lib/luacrypto.h
```diff
@@ -0,0 +1,67 @@
...
+{													\
+	const char *algname = luaL_checkstring(L, 1);							\
+	lunatik_object_t *object = lunatik_newobject(L, &class, 0);					\
+													\
+	T *tfm = alloc(algname, 0, 0);									\
+	if (IS_ERR(tfm)) {										\
+		long err = PTR_ERR(tfm);								\
+		luaL_error(L, "Failed to allocate " #name " transform for %s (err %ld)", algname, err);	\
+	}												\
+	__VA_ARGS__											\
```
lneto: It looks a weird usage of varagargs.. I've made my point in another comment.. as general rule, I don't think it is a good idea to use any macro's argument like this.. I think it's always better to threat like functions if it's not an assignment..

lib/luacrypto_aead.c
jperon: I don’t think so: this would lead to that code:
```C
lunatik_tryret(L, ret, crypto_aead_##name, request);
luacrypto_aead_freerequest(request, iv);
```
Which would leak `request` and `iv` if `crypto_aead_##name` fails.

lib/luacrypto.h
jperon: If so, IMHO it would be better to keep this macro without the last argument for common cases, and to write the full function in specific cases.

lneto: To be clear.. I'm suggesting something like..
```
#define LUACRYPTO_NEW(name, T, alloc, class, priv, new, ...)
...
new(__VAR_ARGS__);
```

jperon: But doing that would imply:
- either seeing again the `, )` at the end of `LUACRYPTO_NEW` calls when there is no extra part;
- or defining a "nop" function in those cases.
Is this what you intend?

jperon: I ended with a solution using two macros for each, with a _CUSTOM version allowing to call a function for extra logic.

lib/luacrypto.h
```diff
@@ -0,0 +1,92 @@
...
+ * SPDX-FileCopyrightText: (c) 2025 jperon <cataclop@hotmail.com>
+ * SPDX-License-Identifier: MIT OR GPL-2.0-only
+ */
+
+#ifndef _LUACRYPTO_H
+#define _LUACRYPTO_H
+
+#include <linux/err.h>
+#include <lua.h>
+#include <lauxlib.h>
```
lneto, line 11: lunatik.h includes both, right?

jperon: Sure. I checked, and this redundancy is found in many other files. Would you mind if I submitted a general PR about it after merge?

lneto: not at all.. it's a leftover from old times.. feel free to update the code base accordingly =)

jperon: So I’ll change it here at the same time.

lib/luacrypto.h
```diff
@@ -0,0 +1,92 @@
...
+ * @class: The lunatik_class_t instance for this object type.
+ * @private: The expression to assign to object->private (e.g., tfm or sdesc).
+ * @define_priv: Optional extra allocation and assignment logic (e.g., for shash_desc).
+ *
+ * This macro generates the `luacrypto_<_name>_new` function, which acts as the
+ * constructor for Lua crypto objects. It handles the common steps of checking
+ * the algorithm name, allocating the Lunatik object, allocating the kernel
+ * crypto transform, handling errors, and assigning the private data.
+ */
+#define LUACRYPTO_NEW_CUSTOM(name, T, alloc, class, define_priv, ...)					\
```
lneto: we can use the same strategy from here (a GCC extension) to get rid of `, `:
https://github.com/luainkernel/lunatik/blob/master/lunatik.h#L50C17-L50C21

lneto: see https://gcc.gnu.org/onlinedocs/gcc-3.0.2/gcc_5.html#SEC82

jperon: I saw this, but it would imply getting both the name of `define_priv` function and its arguments from `__VA_ARGS__`. Moreover, IMHO it’s not bad that one has to explicitly use _CUSTOM when adding custom logic, as it makes it more explicit. What do you think ?

lib/luacrypto_shash.c
```diff
@@ -26,22 +26,18 @@
...
-		tfm = sdesc->tfm;
-		lunatik_free(sdesc);
-		if (tfm)
-			crypto_free_shash(tfm);
-	}
+	if (obj->tfm)
+		crypto_free_shash(obj->tfm);
 }
 
+LUACRYPTO_RELEASE_CUSTOM(shash, struct shash_desc, lunatik_free, luacrypto_shash_release_tfm, obj);
```
lneto: what's `obj` here? it's confusing.. I think a better approach is to pass only a single release callback function (which receives prive) to `LUACRYPTO_RELEASE` and define a inline for this module that ends up calling `lunatik_free()` and `luacrypto_shash_release_tfm`.. IMHO, the code will be way cleaner

lib/luacrypto.h
```diff
@@ -0,0 +1,92 @@
...
+ * @release_priv: Optional extra release logic, which can access `obj` (the casted `T` pointer).
+ *
+ * This macro generates the `luacrypto_<_name>_release` function, which is used
+ * as the `release` callback for Lunatik objects. It handles freeing the primary
+ * private data and allows for additional custom cleanup.
+ */
+#define LUACRYPTO_RELEASE_CUSTOM(name, T, release, release_priv, ...)		\
+static void luacrypto_##name##_release(void *private_ptr)			\
+{										\
+ T *obj = (T *)private_ptr;							\
```
lneto: let's keep the names used in other modules.. we normally call it `private` if I'm not mistaken.. when in doubt, rely on the code base.. if you would like to change, propose the name changing as a whole..

lib/luacrypto.h
```diff
@@ -0,0 +1,92 @@
...
+	object->private = define_priv(__VA_ARGS__);							\
+													\
+	return 1;											\
+}
+
+static inline void *define_priv(void *tfm) {
+	return tfm;
+}
+
+#define LUACRYPTO_NEW(name, T, alloc, class) LUACRYPTO_NEW_CUSTOM(name, T, alloc, class, define_priv, tfm)
```
lneto: I think a better approach is to always expect the inline and let the submodule to define it.. but let's please call it `new` instead of `define_priv`.. so, if a module doesn't need it, it can pass NULL and we just assign tfm to priv.. I think it would be cleaner and more alike the current code base..

lib/luacrypto.h
lneto: btw, we used to suffix this sort of template with `ER` in our code base.. `LUACRYPTO_RELEASER` in this case

lib/luacrypto.h
jperon: You mean that (for example) aead would call it that way ?

```C
LUACRYPTO_NEW(aead, struct crypto_aead, crypto_alloc_aead, luacrypto_aead_class, NULL);
```

But how would I handle that in the code? Calling `NULL(__VA_ARGS__)` would not compile.

lib/luacrypto.h
```diff
@@ -0,0 +1,88 @@
...
+static void luacrypto_##name##_release(void *private)				\
+{										\
+ T *obj = (T *)private;								\
+ if (obj) {									\
+	void (*_release_priv)(T *) = (void (*)(T *))release_priv;		\
+	if (_release_priv)							\
+		_release_priv(obj);						\
+	release(obj);								\
+ }										\
+}
```
lneto: tabs, right?

lib/luacrypto.h
```diff
@@ -0,0 +1,88 @@
...
+{													\
+	const char *algname = luaL_checkstring(L, 1);							\
+	lunatik_object_t *object = lunatik_newobject(L, &class, 0);					\
+													\
+	T *tfm = alloc(algname, 0, 0);									\
+	if (IS_ERR(tfm)) {										\
+		long err = PTR_ERR(tfm);								\
+		luaL_error(L, "Failed to allocate " #name " transform for %s (err %ld)", algname, err);	\
+	}												\
+	void *(*_new)(lua_State *, T *) = (void *(*)(lua_State *, T *))new;				\
```
lneto: it might be worth the typedef to avoid repeating it.. so you can just do this bellow:
```
if (new)
	object->private = (luacrypto_new_t)new(L, tfm);
```
 the same applies to release

lib/luacrypto.h
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * This macro generates the `luacrypto_<_name>_freerequest` function, which handles
+ * freeing the crypto request and the associated IV buffer.
+ */
+#define LUACRYPTO_FREEREQUEST(name, request_T, free_request_fn) \
+static inline void luacrypto_##name##_freerequest(request_T *request, u8 *iv) \
+{ \
+	free_request_fn(request); \
+	lunatik_free(iv); \
+}
```
lneto: <tab>*\

lib/luacrypto.h
jperon: OK. That said, I saw that [kernel coding style](https://www.kernel.org/doc/html/latest/process/coding-style.html#typedefs) quite strongly dislikes typedefs. I think this case would go within their c) or d).

jperon: I tried the suggested fix, but it raised warnings and errors as below. So I added the typedefs, but kept the current pattern otherwise.

```C
In file included from lib/luacrypto_shash.c:29:
lib/luacrypto_shash.c: In function ‘luacrypto_shash_new’:
lib/luacrypto_shash.c:250:86: warning: the address of ‘luacrypto_shash_new_sdesc’ will always evaluate as ‘true’ [-Waddress]
  250 | LUACRYPTO_NEW(shash, struct crypto_shash, crypto_alloc_shash, luacrypto_shash_class, luacrypto_shash_new_sdesc);
      |                                                                                      ^~~~~~~~~~~~~~~~~~~~~~~~~
lib/luacrypto.h:42:13: note: in definition of macro ‘LUACRYPTO_NEW’
   42 |         if (new)                                                                                        \
      |             ^~~
  CC [M]  lib/luacrypto_skcipher.o
In file included from lib/luacrypto_skcipher.c:31:
lib/luacrypto_skcipher.c: In function ‘luacrypto_skcipher_new’:
/usr/src/linux-headers-6.13-common/include/linux/stddef.h:8:14: error: called object is not a function or function pointer
    8 | #define NULL ((void *)0)
      |              ^
lib/luacrypto.h:43:52: note: in definition of macro ‘LUACRYPTO_NEW’
   43 |                 object->private = (luacrypto_new_t)new(L, tfm);                                         \
      |                                                    ^~~
lib/luacrypto_skcipher.c:202:98: note: in expansion of macro ‘NULL’
  202 | LUACRYPTO_NEW(skcipher, struct crypto_skcipher, crypto_alloc_skcipher, luacrypto_skcipher_class, NULL);
      |                                                                                                  ^~~~
```

lib/luacrypto.h
```diff
@@ -0,0 +1,91 @@
...
+	lunatik_object_t *object = lunatik_newobject(L, &class, 0);					\
+													\
+	T *tfm = alloc(algname, 0, 0);									\
+	if (IS_ERR(tfm)) {										\
+		long err = PTR_ERR(tfm);								\
+		luaL_error(L, "Failed to allocate " #name " transform for %s (err %ld)", algname, err);	\
+	}												\
+	luacrypto_new_t _new = (luacrypto_new_t)new;							\
+	if (_new)											\
+		object->private = _new(L, tfm);								\
```
lneto, line 44:
```suggestion
	if (new)											\
		object->private = ((luacrypto_new_t)(new))(L, tfm);								\
```

lib/luacrypto.h
```diff
@@ -0,0 +1,91 @@
...
+ * This macro generates the `luacrypto_<_name>_release` function, which is used
+ * as the `release` callback for Lunatik objects. It handles freeing the primary
+ * private data and allows for additional custom cleanup.
+ */
+#define LUACRYPTO_RELEASER(name, T, obj_free, priv_free)			\
+static void luacrypto_##name##_release(void *private)				\
+{										\
+	T *obj = (T *)private;							\
+	if (obj) {								\
+		luacrypto_free_t _priv_free = (luacrypto_free_t)priv_free;	\
```
lneto, line 67: see https://github.com/luainkernel/lunatik/pull/296/files#r2194880040

lib/luacrypto.h
lneto: > OK. That said, I saw that [kernel coding style](https://www.kernel.org/doc/html/latest/process/coding-style.html#typedefs) quite strongly dislikes typedefs. I think this case would go within their c) or d).

yeah.. on this we are sticking to Lua.. hehe.. sorry about that, but Lunatik uses typedef extensively; personally, I think it really improves readability.. however, we can discuss this change..

lneto: > I tried the suggested fix, but it raised warnings and errors as below. So I added the typedefs, but kept the current pattern otherwise.
> 
> ```c
> In file included from lib/luacrypto_shash.c:29:
> lib/luacrypto_shash.c: In function ‘luacrypto_shash_new’:
> lib/luacrypto_shash.c:250:86: warning: the address of ‘luacrypto_shash_new_sdesc’ will always evaluate as ‘true’ [-Waddress]
>   250 | LUACRYPTO_NEW(shash, struct crypto_shash, crypto_alloc_shash, luacrypto_shash_class, luacrypto_shash_new_sdesc);
>       |                                                                                      ^~~~~~~~~~~~~~~~~~~~~~~~~
> lib/luacrypto.h:42:13: note: in definition of macro ‘LUACRYPTO_NEW’
>    42 |         if (new)                                                                                        \
>       |             ^~~
>   CC [M]  lib/luacrypto_skcipher.o
> In file included from lib/luacrypto_skcipher.c:31:
> lib/luacrypto_skcipher.c: In function ‘luacrypto_skcipher_new’:
> /usr/src/linux-headers-6.13-common/include/linux/stddef.h:8:14: error: called object is not a function or function pointer
>     8 | #define NULL ((void *)0)
>       |              ^
> lib/luacrypto.h:43:52: note: in definition of macro ‘LUACRYPTO_NEW’
>    43 |                 object->private = (luacrypto_new_t)new(L, tfm);                                         \
>       |                                                    ^~~
> lib/luacrypto_skcipher.c:202:98: note: in expansion of macro ‘NULL’
>   202 | LUACRYPTO_NEW(skcipher, struct crypto_skcipher, crypto_alloc_skcipher, luacrypto_skcipher_class, NULL);
>       |                                                                                                  ^~~~
> ```

I've made a suggestion, but the current looks good as well.. I just think using the typecast directly will reduce a bit the code surface making it slightly easier to read..

lib/luacrypto.h
```diff
@@ -0,0 +1,91 @@
...
+#define LUACRYPTO_NEW(name, T, alloc, class, new)							\
+static int luacrypto_##name##_new(lua_State *L)								\
+{													\
+	const char *algname = luaL_checkstring(L, 1);							\
+	lunatik_object_t *object = lunatik_newobject(L, &class, 0);					\
+													\
+	T *tfm = alloc(algname, 0, 0);									\
+	if (IS_ERR(tfm)) {										\
+		long err = PTR_ERR(tfm);								\
+		luaL_error(L, "Failed to allocate " #name " transform for %s (err %ld)", algname, err);	\
```
lneto, line 40: can you review all your error strings? I've just notice that you are starting them with upper case and we don't do this in the other modules.. it would be good to take the opportunity to take a look in the other libraries and align with their error messages if you find similarities on your cases.. whatever it applies.. (not a strong requirement to merge this though).. BTW, I think we should have some error strings defined globally for the most common cases at least in the futuer..

lib/luacrypto_aead.c
```diff
@@ -0,0 +1,256 @@
...
+* Encrypts data using the AEAD transform.
+* The IV (nonce) must be unique for each encryption operation with the same key.
+* @function crypto_aead:encrypt
+* @tparam string iv The Initialization Vector (nonce). Its length must match `ivsize()`.
+* @tparam string combined_data A string containing AAD (Additional Authenticated Data) concatenated with the plaintext (format: AAD || Plaintext).
+* @tparam integer aad_len The length of the AAD part in `combined_data`.
+* @treturn string The encrypted data, formatted as (AAD || Ciphertext || Tag).
+* @raise Error on encryption failure, incorrect IV length, or allocation issues.
+*/
+LUACRYPTO_AEAD_CRYPT_FN(encrypt, ENCRYPT, 1);
```
lneto: we are not suffixing templates with `FN`.. the common pattern in our code base is like this:
```suggestion
LUACRYPTO_AEAD_NEWCRYPT(encrypt, ENCRYPT, 1);
```

lneto (commented):
I think after this round we are good to merge.. I tried to catch memory leaks, but I confess my eyes are a bit tired.. it would be good if @sheharyaar can take a look at it as well.. thank you for this work!

lib/luacrypto.h
jperon: It raises following:

```
In file included from lib/luacrypto_shash.c:29:
lib/luacrypto_shash.c: In function ‘luacrypto_shash_new’:
lib/luacrypto_shash.c:250:86: warning: the address of ‘luacrypto_shash_new_sdesc’ will always evaluate as ‘true’ [-Waddress]
  250 | LUACRYPTO_NEW(shash, struct crypto_shash, crypto_alloc_shash, luacrypto_shash_class, luacrypto_shash_new_sdesc);
      |                                                                                      ^~~~~~~~~~~~~~~~~~~~~~~~~
lib/luacrypto.h:42:13: note: in definition of macro ‘LUACRYPTO_NEW’
   42 |         if (new)                                                                                        \
      |             ^~~
lib/luacrypto.h:43:36: warning: function called through a non-compatible type
   43 |                 object->private = ((luacrypto_new_t)(new))(L, tfm);                                     \
      |                                   ~^~~~~~~~~~~~~~~~~~~~~~~
lib/luacrypto_shash.c:250:1: note: in expansion of macro ‘LUACRYPTO_NEW’
  250 | LUACRYPTO_NEW(shash, struct crypto_shash, crypto_alloc_shash, luacrypto_shash_class, luacrypto_shash_new_sdesc);
      | ^~~~~~~~~~~~~
```

lib/luacrypto.h
jperon: Fixed for aead (BTW I noticed 2 `lunatik_try` use cases). If you don’t mind, I’ll keep this kind of changes for shash and skcipher for a last code style PR after last crypto PR are merged.

lneto (approved):

tests/crypto/aead.lua
```diff
@@ -0,0 +1,99 @@
...
+-- SPDX-FileCopyrightText: (c) 2025 jperon <cataclop@hotmail.com>
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+local aead = require"crypto.aead"
+local util = require("util")
+local test = util.test
+local hex2bin = util.hex2bin
+local bin2hex = util.bin2hex
+local EINVAL = require("linux").errno.INVAL
+local EBADMSG = 74
```
lneto: btw, if it's not present on lualinux, we should add it there, right?

lneto (approved):

jperon:
Sorry for last pushes: I had overseen a duplicate when rebasing.

lib/luacrypto_skcipher.c
```diff
@@ -28,15 +28,11 @@
 #include <lualib.h>
 #include <lauxlib.h>
 #include <lunatik.h>
+#include "luacrypto.h"
```
lneto, line 31:
```suggestion

#include "luacrypto.h"
```

jperon:
Done.

PR #297, opened 2025-07-07
add handler for probe

qrsikno2:
This pull request introduces significant enhancements to the `lib/luahid.c` file to improve HID (Human Interface Device) driver functionality. The changes include adding new dependencies, extending the `luahid_t` structure, implementing a `probe` mechanism for device initialization, and modifying the HID driver registration process to support Lua-based device handling.

### Enhancements to HID driver functionality:

* **Added new dependencies for Lua and kernel utilities**:
  Included `linux/printk.h` and `lua.h` headers to enable logging and Lua integration in the HID driver. 

* **Extended `luahid_t` structure**:
  Added a `devtable` member to store device-specific private data, improving the driver's ability to manage multiple devices. 

### Implementation of device initialization (`probe` mechanism):

* **Introduced `luahid_probe` and `luahid_probe_handler`**:
  Implemented a `probe` function and its Lua-based handler to initialize devices using Lua scripts. The handler retrieves device information, calls an `init` function defined in Lua, and sets device-specific data (`drvdata`).

* **Lua example for device initialization**:
  Added a sample Lua function (`hid_driver:init`) to demonstrate how device-specific features can be set based on `devinfo`.

### Modifications to HID driver registration:

* **Updated `luahid_register` function**:
  Integrated the `probe` mechanism into the HID driver registration process. This includes setting up the Lua registry for device handling and ensuring proper cleanup in case of registration failure.

### DIFF
--- a/lib/luahid.c
+++ b/lib/luahid.c
@@ -49,12 +49,12 @@ static int luahid_register(lua_State *L);
 
 static const luaL_Reg luahid_lib[] = {
 	{"register", luahid_register},
-	{NULL, NULL}
+	{NULL, NULL},
 };
 
 static const luaL_Reg luahid_mt[] = {
 	{"__gc", lunatik_deleteobject},
-	{NULL, NULL}
+	{NULL, NULL},
 };
 
 static const lunatik_class_t luahid_class = {
@@ -136,7 +136,7 @@ static int luahid_register(lua_State *L)
 
 	lunatik_checkfield(L, 1, "id_table", LUA_TTABLE);
 	luaL_argcheck(L, (user_driver->id_table = luahid_setidtable(L, -1)) != NULL,
-			   2, "invaild id_table");
+		      2, "invaild id_table");
 
 	lunatik_setruntime(L, hid, hid);
 	lunatik_getobject(hid->runtime);
--- a/lib/luahid.c
+++ b/lib/luahid.c
@@ -93,6 +93,66 @@ out:
 	return user_table;
 }
 
+#define luahid_setfield(L, idx, obj, field)	\
+do { 						\
+	lua_pushinteger(L, obj->field);		\
+	lua_setfield(L, idx - 1, #field);	\
+} while (0)
+
+static inline void luahid_pushdevid(lua_State *L, int idx, const struct hid_device_id *device_id)
+{
+	lua_newtable(L);
+	luahid_setfield(L, -1, device_id, bus);
+	luahid_setfield(L, -1, device_id, group);
+	luahid_setfield(L, -1, device_id, vendor);
+	luahid_setfield(L, -1, device_id, product);
+	luahid_setfield(L, -1, device_id, driver_data);
+}
+
+#define luahid_checkdriver(L, hid, idx, field) (lunatik_getregistry(L, hid) != LUA_TTABLE || \
+	lua_getfield(L, idx, "ops") != LUA_TTABLE || lua_getfield(L, idx - 1, field) != LUA_TTABLE)
+
+static int luahid_doprobe(lua_State *L, luahid_t *hid, struct hid_device *hdev, const struct hid_device_id *id)
+{
+	if (luahid_checkdriver(L, hid, -1, "_info")) {
+		pr_err("probe: couldn't find driver");
+		return -ENXIO;
+	}
+
+	if (lua_getfield(L, -2, "probe") != LUA_TFUNCTION)
+		return 0;
+
+	lua_pushvalue(L, -3); /* hid.ops */
+	luahid_pushdevid(L, -4, id);
+
+	if (lua_pcall(L, 2, 1, 0) != LUA_OK) {
+		pr_err("probe: %s\n", lua_tostring(L, -1));
+		return -ECANCELED;
+	}
+
+	if (lua_type(L, -1) == LUA_TTABLE) {
+		lua_pushlightuserdata(L, hdev);
+		lua_pushvalue(L, -2); /* returned table */
+		lua_settable(L, -4); /* hid._info[hdev] = returned table */
+
+		hid_set_drvdata(hdev, hid);
+	}
+	return 0;
+}
+
+static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
+{
+	struct hid_driver *driver = hdev->driver;
+	luahid_t *hid = container_of(driver, luahid_t, driver);
+	int ret;
+
+	lunatik_run(hid->runtime, luahid_doprobe, ret, hid, hdev, id);
+	if (ret != 0 || (ret = hid_parse(hdev)) != 0)
+		return ret;
+
+	return hid_hw_start(hdev, HID_CONNECT_DEFAULT);
+}
+
 /***
 * Registers a new HID driver.
 * This function creates a new HID driver object from a Lua table and registers it with the kernel.
@@ -120,11 +180,29 @@ out:
 *	 		{ bus = 2, vendor = 0x4321, product = 0x8765 },
 *	 	}
 *	 }
+*
+*	 function hid_driver:probe(devid)
+*	 	drvdata = {}
+*	 	if devid.bus == 1 then
+*	 		drvdata.usb_feature = true
+*	 	end
+*	 	if devid.driver_data % 2 == 0 then
+*	 		drvdata.even_special_feature = true
+*	 	end
+*	 	return drvdata
+*	 end
+*
 *	 hid.register(hid_driver)
 */
 static int luahid_register(lua_State *L)
 {
 	luaL_checktype(L, 1, LUA_TTABLE);
+
+	lua_newtable(L); /* hid = {} */
+	lua_pushvalue(L, 1);
+	lua_setfield(L, 2, "ops"); /* hid.ops = table */
+	lua_newtable(L);
+	lua_setfield(L, 2, "_info"); /* hid._info = {} */
 
 	lunatik_object_t *object = lunatik_newobject(L, &luahid_class, sizeof(luahid_t));
 	luahid_t *hid = (luahid_t *)object->private;
@@ -138,14 +216,18 @@ static int luahid_register(lua_State *L)
 	luaL_argcheck(L, (user_driver->id_table = luahid_setidtable(L, -1)) != NULL,
 		      2, "invaild id_table");
 
+	user_driver->probe = luahid_probe;
+
 	lunatik_setruntime(L, hid, hid);
 	lunatik_getobject(hid->runtime);
+	lunatik_registerobject(L, 2, object);
 
-	if (__hid_register_driver(user_driver, THIS_MODULE, KBUILD_MODNAME) != 0)
+	if (__hid_register_driver(user_driver, THIS_MODULE, KBUILD_MODNAME) != 0) {
+		lunatik_unregisterobject(L, object);
 		luaL_error(L, "failed to register hid driver: %s", user_driver->name);
+	}
 
 	hid->registered = true;
-	lunatik_registerobject(L, 1, object);
 	return 1; /* object */
 }
 


lib/luahid.c
```diff
@@ -12,6 +12,8 @@
 */
 
 #define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+#include <linux/printk.h>
+#include <lua.h>
```
sheharyaar: Are you sure these two are absolutely needed ?
`lua.h` is already included by `lunatik.h`

sheharyaar:
Also, have you tested this logic with an example lua script ? If yes, please share it with us or add it as an example script.
What do you suggest @lneto ?

lib/luahid.c
qrsikno2: i try to delete them, and found no compilation error.
i will remove them, this is added default by lsp.
done.

sheharyaar: Sure, next time, please take care to remove the lsp additions.

qrsikno2:
> Also, have you tested this logic with an example lua script ? If yes, please share it with us or add it as an example script. What do you suggest @lneto ?

shared in the chat channel

lib/luahid.c
```diff
@@ -93,6 +94,75 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+	lua_pushinteger(L, id->driver_data); lua_setfield(L, -2, "driver_data");
+
+	if (lua_pcall(L, 2, 1, 0) != LUA_OK) {
+		pr_err("Failed to call init function: %s", lua_tostring(L, -1));
+		ret = -EFAULT;
+		goto out;
+	}
+
+	if (lua_type(L, -1) != LUA_TTABLE) {
+		pr_warn("No drvdata will be set.\n");
```
lneto: why do we need a warn here?

qrsikno2: what do u mean? an error or an info

lib/luahid.c
```diff
@@ -93,6 +94,75 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+		ret = 0;
+		goto out;
+	}
+
+	if (lunatik_getregistry(L, luahid_probe) != LUA_TTABLE) {
```
lneto: it doesn't see right for me.. what happens if you have many hid objects? you have only one table for each runtime?

lib/luahid.c
```diff
@@ -93,6 +94,75 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+static int luahid_probe_handler(lua_State *L, luahid_t *hid, struct hid_device *hdev, const struct hid_device_id *id) {
+	int ret = 0;
+
+	if (lunatik_getregistry(L, hid) != LUA_TTABLE) {
+		pr_err("Could not find the hid table");
+		ret = -EFAULT;
+		goto out;
+	}
+
+	if (lua_getfield(L, -1, "init") != LUA_TFUNCTION) {
```
lneto: probe seems a more accurate name here..

lib/luahid.c
```diff
@@ -93,6 +94,75 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+	struct hid_driver *driver = hdev->driver;
+	luahid_t *hid = container_of(driver, luahid_t, driver);
+	int ret;
+
+	lunatik_handle(hid->runtime, luahid_probe_handler, ret, hid, hdev, id);
+	if (ret)
+		return ret;
+
+	ret = hid_parse(hdev);
+	if (ret)
```
```suggestion
	if ((ret = hid_parse(hdev)) != 0)
```

lib/luahid.c
```diff
@@ -138,14 +220,20 @@ static int luahid_register(lua_State *L)
...
+	lunatik_setregistry(L, -1, luahid_probe);
+	user_driver->probe = luahid_probe;
+
 	lunatik_setruntime(L, hid, hid);
 	lunatik_getobject(hid->runtime);
+	lunatik_registerobject(L, 1, object);
 
-	if (__hid_register_driver(user_driver, THIS_MODULE, KBUILD_MODNAME) != 0)
+	if (__hid_register_driver(user_driver, THIS_MODULE, KBUILD_MODNAME) != 0) {
+		lunatik_unregisterobject(L, object);
```
lneto, line 226: can't it be done on release()?

lib/luahid.c
lneto: I meant.. isn't it a common flow? isn't ok for not returning a table from lua? or should we always do this? why warning user about this?

lib/luahid.c
qrsikno2: but what can be a better addr for the key?

lneto:
> Also, have you tested this logic with an example lua script ? If yes, please share it with us or add it as an example script. What do you suggest @lneto ?

I think we should have an example, but we don't need to block this PR for this.. but the sooner is better indeed..

lib/luahid.c
qrsikno2: i used to try about a new variable, but it seems to be a little bit complex for a new lunatik_object

lib/luahid.c
qrsikno2: i think i should use init, just because probe can return error to control the process stream. but this is only return a table.

lib/luahid.c
qrsikno2: i have no idea about this, because i look thought all the lib uses `unregisterobject`, such as notifier, device, and probe.
they choose to call this func at _stop, and the notifiter calls the func when failed. 
i have no stop func so i choose to put it when failed.

lib/luahid.c

lib/luahid.c
lneto: So, please, evaluate the execution flow of doing it on release().. if you don't understand how it works, ask for clarification, instead of just stating that "you have no idea".. when you do so, it really looks like you don't want to fully understand the work you committed to do

lib/luahid.c
qrsikno2: aha, it's ok for not returning a table, but it will do nothing.
i mean the callback "init"(if i haven't change it) is for user to set their private data for each device. so it's need to return a table as the private data "structure".

qrsikno2: do you mean if user's private data is a number i can also provide the number  .

lneto: > aha, it's ok for not returning a table, but it will do nothing. i mean the callback "init"(if i haven't change it) is for user to set their private data for each device. so it's need to return a table as the private data "structure".

Isn't there any usage for probe than returning its private data?

lib/luahid.c
lneto: "init" seems misleading to me.. it's a lua callback that we be called from probe(), I don't see why calling it "init".. also, why shouldn't we return error for controlling the process here?

lneto: I mean what's special in the Lua case? We should be able the same thing as we do in C for implementing device drivers

lib/luahid.c
qrsikno2: i think temporarily no, i would do the standard startup in my C code.and user don't need to care about it.

lib/luahid.c

lib/luahid.c
qrsikno2: i changed the name into init, done.

lib/luahid.c
lneto: I see no point for this warning here.. IMO, or we do return an error or fail safely over this without warn.. @sheharyaar let us know if you think differently..

lib/luahid.c
```diff
@@ -93,6 +94,84 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+
+	if (lua_getfield(L, -1, "driver") != LUA_TTABLE) {
+		pr_err("Driver table is not specified");
+		ret = -EFAULT;
+		goto out;
+	}
+
+	if (lua_getfield(L, -1, "probe") != LUA_TFUNCTION) {
+		pr_warn("Probe function is not specified");
+		ret = 0;
```
lneto: does it make sense to not have this specified? it looks to me we should return error on this..

lib/luahid.c
```diff
@@ -93,6 +94,84 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+	lua_pushinteger(L, id->product); lua_setfield(L, -2, "product");
+	lua_pushinteger(L, id->driver_data); lua_setfield(L, -2, "driver_data");
+}
+
+static int luahid_probe_handler(lua_State *L, luahid_t *hid, struct hid_device *hdev, const struct hid_device_id *id) {
+	int ret = 0;
+
+	if (lunatik_getregistry(L, hid) != LUA_TTABLE) {
+		pr_err("Could not find the hid table");
+		ret = -EFAULT;
```
lneto: why EFAULT? please, take a look at the error conditions raised on luadevice.. you can borrow some ideas from there..

lib/luahid.c
```diff
@@ -93,6 +94,84 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+		goto out;
+	}
+
+	if (lua_type(L, -1) != LUA_TTABLE) {
+		pr_warn("No drvdata will be set.\n");
+		ret = 0;
+		goto out;
+	}
+
+	if (lunatik_getregistry(L, hid) && lua_getfield(L, -1, "device_info") != LUA_TTABLE) {
```
lneto: why getting registry again?

lib/luahid.c
```diff
@@ -120,12 +199,28 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+*	 end
+*
 *	 hid.register(hid_driver)
 */
 static int luahid_register(lua_State *L)
 {
 	luaL_checktype(L, 1, LUA_TTABLE);
 
+	lua_newtable(L); /* main table*/
+	lua_pushvalue(L, 1); lua_setfield(L, 2, "driver"); /* main_table.driver = table */
```
lneto: I would borrow the name from luadevice.. I think we call `ops` there if not mistaken..

lib/luahid.c
```diff
@@ -120,12 +199,28 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+	lua_newtable(L); lua_setfield(L, 2, "device_info"); /* main_table.device_info = {} */
```
lneto: `devices` would be more concise, IMO

lneto: another idea is to prefix with `_` our internal fields.. perhaps `_info`.. what do you think @sheharyaar?

lib/luahid.c
```diff
@@ -120,12 +199,28 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
 *	 		{ bus = 2, vendor = 0x4321, product = 0x8765 },
 *	 	}
 *	 }
+*
+*	 function hid_driver:probe(devinfo)
```
lneto: it's devid, right? let's uniformize the nomenclature

lib/luahid.c
```diff
@@ -93,6 +94,84 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
 	return user_table;
 }
 
+static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id);
+
+static void luahid_pushdevid(lua_State *L, const struct hid_device_id *id) {
```
lneto: btw, please, again, review the code style.. you are doing that in another place as well..

lib/luahid.c
```diff
@@ -93,6 +94,84 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
 	return user_table;
 }
 
+static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id);
```
lneto: it doesn't seem necessary..

lib/luahid.c
```diff
@@ -93,6 +94,84 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+
+static void luahid_pushdevid(lua_State *L, const struct hid_device_id *id) {
+	lua_newtable(L);
+	lua_pushinteger(L, id->bus); lua_setfield(L, -2, "bus");
```
lneto: again.. code style..

lneto: btw, you could have a macro to avoid repeating this pattern.. e.g., `LUAHID_SETINTEGER(L, ix, id, field)` or something

lneto: actually, in this case, it can be an inline..

lib/luahid.c
```diff
@@ -11,6 +11,7 @@
 * @module hid
 */
 
+#include "lua.h"
```
lneto: Why this here?

lib/luahid.c
qrsikno2: a driver can be defined without probe...
https://elixir.bootlin.com/linux/v6.13.7/source/drivers/hid/hid-xiaomi.c#L87

lneto: so, there's no point on the warn..

lib/luahid.c
qrsikno2: which line do you mean?
i think the line :
```c
lua_newtable(L); /* main table*/
lua_pushvalue(L, 1);
lua_setfield(L, 2, "driver"); /* main_table.driver = table */
```
the table which was refered in pushvalue, that is the `main_table.driver` canbe changed into `main_table.ops`
Do I understand correctly?

lib/luahid.c
qrsikno2: i think _info is good for me.

lib/luahid.c
```diff
@@ -49,12 +49,12 @@ static int luahid_register(lua_State *L);
 
 static const luaL_Reg luahid_lib[] = {
 	{"register", luahid_register},
-	{NULL, NULL}
+	{NULL, NULL},
```
lneto, line 52: please do any style change as a separate commit or a fixup in the original one.. don't mix up into new developments.. it makes quite harder to separate what needs to be reviewed..

lib/luahid.c
```diff
@@ -93,6 +93,88 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
 	return user_table;
 }
 
+#define luahid_setfield(L, idx, obj, field)	\
+do { 						\
+	lua_pushinteger(L, obj->field);		\
+	lua_setfield(L, idx, #field);		\
+} while (0)
```
lneto: it's better to have a inline here..

lib/luahid.c
```diff
@@ -93,6 +93,88 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+		pr_err("Could not find the hid table");
+		ret = -ENXIO;
+		goto out;
+	}
+
+	if (lua_getfield(L, -1, "driver") != LUA_TTABLE) {
+		pr_err("Driver table is not specified");
+		ret = -ENXIO;
+		goto out;
+	}
```
lneto: you can simplify the two cases with an `||`..  different error messages don't worth the trouble here.. btw, let's borrow error message style from luadevice as well..

lib/luahid.c
```diff
@@ -93,6 +93,88 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+	if (lua_type(L, -1) != LUA_TTABLE) {
+		ret = 0;
+		goto out;
+	}
+
+	if (lua_getfield(L, -3, "device_info") != LUA_TTABLE) {
+		pr_err("Could not find the device table");
+		ret = -ENXIO;
+		goto out;
+	}
```
lneto: it's weird to not requiring the callback to return a table but returning error if the table is invalid..

lib/luahid.c
```diff
@@ -136,16 +236,20 @@ static int luahid_register(lua_State *L)
 
 	lunatik_checkfield(L, 1, "id_table", LUA_TTABLE);
 	luaL_argcheck(L, (user_driver->id_table = luahid_setidtable(L, -1)) != NULL,
-			   2, "invaild id_table");
+		      2, "invaild id_table");
```
lneto, line 217: see https://github.com/luainkernel/lunatik/pull/297/files#r2192884679

lib/luahid.c
qrsikno2: removed. at least temporarily

lib/luahid.c
qrsikno2: removed.

lib/luahid.c
qrsikno2: changed

lib/luahid.c
qrsikno2: distributed into small pieces of commits

lib/luahid.c
qrsikno2: how can i turn it into inline format?

suchas:
lua_setfield(L, idx, #field);

lib/luahid.c
qrsikno2: distributed

lib/luahid.c
qrsikno2: so what do you mean about this?

lib/luahid.c
qrsikno2: removed

lib/luahid.c
qrsikno2: removed

lib/luahid.c
qrsikno2: changed

lib/luahid.c
qrsikno2: changed

lib/luahid.c
qrsikno2: removed

lib/luahid.c
qrsikno2: changed.

lib/luahid.c
qrsikno2: changed

lib/luahid.c
qrsikno2: changed

lib/luahid.c
lneto: sorry.. I misread it.. missed the `obj->field`.. you can disregard this comment..

lib/luahid.c
```diff
@@ -93,6 +93,83 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+	if (lua_getfield(L, -1, "probe") != LUA_TFUNCTION) {
+		ret = 0;
+		goto out;
+	}
+
+	lua_pushvalue(L, -2); /* driver table */
+	luahid_pushdevid(L, -2, id);
+
+	if (lua_pcall(L, 2, 1, 0) != LUA_OK) {
+		pr_err("Failed to call init function: %s", lua_tostring(L, -1));
```
lneto: please, review all your log messages and try to base them on luadevice.. e.g., https://github.com/luainkernel/lunatik/blob/master/lib/luadevice.c#L109
```suggestion
		pr_err("probe: %s\n", lua_tostring(L, -1));
```

also, please, avoid upper case on the beginning of the log message

lib/luahid.c
```diff
@@ -93,6 +93,83 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+	    lua_getfield(L, -1, "driver") != LUA_TTABLE) {
+		pr_err("probe: couldn't find driver");
+		ret = -ENXIO;
+		goto out;
+	}
+
+	if (lua_getfield(L, -1, "probe") != LUA_TFUNCTION) {
+		ret = 0;
+		goto out;
+	}
```
```suggestion
	if (lua_getfield(L, -1, "probe") != LUA_TFUNCTION)
		goto out;
```

lib/luahid.c
```diff
@@ -93,6 +93,83 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+	if (lua_pcall(L, 2, 1, 0) != LUA_OK) {
+		pr_err("Failed to call init function: %s", lua_tostring(L, -1));
+		ret = -ECANCELED;
+		goto out;
+	}
+
+	if (lua_type(L, -1) != LUA_TTABLE) {
+		ret = 0;
+		goto out;
+	}
```
```suggestion
	if (lua_type(L, -1) != LUA_TTABLE)
		goto out;
```

lib/luahid.c
```diff
@@ -120,12 +197,30 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+*	 	return drvdata
+*	 end
+*
 *	 hid.register(hid_driver)
 */
 static int luahid_register(lua_State *L)
 {
 	luaL_checktype(L, 1, LUA_TTABLE);
 
+	lua_newtable(L); /* main table*/
```
lneto: let's uniformize the nomenclature.. what's the "main table"? "device table"?

lneto:
```suggestion
	lua_newtable(L); /* hid = {} */
```

lneto: or
```suggestion
	lua_newtable(L); /* hidev = {} */
```

up to you.. but let's make it easier to the reader

qrsikno2: changed

lib/luahid.c
qrsikno2: changed

lib/luahid.c
qrsikno2: changed

lib/luahid.c
qrsikno2: changed

lib/luahid.c
qrsikno2: do you suggest to ask the user return a table? or not returning error if the other table table ("device_info" aka "_info" now) is invaild?

qrsikno2: or both?

lib/luahid.c
qrsikno2:
```c
static int luadevice_new(lua_State *L)
{
	lunatik_object_t *object;
	luadevice_t *luadev;
	struct device *device;
	const char *name;
	int ret;

	luaL_checktype(L, 1, LUA_TTABLE); /* driver */

	lunatik_checkfield(L, 1, "name", LUA_TSTRING);
	name = lua_tostring(L, -1);

```
the first table provided by user is "driver" as you wrote...

lib/luahid.c
```diff
@@ -93,6 +93,79 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+		ret = -ENXIO;
+		goto out;
+	}
+
+	lua_pushlightuserdata(L, hdev);
+	lua_pushvalue(L, -3); /* returned table */
+	lua_settable(L, -3); /* _info[hdev] = returned table */
+
+	hid_set_drvdata(hdev, hid);
+out:
```
lneto: I think as we don't have anything to tear down here.. I would just  use returns instead of gotos.. it might simplify a bit your code..

qrsikno2: okay。but what do you refer about tearing down? To destruct some object?

lneto: yeah.. of cleaning up some state.. as you simply return.. it seems that you can do it promptly instead of deferring to here..

qrsikno2: can i reserve it for further possible changes?

lneto: I don't see the point.. if we need it later, we can add back the goto's

lneto: And now the returns will make it simpler and smaller

lib/luahid.c

lib/luahid.c
```diff
@@ -93,6 +93,72 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+	luahid_setfield(L, idx, device_id, group);
+	luahid_setfield(L, idx, device_id, vendor);
+	luahid_setfield(L, idx, device_id, product);
+	luahid_setfield(L, idx, device_id, driver_data);
+}
+
+#define luahid_checkhandlertable(L, table_name)		\
+	lunatik_getregistry(L, hid) != LUA_TTABLE ||	\
+	lua_getfield(L, -1, "ops") != LUA_TTABLE ||	\
+	lua_getfield(L, -2, table_name) != LUA_TTABLE
```
```suggestion
#define luahid_checkdriver(L, hid, idx)		(lunatik_getregistry(L, hid) != LUA_TTABLE ||	\
	lua_getfield(L, idx, "ops") != LUA_TTABLE || lua_getfield(L, idx - 1, "_info") != LUA_TTABLE)
```

qrsikno2: i suggest make "_info" an param.

qrsikno2: i believe we need more private data to be stored. and i have already found [one](https://elixir.bootlin.com/linux/v6.13.7/source/drivers/hid/hid-core.c#L1279).

lneto: You mean you will reuse the macro in another callback? Just changing "_info" for another field? If it's the case, I think it's ok to have it as another parm.. otherwise, I no difference here among "_info" and "ops".. in the sense why one should be a parm and but the other not..

qrsikno2: it's because the ops is always necessary.
if you have an callbacks, it should be like:
```lua
function driver:abcd(xyz, ..., pq, nm)
```
rather than 
```lua
function driver.abcd(xyz, ..., pq, nm)
```

lib/luahid.c
```diff
@@ -93,6 +93,70 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+	lua_newtable(L);
+	luahid_setfield(L, -1, device_id, bus);
+	luahid_setfield(L, -1, device_id, group);
+	luahid_setfield(L, -1, device_id, vendor);
+	luahid_setfield(L, -1, device_id, product);
+	luahid_setfield(L, -1, device_id, driver_data);
+}
+
+#define luahid_checkhandlertable(L, hid, idx, table_name) (lunatik_getregistry(L, hid) != LUA_TTABLE || \
+	lua_getfield(L, idx, "ops") != LUA_TTABLE || lua_getfield(L, idx - 1, table_name) != LUA_TTABLE)
```
```suggestion
#define luahid_checkhandlertable(L, hid, idx, field) (lunatik_getregistry(L, hid) != LUA_TTABLE || \
	lua_getfield(L, idx, "ops") != LUA_TTABLE || lua_getfield(L, idx - 1, field) != LUA_TTABLE)
```

lib/luahid.c
```diff
@@ -93,6 +93,70 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+	}
+
+	if (lua_type(L, -1) != LUA_TTABLE)
+		return 0;
+
+	lua_pushlightuserdata(L, hdev);
+	lua_pushvalue(L, -2); /* returned table */
+	lua_settable(L, -4); /* _info[hdev] = returned table */
+
+	hid_set_drvdata(hdev, hid);
```
```suggestion
	if (lua_type(L, -1) == LUA_TTABLE) {
		lua_pushlightuserdata(L, hdev);
		lua_pushvalue(L, -2); /* returned table */
		lua_settable(L, -4); /* _info[hdev] = returned table */
		hid_set_drvdata(hdev, hid);
	}
```

lib/luahid.c
```diff
@@ -93,6 +93,70 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+{
+	if (luahid_checkhandlertable(L, hid, -1, "_info")) {
+		pr_err("probe: couldn't find driver");
+		return -ENXIO;
+	}
+
+	if (lua_getfield(L, -2, "probe") != LUA_TFUNCTION)
+		return 0;
+
+	lua_pushvalue(L, -3); /* ops table */
```
lneto: wouldn't `hid.ops` be clearer here? let's please use the same nomenclature on all places

lib/luahid.c
```diff
@@ -93,6 +93,69 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+	return 0;
+}
+
+static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
+{
+	struct hid_driver *driver = hdev->driver;
+	luahid_t *hid = container_of(driver, luahid_t, driver);
+	int ret;
+
+	lunatik_handle(hid->runtime, luahid_probe_handler, ret, hid, hdev, id);
```
lneto: why `lunatik_handle` here? it should be `lunatik_run`, right?

lib/luahid.c
```diff
@@ -93,6 +93,69 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+
+		hid_set_drvdata(hdev, hid);
+	}
+	return 0;
+}
+
+static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
+{
+	struct hid_driver *driver = hdev->driver;
+	luahid_t *hid = container_of(driver, luahid_t, driver);
```
lneto, line 146: I think we have a macro to do this on luadevice.. can't you take a look? perhaps we can generalize it and move to lunatik.h

qrsikno2: which one? plz point it out. i can't even find the string `container_of` in `luadevice`

lneto: my bad.. it was "fromfile" not containerof.. please disregard..

lib/luahid.c
qrsikno2: i don't get it, how can you do it?
the `luahid_release()` is to release the priv part of the object. You can make it recycled or manually. 
and the _release only have the ptr to priv, it's not sufficient to do it.

lib/luahid.c

lib/luahid.c

lib/luahid.c

lib/luahid.c

lib/luahid.c

lib/luahid.c
qrsikno2: changed the "driver" to "ops"
done

lib/luahid.c
```diff
@@ -93,6 +93,69 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+	int ret;
+
+	lunatik_run(hid->runtime, luahid_probe_handler, ret, hid, hdev, id);
+	if (ret)
+		return ret;
+
+	if ((ret = hid_parse(hdev)) != 0)
+		return ret;
```
```suggestion
	if (ret != 0 || (ret = hid_parse(hdev)) != 0)
		return ret;
```

lib/luahid.c
```diff
@@ -93,6 +93,66 @@ static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
...
+	luahid_setfield(L, -1, device_id, group);
+	luahid_setfield(L, -1, device_id, vendor);
+	luahid_setfield(L, -1, device_id, product);
+	luahid_setfield(L, -1, device_id, driver_data);
+}
+
+#define luahid_checkhandlertable(L, hid, idx, field) (lunatik_getregistry(L, hid) != LUA_TTABLE || \
+	lua_getfield(L, idx, "ops") != LUA_TTABLE || lua_getfield(L, idx - 1, field) != LUA_TTABLE)
+
+static int luahid_probe_handler(lua_State *L, luahid_t *hid, struct hid_device *hdev, const struct hid_device_id *id)
```
lneto: I would apply the same nomenclature we are using on luadevice..
```suggestion
static int luahid_doprobe(lua_State *L, luahid_t *hid, struct hid_device *hdev, const struct hid_device_id *id)
```

PR #300, opened 2025-07-08
Code style guide

jperon:
Following a discussion with @lneto, here is a proposal for a coding style guide. Submitting as a draft, as it’s open to criticism and modifications!

### DIFF
--- a/Codestyle.md
+++ b/Codestyle.md
@@ -59,7 +59,7 @@ if (condition) {
 #define lunatik_handle(runtime, handler, ret, ...)	\
 do {							\
 	/* ... */					\
-} while(0)
+} while (0)
 ```
 
 ## Naming Conventions
--- a/Codestyle.md
+++ b/Codestyle.md
@@ -1,0 +1,236 @@
+# Lunatik C Code Style Guide
+
+This guide outlines the code style conventions observed in the `lunatik.h` header file, which should be followed for consistency across the Lunatik project. As a general rule, we tend to stick to [Linux kernel rules](https://www.kernel.org/doc/html/latest/process/coding-style.html).
+
+## File Headers
+
+All source files should begin with the SPDX license identifier and copyright information (e.g., [`lunatik.h`](lunatik.h:1-4)).
+
+```c
+/*
+ * SPDX-FileCopyrightText: (c) 2023-2025 Ring Zero Desenvolvimento de Software LTDA
+ * SPDX-License-Identifier: MIT OR GPL-2.0-only
+ */
+```
+
+## Includes
+
+Includes should be grouped, with system/kernel headers first, followed by Lua headers, and then project-specific headers. Use an empty line to separate groups (e.g., [`lunatik.h`](lunatik.h:9-16)).
+
+```c
+#include <linux/mutex.h>
+#include <linux/spinlock.h>
+#include <linux/slab.h>
+#include <linux/kref.h>
+
+#include <lua.h>
+#include <lauxlib.h>
+
+/* #include "project_header.h" (if applicable) */
+```
+
+## Indentation
+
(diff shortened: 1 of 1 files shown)


Codestyle.md
```diff
@@ -0,0 +1,236 @@
...
+
+```c
+if (condition) {
+	/* ... */
+}
+
+#define lunatik_handle(runtime, handler, ret, ...)	\
+do {							\
+	/* ... */					\
+} while(0)
```
qrsikno2: i would have an ' ' after the "while" to make the all the "Control Structures" the same. 
it's always a ' ' after the Control Structures's keyword.

lneto: @qrsikno2 is right here..

jperon: OK. I just created an issue to have a TODO list of changes that should be done to current codebase.

Codestyle.md
```diff
@@ -0,0 +1,236 @@
...
+Break long conditional expressions after logical operators, aligning the continuation (e.g., [`lunatik.h`](lunatik.h:245-246)).
+
+```c
+return class != NULL &&
+       (pobject = luaL_testudata(L, ix, class->name)) != NULL ? *pobject : NULL;
+```
+
+## Line breaks, new lines and file Endings
+
+All source files must end with two empty lines. Otherwise, there mustn’t be multiple consecutive empty lines in the files.
```
lneto, line 235: Just one, right.. one empty line at the end of the file

Codestyle.md
```diff
@@ -0,0 +1,236 @@
...
+Parameters in function declarations should be clearly typed. If a parameter is a pointer, the `*` should be next to the variable name (e.g., [`lunatik.h`](lunatik.h:37), [`lunatik.h`](lunatik.h:101)).
+
+```c
+lua_State *L
+lunatik_object_t **pruntime
+```
+
+## Line Breaking
+
+Staying under 80 characters per line is [preferred](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=bdc48fa11e46f867ea4d75fa59ee87a7f48be144), but [don’t stick to this rule against common sense](https://lkml.org/lkml/2020/5/29/1038).
```
lneto, line 211: We actually don't respect this.. we can agree on a sane number though.. 80 is too little

jperon: In [the reference above](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=bdc48fa11e46f867ea4d75fa59ee87a7f48be144), Linus raised the detected limit in the warning script to 100.
Luacheck by default checks for [lines > 120](https://luacheck.readthedocs.io/en/stable/cli.html).

lneto: Perhaps we should see what's the longest in the current code base?

jperon: [180 characters long](https://github.com/luainkernel/lunatik/blob/master/lib/luanetfilter.c#L126).

lneto: Ohh that's long.. hehe.. any chance of getting the second longest? I think that might be a good number..

jperon: 176, then 135.
Over 105903 lines whose length > 9, only 58 (lines of code, not comments or LDoc) are above 100, only 11 above 120.
With Linus leaving a warning above 100, I think it could be a good rule, as soon as we agree that common sense keeps all its rights.

IMHO, [this style of break](https://github.com/torvalds/linux/blob/master/fs/namei.c#L4074) (as function definitions are likely the longest ones) would be the best: two tabs before the second line, to avoid confusion with function body. [Python follows this convention](https://peps.python.org/pep-0008/#indentation). Aligning with the opening parenthesis would require mixing tabs and spaces for indentation.

lneto: I would say 120 and 1 tab.. my vote, at least ;-)

jperon: I think we could agree on that, then:

```md
Staying under 120 characters per line is _preferred_, but [don’t stick to this rule against common sense](https://lkml.org/lkml/2020/5/29/1038).
```

PR #302, opened 2025-07-11
add callback for report_descriptor

qrsikno2:
This pull request introduces enhancements to the HID driver integration in the `lib/luahid.c` file, including the addition of support for report fixups, a new `luadata_t` structure, and improved Lua integration for HID device information. These changes aim to extend the functionality of the driver and improve its interaction with Lua scripts.

### New functionality for report fixups:
* Added the `luahid_report_fixup` function to handle report fixups for HID devices, with support for kernel versions 6.12.0 and above. This function integrates Lua-based processing of HID reports, allowing modifications to the report data.
* Introduced the `luahid_doreport_fixup` helper function to manage the Lua interaction for report fixups, including creating and handling `luadata_t` objects for report data.

### New data structure and Lua integration:
* Defined a new `luadata_t` structure to encapsulate report data, including a pointer to the data, its size, and additional options.
* Added the `luahid_push_device_info` function to populate a Lua table with detailed HID device information, such as vendor, product, and version.

### Registration enhancements:
* Updated the `luahid_register` function to initialize a `_report_des` table for storing report descriptors and to assign the new `report_fixup` callback to the HID driver.

These changes improve the modularity and extensibility of the HID driver, enabling advanced use cases through Lua scripting.

### DIFF
--- a/lib/luahid.c
+++ b/lib/luahid.c
@@ -103,7 +103,6 @@ do { 						\
 	lua_setfield(L, idx - 1, #field);	\
 } while (0)
 
-static inline void luahid_pushdevid(lua_State *L, int idx, const struct hid_device_id *device_id)
 #define luahid_pcall(L, func, arg) 					\
 do { 									\
 	int n = lua_gettop(L); 						\
@@ -125,13 +124,11 @@ do { 							\
 	luahid_setfield(L, -1, dev, extra); 		\
 } while (0)
 
+static inline int luahid_pushdevid(lua_State *L)
 {
-	lua_newtable(L);
-	luahid_setfield(L, -1, device_id, bus);
-	luahid_setfield(L, -1, device_id, group);
-	luahid_setfield(L, -1, device_id, vendor);
-	luahid_setfield(L, -1, device_id, product);
-	luahid_setfield(L, -1, device_id, driver_data);
+	struct hid_device_id *device_id = (struct hid_device_id *)lua_touserdata(L, 1);
+	luahid_newtable(L, device_id, driver_data);
+	return 1; /* devid table */
 }
 
 static inline int luahid_pushhdev(lua_State *L)
@@ -163,7 +160,7 @@ static int luahid_doprobe(lua_State *L, luahid_t *hid, struct hid_device *hdev, 
 		return 0;
 
 	lua_pushvalue(L, -3); /* hid.ops */
-	luahid_pushdevid(L, -4, id);
+	luahid_pcall(L, luahid_pushdevid, id);
 
 	if (lua_pcall(L, 2, 1, 0) != LUA_OK) {
 		pr_err("probe: %s\n", lua_tostring(L, -1));
--- a/lib/luahid.c
+++ b/lib/luahid.c
@@ -155,7 +155,7 @@ static inline void luahid_pushinfo(lua_State *L, int idx, struct hid_device *hde
 static int luahid_doprobe(lua_State *L, luahid_t *hid, struct hid_device *hdev, const struct hid_device_id *id)
 {
 	if (luahid_checkdriver(L, hid, -1, "_info")) {
-		pr_err("probe: couldn't find driver");
+		pr_err("probe: couldn't find driver\n");
 		return -ENXIO;
 	}
 
--- a/lib/luahid.c
+++ b/lib/luahid.c
@@ -12,10 +12,13 @@
 */
 
 #define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
-#include <string.h>
+#include <linux/version.h>
+#include <linux/string.h>
 #include <linux/hid.h>
 #include <lunatik.h>
 #include <lauxlib.h>
+
+#include "luadata.h"
 
 /***
 * Represents a registered HID driver.
@@ -27,6 +30,7 @@
 */
 typedef struct luahid_s {
 	lunatik_object_t *runtime;
+	lunatik_object_t *descriptor;
 	struct hid_driver driver;
 	bool registered;
 } luahid_t;
@@ -100,6 +104,27 @@ do { 						\
 } while (0)
 
 static inline void luahid_pushdevid(lua_State *L, int idx, const struct hid_device_id *device_id)
+#define luahid_pcall(L, func, arg) 					\
+do { 									\
+	int n = lua_gettop(L); 						\
+	lua_pushcfunction(L, func); 					\
+	lua_pushlightuserdata(L, (void *)arg); 				\
+	if (lua_pcall(L, 1, LUA_MULTRET, 0) != LUA_OK) { 		\
+		pr_warn("%s: %s\n", #func, lua_tostring(L, -1));	\
+		lua_settop(L, n);					\
+	} 								\
+} while (0)
+
+#define luahid_newtable(L, dev, extra)			\
+do { 							\
+	lua_newtable(L); 				\
+	luahid_setfield(L, -1, dev, bus); 		\
+	luahid_setfield(L, -1, dev, group); 		\
+	luahid_setfield(L, -1, dev, vendor); 		\
+	luahid_setfield(L, -1, dev, product); 		\
+	luahid_setfield(L, -1, dev, extra); 		\
+} while (0)
+
 {
 	lua_newtable(L);
 	luahid_setfield(L, -1, device_id, bus);
@@ -107,6 +132,21 @@ static inline void luahid_pushdevid(lua_State *L, int idx, const struct hid_devi
 	luahid_setfield(L, -1, device_id, vendor);
 	luahid_setfield(L, -1, device_id, product);
 	luahid_setfield(L, -1, device_id, driver_data);
+}
+
+static inline int luahid_pushhdev(lua_State *L)
+{
+	struct hid_device *hdev = (struct hid_device *)lua_touserdata(L, 1);
+	luahid_newtable(L, hdev, version);
+	lua_pushstring(L, hdev->name);
+	lua_setfield(L, -2, "name");
+	return 1; /* hdev table */
+}
+
+static inline void luahid_pushinfo(lua_State *L, int idx, struct hid_device *hdev)
+{
+	lua_pushlightuserdata(L, hdev);
+	lua_rawget(L, idx - 1);
 }
 
 #define luahid_checkdriver(L, hid, idx, field) (lunatik_getregistry(L, hid) != LUA_TTABLE || \
@@ -153,6 +193,58 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
 	return hid_hw_start(hdev, HID_CONNECT_DEFAULT);
 }
 
+static inline lunatik_object_t *luahid_getdescriptor(lua_State *L, luahid_t *hid)
+{
+	lunatik_object_t *data;
+	if (lunatik_getregistry(L, hid->descriptor) != LUA_TUSERDATA || unlikely((data = (lunatik_object_t *)lunatik_toobject(L, -1)) == NULL)) {
+		pr_err("could not find descriptor\n");
+		return NULL;
+	}
+	return data;
+}
+
+static int luahid_doreport_fixup(lua_State *L, luahid_t *hid, struct hid_device *hdev, __u8 *rdesc, unsigned int rsize)
+{
+	if (luahid_checkdriver(L, hid, -1, "_info") || lua_getfield(L, -2, "report_fixup") != LUA_TFUNCTION) {
+		pr_warn("report_fixup: invaild driver\n");
+		goto out;
+	}
+
+	lua_pushvalue(L, -3); /* hid.ops */
+	luahid_pcall(L, luahid_pushhdev, hdev);
+	luahid_pushinfo(L, -4, hdev);
+
+	lunatik_object_t *data;
+	if ((data = luahid_getdescriptor(L, hid)) == NULL) {
+		pr_warn("report_fixup: descriptor not found\n");
+		goto out;
+	}
+	luadata_reset(data, rdesc, rsize, LUADATA_OPT_NONE);
+
+	if (lua_pcall(L, 4, 0, 0) != LUA_OK)
+		pr_warn("report_fixup: %s\n", lua_tostring(L, -1));
+
+	luadata_clear(data);
+out:
+	return 0; /* unused  */
+}
+
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 12, 0))
+typedef const __u8* luahid_ret_t;
+#else
+typedef __u8* luahid_ret_t;
+#endif
+
+static luahid_ret_t luahid_report_fixup(struct hid_device *hdev, __u8 *rdesc, unsigned int *rsize)
+{
+	struct hid_driver *driver = hdev->driver;
+	luahid_t *hid = container_of(driver, luahid_t, driver);
+	int ret;
+
+	lunatik_run(hid->runtime, luahid_doreport_fixup, ret, hid, hdev, rdesc, *rsize);
+	return rdesc;
+}
+
 /***
 * Registers a new HID driver.
 * This function creates a new HID driver object from a Lua table and registers it with the kernel.
@@ -192,6 +284,13 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
 *	 	return drvdata
 *	 end
 *
+*	 function hid_driver:report_fixup(hdev, priv_data, descriptor)
+*	 	if hdev.vendor == 0x1234 and hdev.product == 0x5678 then
+*	 		descriptor[1] = 0x05
+*	 		descriptor[descriptor.size] = 0x0A
+*	 	end
+*	 end
+*
 *	 hid.register(hid_driver)
 */
 static int luahid_register(lua_State *L)
@@ -207,6 +306,7 @@ static int luahid_register(lua_State *L)
 	lunatik_object_t *object = lunatik_newobject(L, &luahid_class, sizeof(luahid_t));
 	luahid_t *hid = (luahid_t *)object->private;
 	memset(hid, 0, sizeof(luahid_t));
+	luadata_attach(L, hid, descriptor);
 
 	struct hid_driver *user_driver = &(hid->driver);
 	user_driver->name = lunatik_checkalloc(L, NAME_MAX);
@@ -217,6 +317,7 @@ static int luahid_register(lua_State *L)
 		      2, "invaild id_table");
 
 	user_driver->probe = luahid_probe;
+	user_driver->report_fixup = luahid_report_fixup;
 
 	lunatik_setruntime(L, hid, hid);
 	lunatik_getobject(hid->runtime);


lib/luahid.c
```diff
@@ -109,6 +117,18 @@ static inline void luahid_pushdevid(lua_State *L, int idx, const struct hid_devi
 	luahid_setfield(L, -1, device_id, driver_data);
 }
 
+static inline void luahid_push_device_info(lua_State *L, struct hid_device *hdev)
```
lneto: please, let's keep consistency on the name space.. you already have `pushdevid`, I would say this might be something like `pushdevinfo`(?).. otherwise, let's update your current name space accordingly..

lib/luahid.c
```diff
@@ -153,6 +173,64 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
 }
 
+static int luahid_doreport_fixup(lua_State *L, luahid_t *hid,
+				 struct hid_device *hdev, __u8 *buf, unsigned int *size,
+				 __u8 **ret_ptr)
+{
+	if (luahid_checkdriver(L, hid, -1, "_report_des")) {
+		pr_err("report_fixup: couldn't find driver\n");
+		return -ENXIO;
+	}
```
lneto: this can be turned into a inline that returns 0 or -ENXIO, as you are using this pattern multiple times..

lib/luahid.c
```diff
@@ -153,6 +173,64 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+
+	if (lua_getfield(L, -2, "report_fixup") != LUA_TFUNCTION)
+		return -ENXIO;
+
+	lua_pushvalue(L, -3); /* hid.ops */
+	luahid_push_device_info(L, hdev);
+	lunatik_object_t *original_data = luadata_new(buf, *size, hid->runtime->sleep, LUADATA_OPT_READONLY);
```
lneto: if we don't have a practical case on increasing the size of the buffer, let's just let the user change the original data in place.. instead of creating and returning a new data object.. (even if we have a practical case, but it's not common, I would implement the simplified version firstly and create a issue for handling this case later)

lib/luahid.c
```diff
@@ -218,6 +299,8 @@ static int luahid_register(lua_State *L)
 
 	user_driver->probe = luahid_probe;
```
```suggestion
```

lib/luahid.c
```diff
@@ -153,6 +173,64 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+	luadata_t *data = (luadata_t *)returned_object->private;
+	*size = data->size;
+	*ret_ptr = data->ptr;
+
+	return 0;
+}
+
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 12, 0))
+static const __u8 *luahid_report_fixup(struct hid_device *hdev, __u8 *buf, unsigned int *size)
+{
```
lneto: ?

lib/luahid.c
qrsikno2: it's just because the param, the report_fixup can only get the hid_device rather than hid_device_id

lib/luahid.c
qrsikno2: as you can see the higher version of kernel changed the returned type of the callbacks

lneto: I see that.. I just don't get why you do need to repeat `{` on both cases.. it doesn't make sense to me..

lib/luahid.c
qrsikno2: i believe it's different for two types

lib/luahid.c
qrsikno2: fixed

lib/luahid.c

lib/luahid.c

lib/luahid.c
qrsikno2: this is [one](https://elixir.bootlin.com/linux/v6.13.7/source/drivers/hid/hid-asus.c#L1191), easy to be found.
```c
	/*
		 * Change Usage (76h) to Usage Minimum (00h), Usage Maximum
		 * (FFh) and clear the flags in the Input() byte.
		 * Note the descriptor has a bogus 0 byte at the end so we
		 * only need 1 extra byte.
		 */
		if (*rsize == rsize_orig &&
			rdesc[offs] == 0x09 && rdesc[offs + 1] == 0x76) {
			*rsize = rsize_orig + 1;
			rdesc = kmemdup(rdesc, *rsize, GFP_KERNEL);
			if (!rdesc)
				return NULL;

			hid_info(hdev, "Fixing up %s keyb report descriptor\n",
				drvdata->quirks & QUIRK_T100CHI ?
				"T100CHI" : "T90CHI");
			memmove(rdesc + offs + 4, rdesc + offs + 2, 12);
			rdesc[offs] = 0x19;
			rdesc[offs + 1] = 0x00;
			rdesc[offs + 2] = 0x29;
			rdesc[offs + 3] = 0xff;
			rdesc[offs + 14] = 0x00;
		}
	}

```

qrsikno2: it extended the length: `*rsize = rsize_orig + 1`

that's what i worried, so i ask user to return a new `luadata` object at first. they can feel free to define the length

lib/luahid.c
qrsikno2: i change it to `luahid_pushhdev`, actually it's try to push the `hdev`( as the type of `hid_device`)
and the `devid` is try to push the `device_id`

lib/luahid.c
lneto: we should add support on luadata for this sort of dupping.. it's kinda straightforward to implement.. something as, `data:dup([size])`

lneto: and I don't think it's necessary to the original to be READONLY, right? if it's not necessary to change its size, we can just return the same object, right/

lib/luahid.c
```diff
@@ -204,6 +277,9 @@ static int luahid_register(lua_State *L)
 	lua_newtable(L);
 	lua_setfield(L, 2, "_info"); /* hid._info = {} */
 
+	lua_newtable(L);
+	lua_setfield(L, 2, "_report_des"); /* hid._report_des = {} */
```
```suggestion
	lua_setfield(L, 2, "_report"); /* hid._report = {} */
```

lib/luahid.c
```diff
@@ -12,11 +12,14 @@
 */
 
 #define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+#include <linux/version.h>
 #include <string.h>
```
lneto: I don't think you should include <string.h> here.. it's more likely you want to include the one in linux/

lib/luahid.c
qrsikno2: why? its task is to fix the "report descriptor" not the "report"

lib/luahid.c
qrsikno2: fixed

lib/luahid.c
lneto: Because you don't have any other table related to "report".. then, it's clear enough that a table name "report" is a "report descriptor".. there's no clash, therefore you don't need to be that verbose..

qrsikno2: yes. fixed

lib/luahid.c
qrsikno2: added and changed please check it.

lib/luadata.c
```diff
@@ -41,6 +41,8 @@ typedef struct luadata_s {
 
 static int luadata_lnew(lua_State *L);
 
+static int luadata_dup(lua_State *L);
```
lneto: We don't need this.. just move the implementation to before the lib table..

lib/luadata.c
```diff
@@ -354,6 +357,31 @@ static const lunatik_class_t luadata_class = {
...
+* @tparam data orig_data The original data object to duplicate.
+* @tparam[opt] integer new_size The size of the new data object. If omitted, uses the size of `orig_data
+* @treturn data A new data object with the same content as `orig_data`, but with the specified size.
+*/
+static int luadata_dup(lua_State *L)
+{
+	luadata_t *orig_data = luadata_check(L, 1);
+	lua_Integer new_size = luaL_optinteger(L, 2, orig_data->size);
+
+	lunatik_object_t *new_object = lunatik_newobject(L, &luadata_class, sizeof(luadata_t));
```
lneto: We shouldn't replicate the creation code.. instead, we should use an auxiliar function that can be used both here and on new()

lib/luahid.c
```diff
@@ -153,6 +175,57 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+
+	luadata_t *data = (luadata_t *)returned_object->private;
+	*size = data->size;
+	*ret_ptr = data->ptr;
+
+	return 0;
+}
+
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 12, 0))
+static const __u8 *luahid_report_fixup(struct hid_device *hdev, __u8 *buf, unsigned int *size)
```
lneto: If the diff here is only "const" you can put this in a macro and define it accordingly.. let's say LUAHID_RET

lneto: or even a typedef.. e.g., luahid_report_t.. whatever suites better to you.. so we avoid repeat the entire prototype and make the code a bit cleaner

lib/luahid.c
```diff
@@ -109,26 +117,40 @@ static inline void luahid_pushdevid(lua_State *L, int idx, const struct hid_devi
...
+	lua_setfield(L, -2, "name");
+}
+
 #define luahid_checkdriver(L, hid, idx, field) (lunatik_getregistry(L, hid) != LUA_TTABLE || \
 	lua_getfield(L, idx, "ops") != LUA_TTABLE || lua_getfield(L, idx - 1, field) != LUA_TTABLE)
 
+#define luahid_error(code, op, msg)		\
+do {						\
+	pr_err("%s: %s\n", (op), (msg)); 	\
+	return code; 				\
```
lneto: I wouldn't put return in a macro like this.. it's not clear to the outside we are returning

lib/luahid.c
```diff
@@ -153,6 +175,57 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+	lunatik_object_t *original_data = luadata_new(buf, *size, hid->runtime->sleep, LUADATA_OPT_FREE);
+	if (!original_data)
+		luahid_error(-ENOMEM, "report_fixup", "failed to create luadata for original report");
+	lunatik_pushobject(L, original_data);
+
+	if (lua_pcall(L, 3, 1, 0) != LUA_OK)
+		luahid_error(-ECANCELED, "report_fixup", lua_tostring(L, -1));
+
+	if (lua_isnil(L, -1))
+		return -ENXIO;
```
lneto: Why returning ENXIO in this case?

lib/luahid.c
```diff
@@ -153,6 +175,57 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+
+	lunatik_object_t *returned_object = lunatik_checkobject(L, -1);
+	lua_pushlightuserdata(L, hdev);
+	lua_pushvalue(L, -2); /* luadata */
+	lua_settable(L, -4); /* hid._report[hdev] = luadata */
+
+	luadata_t *data = (luadata_t *)returned_object->private;
```
lneto: What happens here if the returned object is another lunatik object, instead of luadata?

lib/luahid.c
```diff
@@ -153,6 +175,57 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+{
+	if (luahid_checkdriver(L, hid, -1, "_report"))
+		luahid_error(-ENXIO, "report_fixup", "couldn't find driver");
+
+	if (lua_getfield(L, -2, "report_fixup") != LUA_TFUNCTION)
+		return -ENXIO;
+
+	lua_pushvalue(L, -3); /* hid.ops */
+	luahid_pushhdev(L, hdev);
+	lunatik_object_t *original_data = luadata_new(buf, *size, hid->runtime->sleep, LUADATA_OPT_FREE);
```
lneto: Why using OPT_FREE here?

lneto: Did you test this code?

lib/luahid.c
```diff
@@ -109,26 +117,40 @@ static inline void luahid_pushdevid(lua_State *L, int idx, const struct hid_devi
...
+} while (0)
+
 static int luahid_doprobe(lua_State *L, luahid_t *hid, struct hid_device *hdev, const struct hid_device_id *id)
 {
-	if (luahid_checkdriver(L, hid, -1, "_info")) {
-		pr_err("probe: couldn't find driver");
-		return -ENXIO;
-	}
+	if (luahid_checkdriver(L, hid, -1, "_info"))
+		luahid_error(-ENXIO, "probe", "couldn't find driver");
```
lneto: I wouldn't change this pattern.. unless you propose it as a whole for all lunatik modules.. please, always mind our code base and how we are doing things.. if you want to change something.. please, propose, then we can discuss it alongside the other developers..

lib/luadata.c
qrsikno2: done

lib/luahid.c

lib/luahid.c
qrsikno2: get recorded in [link](https://github.com/luainkernel/lunatik/issues/306#issue-3236541567)

lib/luahid.c
qrsikno2: changed into -EINVAL

lib/luahid.c
qrsikno2: get recorded in [link](https://github.com/luainkernel/lunatik/issues/306#issue-3236541567)

lib/luahid.c
qrsikno2: fixed

lib/luahid.c
qrsikno2: fixed as OPT_NONE

lib/luadata.c
qrsikno2: done

lib/luahid.c
```diff
@@ -153,6 +166,87 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+		}
+		lua_pop(L, 1);
+	}
+
+	*ret_ptr = buf;
+	return 0;
+}
+
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 12, 0))
+#define LUAHID_RET const __u8 *
```
lneto: typedef seems more adequate here
```suggestion
typedef const __u8 * luahid_ret_t;
```

lib/luahid.c
```diff
@@ -153,6 +166,87 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+
+static LUAHID_RET luahid_report_fixup(struct hid_device *hdev, __u8 *buf, unsigned int *size)
+{
+	struct hid_driver *driver = hdev->driver;
+	luahid_t *hid = container_of(driver, luahid_t, driver);
+	__u8 *ret_ptr = buf;
+	int ret;
+
+	lunatik_run(hid->runtime, luahid_doreport_fixup, ret, hid, hdev, buf, size, &ret_ptr);
+
```
```suggestion
```

lib/luahid.c
```diff
@@ -153,6 +166,87 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+	lua_pushvalue(L, -3); /* hid.ops */
+	luahid_pushhdev(L, hdev);
+
+	lua_newtable(L); /* original descriptor */
+	for (unsigned int i = 0; i < *size; i++) {
+		lua_pushinteger(L, buf[i]);
+		lua_seti(L, -2, i + 1);
+	}
+	lua_pushinteger(L, *size);
+	lua_setfield(L, -2, "size");
```
lneto: it wasn't my suggestion.. my recommendation wat to use a luadata object whithout allowing the Lua script to resize it by now.. and let the script modify the original one in place whitout dupping it.. I don't see why we should use Lua tables instead

lib/luahid.c
```diff
@@ -153,6 +166,87 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+				 struct hid_device *hdev, __u8 *buf, unsigned int *size,
+				 __u8 **ret_ptr)
+{
+	if (luahid_checkdriver(L, hid, -1, "_info")) {
+		pr_err("report_fixup: couldn't find driver\n");
+		return -ENXIO;
+	}
+
+	if (lua_getfield(L, -2, "report_fixup") != LUA_TFUNCTION)
+		return -ENXIO;
```
lneto: I would just use an `||` here.. it will make the code way smaller and simpler and more readable

lib/luahid.c
```diff
@@ -153,6 +166,87 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
 	return hid_hw_start(hdev, HID_CONNECT_DEFAULT);
 }
 
+static int luahid_doreport_fixup(lua_State *L, luahid_t *hid,
+				 struct hid_device *hdev, __u8 *buf, unsigned int *size,
+				 __u8 **ret_ptr)
```
lneto: I don't see why breaking lines like this.. please, mind the code base for long parm lines

lib/luahid.c

lib/luahid.c

lib/luahid.c
qrsikno2: changed

lib/luahid.c
qrsikno2: changed

lib/luahid.c
qrsikno2: changed

lib/luahid.c
```diff
@@ -153,6 +174,48 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+typedef const __u8* luahid_ret_t;
+#else
+typedef  __u8* luahid_ret_t;
+#endif
+
+static luahid_ret_t luahid_report_fixup(struct hid_device *hdev, __u8 *buf, unsigned int *size)
+{
+	struct hid_driver *driver = hdev->driver;
+	luahid_t *hid = container_of(driver, luahid_t, driver);
+	__u8 *ret_ptr = buf;
```
lneto: Why you need this?

lib/luahid.c
```diff
@@ -109,13 +112,31 @@ static inline void luahid_pushdevid(lua_State *L, int idx, const struct hid_devi
 	luahid_setfield(L, -1, device_id, driver_data);
 }
 
+static inline void luahid_pushhdev(lua_State *L, struct hid_device *hdev)
```
lneto: idx?

qrsikno2: i see, there is no need to have an `idx`, the `pushhdev` is to construct a new table based on the `hid_device *hdev`, so we have no need of using `idx` to visit other data.

lib/luahid.c
qrsikno2: i see... fixed it.

lib/luahid.c
```diff
@@ -153,6 +174,47 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+		return -ECANCELED;
+	}
+
+	return 0;
+}
+
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 12, 0))
+typedef const __u8* luahid_ret_t;
+#else
+typedef  __u8* luahid_ret_t;
```
```suggestion
typedef __u8* luahid_ret_t;
```
please, mind white spaces

lib/luahid.c
```diff
@@ -153,6 +174,47 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
 	return hid_hw_start(hdev, HID_CONNECT_DEFAULT);
 }
 
+static int luahid_doreport_fixup(lua_State *L, luahid_t *hid, struct hid_device *hdev, __u8 *buf, unsigned int *size)
+{
+	if (luahid_checkdriver(L, hid, -1, "_info") || lua_getfield(L, -2, "report_fixup") != LUA_TFUNCTION) {
+		pr_err("report_fixup: couldn't find driver or the function\n");
```
lneto: "the function" is too generic here.. IMO, a better error message would be something like
```suggestion
		pr_err("report_fixup: invalid driver\n");
```

lib/luahid.c
```diff
@@ -153,6 +174,47 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+		return -ENXIO;
+	}
+
+	lua_pushvalue(L, -3); /* hid.ops */
+	luahid_pushhdev(L, hdev);
+	luahid_pushinfo(L, -4, hdev);
+	lunatik_object_t *original_data = luadata_new(buf, *size, hid->runtime->sleep, LUADATA_OPT_NONE);
```
lneto: now you only have one data object.. no need for calling it "original".. I would take a look at other drivers to see how they call `buf` there.. `data` sounds too generic in this case..

lib/luahid.c

lib/luahid.c

lib/luahid.c
qrsikno2: i rename it as "descriptor"。 It's very straightforward, just call it what it should be called

lib/luahid.c
```diff
@@ -153,6 +174,47 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+typedef __u8* luahid_ret_t;
+#endif
+
+static luahid_ret_t luahid_report_fixup(struct hid_device *hdev, __u8 *buf, unsigned int *size)
+{
+	struct hid_driver *driver = hdev->driver;
+	luahid_t *hid = container_of(driver, luahid_t, driver);
+	int ret;
+
+	lunatik_run(hid->runtime, luahid_doreport_fixup, ret, hid, hdev, buf, size);
```
lneto: you are not checking `ret` after returning.. why are you setting it at all if you are only using buf.. and what happens if you have an error condition? shouldn't you let the caller know it has failed?

lib/luahid.c
```diff
@@ -153,6 +174,47 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+	return 0;
+}
+
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 12, 0))
+typedef const __u8* luahid_ret_t;
+#else
+typedef __u8* luahid_ret_t;
+#endif
+
+static luahid_ret_t luahid_report_fixup(struct hid_device *hdev, __u8 *buf, unsigned int *size)
```
lneto: other modules call `buf` as `rdesc`.. I would use the same nomenclature.. please, be attemptive to this..

lib/luahid.c
```diff
@@ -153,6 +174,47 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+		pr_err("report_fixup: failed to create the buffer of descriptor\n");
+		return -ENOMEM;
+	}
+	lunatik_pushobject(L, descriptor);
+
+	if (lua_pcall(L, 4, 0, 0) != LUA_OK) {
+		pr_err("report_fixup: %s\n", lua_tostring(L, -1));
+		return -ECANCELED;
+	}
+
```
lneto, line 223:
```suggestion
```

lib/luahid.c
```diff
@@ -153,6 +174,47 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
 	return hid_hw_start(hdev, HID_CONNECT_DEFAULT);
 }
 
+static int luahid_doreport_fixup(lua_State *L, luahid_t *hid, struct hid_device *hdev, __u8 *buf, unsigned int *size)
+{
+	if (luahid_checkdriver(L, hid, -1, "_info") || lua_getfield(L, -2, "report_fixup") != LUA_TFUNCTION) {
+		pr_err("report_fixup: invaild driver\n");
+		return -ENXIO;
```
lneto: if you are not effectivaly raising an error no where.. there's no point on returning an error code.. nor printing a error.. you should just print a warn

lib/luahid.c
qrsikno2: The pattern is designed with no return error value. In normal implementation of `report_fixup` they usually changed directly or return a new buffer. But my callback of `report_fixup` just let user changes the original buffer. Even if an error occurs i cannot detect where has been changed.... so i only try to return the buf.....

lib/luahid.c
qrsikno2: fixed

lib/luahid.c
qrsikno2: fixed

lib/luahid.c
```diff
@@ -153,6 +174,46 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
 	return hid_hw_start(hdev, HID_CONNECT_DEFAULT);
 }
 
+static int luahid_doreport_fixup(lua_State *L, luahid_t *hid, struct hid_device *hdev, __u8 *rdesc, unsigned int *rsize)
```
lneto: if you are not updating rsize, you don't need a pointer..

lib/luahid.c
```diff
@@ -109,13 +112,31 @@ static inline void luahid_pushdevid(lua_State *L, int idx, const struct hid_devi
 	luahid_setfield(L, -1, device_id, driver_data);
 }
 
+static inline void luahid_pushhdev(lua_State *L, struct hid_device *hdev)
+{
+	lua_newtable(L);
```
lneto: what happens if this (or any other call below) raises an error? (you propably will need to use lua_pcall)

lib/luahid.c
```diff
@@ -153,6 +174,48 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
 	return hid_hw_start(hdev, HID_CONNECT_DEFAULT);
 }
 
+static int luahid_doreport_fixup(lua_State *L, luahid_t *hid, struct hid_device *hdev, __u8 *rdesc, unsigned int *rsize)
+{
+	if (luahid_checkdriver(L, hid, -1, "_info") || lua_getfield(L, -2, "report_fixup") != LUA_TFUNCTION) {
+		pr_warn("report_fixup: invaild driver\n");
+		return -ENXIO;
```
lneto: there's no point to return error codes if you are not using them..

lib/luahid.c
```diff
@@ -153,6 +174,48 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+
+	lua_pushvalue(L, -3); /* hid.ops */
+	luahid_pushhdev(L, hdev);
+	luahid_pushinfo(L, -4, hdev);
+	lunatik_object_t *descriptor = luadata_new(rdesc, *rsize, hid->runtime->sleep, LUADATA_OPT_NONE);
+	if (!descriptor) {
+		pr_warn("report_fixup: failed to create the buffer of descriptor\n");
+		return -ENOMEM;
+	}
+	lunatik_pushobject(L, descriptor);
```
lneto: this can raise a error as well..

lib/luahid.c
qrsikno2: fixed

lneto: how?

lib/luahid.c
```diff
@@ -109,13 +125,36 @@ static inline void luahid_pushdevid(lua_State *L, int idx, const struct hid_devi
...
+	luahid_setfield(L, -1, hdev, product);
+	luahid_setfield(L, -1, hdev, version);
+	lua_pushstring(L, hdev->name);
+	lua_setfield(L, -2, "name");
+	return 1; /* hdev table */
+}
+
+static inline int luahid_pushinfo(lua_State *L)
+{
+	int idx = lua_tointeger(L, 1);
```
lneto: it doesn't make sense.. just  pass the actual table as an argument for the call in the stack..

qrsikno2: fixed

lib/luahid.c
qrsikno2: fixed

lib/luahid.c
qrsikno2: fixed

lib/luahid.c
qrsikno2: by removing it

lib/luahid.c
qrsikno2: fixed by `luahid_pcall`

lib/luahid.c
qrsikno2: fixed

lib/luahid.c
```diff
@@ -99,14 +103,50 @@ do { 						\
 	lua_setfield(L, idx - 1, #field);	\
 } while (0)
 
-static inline void luahid_pushdevid(lua_State *L, int idx, const struct hid_device_id *device_id)
+#define luahid_pcall(L, func, idx, arg) 			\
+do { 								\
+	int n = lua_gettop(L); 					\
+	lua_pushcfunction(L, func); 				\
+	lua_pushvalue(L, idx - 1); 				\
```
lneto: why `idx - 1`? it's confusing as an API.. just pass `idx - 1` on the caller side

lib/luahid.c
```diff
@@ -99,14 +103,50 @@ do { 						\
...
+	lua_pushlightuserdata(L, (void *)arg); 			\
+	if (lua_pcall(L, 2, LUA_MULTRET, 0) != LUA_OK) { 	\
+		pr_err("%s: %s\n", #func, lua_tostring(L, -1));	\
```
lneto: you aren't returning error or doing anything different in error case, thus it should be the case of pr_warn instead

lib/luahid.c
```diff
@@ -99,14 +103,50 @@ do { 						\
 	lua_setfield(L, idx - 1, #field);	\
 } while (0)
 
-static inline void luahid_pushdevid(lua_State *L, int idx, const struct hid_device_id *device_id)
+#define luahid_pcall(L, func, idx, arg) 			\
+do { 								\
+	int n = lua_gettop(L); 					\
+	lua_pushcfunction(L, func); 				\
+	lua_pushvalue(L, idx - 1); 				\
+	lua_pushlightuserdata(L, (void *)arg); 			\
```
lneto: you don't need this typecast here, right?

lib/luahid.c
qrsikno2: it's because the `lua_pushcfuntion` make the top++

lib/luahid.c
qrsikno2: no, it will cause compilation error when passing a `const sometype *`

lib/luahid.c
```diff
@@ -99,14 +103,50 @@ do { 						\
...
+} while (0)
+
+static inline int luahid_pushdevid(lua_State *L)
 {
+	struct hid_device_id *device_id = (struct hid_device_id *)lua_touserdata(L, 2);
 	lua_newtable(L);
 	luahid_setfield(L, -1, device_id, bus);
 	luahid_setfield(L, -1, device_id, group);
 	luahid_setfield(L, -1, device_id, vendor);
 	luahid_setfield(L, -1, device_id, product);
```
lneto, line 125: this entire block can be moved to a macro.. you have the exact same pattern in pushdev()

lib/luahid.c
```diff
@@ -99,14 +103,50 @@ do { 						\
...
 	luahid_setfield(L, -1, device_id, driver_data);
```
lneto: this entire block can be moved to a macro.. you have the exact same pattern in pushdev()

lib/luahid.c
qrsikno2: fixed

lib/luahid.c
```diff
@@ -153,6 +193,59 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
 }
 
+static inline lunatik_object_t *luahid_getdescriptor(lua_State *L, luahid_t *hid)
+{
+	if (lunatik_getregistry(L, hid->descriptor) != LUA_TUSERDATA) {
+		pr_err("could not find descriptor\n");
+		return NULL;
+	}
+	lunatik_object_t *data = (lunatik_object_t *)lunatik_toobject(L, -1);
+	if (unlikely(data == NULL)) {
```
lneto: you can just use an `||` here

lib/luahid.c
lneto: > just pass idx - 1 on the caller side

^

lneto: the point of passing idx is to not assume anything about the stack on the caller side..

lib/luahid.c
```diff
@@ -99,14 +103,50 @@ do { 						\
...
+	luahid_setfield(L, -1, hdev, product);
+	luahid_setfield(L, -1, hdev, version);
+	lua_pushstring(L, hdev->name);
+	lua_setfield(L, -2, "name");
+	return 1; /* hdev table */
+}
+
+static inline int luahid_pushinfo(lua_State *L)
+{
+	struct hid_device *hdev = (struct hid_device *)lua_touserdata(L, 2);
```
lneto: why not passing the lightuserdata as the first argument?

lneto: it's a bit weird here to access first the second arg and second the first.. unless you have a reason to..

lib/luahid.c

lib/luahid.c
qrsikno2: fixed

lib/luahid.c
qrsikno2: fixed

lib/luahid.c
```diff
@@ -99,14 +103,48 @@ do { 						\
...
+{
+	struct hid_device *hdev = (struct hid_device *)lua_touserdata(L, 1);
+	luahid_pushcommonfields(L, hdev);
+	luahid_setfield(L, -1, hdev, version);
+	lua_pushstring(L, hdev->name);
+	lua_setfield(L, -2, "name");
+	return 1; /* hdev table */
+}
+
+static inline void luahid_pushinfo(lua_State *L, int idx, struct hid_device *hdev)
```
lneto, line 143: I don't think we need this auxiliary function here.. if you are not using this pattern anywhere else, I think it would be cleaner to have the two calls bellow inline in luahid_doreport_fixup

lib/luahid.c
```diff
@@ -99,14 +103,48 @@ do { 						\
...
 } while (0)
 
-static inline void luahid_pushdevid(lua_State *L, int idx, const struct hid_device_id *device_id)
+#define luahid_pcall(L, func, arg) 				\
+do { 								\
+	int n = lua_gettop(L); 					\
+	lua_pushcfunction(L, func); 				\
+	lua_pushlightuserdata(L, (void *)arg); 			\
+	if (lua_pcall(L, 1, LUA_MULTRET, 0) != LUA_OK) { 	\
+		pr_warn("%s: %s\n", #func, lua_tostring(L, -1));\
```
lneto: you miss a tab before `\` here

lib/luahid.c
qrsikno2: i think it would be "using this pattern anywhere else". Although this is not the case now. In fact, this user-defined priv_data is available at any time after probing, so I think it is necessary to retain it so that it can be provided to users in the callback of subsequent development. Or do you have any good ideas to deal with this situation?

lib/luahid.c
qrsikno2: fixed

lib/luahid.c
```diff
@@ -99,14 +103,48 @@ do { 						\
...
+		lua_settop(L, n);				\
+	} 							\
+} while (0)
+
+#define luahid_pushcommonfields(L, udata)		\
```
```suggestion
#define luahid_newtable(L, dev, extra)		\
```

lib/luahid.c
```diff
@@ -99,14 +103,48 @@ do { 						\
...
+do { 							\
+	lua_newtable(L); 				\
+	luahid_setfield(L, -1, udata, bus); 		\
+	luahid_setfield(L, -1, udata, group); 		\
+	luahid_setfield(L, -1, udata, vendor); 		\
+	luahid_setfield(L, -1, udata, product); 	\
```
```suggestion
	luahid_setfield(L, -1, dev, product); 	\
	luahid_setfield(L, -1, dev, extra); 	\
```

lib/luahid.c
```diff
@@ -99,14 +103,48 @@ do { 						\
 	lua_setfield(L, idx - 1, #field);	\
 } while (0)
 
-static inline void luahid_pushdevid(lua_State *L, int idx, const struct hid_device_id *device_id)
+#define luahid_pcall(L, func, arg) 					\
```
lneto, line 106: it's more like a `prun` then a `pcall`, as you are not passing a CFunction..

lneto: let's focus on this later..

lib/luahid.c
qrsikno2: finished

lib/luahid.c
qrsikno2: finished

lib/luahid.c
```diff
@@ -153,6 +190,57 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+static int luahid_doreport_fixup(lua_State *L, luahid_t *hid, struct hid_device *hdev, __u8 *rdesc, unsigned int rsize)
+{
+	if (luahid_checkdriver(L, hid, -1, "_info") || lua_getfield(L, -2, "report_fixup") != LUA_TFUNCTION) {
+		pr_warn("report_fixup: invaild driver\n");
+		return 0;
+	}
+
+	lua_pushvalue(L, -3); /* hid.ops */
+	luahid_pcall(L, luahid_pushhdev, hdev);
+	luahid_pushinfo(L, -4, hdev);
```
lneto, line 212:
```suggestion
	lua_pushlightuserdata(L, hdev);
	lua_rawget(L, -5);
```

qrsikno2: so i would mark this as solved due to our discussion.

lib/luahid.c
```diff
@@ -99,14 +103,47 @@ do { 						\
...
+} while (0)
+
+#define luahid_newtable(L, dev, extra)			\
+do { 							\
+	lua_newtable(L); 				\
+	luahid_setfield(L, -1, dev, bus); 		\
+	luahid_setfield(L, -1, dev, group); 		\
+	luahid_setfield(L, -1, dev, vendor); 		\
+	luahid_setfield(L, -1, dev, product); 		\
+	luahid_setfield(L, -1, dev, extra); 		\
```
lneto, line 124: please, fix the tabs

lib/luahid.c
```diff
@@ -153,6 +190,57 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+		return NULL;
+	}
+	return data;
+}
+
+static int luahid_doreport_fixup(lua_State *L, luahid_t *hid, struct hid_device *hdev, __u8 *rdesc, unsigned int rsize)
+{
+	if (luahid_checkdriver(L, hid, -1, "_info") || lua_getfield(L, -2, "report_fixup") != LUA_TFUNCTION) {
+		pr_warn("report_fixup: invaild driver\n");
+		return 0;
```
```suggestion
		goto out;
```

lib/luahid.c
```diff
@@ -153,6 +190,57 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+	}
+
+	lua_pushvalue(L, -3); /* hid.ops */
+	luahid_pcall(L, luahid_pushhdev, hdev);
+	luahid_pushinfo(L, -4, hdev);
+
+	lunatik_object_t *data;
+	if ((data = luahid_getdescriptor(L, hid)) == NULL) {
+		pr_warn("report_fixup: descriptor not found\n");
+		return 0;
```
```suggestion
		goto out;
```

lib/luahid.c
```diff
@@ -153,6 +190,57 @@ static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
+	}
+	luadata_reset(data, rdesc, rsize, LUADATA_OPT_NONE);
+
+	if (lua_pcall(L, 4, 0, 0) != LUA_OK)
+		pr_warn("report_fixup: %s\n", lua_tostring(L, -1));
+
+	luadata_clear(data);
+	return 0;
```
```suggestion
out:
	return 0; /* unused */
```

PR #304, opened 2025-07-13
add support for require("mylib.sub")

lneto:
* where adding it to lib/luamylib_sub.c


jperon (approved):
Just tested, it works as expected.

PR #311, opened 2025-07-24
crypto_acompress

### LINKED ISSUE #307: Build failing in latest release
Build logs:

```
make -C /lib/modules/6.15.7-arch1-1/build M=/home/wazir/workspace/repos/lunatik CONFIG_LUNATIK=m	\
CONFIG_LUNATIK_RUN=m CONFIG_LUNATIK_RUNTIME=y CONFIG_LUNATIK_DEVICE=m	\
CONFIG_LUNATIK_LINUX=m CONFIG_LUNATIK_NOTIFIER=m CONFIG_LUNATIK_SOCKET=m \
CONFIG_LUNATIK_RCU=m CONFIG_LUNATIK_THREAD=m CONFIG_LUNATIK_FIB=m \
CONFIG_LUNATIK_DATA=m CONFIG_LUNATIK_PROBE=m CONFIG_LUNATIK_SYSCALL=m \
CONFIG_LUNATIK_XDP=m CONFIG_LUNATIK_FIFO=m CONFIG_LUNATIK_XTABLE=m \
CONFIG_LUNATIK_CRYPTO_COMP=m
make[1]: Entering directory '/usr/lib/modules/6.15.7-arch1-1/build'
make[2]: Entering directory '/home/wazir/workspace/repos/lunatik'
  CC [M]  lua/lapi.o
  CC [M]  lua/lcode.o
  CC [M]  lua/lctype.o
  CC [M]  lua/ldebug.o
  CC [M]  lua/ldo.o
  CC [M]  lua/ldump.o
  CC [M]  lua/lfunc.o
  CC [M]  lua/lgc.o
  CC [M]  lua/llex.o
  CC [M]  lua/lobject.o
  CC [M]  lua/lmem.o
  CC [M]  lua/lopcodes.o
  CC [M]  lua/lparser.o
  CC [M]  lua/lstate.o
  CC [M]  lua/lstring.o
  CC [M]  lua/ltable.o
  CC [M]  lua/ltm.o
  CC [M]  lua/lundump.o
  CC [M]  lua/lvm.o
  CC [M]  lua/lzio.o
  CC [M]  lua/lauxlib.o
  CC [M]  lua/lbaselib.o
  CC [M]  lua/lcorolib.o
  CC [M]  lua/ldblib.o
  CC [M]  lua/lstrlib.o
  CC [M]  lua/ltablib.o
  CC [M]  lua/lutf8lib.o
  CC [M]  lua/lmathlib.o
  CC [M]  lua/linit.o
  CC [M]  lua/loadlib.o
  AS      klibc/usr/klibc/arch/x86_64/setjmp.o
  CC [M]  lunatik_aux.o
  CC [M]  lunatik_obj.o
  CC [M]  lunatik_core.o
  CC [M]  lunatik_run.o
  CC [M]  lib/luadevice.o
  CC [M]  lib/lualinux.o
  CC [M]  lib/luasocket.o
  CC [M]  lib/luanotifier.o
  CC [M]  lib/luarcu.o
  CC [M]  lib/luathread.o
  CC [M]  lib/luafib.o
  CC [M]  lib/luadata.o
  CC [M]  lib/luaprobe.o
  CC [M]  lib/luasyscall.o
  CC [M]  lib/luaxdp.o
  CC [M]  lib/luafifo.o
  CC [M]  lib/luaxtable.o
  CC [M]  lib/luanetfilter.o
  CC [M]  lib/luacompletion.o
  CC [M]  lib/luacrypto_shash.o
  CC [M]  lib/luacrypto_skcipher.o
  CC [M]  lib/luacrypto_aead.o
  CC [M]  lib/luacrypto_rng.o
  CC [M]  lib/luacrypto_comp.o
  LD [M]  lunatik.o
lunatik.o: warning: objtool: luaV_execute+0x71: sibling call from callable instruction with modified stack frame
In file included from lib/luacrypto_comp.c:30:
lib/luacrypto_comp.c: In function ‘luacrypto_comp_release’:
lib/luacrypto_comp.c:34:46: error: implicit declaration of function ‘crypto_free_comp’; did you mean ‘crypto_free_tfm’? [-Wimplicit-function-declaration]
   34 | LUACRYPTO_RELEASER(comp, struct crypto_comp, crypto_free_comp, NULL);
      |                                              ^~~~~~~~~~~~~~~~
lib/luacrypto.h:70:17: note: in definition of macro ‘LUACRYPTO_RELEASER’
   70 |                 obj_free(obj);                                                  \
      |                 ^~~~~~~~
In file included from lib/luacrypto_comp.c:28:
lib/luacrypto_comp.c: In function ‘luacrypto_comp_compress’:
lib/luacrypto_comp.c:54:24: error: implicit declaration of function ‘crypto_comp_compress’; did you mean ‘luacrypto_comp_compress’? [-Wimplicit-function-declaration]
   54 |         lunatik_try(L, crypto_comp_##name, tfm, data, (unsigned int)datalen, output_buf, &max_len);     \
      |                        ^~~~~~~~~~~~
././lunatik.h:126:20: note: in definition of macro ‘lunatik_tryret’
  126 |         if ((ret = op(__VA_ARGS__)) < 0) {                      \
      |                    ^~
lib/luacrypto_comp.c:54:9: note: in expansion of macro ‘lunatik_try’
   54 |         lunatik_try(L, crypto_comp_##name, tfm, data, (unsigned int)datalen, output_buf, &max_len);     \
      |         ^~~~~~~~~~~
lib/luacrypto_comp.c:70:1: note: in expansion of macro ‘LUACRYPTO_COMP_OPERATION’
   70 | LUACRYPTO_COMP_OPERATION(compress);
      | ^~~~~~~~~~~~~~~~~~~~~~~~
lib/luacrypto_comp.c: In function ‘luacrypto_comp_decompress’:
lib/luacrypto_comp.c:54:24: error: implicit declaration of function ‘crypto_comp_decompress’; did you mean ‘luacrypto_comp_decompress’? [-Wimplicit-function-declaration]
   54 |         lunatik_try(L, crypto_comp_##name, tfm, data, (unsigned int)datalen, output_buf, &max_len);     \
      |                        ^~~~~~~~~~~~
././lunatik.h:126:20: note: in definition of macro ‘lunatik_tryret’
  126 |         if ((ret = op(__VA_ARGS__)) < 0) {                      \
      |                    ^~
lib/luacrypto_comp.c:54:9: note: in expansion of macro ‘lunatik_try’
   54 |         lunatik_try(L, crypto_comp_##name, tfm, data, (unsigned int)datalen, output_buf, &max_len);     \
      |         ^~~~~~~~~~~
lib/luacrypto_comp.c:83:1: note: in expansion of macro ‘LUACRYPTO_COMP_OPERATION’
   83 | LUACRYPTO_COMP_OPERATION(decompress);
      | ^~~~~~~~~~~~~~~~~~~~~~~~
lib/luacrypto_comp.c: In function ‘luacrypto_comp_new’:
lib/luacrypto_comp.c:114:41: error: implicit declaration of function ‘crypto_alloc_comp’; did you mean ‘crypto_alloc_base’? [-Wimplicit-function-declaration]
  114 | LUACRYPTO_NEW(comp, struct crypto_comp, crypto_alloc_comp, luacrypto_comp_class, NULL);
      |                                         ^~~~~~~~~~~~~~~~~
lib/luacrypto.h:37:18: note: in definition of macro ‘LUACRYPTO_NEW’
   37 |         T *tfm = alloc(algname, 0, 0);                                                                  \
      |                  ^~~~~
lib/luacrypto_comp.c:114:41: error: initialization of ‘struct crypto_comp *’ from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
  114 | LUACRYPTO_NEW(comp, struct crypto_comp, crypto_alloc_comp, luacrypto_comp_class, NULL);
      |                                         ^~~~~~~~~~~~~~~~~
lib/luacrypto.h:37:18: note: in definition of macro ‘LUACRYPTO_NEW’
   37 |         T *tfm = alloc(algname, 0, 0);                                                                  \
      |                  ^~~~~
make[4]: *** [/usr/lib/modules/6.15.7-arch1-1/build/scripts/Makefile.build:203: lib/luacrypto_comp.o] Error 1
make[3]: *** [/usr/lib/modules/6.15.7-arch1-1/build/Makefile:2003: .] Error 2
make[2]: *** [/usr/lib/modules/6.15.7-arch1-1/build/Makefile:248: __sub-make] Error 2
make[2]: Leaving directory '/home/wazir/workspace/repos/lunatik'
make[1]: *** [Makefile:248: __sub-make] Error 2
make[1]: Leaving directory '/usr/lib/modules/6.15.7-arch1-1/build'
make: *** [Makefile:24: all] Error 2
```

It builds successfully for this commit: https://github.com/luainkernel/lunatik/commit/8e79d4ed96dcfc7d49edf748144366edbbd77c94

uname: 

```
Linux virtme-ng 6.15.7-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 17 Jul 2025 21:05:29 +0000 x86_64 GNU/Linux
```

jperon:
Fixes #307


config.ld
```diff
@@ -15,7 +15,8 @@ title = 'Lunatik API Documentation'
 file = {
 	'./lib/luacompletion.c',
 	'./lib/crypto/aead.lua',
-	'./lib/luacrypto_comp.c',
+	'./lib/luacrypto_acompress.c',
+	'./lib/crypto/comp.lua',
```
lneto, line 19: I would call this compress

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,225 @@
...
+
+static void __exit luacrypto_acompress_exit(void)
+{
+}
+
+module_init(luacrypto_acompress_init);
+module_exit(luacrypto_acompress_exit);
+MODULE_LICENSE("Dual MIT/GPL");
+MODULE_AUTHOR("jperon <cataclop@hotmail.com>");
+MODULE_DESCRIPTION("Lunatik low-level Linux Crypto API interface (ACOMPRESS)");
```
lneto: why caps here?

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,224 @@
...
+* @within acompress
+*/
+static int luacrypto_acompress_new(lua_State *L)
+{
+	const char *algname = luaL_checkstring(L, 1);
+
+	lunatik_object_t *object = lunatik_newobject(L, &luacrypto_acompress_class, sizeof(luacrypto_acomp_req_t));
+	luacrypto_acomp_req_t *req = (luacrypto_acomp_req_t *)object->private;
+
+	memset(req, 0, sizeof(*req));
```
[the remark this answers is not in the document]
lneto: er.. no.

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,225 @@
...
```
lneto: I prefer to be explicity re the type.. but it's up to you..
```suggestion
	memset(req, 0, sizeof(luacrypto_acomp_req_t));
```

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,225 @@
...
+#include <linux/version.h>
+
+#include <lua.h>
+#include <lualib.h>
+#include <lauxlib.h>
+#include <lunatik.h>
+
+#include "luacrypto.h"
+
+typedef struct luacrypto_acomp_req_s {
```
lneto: please, let's keep the same prefix for everything in the module..

lneto:
```suggestion
typedef struct luacrypto_acompress_s {
```

I don't think we need `req` here..

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,225 @@
...
+	obj->sg_out = lunatik_checkalloc(L, sizeof(struct scatterlist));
+	sg_init_one(obj->sg_out, obj->outbuf_data, out_len);
+
+	acomp_request_set_params(obj->req, obj->sg_in, obj->sg_out, in_len, out_len);
+
+	obj->cb_ref = cb_ref;
+	obj->L = L;
+
+	acomp_request_set_callback(obj->req, 0, luacrypto_acompress_lua_callback, obj);
+
```
lneto, line 165:
```suggestion
```

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,225 @@
...
+}
+
+#define LUACRYPTO_ACOMPRESS_OPERATION(name)							\
+static int luacrypto_acompress_##name(lua_State *L)						\
+{												\
+	luacrypto_acomp_req_t *obj = luacrypto_acompress_check(L, 1);				\
+	size_t in_len;										\
+	const char *in_buf = luaL_checklstring(L, 2, &in_len);					\
+	lunatik_checkbounds(L, 2, in_len, 1, UINT_MAX);						\
+	unsigned int out_len = lunatik_checkuint(L, 3);						\
```
lneto: \n

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,225 @@
...
+#include <lunatik.h>
+
+#include "luacrypto.h"
+
+typedef struct luacrypto_acomp_req_s {
+	struct crypto_acomp *tfm;
+	struct acomp_req *req;
+	struct scatterlist *sg_in;
+	struct scatterlist *sg_out;
+	int cb_ref;
```
lneto:
```suggestion
	int cb;
```

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,225 @@
...
+* @see new
+* @type ACOMPRESS
+*/
+
+static void luacrypto_acompress_lua_callback(void *data, int err)
+{
+	luacrypto_acomp_req_t *obj = data;
+
+	lua_rawgeti(obj->L, LUA_REGISTRYINDEX, obj->cb_ref);
+	lua_pushinteger(obj->L, err);
```
lneto: why not using lunatik_register?

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,225 @@
...
+
+	if (err == 0)
+		lua_pushlstring(obj->L, (const char *)obj->outbuf_data, obj->req->dlen);
+	else
+		lua_pushnil(obj->L);
+
+	lua_call(obj->L, 2, 0);
```
lneto: why not pcall?

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,225 @@
...
+	sg_init_one(obj->sg_in, in_buf, in_len);
+	obj->sg_out = lunatik_checkalloc(L, sizeof(struct scatterlist));
+	sg_init_one(obj->sg_out, obj->outbuf_data, out_len);
+
+	acomp_request_set_params(obj->req, obj->sg_in, obj->sg_out, in_len, out_len);
+
+	obj->cb_ref = cb_ref;
+	obj->L = L;
+
+	acomp_request_set_callback(obj->req, 0, luacrypto_acompress_lua_callback, obj);
```
lneto: I don't think you should let it simply call Lua directly.. how are you planning to sync? why not using lunatik_run?

lib/luacrypto_acompress.c
jperon: Same logic as all other crypto wrappers (shash, rng etc.).

lib/luacrypto_acompress.c
jperon: The idea is to sync with completion (like in the lua wrapper).

lneto: I meant, this callback should be called assynchronously from another control flow, right? in this case, we need to synchronize its execution (e.g., using mutex or spinlock).. we usually do this `using lunatik_run`.. for instance, what happens if the same runtime is used concurrently by the acomp module and a kthread or a chardev?

jperon: I think I don’t totally get the point. If one uses acompress [this way](https://github.com/luainkernel/lunatik/blob/jperon_crypto_acompress/lib/crypto/comp.lua#L33-L42), I think the runtime can’t be used by anything else while `done:wait()` is waiting? If so, shouldn’t it be the "responsibility" of the caller to spawn a new runtime using lunatik lua api when needed?

lneto: that's the thing, we can not let synchronization to be handled by the Lua scripts.. our bindings need to enforce it.. what happens if the script fails to call wait()? it's responsibility of the binding to properly synchronize if it let the kernel call the runtime assynchronously.. think about luanetfilter, for instance, if we define callbacks there to be called on netfilter hooks and we always enforce synchronization using spinlocks on their calls.. using lunatik_run()

lneto: I tried to explain this on the Section "Kernel Calling" [here](https://medium.com/@lourival.neto/from-the-kernel-to-the-moon-a-journey-into-lunatik-bindings-c2cac8816f9e)

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,247 @@
...
+		pr_err("Lua callback error: %s\n", lua_tostring(L, -1));
+		lua_pop(L, 1);
+	}
+	lua_pop(L, 1);
+	return 0;
+}
+
+static void luacrypto_acompress_docall(void *data, int err)
+{
+	lunatik_object_t *runtime = lunatik_toruntime(((luacrypto_acompress_t *)data)->L);
```
lneto: you should actually save the runtime in your struct instead of the Lua state.. you have no guarantee that such state will be still valid when your callback is called.. you can just follow the pattern we are using in other modules

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,247 @@
...
+	if (lunatik_getregistry(L, lunatik_toobject(L, 1)) != LUA_TFUNCTION) {
+		pr_err("No callback function found for acomp object\n");
+		lua_pop(L, 1);
+		return 0;
+	}
+
+	lua_pushinteger(L, err);
+
+	if (err == 0)
+		lua_pushlstring(L, (const char *)obj->outbuf_data, obj->req->dlen);
```
lneto: it might cause a kernel panic due to ENOMEM.. in this case, you are not running in a pcall.. you should move everything need that might raise an error to a pcall on a callback called from the kernel

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,247 @@
...
+	lunatik_free(obj->outbuf_data);
+	acomp_request_free(obj->req);
+	crypto_free_acomp(obj->tfm);
+}
+
+static int luacrypto_acompress_lua_cb(lua_State *L, void *data, int err)
+{
+	luacrypto_acompress_t *obj = data;
+
+	if (lunatik_getregistry(L, lunatik_toobject(L, 1)) != LUA_TFUNCTION) {
```
lneto: this pattern looks weird to me.. we shouldn't neve use toobject unless in the cases we are perfectly sure it's our object (e.g., after checking it).. it's not clear to me why you need it here anyway.. I would take a look in the patter we are using for acquiring Lua values in kernel callbacks, for instance, luanetfilter and luaxdp

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,247 @@
...
+#include <linux/slab.h>
+#include <linux/limits.h>
+#include <linux/completion.h>
+#include <linux/scatterlist.h>
+#include <linux/version.h>
+
+#include <lua.h>
+#include <lualib.h>
+#include <lauxlib.h>
+#include <lunatik.h>
```
lneto:
```suggestion

#include <lunatik.h>
```

lib/luacrypto_acompress.c
lneto: see this.. https://github.com/luainkernel/lunatik/blob/master/lib/luanetfilter.c#L187

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,373 @@
...
+
+	object = lunatik_newobject(L, &luacrypto_acomp_req_class, sizeof(luacrypto_acomp_req_t));
+	req = (luacrypto_acomp_req_t *)object->private;
+
+	memset(req, 0, sizeof(luacrypto_acomp_req_t));
+	req->runtime = lunatik_toruntime(L);
+	lunatik_getobject(req->runtime);
+	req->cb = LUA_NOREF;
+	req->buf = LUA_NOREF;
+	req->self = LUA_NOREF;
```
lneto: I think using a lua table for all you need will simplify a bit the implementation..

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,373 @@
...
+												\
+	lua_pushvalue(L, 4);									\
+	int cb = luaL_ref(L, LUA_REGISTRYINDEX);						\
+	lua_pushvalue(L, 2);									\
+	int buf = luaL_ref(L, LUA_REGISTRYINDEX);						\
+												\
+	lunatik_try(L, luacrypto_acomp_req_prepare, obj, L, in_buf, in_len, out_len, cb, buf);	\
+	int ret = crypto_acomp_##name(obj->req);						\
+	if (ret != -EINPROGRESS)								\
+		luacrypto_acomp_req_docall(obj, ret);						\
```
lneto: I don't get this.. shouldn't this function be called by the crypto_acomp request itself? why you are calling synchronously here?

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,373 @@
...
+
+	memset(req, 0, sizeof(luacrypto_acomp_req_t));
+	req->runtime = lunatik_toruntime(L);
+	lunatik_getobject(req->runtime);
+	req->cb = LUA_NOREF;
+	req->buf = LUA_NOREF;
+	req->self = LUA_NOREF;
+	
+	lua_pushvalue(L, 1); /* push TFM object */
+	req->tfm = luaL_ref(L, LUA_REGISTRYINDEX);
```
lneto: we have a macro for this pattern on lunatik.h, right? couldn't we use it?

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,373 @@
...
+
+	if (obj->req) {
+		acomp_request_free(obj->req);
+		obj->req = NULL;
+	}
+
+	if (obj->outbuf) {
+		lunatik_free(obj->outbuf);
+		obj->outbuf = NULL;
+	}
```
lneto, line 62: shouldn't this be done on the obj release callback?

lib/luacrypto_acompress.c
lneto: it looks to me that we are trying to turn an async API into sync.. I would follow the same approach the clients of this kernel API are doing.. can you point to kernel usage example, btw? my bet is that they have an API for checking result later on.. we could do the same..

jperon: We call it synchronously only in the case when crypto_acomp_… has already completed (synchronously), in which case the callback would never be invoked automatically. If ret == -EINPROGRESS, the behavior remains asynchronous.

This follows the standard pattern used in the kernel. For example:

In xfrm_ipcomp.c: https://github.com/jonmason/ntb/blob/09ca5e67c4ab3c8bff7aef6085ee0cbb4337c3f8/net/xfrm/xfrm_ipcomp.c#L272
In crypto/acompress.c: https://github.com/torvalds/linux/blob/master/crypto/acompress.c#L278

lib/luacrypto_acompress.c
lneto: sorry.. it was my tired eyes.. it is the release callback.. sorry for the noise..

Kbuild
```diff
@@ -51,6 +51,6 @@ obj-$(CONFIG_LUNATIK_CRYPTO_SHASH) += lib/luacrypto_shash.o
 obj-$(CONFIG_LUNATIK_CRYPTO_SKCIPHER) += lib/luacrypto_skcipher.o
 obj-$(CONFIG_LUNATIK_CRYPTO_AEAD) += lib/luacrypto_aead.o
 obj-$(CONFIG_LUNATIK_CRYPTO_RNG) += lib/luacrypto_rng.o
-obj-$(CONFIG_LUNATIK_CRYPTO_COMP) += lib/luacrypto_comp.o
+obj-$(CONFIG_LUNATIK_CRYPTO_ACOMPRESS) += lib/luacrypto_acompress.o
```
lneto: shouldn't we keep comp for older kernels?

jperon: OK, but that means conditionnally, as it won’t compile on recent ones.

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,350 @@
...
+* The callback is invoked when the operation completes (either synchronously or asynchronously).
+* @function decompress
+* @tparam string data The compressed data to decompress.
+* @tparam integer output_size The maximum size of the output buffer.
+* @tparam function callback The callback function to invoke upon completion. It receives two arguments: `err` (integer error code, 0 on success) and `data` (string containing decompressed data, or nil on error).
+* @raise Error if the request object is busy, or if parameters are invalid.
+*/
+LUACRYPTO_ACOMP_REQ_OPERATION(decompress)
+
+static const luaL_Reg luacrypto_acomp_req_mt[] = {
```
lneto: let's keep the pattern of prefixing with the full module name.. `luacrypto_acompress_*`

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,363 @@
...
+
+static const luaL_Reg luacrypto_acompress_lib[] = {
+	{"new", luacrypto_acompress_new},
+	{NULL, NULL}
+};
+
+static const lunatik_class_t luacrypto_acompress_classes[] = {
+	luacrypto_acompress_class,
+	luacrypto_acompress_req_class,
+	{NULL}
```
lneto:
```suggestion
	NULL
```
shouldn't be just `NULL`, here? the array element is a pointer, not a struct, right?

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,363 @@
...
+	lua_pushvalue(L, 4);
+	lua_setfield(L, -2, "cb");
+	
+	lua_pushvalue(L, 2);
+	lua_setfield(L, -2, "buf");
+	
+	lua_pushvalue(L, 1);
+	lua_setfield(L, -2, "self");
+	
+	obj->refs = luaL_ref(L, LUA_REGISTRYINDEX);
```
lneto: I think moving this to an auxiliary function, also renaming `ref` to `request` (or another name for this table), would make it more readable.. btw, can't we use the object pointer to register this table instead of using a new ref on Lua registry? I would also move the call to this aux function to the macro bellow as it accesses argument indexes directly.. we are trying to reference the argument numbers on the lua_CFunction itself or passing the index to auxiliary functions to make easier to understand the context of the stack

lunatik.h
```diff
@@ -259,21 +267,46 @@ static inline void lunatik_newnamespaces(lua_State *L, const lunatik_namespace_t
...
+
+#define LUNATIK_NEWLIB_MULTICLASS(libname, funcs, classes, namespaces)	\
+int luaopen_##libname(lua_State *L);					\
+int luaopen_##libname(lua_State *L)					\
+{									\
+	luaL_newlib(L, funcs);						\
+	lunatik_register_classes(L, classes);				\
+	lunatik_register_namespaces(L, namespaces);			\
+	return 1;							\
+}									\
```
lneto: sorry, I know that I advised otherwise.. but now that we can see it implemented, I think it would be simpler to have just one API and to just add the class array to LUANATIK_NEWLIB.. I can help to update the other modules if needed.. it will be more alike the flags case..  we also need to keep in mind that we will need to update the README

lib/crypto/compress.lua
```diff
@@ -0,0 +1,77 @@
...
+-- This module provides a synchronous, high-level API, but implemented on top
+-- of the low-level `crypto.acompress` and `completion` modules.
+--
+-- @module crypto.compress
+
+local acomp = require("crypto.acompress")
+local completion = require("completion")
+
+--- @type COMPRESS
+local COMPRESS = {}
```
lneto: I'm not sure we are using this pattern (with all caps) elsewhere.. it looked a bit odd to me.. but no strong feelings.. just a note..

lunatik.h
lneto: I'd also separate this API change in another commit.. before the acompress one

tests/crypto/comp.lua
```diff
@@ -3,7 +3,11 @@
 -- SPDX-License-Identifier: MIT OR GPL-2.0-only
 --
 
-local comp = require"crypto.comp"
+local status, comp = pcall(require, "crypto.comp")
+if not status then
+	print("crypto.comp not available, skipping tests")
+	return
+end
```
lneto: instead of doing this, we can register an empty library in this case.. we already do it on luaxdp if I'm not mistaken..

lib/luacrypto_acompress.c
```diff
@@ -0,0 +1,363 @@
...
+/* ACOMPRESS Request Object */
+
+typedef struct luacrypto_acompress_req_s {
+	struct acomp_req *req;
+	struct scatterlist sg_in;
+	struct scatterlist sg_out;
+	int refs;
+	lunatik_object_t *runtime;
+	u8 *outbuf;
+	size_t outbuf_len;
```
lneto, line 34: this is a very repetitive pattern, we use it on luadata (perhaps on other modules as well).. wondering if it wouldn't worth to have a `lunatik_buffer_t` or similar..

jperon:
@lneto Just updated this PR to match current master. Added a compatibility wrapper for `require"crypto.comp"`.

PR #312, opened 2025-07-25
add callback for raw_event

qrsikno2:
This pull request introduces significant updates to the `lib/luahid.c` file, primarily focused on enhancing the handling of HID (Human Interface Device) events and improving the extensibility of the HID driver through Lua scripting. The changes include adding new functions for processing raw events, extending the Lua HID driver interface, and improving data encapsulation.

### Enhancements to HID event handling:

* Added the `luahid_raw_event` function to handle raw HID events, enabling Lua scripts to process incoming HID reports and raw data. This involves invoking a Lua callback (`raw_event`) and passing detailed information about the HID device, report, and data.
* Introduced helper functions (`luahid_pushhdev`, `luahid_pushinfo`, and `luahid_pushreport`) to encapsulate HID device and report data into Lua tables for easier manipulation within Lua scripts.

### Extensions to Lua HID driver interface:

* Updated the `luahid_register` function to include the new `raw_event` callback in the Lua HID driver interface, allowing Lua scripts to define custom behavior for handling raw HID events.
* Added documentation and examples for the new `raw_event` callback in the Lua HID driver interface, demonstrating how to process raw HID events.

### Codebase improvements:

* Included the `luadata.h` header file to facilitate the creation and manipulation of Lua data objects, ensuring better integration with the existing Lua runtime.

### DIFF
--- a/lib/luahid.c
+++ b/lib/luahid.c
@@ -269,7 +269,7 @@ static int luahid_do##NAME(lua_State *L)								\
 
 LUAHID_CALLBACK(probe, 2, 1, err);
 LUAHID_CALLBACK(report_fixup, 4, 0, warn);
-LUAHID_CALLBACK(raw_event, 5, 1, warn);
+LUAHID_CALLBACK(raw_event, 5, 1, err);
 
 static int luahid_runprobe(lua_State *L, struct hid_device *hdev, luahid_probe_arg_t *arg)
 {


lib/luaxtable.c
```diff
@@ -154,7 +154,9 @@ static T luaxtable_##hook(U skb, V par)					\
 	const luaxtable_info_t *info = (const luaxtable_info_t *)par->huk##info;	\
 	luaxtable_t *xtable = info->data;				\
 									\
+	local_bh_disable();						\
 	lunatik_run(xtable->runtime, luaxtable_do##hook, ret, xtable, skb, par, luaxtable_hooks.hook##_fallback);	\
+	local_bh_enable(); 						\
```
lneto, line 159: create macros for this.. lunatik_runbh and lunatik_runirq

lib/luahid.c
```diff
@@ -242,6 +255,64 @@ static luahid_ret_t luahid_report_fixup(struct hid_device *hdev, __u8 *rdesc, un
...
+}
+
+static int luahid_doraw_event(lua_State *L, luahid_t *hid, struct hid_device *hdev, struct hid_report *report, u8 *data, int size, int *ret)
+{
+	if (luahid_checkdriver(L, hid, -1, "_info")) {
+		pr_err("raw_event: couldn't find driver");
+		return -ENXIO;
+	}
+
+	if (lua_getfield(L, -2, "raw_event") != LUA_TFUNCTION)
```
lneto: https://www.lua.org/manual/5.4/manual.html#lua_getfield

please, again, be more careful..

lneto: also, check if any other API can raise errors here..

lib/luahid.c
```diff
@@ -304,6 +380,7 @@ static int luahid_register(lua_State *L)
 	luahid_t *hid = (luahid_t *)object->private;
 	memset(hid, 0, sizeof(luahid_t));
 	luadata_attach(L, hid, descriptor);
+	luadata_attach(L, hid, raw_event);
```
lneto: why do you need another buffer?

lib/luaxtable.c
qrsikno2: checked and fixed. added in `lunatik.h`

lib/luahid.c
qrsikno2: checked and found `getfield` & `setfield` may raise errors.

created `pxxxfield` for protecting/wrapping it. created `safexxxfield` to make `pxxxfield` fits the original signature.

lib/luahid.c
qrsikno2: just because the `raw_event` will pass a small piece of memory to the user's callback.

it's the raw_event's data itself. 
so i created it for the same reason as `descriptor`'s creation

lneto: my  point is "why not reusing the same", you are going to reset it everytime anyway..

qrsikno2: fixed

lib/luahid.c
```diff
@@ -31,6 +31,7 @@
 typedef struct luahid_s {
 	lunatik_object_t *runtime;
 	lunatik_object_t *descriptor;
+	lunatik_object_t *raw_event;
```
lneto: is it being used anywhere?

lunatik.h
```diff
@@ -61,6 +61,31 @@ do {									\
...
 
+#define lunatik_runbh(runtime, handler, ret, ...)			\
+do {									\
+	local_bh_disable();						\
+	lunatik_lock(runtime);						\
+	if (unlikely(!lunatik_getstate(runtime)))			\
+		ret = -ENXIO;						\
+	else								\
+		lunatik_handle(runtime, handler, ret, ## __VA_ARGS__);	\
+	lunatik_unlock(runtime);					\
```
lneto, line 72: why copying this block instead of just calling `lunatik_run`?

lib/luahid.c
qrsikno2: removed

lunatik.h
qrsikno2: fixed.

lib/luahid.c
```diff
@@ -237,11 +290,88 @@ static luahid_ret_t luahid_report_fixup(struct hid_device *hdev, __u8 *rdesc, un
...
+static int luahid_runraw_event(lua_State *L, luahid_arg_t *arg)
+{
+	luahid_pcall(L, luahid_doraw_event, arg);
+	return lua_tointeger(L, -1);
+}
+
+static int luahid_raw_event(struct hid_device *hdev, struct hid_report *report, u8 *data, int size)
+{
+	struct hid_driver *driver = hdev->driver;
+	luahid_t *hid = container_of(driver, luahid_t, driver);
```
lneto: Do all your callbacks do this? If yes, you don't need to pass luahid pointer in the args, you can retrieve it from driver.. you should think on the minimum you need to actual pass through arguments and the most you can retrieve in the CFunction to clean up your interface.

lneto: you should check if it's enough to pass hdev.. the point is to simplify the most you can..

lib/luahid.c
```diff
@@ -146,25 +137,64 @@ static inline void luahid_pushinfo(lua_State *L, int idx, struct hid_device *hde
...
+		struct {
+			__u8 *rdesc;
+			unsigned int rsize;
+		} report;
+		struct {
+			struct hid_report *report;
+			u8 *data;
+			int size;
+			int ret_bool;
+		} raw;
```
lneto: Can't you return this bool using the Lua stack as well?

lib/luahid.c
```diff
@@ -146,25 +137,64 @@ static inline void luahid_pushinfo(lua_State *L, int idx, struct hid_device *hde
...
+	luahid_setfield(L, -1, report, application);
+	luahid_setfield(L, -1, report, maxfield);
+}
+
 #define luahid_checkdriver(L, hid, idx, field) (lunatik_getregistry(L, hid) != LUA_TTABLE || \
 	lua_getfield(L, idx, "ops") != LUA_TTABLE || lua_getfield(L, idx - 1, field) != LUA_TTABLE)
 
-static int luahid_doprobe(lua_State *L, luahid_t *hid, struct hid_device *hdev, const struct hid_device_id *id)
+typedef struct {
+	luahid_t *hid;
```
lneto: Instead of having this complex structure, I would just pass 2 lightuserdata, 1 is hdev and the other is the specifics args.. e.g., id or a struct with the buffer and size..

lib/luahid.c
qrsikno2: i think it's yes, if you don't accept the `hdev` param. it will cause compilation error. since the kernel calls in that way.

lneto: My point is that that if it's accepted or not, but if it is present on every callback prototype, if so I think you should just implement it as indicated. (After you finish the working version as we discussed on matrix)

lib/luahid.c
qrsikno2: fixed

lib/luahid.c
qrsikno2: how can i retrieve hid from driver? i don't know and im confused about that...

hid's the key of the "main table"(which stores devices' info) for each instance of driver.

qrsikno2: do you mean i could get it by hdev?

qrsikno2: im sorry, i think i get your idea.

qrsikno2: okaayy! i have fix it.. just add a new param for hdev and i add an inline for getting hid.

lib/luahid.c
qrsikno2: fixed as what i described above.

PR #314, opened 2025-08-04
Ifdef comp kernelverion

qrsikno2:
This pull request updates the `luacrypto_comp.c` file to ensure compatibility with different Linux kernel versions. The changes introduce conditional compilation to handle the deprecation of the `crypto_comp` API in Linux kernel version 6.15 and later.

### Compatibility updates for Linux kernel versions:

* Added a version check using `LINUX_VERSION_CODE` to conditionally include the `crypto_comp` module only for kernels below version 6.15. 
* Introduced a preprocessor warning to notify developers that the `crypto_comp` API is removed in Linux kernel versions 6.15 and above.

### Code maintenance:

* Included `<linux/version.h>` to access the `LINUX_VERSION_CODE` macro for kernel version checks.

### DIFF
--- a/lib/luacrypto_comp.c
+++ b/lib/luacrypto_comp.c
@@ -21,6 +21,7 @@
 #include <linux/err.h>
 #include <linux/slab.h>
 #include <linux/limits.h>
+#include <linux/version.h>
 
 #include <lua.h>
 #include <lualib.h>
@@ -29,6 +30,9 @@
 
 #include "luacrypto.h"
 
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0))
+#warning "crypto_comp API was removed in Linux 6.15, skip COMP module"
+#else
 LUNATIK_PRIVATECHECKER(luacrypto_comp_check, struct crypto_comp *);
 
 LUACRYPTO_RELEASER(comp, struct crypto_comp, crypto_free_comp, NULL);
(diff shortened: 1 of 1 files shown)


jperon:
That would be OK, but with the work on #311, it’s not very useful; moreover, it would induce merge conflicts when acompress gets done.

lib/luacrypto_comp.c
```diff
@@ -29,6 +30,10 @@
 
 #include "luacrypto.h"
 
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0))
+#warning "crypto_comp API removed in Linux ≥6.15, skip COMP module"
+#else
+
```
lneto:
```suggestion
```

lib/luacrypto_comp.c
```diff
@@ -29,6 +30,10 @@
 
 #include "luacrypto.h"
 
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0))
+#warning "crypto_comp API removed in Linux ≥6.15, skip COMP module"
```
lneto:
```suggestion
#warning "crypto_comp API was removed in Linux 6.15, skip COMP module"
```

lneto:
> That would be OK, but with the work on #311, it’s not very useful; moreover, it would induce merge conflicts when acompress gets done.

it's just a temporary fix.. you can remove it completely on acompress branch.. if it's alright to you, I would merge it..

PR #315, opened 2025-08-07
add luadata_clear for netfilter & xtable

qrsikno2:
This pull request introduces a small but important change to the `luanetfilter_hook_cb` function in `lib/luanetfilter.c`. The update ensures that the `data` structure is cleared before further processing, which can help prevent issues caused by stale or leftover data.

* Added a call to `luadata_clear(data);` at the beginning of the function to ensure the `data` structure is reset before use.

### DIFF
--- a/lib/luanetfilter.c
+++ b/lib/luanetfilter.c
@@ -83,6 +83,7 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
 
 	if (lua_pcall(L, 1, 2, 0) != LUA_OK) {
 		pr_err("%s\n", lua_tostring(L, -1));
+		luadata_clear(data);
 		return -1;
 	}
 
--- a/lib/luaxtable.c
+++ b/lib/luaxtable.c
@@ -124,8 +124,13 @@ static int luaxtable_pushparams(lua_State *L, const struct xt_action_param *par,
 	return 0;
 }
 
+static inline int luaxtable_cleanup(luaxtable_t *xtable)
+{
+	return luadata_clear(xtable->skb);
+}
+
 #define luaxtable_call(L, op, xtable, skb, par, info, opt)	\
-	((luaxtable_pushparams(L, par, xtable, skb, opt) == -1) || (luaxtable_docall(L, xtable, info, op, 2, 1) == -1))
+	((luaxtable_pushparams(L, par, xtable, skb, opt) == -1) || (luaxtable_docall(L, xtable, info, op, 2, 1) == -1) || luaxtable_cleanup(xtable))
 
 static int luaxtable_domatch(lua_State *L, luaxtable_t *xtable, const struct sk_buff *skb, struct xt_action_param *par, int fallback)
 {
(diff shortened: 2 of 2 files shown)


lib/luanetfilter.c
```diff
@@ -86,6 +86,8 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
 		return -1;
```
lneto, line 87: it's also necessary to clear on this case

lib/luaxtable.c
```diff
@@ -135,6 +135,7 @@ static int luaxtable_domatch(lua_State *L, luaxtable_t *xtable, const struct sk_
 	int ret = lua_toboolean(L, -1);
 	lua_getfield(L, -2, "hotdrop");
 	par->hotdrop = lua_toboolean(L, -1);
+	luadata_clear(xtable->skb);
```
lneto: you should clear just after the call.. on either case.. success or error.. perhaps a better approach is to move the cleanup to luaxtable_call..

lib/luaxtable.c
```diff
@@ -144,6 +145,7 @@ static int luaxtable_dotarget(lua_State *L, luaxtable_t *xtable, struct sk_buff
 		return fallback;
 
 	int ret = lua_tointeger(L, -1);
+	luadata_clear(xtable->skb);
```
lneto: the same applies here

lib/luaxtable.c
qrsikno2: fixed. but there has some typecast in original commit. so i have to add an extra function to typecast it back, otherwise the `luadata_clear` wouldn't accept it.

lib/luaxtable.c
qrsikno2: fixed

lib/luanetfilter.c
qrsikno2: fixed

lib/luaxtable.c
lneto: why do you need a function? just put a typecast inline..

lib/luaxtable.c
```diff
@@ -124,8 +124,13 @@ static int luaxtable_pushparams(lua_State *L, const struct xt_action_param *par,
...
 }
 
+static inline int luaxtable_cleanup(luaxtable_t *xtable)
+{
+	return luadata_clear(xtable->skb);
+}
+
 #define luaxtable_call(L, op, xtable, skb, par, info, opt)	\
-	((luaxtable_pushparams(L, par, xtable, skb, opt) == -1) || (luaxtable_docall(L, xtable, info, op, 2, 1) == -1))
+	((luaxtable_pushparams(L, par, xtable, skb, opt) == -1) || (luaxtable_docall(L, xtable, info, op, 2, 1) == -1) || luaxtable_cleanup(xtable))
```
lneto, line 133: it will only be called in case of success, right? as I mentioned, both cases should be handled

lneto: IMHO, the best approach is to move the luadata_clear() call to luaxtable_docall() body..

lneto:
https://github.com/luainkernel/lunatik/pull/401

PR #318, opened 2025-11-17
Add luacpu module

endersonmaia:
This is my first attempt to collaborate with lunatik in an topic I have interest that's metrics collection.

I'm starting with a simple  `luacpu`, so I can get some feedback.


lib/lualinux.c
```diff
@@ -255,6 +256,94 @@ static int lualinux_ifindex(lua_State *L)
...
+*/
+static int lualinux_cpustat(lua_State *L)
+{
+	int nproc = num_online_cpus();
+	int cpu = luaL_checkinteger(L, 1);
+	struct kernel_cpustat kcs;
+	int table;
+
+	if (cpu < 0 || cpu >= nproc) {
+		return luaL_error(L, "invalid CPU number: %d (valid range: 0-%d)",
```
lneto: we ca use argchecker here

examples/cpustat.lua
```diff
@@ -0,0 +1,29 @@
...
+	print(string.format("  nice:       %d", stats.nice))
+	print(string.format("  system:     %d", stats.system))
+	print(string.format("  idle:       %d", stats.idle))
+	print(string.format("  iowait:     %d", stats.iowait))
+	print(string.format("  irq:        %d", stats.irq))
+	print(string.format("  softirq:    %d", stats.softirq))
+	print(string.format("  steal:      %d", stats.steal))
+	print(string.format("  guest:      %d", stats.guest))
+	print(string.format("  guest_nice: %d", stats.guest_nice))
+	print("")
```
lneto: I think it would be cleaner to have a multiline string and call format and print only once.. e.g.,
```
local fmt = [[
CPU %d statistics:
  user:       %d
  nice:       %d
  ...
]]
...
for i = 0, num_cpus - 1 do
	local stats = linux.cpustat(i)
	print(string.format(fmt, i, stats.user, stats.nice, ...)
end
```

examples/procstat.lua
```diff
@@ -0,0 +1,29 @@
...
+-- Simple example demonstrating linux.cpustat() usage
+
+local linux = require("linux")
+
+-- Get number of CPUs
+local num_cpus = linux.nproc()
+print(string.format("System has %d online CPUs\n", num_cpus))
+
+-- Get stats for each CPU
+for i = 0, num_cpus - 1 do
```
lneto: I'm not sure it's the right way to do it.. I think we can have gaps.. for instance, have 3 CPU online, out of 4 existent.. I would take a look at foreach_online_cpu macro in the kernel source to make sure..

lib/lualinux.c
```diff
@@ -255,6 +256,91 @@ static int lualinux_ifindex(lua_State *L)
...
+
+	luaL_argcheck(L, cpu >= 0 && cpu < nproc, 1, "invalid CPU number");
+
+	kcpustat_cpu_fetch(&kcs, cpu);
+
+	lua_createtable(L, 0, 10);
+	table = lua_gettop(L);
+
+	lua_pushinteger(L, (lua_Integer)kcs.cpustat[CPUTIME_USER]);
+	lua_setfield(L, table, "user");
```
lneto: it seems a very repetitive pattern.. I would create a macro for this.. e.g., `lualinux_setcpustat(L, -1, kcs, name, NAME)`

```
#define lualinux_setcpustat(L, idx, kcs, name, NAME)	\
do {								\
	lua_pushinteger(L, (lua_Integer)kcs.cpustat[CPUTIME_##NAME]);	\
	lua_setfield(L, idx - 1, #name);	\
} while (0)
```

lib/lualinux.c
```diff
@@ -255,6 +256,91 @@ static int lualinux_ifindex(lua_State *L)
...
+static int lualinux_cpustat(lua_State *L)
+{
+	int nproc = num_online_cpus();
+	int cpu = luaL_checkinteger(L, 1);
+	struct kernel_cpustat kcs;
+	int table;
+
+	luaL_argcheck(L, cpu >= 0 && cpu < nproc, 1, "invalid CPU number");
+
+	kcpustat_cpu_fetch(&kcs, cpu);
```
lneto: what happens if cpu is invalid here? is it safe? doesn't this function return anything?

lib/lualinux.c
```diff
@@ -306,37 +306,23 @@ static int lualinux_cpustat(lua_State *L)
...
-	lua_pushinteger(L, (lua_Integer)kcs.cpustat[CPUTIME_GUEST]);
-	lua_setfield(L, table, "guest");
-
-	lua_pushinteger(L, (lua_Integer)kcs.cpustat[CPUTIME_GUEST_NICE]);
-	lua_setfield(L, table, "guest_nice");
+#define lualinux_setcpustat(L, idx, kcs, name, NAME)			\
+do {									\
+	lua_pushinteger(L, (lua_Integer)kcs.cpustat[CPUTIME_##NAME]);	\
+	lua_setfield(L, idx - 1, #name);				\
+} while (0)
```
lneto: we usually do this outside of the function declaration..

lib/lualinux.c
```diff
@@ -255,6 +256,77 @@ static int lualinux_ifindex(lua_State *L)
...
+*   local cpu0 = linux.cpustat(0)
+*   print("CPU 0 user time:", cpu0.user)
+*   print("CPU 0 idle time:", cpu0.idle)
+*/
+static int lualinux_cpustat(lua_State *L)
+{
+	int nproc = num_online_cpus();
+	int cpu = luaL_checkinteger(L, 1);
+	struct kernel_cpustat kcs;
+	int table;
```
lneto: not used anymore

lib/lualinux.c
```diff
@@ -255,6 +256,77 @@ static int lualinux_ifindex(lua_State *L)
...
+* @raise Error if CPU number is invalid
+* @usage
+*   local cpu0 = linux.cpustat(0)
+*   print("CPU 0 user time:", cpu0.user)
+*   print("CPU 0 idle time:", cpu0.idle)
+*/
+static int lualinux_cpustat(lua_State *L)
+{
+	int nproc = num_online_cpus();
+	int cpu = luaL_checkinteger(L, 1);
```
lneto: I would invert this two lines.. just a small detail, but we don't need to call `num_online_cpus()` in the case cpu is not an integer

lib/lualinux.c
endersonmaia: Does it need to be on top of the file, or just  before the function definition is enough ?

lneto: just before the func is good enough..

examples/cpustat.lua
```diff
@@ -0,0 +1,45 @@
...
+		stats.nice,
+		stats.system,
+		stats.idle,
+		stats.iowait,
+		stats.irq,
+		stats.softirq,
+		stats.steal,
+		stats.guest,
+		stats.guest_nice
+	))
```
lneto: I think we don't need to break this in too many lines as this, right?
```suggestion
	print(string.format(fmt, i, stats.user, stats.nice, stats.system, stats.idle,
		stats.iowait, stats.irq, stats.softirq, stats.steal, stats.guest, stats.guest_nice))
```

examples/procstat.lua
lneto: I believe this is the right way to do it.. https://elixir.bootlin.com/linux/v6.1.80/source/drivers/cpufreq/cpufreq_ondemand.c#L428

you can use this pattern to create a lualinux_foreach_cpu.. that would end up calling a function for each online cpu.. e.g.,
```
lualinux.foreach_online_cpu(function (cpu)
	local stats = linux.cpustat(cpu)

	print(string.format(fmt, cpu, stats.user, stats.nice, stats.system, stats.idle,
		stats.iowait, stats.irq, stats.softirq, stats.steal, stats.guest, stats.guest_nice))
end
```

https://github.com/luainkernel/lunatik/blob/master/lib/luarcu.c#L335

I think I would make it more generic though, having a template and export the other sibling functions.. e.g., `present`, `possible`.

btw, as we grow this binding complexity.. I'm wondering if it wouldn't be better to separate this into a luacpumask binding.. so we could write..

```
local cpu = require("cpumask")

cpumask.foreach_online(function (cpu)
	local stats = cpumask.stats(cpu)

	print(string.format(fmt, cpu, stats.user, stats.nice, stats.system, stats.idle,
		stats.iowait, stats.irq, stats.softirq, stats.steal, stats.guest, stats.guest_nice))
end
```

perhaps it would be cleaner.. what do you think?

endersonmaia: that escalated quickly

I agree it should be separate binding, will work on this later

I may need help with other stuff I didn't have to worry while reusing this existing binding, let's see how it goes

examples/cpustat.lua
```diff
@@ -0,0 +1,34 @@
...
+  nice:       %d
+  system:     %d
+  idle:       %d
+  iowait:     %d
+  irq:        %d
+  softirq:    %d
+  steal:      %d
+  guest:      %d
+  guest_nice: %d
+
```
lneto: this will add two newlines.. is it intentional?

lib/lualinux.c
```diff
@@ -255,6 +256,84 @@ static int lualinux_ifindex(lua_State *L)
...
+*   print("CPU 0 user time:", cpu0.user)
+*   print("CPU 0 idle time:", cpu0.idle)
+*/
+static int lualinux_cpustat(lua_State *L)
+{
+	int cpu = luaL_checkinteger(L, 1);
+	int nproc = num_online_cpus();
+	struct kernel_cpustat kcs;
+
+	luaL_argcheck(L, cpu >= 0 && cpu < nproc, 1, "invalid CPU number");
```
lneto: are you sure you only can retrieve status of online CPUs? what happens with present, for instance? also, keep the gaps in mind here.. what happens in case of having CPU that is in that range but isn't online? wouldn't be better to just fail safe over kcpustat_cpu_fetch call?

examples/procstat.lua
endersonmaia: I'll send a fixup where I already introduce a new module called `luaprocstat`

Not sure about the name yet, but was an attempt to narrow the scope, we can discuss it later.

I'm just in doubt if I create a new PR, or keep with this one.

And if keeping this one, if I should squash so we can start a ne round of reviews, 'cause I think it's starting to get messy.

You tell me.

lneto: > I'll send a fixup where I already introduce a new module called luaprocstat
>
> Not sure about the name yet, but was an attempt to narrow the scope, we can discuss it later.

I don't find this a good name; IMO, we shouldn't tie the name to this particular application. This binding should grow and handle other cpu/cpumask APIs. Thus, I would call this `luacpumask` or `luacpu`.

lneto: > I'm just in doubt if I create a new PR, or keep with this one.
>
> And if keeping this one, if I should squash so we can start a ne round of reviews, 'cause I think it's starting to get messy.
>
> You tell me.

It's up to you.. I don't mind both ways =)

lib/luaprocstat.c
```diff
@@ -0,0 +1,124 @@
...
+*   print("CPU 0 user time:", cpu0.user)
+*   print("CPU 0 idle time:", cpu0.idle)
+*/
+static int luaprocstat_stat(lua_State *L)
+{
+	int cpu = luaL_checkinteger(L, 1);
+	int nproc = num_online_cpus();
+	struct kernel_cpustat kcs;
+
+	luaL_argcheck(L, cpu >= 0 && cpu < nproc, 1, "invalid CPU number");
```
lneto: please, mind my original comment here.

lib/luaprocstat.c
```diff
@@ -0,0 +1,124 @@
...
+
+	kcpustat_cpu_fetch(&kcs, cpu);
+
+	lua_createtable(L, 0, 10);
```
lneto: you can use the array size here instead of a magic number, right?

lib/luaprocstat.c
```diff
@@ -0,0 +1,124 @@
...
+
+static void __exit luaprocstat_exit(void)
+{
+}
+
+module_init(luaprocstat_init);
+module_exit(luaprocstat_exit);
+MODULE_LICENSE("Dual MIT/GPL");
+MODULE_AUTHOR("Enderson Maia <endersonmaia@gmail.com");
+MODULE_DESCRIPTION("Lunatik CPU statistics interface");
```
lneto: I would not restrict this module to cpustat

lneto: \n

lneto: btw, `foreach_*cpu` family API is a good example of other interfaces that should live in this module and can be use to way more things than just cpustat

lib/lualinux.c
endersonmaia: I tried calling kcpstat_cpu_fetch on an offline cpu and got a kernel panic.

So i put a check if the cpu is online.

lib/luacpu.c
```diff
@@ -0,0 +1,182 @@
...
+*   local cpu0 = cpu.stats(0)
+*   print("CPU 0 user time:", cpu0.user)
+*   print("CPU 0 idle time:", cpu0.idle)
+*/
+static int luacpu_stats(lua_State *L)
+{
+	unsigned int cpu = luaL_checkinteger(L, 1);
+	struct kernel_cpustat kcs;
+
+	luaL_argcheck(L, cpu_online(cpu), 1, "CPU is offline");
```
lneto, line 99: I was about to suggest that.. you were faster ;-)

bin/lunatik
```diff
@@ -10,7 +10,7 @@ local lunatik = {
 	modules = {"lunatik", "luadevice", "lualinux", "luanotifier", "luasocket", "luarcu",
 		"luathread", "luafib", "luadata", "luaprobe", "luasyscall", "luaxdp", "luafifo", "luaxtable",
 		"luanetfilter", "luacompletion", "luacrypto_shash", "luacrypto_skcipher", "luacrypto_aead",
-		"luacrypto_rng", "luacrypto_comp", "lunatik_run"}
+		"luacrypto_rng", "luacrypto_comp", "lunatik_run", "luacpu"},
```
lneto: we should keep `lunatik_run` on the tail

lib/luacpu.c
```diff
@@ -0,0 +1,182 @@
...
+	unsigned int cpu;				\
+							\
+	luaL_checktype(L, 1, LUA_TFUNCTION);		\
+							\
+	for_each_##name##_cpu(cpu) {			\
+		lua_pushvalue(L, 1);			\
+		lua_pushinteger(L, cpu);		\
+		lua_call(L, 1, 0);			\
+	}						\
+							\
```
lneto:
```suggestion
	luaL_checktype(L, 1, LUA_TFUNCTION);		\
	for_each_##name##_cpu(cpu) {			\
		lua_pushvalue(L, 1);			\
		lua_pushinteger(L, cpu);		\
		lua_call(L, 1, 0);			\
	}						\
```

lib/luacpu.c
```diff
@@ -0,0 +1,182 @@
...
+	luacpu_setstat(L, -1, kcs, iowait, IOWAIT);
+	luacpu_setstat(L, -1, kcs, irq, IRQ);
+	luacpu_setstat(L, -1, kcs, softirq, SOFTIRQ);
+	luacpu_setstat(L, -1, kcs, steal, STEAL);
+	luacpu_setstat(L, -1, kcs, guest, GUEST);
+	luacpu_setstat(L, -1, kcs, guest_nice, GUEST_NICE);
+#ifdef CONFIG_SCHED_CORE
+	luacpu_setstat(L, -1, kcs, forceidle, FORCEIDLE);
+#endif
+
```
lneto:
```suggestion
```

lib/luacpu.c
```diff
@@ -0,0 +1,182 @@
...
+/***
+* Gets the number of available processing units (CPUs).
+* @function nproc
+* @treturn integer The number of CPUs available in the system.
+* @usage
+*   local num_cpus = cpustat.nproc()
+*   print("Number of CPUs:", num_cpus)
+*/
+static int luacpu_nproc(lua_State *L)
```
lneto: I think we should use the same template approach of `foreach` here and add the sibling APIs as well.. also, I would use a more descriptive name for these functions.. e.g., `num_online`, `num_present`, etc

examples/cpu.lua
```diff
@@ -0,0 +1,33 @@
...
+  guest_nice: %d
+]]
+
+-- Get stats for each online CPU
+cpu.foreach_online(function(cpu_id)
+	local s = cpu.stats(cpu_id)
+
+	print(string.format(fmt, cpu_id, s.user, s.nice, s.system, s.idle,
+		s.iowait, s.irq, s.softirq, s.steal, s.guest, s.guest_nice))
+end)
```
lneto: \n

examples/cpu.lua
```diff
@@ -0,0 +1,33 @@
+--
```
lneto: I'm not sure we should add an example as simple as this.. perhaps if you add a chardev it could look more alike our other examples.. also, notice that it's not mandatory to have such example.. it's up to you, but if we gonna add an example, it should be more useful, IMO.. also, it would be good to implement something that it's not available on stock kernel.. perhaps I would leave the example for a new PR, having a simple loop pushing this stats to influx.. but I'm not against on turning this into `/dev/cpustat`.. again, up to you ;-)

lneto: also, you should find a better name for the script.. `cpu.lua` is way too generic for an example.. also, if we gonna have a new example, it should be included on README

endersonmaia: gonna remove the example for now, may reintroduce when I have something more useful

PR #321, opened 2025-11-23
add support for AF_UNIX and AF_LOCAL on luasocket

lneto:


lneto:
@endersonmaia we could use this on the cpustats example.. e.g., retrieving cpustats using a unix domain socket..

PR #324, opened 2025-11-24
Add DNS rewrite example with packet expansion support

ArifAlam:
Introduce a new DNS rewrite example that demonstrates kernel-level DNS interception and response generation using Lunatik's netfilter hooks. The example intercepts DNS queries for 'test.internal' and responds directly with 127.0.0.1, preventing queries from reaching external DNS servers.

Changes include:

- Add dnsrewrite example with netfilter hook implementation that intercepts outgoing and incoming DNS packets at LOCAL_OUT and PRE_ROUTING points

- Implement DNS packet parsing, response building, and checksum recalculation

- Extend luadata API with expand() method to support dynamic packet buffer expansion using skb_put(), enabling in-place DNS response construction

### DIFF
--- a/examples/dnsrewrite/Makefile
+++ b/examples/dnsrewrite/Makefile
@@ -1,0 +1,16 @@
+# SPDX-FileCopyrightText: (c) 2025 Ring Zero Desenvolvimento de Software LTDA
+# SPDX-License-Identifier: MIT OR GPL-2.0-only
+
+# This is a simple example with only Lua files
+# Installation is handled by the main Makefile's examples_install target
+
+all:
+	@echo "dnsrewrite example - Lua only, no build required"
+
+install:
+	@echo "Use 'make install' from the main lunatik directory"
+
+clean:
+	@echo "Nothing to clean"
+
+.PHONY: all install clean
--- a/examples/dnsrewrite/common.lua
+++ b/examples/dnsrewrite/common.lua
@@ -1,0 +1,220 @@
+local linux = require("linux")
+local string = require("string")
+
+local common = {}
+
+local udp = 0x11
+local dns = 0x35  -- DNS port 53
+local eth_len = 14
+
+-- Target domain: test.internal
+local target_dns = string.pack("s1s1", "test", "internal")
+
+-- Target IP: 127.0.0.1
+local target_ip = 0x7F000001  -- 127.0.0.1 in network byte order (big-endian)
+
+local function get_domain(skb, off)
+	local _, nameoff, name = skb:getstring(off):find("([^\0]*)")
+	return name, nameoff + 1
+end
+
+local function calculate_ip_checksum(skb, ip_off, ip_len)
+	-- Zero out the checksum field first
+	skb:setuint16(ip_off + 10, 0)
+
+	local sum = 0
+	-- Sum all 16-bit words in the IP header
+	for i = 0, ip_len - 1, 2 do
+		sum = sum + skb:getuint16(ip_off + i)
+	end
+
+	-- Fold 32-bit sum to 16 bits
+	while (sum >> 16) > 0 do
+		sum = (sum & 0xFFFF) + (sum >> 16)
+	end
+
+	return ~sum & 0xFFFF
+end
+
+local function calculate_udp_checksum(skb, ip_off, ihl, udp_off, udp_len)
+	-- Get source and destination IP addresses
+	local src_ip = skb:getuint32(ip_off + 12)
+	local dst_ip = skb:getuint32(ip_off + 16)
+
+	-- Zero out the UDP checksum field first
+	skb:setuint16(udp_off + 6, 0)
+
+	local sum = 0
+
+	-- Add pseudo-header
+	sum = sum + ((src_ip >> 16) & 0xFFFF)
+	sum = sum + (src_ip & 0xFFFF)
+	sum = sum + ((dst_ip >> 16) & 0xFFFF)
+	sum = sum + (dst_ip & 0xFFFF)
+	sum = sum + udp  -- Protocol
+	sum = sum + udp_len  -- UDP length
+
+	-- Add UDP header and data
+	for i = 0, udp_len - 1, 2 do
+		sum = sum + skb:getuint16(udp_off + i)
+	end
+
+	-- Fold 32-bit sum to 16 bits
+	while (sum >> 16) > 0 do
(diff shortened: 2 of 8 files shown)


examples/dnsrewrite/Makefile
```diff
@@ -0,0 +1,16 @@
+# SPDX-FileCopyrightText: (c) 2025 Ring Zero Desenvolvimento de Software LTDA
```
lneto, line 1: I don't think we need this makefile.. we only have it on the other dns filter examples for using iptables (it requires a userspace plugin)..  if we this only uses netfilter, we can avoid this file..

examples/dnsrewrite/common.lua
```diff
@@ -0,0 +1,220 @@
+local linux = require("linux")
+local string = require("string")
```
lneto, line 2: we don't need to require Lua standard libs.. it's safe to just use `string.*`

examples/dnsrewrite/common.lua
```diff
@@ -0,0 +1,220 @@
+local linux = require("linux")
```
lneto, line 1: you should add your copyright and license on top =)

examples/dnsrewrite/nf_dnsrewrite.lua
```diff
@@ -0,0 +1,26 @@
+local nf = require("netfilter")
+local common = require("examples.dnsrewrite.common")
```
lneto, line 2: I don't think it's necessary to split this module in two parts.. common and nf_*.. we used this pattern only for the ones we have support for both iptables and netfilter.. I would keep it in just one single file..

lib/luadata.c
```diff
@@ -35,6 +37,7 @@ typedef struct luadata_s {
 	char *ptr;
 	size_t size;
 	uint8_t opt;
+	struct sk_buff *skb;  /* optional sk_buff pointer for packet expansion */
```
lneto, line 40: I think a better approach is to introduce a new option.. e.g., LUADATA_OPT_SKB.. then, we can use container_of to get this struct, right?

lib/luadata.c
```diff
@@ -410,6 +452,17 @@ int luadata_reset(lunatik_object_t *object, void *ptr, size_t size, uint8_t opt)
...
+void luadata_set_skb(lunatik_object_t *object, struct sk_buff *skb)
+{
+	luadata_t *data;
+
+	lunatik_lock(object);
+	data = (luadata_t *)object->private;
+	data->skb = skb;
+	lunatik_unlock(object);
+}
+EXPORT_SYMBOL(luadata_set_skb);
```
lneto, line 464: we could skip this if using a new option, we would just need to set the flag

lib/luadata.c
```diff
@@ -344,6 +384,7 @@ static const luaL_Reg luadata_mt[] = {
 #endif
 	{"getstring", luadata_getstring},
 	{"setstring", luadata_setstring},
+	{"expand", luadata_expand},
```
lneto, line 387: I would provide a more generic API, e.g. `resize`, allowing both shrink and growth, both skb and raw buffers, what do you think? If you think it's better to have only `expand`, I would at least support both skb and raw buffer, making it more ergonomic..

examples/dnsrewrite/common.lua
```diff
@@ -0,0 +1,220 @@
...
+end
+
+function common.hook(skb, action)
+	-- Get protocol from IP header (offset 9)
+	local proto = skb:getuint8(eth_len + 9)
+
+	-- Only process UDP packets
+	if proto ~= udp then
+		return action.ACCEPT
+	end
```
lneto, line 83: it's totally alright to do it at this level, but we can also leverage the nft integration, so you can match the UDP protocol on nft and mark only the packets you want to call this hook.. you just need to add `mark = X` on your hook.. it usually makes the lua portion cleaner and the overall flow more efficient =), as we are not calling the script for every packet..

examples/dnsrewrite/common.lua
```diff
@@ -0,0 +1,220 @@
...
+
+	-- Get IP header length (IHL field in first byte of IP header)
+	local ihl = skb:getuint8(eth_len) & 0x0F
+	local thoff = eth_len + ihl * 4  -- Transport header offset
+
+	-- Check if destination port is DNS (53) - this means it's a DNS query
+	local dstport = linux.ntoh16(skb:getuint16(thoff + 2))
+	if dstport ~= dns then
```
lneto, line 91: we can decide this at nft level as well.. (not mandatory though)

examples/dnsrewrite/common.lua
```diff
@@ -0,0 +1,220 @@
...
+		local udp_len = 8 + dns_len
+		local ip_len = ihl * 4 + udp_len
+
+		-- Update UDP length
+		skb:setuint16(thoff + 4, linux.hton16(udp_len))
+
+		-- Update IP total length
+		skb:setuint16(eth_len + 2, linux.hton16(ip_len))
+
+		-- Recalculate UDP checksum
```
lneto, line 204: we would need to check, but I have the impression that when we change the SKB at INET level, netfilter will recalc the checksums by itself.. but now sure.. (sorry if memory is not that reliable ;-)

lneto: @jperon I think you have played with it recently, right?

lib/luadata.h
```diff
@@ -17,8 +17,11 @@ LUNATIK_LIB(data);
 
 #define luadata_clear(o)	(luadata_reset((o), NULL, 0, LUADATA_OPT_KEEP))
 
+struct sk_buff;
```
lneto, line 20: it's ok for us to include the skb header in luadata.h..

examples/dnsrewrite/common.lua
sheharyaar: I too think handling these at nft level will be better, it will skip the lua C API calls.

examples/dnsrewrite/common.lua
sheharyaar: I think we should do it in the script itself, some configs can silently drop the packets if the checksum is not correct unless checksum offload is enabled. But since you mentioned changing skb at the inet level, I am also unsure if netfilter will recalc the checksum.

sheharyaar:
Did one overview review, haven't tested the patch yet.

PR #327, opened 2025-11-26
add cpuexporter example

endersonmaia:
Basing on top of lneto_unixsock branch since I need it for longer filename.

I'm still planning to test it a little more comparing with other cpu collectors like telegraf and fluent-bit to make sure my measurements are correct.

I'm asking for review anyway since I need feedback on code quality and style, since I'm not Lua fluent yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,11 +469,30 @@ sudo make examples_install                  # installs examples
 ip link add veth0 type veth peer name veth1
 ip link set veth0 up
 ip link set veth1 up
+```
 
 sudo lunatik spawn examples/lldpd           # runs lldpd
 
 # verify LLDP frames are being transmitted
 sudo tcpdump -i veth0 -e ether proto 0x88cc -vv
+
+### cpuexporter
+
+[cpuexporter](examples/cpuexporter.lua) will gather CPU usage statistics and expose using [OpenMetrics text format](https://github.com/prometheus/OpenMetrics/blob/main/specification/OpenMetrics.md#text-format) at a UNIX socket file.
+
+#### Usage
+
+```shell
+sudo make examples_install         	# installs examples
+sudo lunatik spawn examples/cpuexporter # runs cpuexporter
+sudo socat - UNIX-CONNECT:/tmp/cpuexporter.sock <<<""
+# TYPE cpu_usage_system gauge
+cpu_usage_system{cpu="cpu1"} 0.0000000000000000 1764094519529162
+cpu_usage_system{cpu="cpu0"} 0.0000000000000000 1764094519529162
+# TYPE cpu_usage_idle gauge
+cpu_usage_idle{cpu="cpu1"} 100.0000000000000000 1764094519529162
+cpu_usage_idle{cpu="cpu0"} 100.0000000000000000 1764094519529162
+...
 ```
 
 ## References
--- a/examples/cpuexporter.lua
+++ b/examples/cpuexporter.lua
@@ -1,0 +1,170 @@
+--
+-- SPDX-FileCopyrightText: (c) 2025-2026 Enderson Maia <endersonmaia@gmail.com>
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+local lunatik = require("lunatik")
+local thread  = require("thread")
+local unix    = require("socket.unix")
+local linux   = require("linux")
+local cpu     = require("cpu")
+
+local shouldstop = thread.shouldstop
+
+local server = unix.bind("/tmp/cpuexporter.sock", "STREAM")
+server:listen()
+
+local last_stats = {}
+local last_total_stats = {}
+
+-- Scale integer to decimal with high precision
+-- Converts a ratio (metric/total) to percentage format with 16 decimal places
+local function format_percentage(metric, total)
+	if total == 0 then
+		return "0.0000000000000000"
+	end
+
+	local int_part = (metric * 100) // total
+	local remainder = (metric * 100) % total
+	local frac_high = (remainder * 100000000) // total
+	local remainder2 = (remainder * 100000000) % total
+	local frac_low = (remainder2 * 100000000) // total
+
+	return string.format("%d.%08d%08d", int_part, frac_high, frac_low)
+end
+
+-- Helper function to sum all stats values in a table
+-- Note: guest and guest_nice are already included in user and nice respectively
+-- according to the Linux kernel documentation, so we must exclude them from the total
+local function sum_stats(stats)
+	local sum = 0
+	stats["guest"] = 0
+	stats["guest_nice"] = 0
+	for key, value in pairs(stats) do
+		sum = sum + value
+	end
+	return sum
+end
+
+-- returns per cpu stats, total_stats tables
+local function cpu_stats()
+	--TODO: add cpu-total with accumulated values for all cpus
+	local stats = {}
+	local total_stats = {}
+	cpu.foreach_online(function(id)
+		stats[id] = {}
+		stats[id] = cpu.stats(id)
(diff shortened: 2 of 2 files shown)


examples/cpuexporter.lua
```diff
@@ -0,0 +1,164 @@
...
+
+--TODO: add cpu-total with accumulated values for all cpus
+	cpu.foreach_online(function(cpu_id)
+
+		-- First sample
+		if last_stats[cpu_id] == nil then
+			last_stats[cpu_id] = {}
+			last_stats[cpu_id] = cpu.stats(cpu_id)
+			last_stats[cpu_id]._total = _sum_table(last_stats[cpu_id])
+			linux.schedule(1000) --FIXME: wait 1s for each CPU is bad, even if it's just for the first run
```
lneto: btw, it doesn't guarantee 1s, it's "at least".. this API tells the scheduler that your process it's ok to wait at least that to run again..

lneto: btw, isn't that possible on that first sample to just return an error (e.g., not ready or something) to prometheus?

endersonmaia: I could rework the code so that I collect a sample at start, no matter if it was triggered by the connection

but I sent some fixups that already fix the issue to "sleep" for each cpu, it looks better now, IMO

README.md
```diff
@@ -545,6 +545,26 @@ sudo lunatik unload
...
+# TYPE cpu_usage_system gauge
+cpu_usage_system{cpu="cpu1"} 0.0000000000000000 1764094519529162
+cpu_usage_system{cpu="cpu0"} 0.0000000000000000 1764094519529162
+# TYPE cpu_usage_idle gauge
+cpu_usage_idle{cpu="cpu1"} 100.0000000000000000 1764094519529162
+cpu_usage_idle{cpu="cpu0"} 100.0000000000000000 1764094519529162
+...
+```
+
+
```
lneto:
```suggestion
```

examples/cpuexporter.lua
```diff
@@ -0,0 +1,173 @@
+--
+-- SPDX-FileCopyrightText: (c) 2025 Enderson Maia <endersonmaia@gmail.com>
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+local lunatik = require("lunatik")
+local thread  = require("thread")
+local socket  = require("socket")
```
lneto: it's not needed to require this module for using the module below

examples/cpuexporter.lua
```diff
@@ -0,0 +1,173 @@
...
+end
+
+-- Helper function to sum all stats values in a table
+-- Note: guest and guest_nice are already included in user and nice respectively
+-- according to the Linux kernel documentation, so we must exclude them from the total
+local function _sum_cpu_stats(stats)
+	local sum = 0
+	for key, value in pairs(stats) do
+		-- Skip guest times as they're already included in user/nice
+		if not (key == "guest" or key == "guest_nice") then
```
lneto: couldn't you just remove them from stats beforehand? e.g., `stats.guest = nil`

examples/cpuexporter.lua
```diff
@@ -0,0 +1,173 @@
...
+	local remainder2 = (remainder * 100000000) % total
+	local frac_low = (remainder2 * 100000000) // total
+
+	return string.format("%d.%08d%08d", int_part, frac_high, frac_low)
+end
+
+-- Helper function to sum all stats values in a table
+-- Note: guest and guest_nice are already included in user and nice respectively
+-- according to the Linux kernel documentation, so we must exclude them from the total
+local function _sum_cpu_stats(stats)
```
lneto: why prefixing with `_`? it seems a bit odd..

examples/cpuexporter.lua
```diff
@@ -0,0 +1,173 @@
...
+end
+
+-- returns cpu_stats table
+local function cpu_stats()
+	--TODO: add cpu-total with accumulated values for all cpus
+	local cpu_stats = {}
+	cpu.foreach_online(function(cpu_id)
+		cpu_stats[cpu_id] = {}
+		cpu_stats[cpu_id] = cpu.stats(cpu_id)
+		cpu_stats[cpu_id]._total = _sum_cpu_stats(cpu_stats[cpu_id])
```
lneto: I think it would be clearer to have a `total_stats` table in the same way you have current, than adding a new field here..

examples/cpuexporter.lua
```diff
@@ -0,0 +1,173 @@
...
+
+	for cpu_id, _ in pairs(current_stats) do
+		local total_delta = current_stats[cpu_id]._total - (last_stats[cpu_id]._total or 0)
+		local guest_delta = current_stats[cpu_id].guest - (last_stats[cpu_id].guest or 0)
+		local guest_nice_delta = current_stats[cpu_id].guest_nice - (last_stats[cpu_id].guest_nice or 0)
+
+		usage[cpu_id] = {}
+		for metric, value in pairs(current_stats[cpu_id]) do
+			-- Skip _total metric
+			if metric ~= "_total" then
```
lneto: having a separate table for total, you could get rid of this verification

examples/cpuexporter.lua
```diff
@@ -0,0 +1,173 @@
...
+local sock = socket.sock
+
+local server = unix.bind("/tmp/cpuexporter.sock", "STREAM")
+server:listen()
+
+local last_stats = {}
+
+-- Scale integer to decimal with high precision
+-- Converts a ratio (value/total) to percentage format with 16 decimal places
+local function to_percent_float(value, total)
```
lneto: I think this name is a bit misnomer.. I would call this something like `format_percentage` or something.. you're not converting to float, but formatting it..

examples/cpuexporter.lua
```diff
@@ -0,0 +1,173 @@
...
+	if total == 0 then
+		return "0.0000000000000000"
+	end
+
+	local int_part = (value * 100) // total
+	local remainder = (value * 100) % total
+	local frac_high = (remainder * 100000000) // total
+	local remainder2 = (remainder * 100000000) % total
```
lneto, line 30: `remainder_{high, low}`?

examples/cpuexporter.lua
lneto: s/value/metric/?

examples/cpuexporter.lua
```diff
@@ -0,0 +1,173 @@
...
+	local frac_low = (remainder2 * 100000000) // total
```
lneto, line 31: btw, it's not clear to me why we need to do this twice.. I would move the common pattern to a function anyway..

examples/cpuexporter.lua
endersonmaia: personal convetion :sweat_smile: 

I'll remove it.

examples/cpuexporter.lua
endersonmaia: So I don't overflow 64bit integer and get a 16 decimals precision, I split the operation in two parts (high, low).

lneto: wouldn't you have enough precision using difftime?

lneto:
hey @endersonmaia are you still willing to work on this?

endersonmaia:
> hey @endersonmaia are you still willing to work on this?

Actually, I think this is DONE.

I can work on rebase/squash later today, but from my last tests, the metrics were very consistent with telegraf collected
.

lneto:
> > hey @endersonmaia are you still willing to work on this?
> 
> Actually, I think this is DONE.
> 
> I can work on rebase/squash later today, but from my last tests, the metrics were very consistent with telegraf collected .

I will include this on 4.0 then =); I can rebase/squash it if you don't mind..

PR #330, opened 2025-12-09
Add senddata(luadata) & receivedata

Physic69:
add senddata, receivedata methods that operate directly on luadata buffer memory without Lua string copies.


lib/luasocket.c
```diff
@@ -113,108 +115,223 @@ LUNATIK_PRIVATECHECKER(luasocket_check, struct socket *);
...
-	size_t len;
+	int nargs = lua_gettop(L);
+
 	struct kvec vec;
 	struct msghdr msg;
 	struct sockaddr_storage addr;
-	int nargs = lua_gettop(L);
 	int ret;
-
+
```
lneto:
```suggestion

```

lib/luasocket.c
```diff
@@ -113,108 +115,223 @@ LUNATIK_PRIVATECHECKER(luasocket_check, struct socket *);
...
+			size_t size = luasocket_checkaddr(L, socket, &addr, 3);
+			luasocket_msgaddr(msg, addr, size);
+		}
 
-	lunatik_tryret(L, ret, kernel_sendmsg, socket, &msg, &vec, 1, len);
-	lua_pushinteger(L, ret);
-	return 1;
+		lunatik_tryret(L, ret, kernel_sendmsg, socket, &msg, &vec, 1, len);
+		lua_pushinteger(L, ret);
+		return 1;
```
lneto: this is common code, we can just move it out from the if-then-else blocks

lneto: please mind such repetitions and avoid it.. we try to keep the code base as simple as possible =)

lib/luasocket.c
```diff
@@ -113,106 +115,198 @@ LUNATIK_PRIVATECHECKER(luasocket_check, struct socket *);
...
+    int from;
+    int ret;
+    size_t len = 0;
+    luaL_Buffer B;
+    luadata_t *data = NULL;
+    lua_Integer offset = 0;
+
+    luasocket_setmsg(msg);
+
+    flags = luaL_optinteger(L, t2 != LUA_TUSERDATA && t2 != LUA_TLIGHTUSERDATA ? 3 : 5, 0);
```
Physic69: Should i put these (flags& from)  like this or 
should i move them to if and else

lib/luasocket.c
```diff
@@ -113,106 +115,198 @@ LUNATIK_PRIVATECHECKER(luasocket_check, struct socket *);
...
-	lunatik_tryret(L, ret, kernel_sendmsg, socket, &msg, &vec, 1, len);
-	lua_pushinteger(L, ret);
-	return 1;
+    struct socket *socket = luasocket_check(L, 1);
+    int nargs = lua_gettop(L);
+
+    struct kvec vec;
+    struct msghdr msg;
+    struct sockaddr_storage addr;
+    int ret;
```
lneto: why changing this order? please, don't change what's not relevant for your patch.. if you would like to propose cosmetic changes, you are welcome, but please do it as a separate patch.. otherwise, it get unnecessarily harder for review ;-)

lib/luasocket.c
```diff
@@ -113,106 +115,198 @@ LUNATIK_PRIVATECHECKER(luasocket_check, struct socket *);
...
+
+    luasocket_setmsg(msg);
+
+    if (lua_type(L, 2) == LUA_TSTRING) {
+        const char *s = luaL_checklstring(L, 2, &len);
+        vec.iov_base = (void *)s;
+        vec.iov_len  = len;
+        addr_arg_index = 3;
+    } else {
+        data = luadata_check(L, 2);
```
lneto:
```suggestion
        luadata_t *ddata = luadata_check(L, 2);
```

lib/luasocket.c
```diff
@@ -113,106 +115,198 @@ LUNATIK_PRIVATECHECKER(luasocket_check, struct socket *);
...
+        offset = luaL_checkinteger(L, 3);
```
lneto:
```suggestion
        lua_Integer offset = luaL_checkinteger(L, 3);
```

lib/luasocket.c
```diff
@@ -113,106 +115,198 @@ LUNATIK_PRIVATECHECKER(luasocket_check, struct socket *);
...
+    int nargs = lua_gettop(L);
+
+    struct kvec vec;
+    struct msghdr msg;
+    struct sockaddr_storage addr;
+    int ret;
+    size_t len = 0;
+    lua_Integer offset = 0;
+    luadata_t *data = NULL;
+    int addr_arg_index = 0;
```
lneto:
```suggestion
    int ixadrr = 3;
```

lib/luasocket.c
```diff
@@ -113,106 +115,198 @@ LUNATIK_PRIVATECHECKER(luasocket_check, struct socket *);
...
+
+    luasocket_setmsg(msg);
+
+    if (lua_type(L, 2) == LUA_TSTRING) {
+        const char *s = luaL_checklstring(L, 2, &len);
+        vec.iov_base = (void *)s;
+        vec.iov_len  = len;
+        addr_arg_index = 3;
+    } else {
```
lneto: please, mind our code style..

```suggestion
    }
    else {
```

lib/luasocket.c
```diff
@@ -113,106 +115,198 @@ LUNATIK_PRIVATECHECKER(luasocket_check, struct socket *);
...
+    luadata_t *data = NULL;
+    int addr_arg_index = 0;
+
+    luasocket_setmsg(msg);
+
+    if (lua_type(L, 2) == LUA_TSTRING) {
+        const char *s = luaL_checklstring(L, 2, &len);
+        vec.iov_base = (void *)s;
+        vec.iov_len  = len;
+        addr_arg_index = 3;
```
lneto:
```suggestion
    if (lua_type(L, 2) == LUA_TSTRING)
        vec.iov_base = luaL_checklstring(L, 2, &len);
    else {
        luadata_t *data = luadata_check(L, 2);
        lua_Integer offset = luaL_checkinteger(L, 3);
        len = (size_t)luaL_checkinteger(L, 4);
        luadata_checkbounds(L, 3, data->size, offset, len);
        vec.iov_base = (void *)(data->ptr + offset);
        ixaddr = 5;
    }
    vec.iov_len  = len;
```

lib/luasocket.c
Physic69: Oops, noted , Ill keep this in mind
.

lib/luasocket.c
```diff
@@ -113,108 +115,192 @@ LUNATIK_PRIVATECHECKER(luasocket_check, struct socket *);
...
+        size_t size = luasocket_checkaddr(L, socket, &addr, ixadrr);
+        luasocket_msgaddr(msg, addr, size);
+    }
 
-	lunatik_tryret(L, ret, kernel_sendmsg, socket, &msg, &vec, 1, len);
-	lua_pushinteger(L, ret);
-	return 1;
+    lunatik_tryret(L, ret, kernel_sendmsg, socket, &msg, &vec, 1, len);
+    lua_pushinteger(L, ret);
+    return 1;
```
lneto: why changing this?

lib/luasocket.c
```diff
@@ -113,108 +115,192 @@ LUNATIK_PRIVATECHECKER(luasocket_check, struct socket *);
...
 	vec.iov_len = len;
 
-	if (unlikely(nargs >= 3)) {
-		size_t size = luasocket_checkaddr(L, socket, &addr, 3);
-		luasocket_msgaddr(msg, addr, size);
-	}
+    if (unlikely(nargs >= ixadrr)) {
+        size_t size = luasocket_checkaddr(L, socket, &addr, ixadrr);
+        luasocket_msgaddr(msg, addr, size);
+    }
```
lneto: why changing indentation here? please, make sure you are using tabs..

lib/luasocket.c
```diff
@@ -113,108 +115,192 @@ LUNATIK_PRIVATECHECKER(luasocket_check, struct socket *);
...
 
 	luasocket_setmsg(msg);
 
-	vec.iov_base = (void *)luaL_buffinitsize(L, &B, len);
-	vec.iov_len = len;
+	if (t2 != LUA_TUSERDATA && t2 != LUA_TLIGHTUSERDATA) {
+		len = (size_t)luaL_checkinteger(L, 2);
+		vec.iov_base = (void *)luaL_buffinitsize(L, &B, len);
+		vec.iov_len = len;
+	} else {
```
lneto: please, mind our code style on the whole patch

lib/luasocket.c
Physic69: The diff is accidental; my IDE didn’t show it. I’ll remove it.

lib/luasocket.c
Physic69: I missed it, my bad. Will fix this up.

lib/luadata.h
```diff
@@ -17,6 +17,21 @@ LUNATIK_LIB(data);
...
+    size_t size;
+    uint8_t opt;
+} luadata_t;
+
+LUNATIK_PRIVATECHECKER(luadata_check, luadata_t *);
+
+static inline void luadata_checkbounds(lua_State *L, int ix, size_t size,
+                                       lua_Integer offset, lua_Integer length)
+{
+    int bounds = offset >= 0 && length > 0 && offset + length <= size;
```
lneto:
```suggestion
static inline void luadata_checkbounds(lua_State *L, int ix, luadata_t *data,
                                       lua_Integer offset, lua_Integer length)
{
    int bounds = offset >= 0 && length > 0 && offset + length <= data->size;
```

lib/luasocket.c
```diff
@@ -147,14 +170,24 @@ static int luasocket_send(lua_State *L)
...
 
-	vec.iov_base = (void *)luaL_checklstring(L, 2, &len);
+    if (lua_type(L, 2) == LUA_TSTRING)
+		vec.iov_base = (void *)luaL_checklstring(L, 2, &len);
+    else {
+        luadata_t *data = luadata_check(L, 2);
+        lua_Integer offset = luaL_checkinteger(L, 3);
+        len = (size_t)luaL_checkinteger(L, 4);
+        luadata_checkbounds(L, 3, data->size, offset, len);
+        vec.iov_base = (void *)(data->ptr + offset);
```
lneto: perhaps we can add a macro for that?

```
#define luadata_topointer(data, offset)    ((data)->ptr + offset)
```

or even better, we could return the result of this macro on checkbounds (and update the code base accordingly in a separate commit before this one ;-).

what do you think?

lib/luasocket.c
```diff
@@ -164,53 +197,106 @@ static int luasocket_send(lua_State *L)
...
+		vec.iov_base = (void *)luaL_buffinitsize(L, &B, len);
+		vec.iov_len = len;
+	}
+	else {
+		data = luadata_check(L, 2);
+		lua_Integer offset = luaL_checkinteger(L, 3);
+		len = (size_t)luaL_checkinteger(L, 4);
+
+		flags = luaL_optinteger(L, 5, 0);
+		from = lua_toboolean(L, 6);
```
lneto: let's use the same ixaddr approach here.. having ixflags and ixflags + 1.. perhaps ixopt is a better name.. not sure.. then we move this code out from this block and leave here only the strictly necessary for the luadata case..

lib/luasocket.c
```diff
@@ -164,53 +197,106 @@ static int luasocket_send(lua_State *L)
...
-	vec.iov_len = len;
+	if (t2 != LUA_TUSERDATA && t2 != LUA_TLIGHTUSERDATA) {
+		len = (size_t)luaL_checkinteger(L, 2);
+		vec.iov_base = (void *)luaL_buffinitsize(L, &B, len);
+		vec.iov_len = len;
+	}
+	else {
+		data = luadata_check(L, 2);
+		lua_Integer offset = luaL_checkinteger(L, 3);
+		len = (size_t)luaL_checkinteger(L, 4);
```
lneto: this plus checkbounds seems like a pattern we can move to a new inline.. and perhaps, this new inline should be the one exported by luadata C API..

lib/luasocket.c
lneto: please, always check the ultimate resulto on the github PR

lib/luasocket.c
Physic69: something like this :
`  
   luadata_t *d = luadata_check(L, did);

    lua_Integer off = luaL_checkinteger(L, oid);

    size_t len = (size_t)luaL_checkinteger(L, lid);

    vec.iov_base = luadata_checkbounds(L, 3, data->size, offset, len);  //after updating of  checkbounds

    vec.iov_len  = len;

    return (void *)(d->ptr + off);

`
will update this

lib/luasocket.c
Physic69: I agree, updating checkbounds might be better, Ill update the luadata and make a PR regarding this.
Does that sound good?

lib/luadata.h
```diff
@@ -17,6 +17,44 @@ LUNATIK_LIB(data);
...
+{
+	int bounds = offset >= 0 && length > 0 && offset + length <= data->size;
+	luaL_argcheck(L, bounds, ix, "out of bounds");
+	return (void *)(data->ptr + offset);
+}
+
+static inline void *luadata_checkrange(lua_State *L, int idata, int ioffset, int ilen, size_t *len_out)
+{
+    luadata_t *d      = luadata_check(L, idata);
+    lua_Integer off   = luaL_checkinteger(L, ioffset);
```
lneto: please mind our code style..
```suggestion
    luadata_t *d = luadata_check(L, idata);
    lua_Integer off = luaL_checkinteger(L, ioffset);
```

lib/luadata.h
```diff
@@ -17,6 +17,44 @@ LUNATIK_LIB(data);
...
+ * @return Valid pointer within bounds
+ */
+static inline void *luadata_checkbounds(lua_State *L, int ix, luadata_t *data, lua_Integer offset, lua_Integer length)
+{
+	int bounds = offset >= 0 && length > 0 && offset + length <= data->size;
+	luaL_argcheck(L, bounds, ix, "out of bounds");
+	return (void *)(data->ptr + offset);
+}
+
+static inline void *luadata_checkrange(lua_State *L, int idata, int ioffset, int ilen, size_t *len_out)
```
lneto: I think we should pass just one index on the stack and assume them sequential.. like a CFunction..

lib/luadata.c
```diff
@@ -433,4 +409,3 @@ module_init(luadata_init);
 module_exit(luadata_exit);
 MODULE_LICENSE("Dual MIT/GPL");
 MODULE_AUTHOR("Lourival Vieira Neto <lourival.neto@ring-0.io>");
-
```
lneto: ?

lib/luadata.h
lneto: I'm not fond of calling this function checkrange because it's too similar to checkbounds.. I would call it something like checkpointer (if we can't figure a better name)

lib/luasocket.c
```diff
@@ -164,53 +193,102 @@ static int luasocket_send(lua_State *L)
...
 	luasocket_setmsg(msg);
 
-	vec.iov_base = (void *)luaL_buffinitsize(L, &B, len);
-	vec.iov_len = len;
+	if (t2 != LUA_TUSERDATA && t2 != LUA_TLIGHTUSERDATA) {
+		len = (size_t)luaL_checkinteger(L, 2);
+		vec.iov_base = (void *)luaL_buffinitsize(L, &B, len);
+	}
+	else {
+		data = luadata_check(L, 2);
```
lneto: you don't need this line anymore, right?

lib/luadata.h
lneto: I don't think we need this need to be an inline.. I would have a regular function and export its prototype here, keeping the struct and checkbounds on the .c file

lib/luasocket.c
Physic69: data here is  a flag for us here. If we remove it, we need to replace with something.

lib/luadata.h
Physic69: getrange..How does it sound?

lib/luadata.c
Physic69: will fix

lib/luasocket.c
```diff
@@ -164,53 +193,102 @@ static int luasocket_send(lua_State *L)
...
 	int ret;
+	luadata_t *data = NULL;
+	int ixopt = 3;
+	int t2 = lua_type(L, 2);
 
 	luasocket_setmsg(msg);
 
-	vec.iov_base = (void *)luaL_buffinitsize(L, &B, len);
-	vec.iov_len = len;
+	if (t2 != LUA_TUSERDATA && t2 != LUA_TLIGHTUSERDATA) {
```
lneto: that's weird.. why that?

lib/luasocket.c
```diff
@@ -164,53 +193,102 @@ static int luasocket_send(lua_State *L)
...
+	}
+	else {
+		data = luadata_check(L, 2);
+		vec.iov_base = luadata_checkrange(L, 2, 3, 4, &len);
+		ixopt = 5;
+	}
+	
+	vec.iov_len  = len;
+	flags = luaL_optinteger(L, ixopt, 0);
+    from = lua_toboolean(L, ixopt + 1);
```
lneto: as we are using C99, let's declare these vars in-here..

lneto: .. removing their declaration from the beginning of the function block

lib/luadata.h
lneto: I think it should be check* to be clear it can raise an error and validate the args..

lib/luasocket.c
Physic69: I thought we were following C89, so i made the early declerations. Will fix it up

lib/luasocket.c
Physic69: will fix it, i thought( if not data) use string, else data. will make it similar like send function

lneto: but why checking for light userdata? I would just use the same pattern you used on send() indeed..

lneto: then, you can use type the type as a flag to take the decision later on to pushbuffer

lib/luadata.h
lneto: checkbuffer will do as well.. I think.. couldn't figure anything better.. heh

Physic69: checkbuffer works for me. Sounds good :+1:

lib/luadata.h
```diff
@@ -15,10 +15,12 @@ LUNATIK_LIB(data);
 #define	LUADATA_OPT_FREE	0x02
 #define	LUADATA_OPT_KEEP  	0x80
 
+
```
lneto, line 18:
```suggestion
```

please review your patch regarding things like this..

PR #331, opened 2025-12-13
Refactor luadata_checkbounds function usage

Physic69:


lib/luadata.c
```diff
@@ -70,8 +80,8 @@ static int luadata_get##T(lua_State *L) \
 {					\
 	luadata_t *data = luadata_check(L, 1);					\
 	lua_Integer offset = luaL_checkinteger(L, 2);				\
-	luadata_checkbounds(L, 2, data->size, offset, sizeof(T##_t));		\
-	lua_pushinteger(L, (lua_Integer)*(T##_t *)(data->ptr + offset));	\
+	T##_t *ptr = luadata_checkbounds(L, 2, data, offset, sizeof(T##_t));		\
+	lua_pushinteger(L, (lua_Integer)*ptr);	\
```
```suggestion
	T##_t value = (T##_t *)luadata_checkbounds(L, 2, data, offset, sizeof(T##_t));		\
	lua_pushinteger(L, (lua_Integer)value);	\
```

as we are not going to use the actual pointer anywhere else, we could store the actual value in the var, right?

lib/luadata.c
```diff
@@ -208,9 +218,9 @@ static int luadata_getstring(lua_State *L)
 	luadata_t *data = luadata_check(L, 1);
 	lua_Integer offset = luaL_checkinteger(L, 2);
 	lua_Integer length = luaL_optinteger(L, 3, data->size - offset);
-	luadata_checkbounds(L, 2, data->size, offset, length);
+	char *ptr = luadata_checkbounds(L, 2, data, offset, length);
```
```suggestion
	char *str = (char *)luadata_checkbounds(L, 2, data, offset, length);
```

lib/luadata.c
```diff
@@ -227,10 +237,10 @@ static int luadata_setstring(lua_State *L)
 	luadata_t *data = luadata_check(L, 1);
 	lua_Integer offset = luaL_checkinteger(L, 2);
 	const char *str = luaL_checklstring(L, 3, &length);
-	luadata_checkbounds(L, 2, data->size, offset, length);
+	char *ptr = luadata_checkbounds(L, 2, data, offset, length);
```
```suggestion
	void *ptr = luadata_checkbounds(L, 2, data, offset, length);
```

lib/luadata.c
lneto: (this decision is up to you.. both ways look good to me ;-)

Physic69: Yup,  you are right, changed it. ;)

lib/luadata.c
```diff
@@ -208,9 +218,9 @@ static int luadata_getstring(lua_State *L)
 	luadata_t *data = luadata_check(L, 1);
 	lua_Integer offset = luaL_checkinteger(L, 2);
 	lua_Integer length = luaL_optinteger(L, 3, data->size - offset);
-	luadata_checkbounds(L, 2, data->size, offset, length);
+	void *str = (char *)luadata_checkbounds(L, 2, data, offset, length);
```
```suggestion
	char *str = (char *)luadata_checkbounds(L, 2, data, offset, length);
```

PR #333, opened 2025-12-18
Replace static errno table with dynamic errname() 

### LINKED ISSUE #320: replace lualinux_errno with errname()
see https://elixir.bootlin.com/linux/v6.18-rc6/source/include/linux/errname.h#L8

Aditya-Sarna:
- Add includes for linux/string.h and linux/errname.h
- Implement lualinux_errno_index() to dynamically lookup errno values using errname()
- Keep old static errno table as fallback when CONFIG_SYMBOLIC_ERRNAME is not enabled
- Update luaopen_linux() to create errno table with metatable for dynamic lookup

When CONFIG_SYMBOLIC_ERRNAME is enabled, errno values are resolved dynamically via errname() function. Otherwise, the static table is used as before.fixes #320

### DIFF
--- a/lib/lualinux.c
+++ b/lib/lualinux.c
@@ -25,6 +25,8 @@
 #include <linux/pid.h> 
 #include <linux/signal.h>
 #include <linux/errno.h>
+#include <linux/string.h>
+#include <linux/errname.h>
 
 #include <lua.h>
 #include <lauxlib.h>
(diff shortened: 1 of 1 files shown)


lneto (changes requested):
I don't think we should use this path.. IMO, it's unnecessarily complicated. I would just add a new function to return errname, instead of carrying the burden of creating a new lib like this.. if we feel we *really* want a table for that, we can do it on Lua side (in the same way we do with socket/inet, for instance).

lneto:
https://github.com/luainkernel/lunatik/pull/343

PR #334, opened 2025-12-19
implement generic resize method with SKB support

abhijeetw035:
This PR implements a generic resize() method for luadata objects, specifically enabling dynamic packet expansion and shrinking for network buffers.

Key Changes
- Architectural Refactor: Replaced the previous union with a generic void *ptr in luadata_t to simplify the data model and streamline pointer assignments.
- Abstraction Macros: Introduced LUADATA_TOPTR and LUADATA_TOSKB to cleanly handle memory access and type casting for both raw buffers and sk_buff objects.
- Resize Method: Implemented data:resize(new_size) in luadata.c.
       - For SKB-tagged data, it manages the buffer using skb_put() and skb_trim() after verifying skb_tailroom for expansion.
       - For allocated data, it utilizes lunatik_realloc() with standard OOM error handling via lunatik_checknull().
- Netfilter Integration: Updated luanetfilter.c to properly tag intercepted packets

Testing
Verified the logic using the Lunatik REPL by expanding and shrinking buffers. Confirmed via dmesg that the kernel reflects the correct size updates and that memory remains stable under both raw buffer and SKB contexts.

### DIFF
--- a/lib/luadata.c
+++ b/lib/luadata.c
@@ -22,6 +22,7 @@
 #define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
 #include <linux/kernel.h>
 #include <linux/module.h>
+#include <linux/skbuff.h>
 
 #include <lua.h>
 #include <lualib.h>
@@ -31,8 +32,11 @@
 
 #include "luadata.h"
 
+#define LUADATA_TOPTR(d)  ((d)->opt & LUADATA_OPT_SKB ? (((struct sk_buff *)(d)->ptr)->data) : (d)->ptr)
+#define LUADATA_TOSKB(d)  ((struct sk_buff *)(d)->ptr)
+
 typedef struct luadata_s {
-	char *ptr;
+	void *ptr;
 	size_t size;
 	uint8_t opt;
 } luadata_t;
@@ -56,7 +60,7 @@ static inline void *luadata_checkbounds(lua_State *L, int ix, luadata_t *data, l
 {
 	int bounds = offset >= 0 && length > 0 && offset + length <= data->size;
 	luaL_argcheck(L, bounds, ix, "out of bounds");
-	return (void *)(data->ptr + offset);
+	return (LUADATA_TOPTR(data) + offset);
 }
 
 #define luadata_checkwritable(L, data)	luaL_argcheck((L), !((data)->opt & LUADATA_OPT_READONLY), 1, "read only")
@@ -245,6 +249,54 @@ static int luadata_setstring(lua_State *L)
 }
 
 /***
+* Resizes an SKB (socket buffer) to the specified size.
+* Expands the buffer using skb_put() if new_size > current size,
+* or shrinks it using skb_trim() if new_size < current size.
+* @param L Lua state for error reporting
+* @param data luadata object wrapping the SKB
+* @param new_size The desired size in bytes
+* @raise Error if insufficient tailroom is available for expansion
+*/
+static void luadata_skb_resize(lua_State *L, luadata_t *data, size_t new_size)
+{
+	struct sk_buff *skb = LUADATA_TOSKB(data);
+	if (new_size > data->size) {
+		size_t needed = new_size - data->size;
+		if (skb_tailroom(skb) < needed)
+			luaL_error(L, "insufficient tailroom for resize");
+		skb_put(skb, needed);
+	}
+	else if (new_size < data->size)
+		skb_trim(skb, new_size);
+}
+
+/***
+* Resizes the memory block represented by the data object.
+* If the object is a network packet (SKB), it uses skb_put() to expand 
+* or skb_trim() to shrink the buffer. For raw buffers, it updates the size.
+* @function resize
+* @tparam integer new_size The desired size of the memory block in bytes.
+* @raise Error if the data object is read-only or if resize fails.
+*/
+static int luadata_resize(lua_State *L)
+{
+    luadata_t *data = luadata_check(L, 1);
+    size_t new_size = (size_t)luaL_checkinteger(L, 2);
+    
+    luadata_checkwritable(L, data);
+
+    if (data->opt & LUADATA_OPT_SKB)
+		luadata_skb_resize(L, data, new_size); 
+	else if (data->opt & LUADATA_OPT_FREE)
+		data->ptr = lunatik_checknull(L, lunatik_realloc(L, data->ptr, new_size)); 
+	else
+		luaL_error(L, "cannot resize external memory");
+
+    data->size = new_size;
+    return 0;
+}
+
+/***
 * Returns the length of the data object in bytes.
 * This is the Lua `__len` metamethod, allowing use of the `#` operator.
 * @function __len
@@ -266,7 +318,7 @@ static int luadata_length(lua_State *L)
 static int luadata_tostring(lua_State *L)
 {
 	luadata_t *data = luadata_check(L, 1);
-	lua_pushlstring(L, data->ptr, data->size);
+	lua_pushlstring(L, (char *)LUADATA_TOPTR(data), data->size);
 	return 1;
 }
 
@@ -354,6 +406,7 @@ static const luaL_Reg luadata_mt[] = {
 #endif
 	{"getstring", luadata_getstring},
 	{"setstring", luadata_setstring},
+	{"resize", luadata_resize},
 	{NULL, NULL}
 };
 
--- a/lib/luadata.h
+++ b/lib/luadata.h
@@ -13,6 +13,7 @@ LUNATIK_LIB(data);
 #define	LUADATA_OPT_NONE	0x00
 #define	LUADATA_OPT_READONLY	0x01
 #define	LUADATA_OPT_FREE	0x02
+#define LUADATA_OPT_SKB  	0x04
 #define	LUADATA_OPT_KEEP  	0x80
 
 #define luadata_clear(o)	(luadata_reset((o), NULL, 0, LUADATA_OPT_KEEP))
--- a/lib/luanetfilter.c
+++ b/lib/luanetfilter.c
@@ -77,9 +77,9 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
 		return -1;
 
 	if (skb_mac_header_was_set(skb))
-		luadata_reset(data, skb_mac_header(skb), skb_headlen(skb) + skb_mac_header_len(skb), LUADATA_OPT_NONE);
+		luadata_reset(data, skb, skb_headlen(skb) + skb_mac_header_len(skb), LUADATA_OPT_SKB);
 	else
-		luadata_reset(data, skb->data, skb_headlen(skb), LUADATA_OPT_NONE);
+		luadata_reset(data, skb, skb_headlen(skb), LUADATA_OPT_SKB);
 
 	if (lua_pcall(L, 1, 2, 0) != LUA_OK) {
 		pr_err("%s\n", lua_tostring(L, -1));
--- a/lib/luaxtable.c
+++ b/lib/luaxtable.c
@@ -129,7 +129,7 @@ static int luaxtable_pushparams(lua_State *L, const struct xt_action_param *par,
 
 static int luaxtable_domatch(lua_State *L, luaxtable_t *xtable, const struct sk_buff *skb, struct xt_action_param *par, int fallback)
 {
-	if (luaxtable_call(L, "match", xtable, (struct sk_buff *)skb, par, (luaxtable_info_t *)par->matchinfo, LUADATA_OPT_READONLY) != 0)
+	if (luaxtable_call(L, "match", xtable, (struct sk_buff *)skb, par, (luaxtable_info_t *)par->matchinfo, LUADATA_OPT_READONLY | LUADATA_OPT_SKB) != 0)
 		return fallback;
 
 	int ret = lua_toboolean(L, -1);
@@ -140,7 +140,7 @@ static int luaxtable_domatch(lua_State *L, luaxtable_t *xtable, const struct sk_
 
 static int luaxtable_dotarget(lua_State *L, luaxtable_t *xtable, struct sk_buff *skb, const struct xt_action_param *par, int fallback)
 {
-	if (luaxtable_call(L, "target", xtable,  skb, par, (luaxtable_info_t *)par->targinfo, LUADATA_OPT_NONE) != 0)
+	if (luaxtable_call(L, "target", xtable, skb, par, (luaxtable_info_t *)par->targinfo, LUADATA_OPT_SKB) != 0)
 		return fallback;
 
 	int ret = lua_tointeger(L, -1);


lib/luadata.c
```diff
@@ -31,10 +31,13 @@
 
 #include "luadata.h"
 
+#include <linux/skbuff.h>
```
lneto: This include should be placed above, along with the other kernel headers

lib/luadata.c
```diff
@@ -31,10 +31,13 @@
...
+
 typedef struct luadata_s {
 	char *ptr;
 	size_t size;
 	uint8_t opt;
+	struct sk_buff *skb;
```
lneto: The whole point of the flag is avoiding this field ;-).. with the flag we can identify if ptr is an SKB or not.. if we really need the SKB in this struct by any reason, it should be placed in a union along with ptr

abhijeetw035:
Made required changes in the latest commit please review.

abhijeetw035:
> Did you test this patch? Is it running properly?

Yes, I have verified the patch. I tested both expansion (skb_put) and shrinking (skb_trim) using Netfilter hooks in an atomic context (sleep=false).

the dmesg logs shows that the packet buffer sizes are being correctly updated in the kernel:

Expansion Test: Confirmed skb_put works and respects skb_tailroom safety checks.
image

Shrink Test: Confirmed skb_trim successfully reduces the buffer size.
image

lib/luadata.h
```diff
@@ -13,10 +13,13 @@ LUNATIK_LIB(data);
 #define	LUADATA_OPT_NONE	0x00
 #define	LUADATA_OPT_READONLY	0x01
 #define	LUADATA_OPT_FREE	0x02
+#define LUADATA_OPT_SKB  	0x04
 #define	LUADATA_OPT_KEEP  	0x80
 
 #define luadata_clear(o)	(luadata_reset((o), NULL, 0, LUADATA_OPT_KEEP))
 
+struct sk_buff;
```
lneto: why this?

lib/luadata.c
```diff
@@ -244,6 +253,43 @@ static int luadata_setstring(lua_State *L)
...
+        if (new_size > data->size) {
+            size_t needed = new_size - data->size;
+            if (skb_tailroom(data->skb) < needed) {
+                return luaL_error(L, "insufficient tailroom for resize");
+            }
+            skb_put(data->skb, needed);
+        } else if (new_size < data->size) {
+            skb_trim(data->skb, new_size);
+        }
+    } else {
```
lneto: please, mind our code style..

lib/luadata.c
```diff
@@ -244,6 +253,43 @@ static int luadata_setstring(lua_State *L)
...
+{
+    luadata_t *data = luadata_check(L, 1);
+    size_t new_size = (size_t)luaL_checkinteger(L, 2);
+    
+    luadata_checkwritable(L, data);
+
+    if (data->opt & LUADATA_OPT_SKB) {
+        if (!data->skb) {
+            return luaL_error(L, "invalid skb");
+        }
```
lneto: I'm not sure it's necessary.. how skb can be NULL and and what happens bellow if it is?

lib/luadata.c
```diff
@@ -244,6 +253,43 @@ static int luadata_setstring(lua_State *L)
...
+            size_t needed = new_size - data->size;
+            if (skb_tailroom(data->skb) < needed) {
+                return luaL_error(L, "insufficient tailroom for resize");
+            }
+            skb_put(data->skb, needed);
+        } else if (new_size < data->size) {
+            skb_trim(data->skb, new_size);
+        }
+    } else {
+        data->size = new_size;
```
lneto: it's not enough.. you need to check if the options to verify if it was allocated by lunatik or not.. so, you would need to realloc

lib/luadata.c
```diff
@@ -56,6 +61,10 @@ static inline void *luadata_checkbounds(lua_State *L, int ix, luadata_t *data, l
 {
 	int bounds = offset >= 0 && length > 0 && offset + length <= data->size;
 	luaL_argcheck(L, bounds, ix, "out of bounds");
+
+	if (data->opt & LUADATA_OPT_SKB) {
+        return (void *)(data->skb->data + offset);
```
lneto: I would create macro for that.. e.g.,

```
#define LUADATA_TOPOINTER(d)  ((d)->opt & LUADATA_OPT_SKB ? (d)->skb->data : (d)->ptr)
```

lib/luadata.c
```diff
@@ -354,6 +400,7 @@ static const luaL_Reg luadata_mt[] = {
 #endif
 	{"getstring", luadata_getstring},
 	{"setstring", luadata_setstring},
+	{"resize", luadata_resize},
```
lneto, line 409: you should handle the creation as well, right? what happens when you create a new data object with `OPT_SKB`?

lib/luanetfilter.c
```diff
@@ -76,10 +76,7 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
 	if (!luanetfilter_pushcb(L, luanf) || (data = luanetfilter_pushskb(L, luanf, skb)) == NULL)
 		return -1;
 
-	if (skb_mac_header_was_set(skb))
-		luadata_reset(data, skb_mac_header(skb), skb_headlen(skb) + skb_mac_header_len(skb), LUADATA_OPT_NONE);
```
lneto, line 80: why removing this?

lib/luadata.h
abhijeetw035: I've removed the redundant forward declaration of struct sk_buff. Since luadata.c already includes <linux/skbuff.h>, it is no longer necessary in the header.

lib/luadata.c
abhijeetw035: yes implemented the LUADATA_PTR macro as suggested. This unifies data access across checkbounds, tostring, and getstring, ensuring the correct pointer is used depending on the OPT_SKB flag.

lib/luadata.c
abhijeetw035: Sorry was testing with it while running, removed the NULL check for data->skb.

lib/luadata.c
abhijeetw035: Updated the resize logic for raw buffers. It now uses krealloc() with GFP_ATOMIC for objects with the OPT_FREE flag. For external references that cannot be safely resized, it now returns an explicit error.

lib/luanetfilter.c
abhijeetw035: Ah,  during the union refactor I accidentally dropped the case where the MAC header gets set. I’ve put the original logic back so luadata now counts the MAC header length when it’s present.

lib/luadata.c
abhijeetw035: Updated luadata_lnew to ensure that objects created from user space are strictly initialized with OPT_FREE. The OPT_SKB flag is now strictly managed by internal kernel hooks to prevent unauthorized access to packet structures.

lneto: I don't get it; objects are not created from userspace. Everything runs in the kernel. What I meant was re the data obj creation using the C API. Imagine you want to create a new data object passing an SKB. It should be handled properly; not only reset.

lib/luadata.c
```diff
@@ -244,6 +252,45 @@ static int luadata_setstring(lua_State *L)
...
+    luadata_checkwritable(L, data);
+
+    if (data->opt & LUADATA_OPT_SKB) {
+		if (new_size > data->size) {
+			size_t needed = new_size - data->size;
+			if (skb_tailroom(data->skb) < needed)
+				return luaL_error(L, "insufficient tailroom for resize");
+
+			skb_put(data->skb, needed);
+		} else if (new_size < data->size) {
```
lneto: Please, read our code base and mind our code style on if-then-else blocks =).

lib/luadata.c
```diff
@@ -244,6 +252,45 @@ static int luadata_setstring(lua_State *L)
...
+			skb_trim(data->skb, new_size);
+		}
+	} else if (data->opt & LUADATA_OPT_FREE) {
+		void *ptr = krealloc(data->ptr, new_size, GFP_ATOMIC);
+		if (unlikely(!ptr))
+			return luaL_error(L, "resize: out of memory");
```
lneto: Please do not prefix log like this. Mind the other places we raise OOM and replicate it. You are welcome to suggest any change in our code style; but let's not do It adhoc'ly.

lib/luadata.c
```diff
@@ -244,6 +252,45 @@ static int luadata_setstring(lua_State *L)
...
+			size_t needed = new_size - data->size;
+			if (skb_tailroom(data->skb) < needed)
+				return luaL_error(L, "insufficient tailroom for resize");
+
+			skb_put(data->skb, needed);
+		} else if (new_size < data->size) {
+			skb_trim(data->skb, new_size);
+		}
+	} else if (data->opt & LUADATA_OPT_FREE) {
+		void *ptr = krealloc(data->ptr, new_size, GFP_ATOMIC);
```
lneto: We shouldn't call realloc directly here. Instead, we should use Lunatik APIs that will handle that for us on the right context. Please, read our code base. It's named as lunatik_realloc if I'm not mistaken.

lib/luadata.c
```diff
@@ -22,6 +22,8 @@
 #define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
 #include <linux/kernel.h>
 #include <linux/module.h>
+#include <linux/skbuff.h>
+#include <linux/slab.h>
```
lneto: You don't need to include slab explicitly, the core should handle it for you.

lib/luadata.c
abhijeetw035: I checked lunatik.h and saw that lunatik_checknull raises the standard 'not enough memory' error if the pointer is NULL.

So I can just simplify it to this, correct?
`data->ptr = lunatik_checknull(L, lunatik_realloc(L, data->ptr, new_size));`

abhijeetw035: Also for tailroom and external memory this are fine right ??

``` c

tailroom ->
if (skb_tailroom(data->skb) < needed)
      return luaL_error(L, "insufficient tailroom for resize");

external memory ->
else
	return luaL_error(L, "cannot resize external memory");

Comment (lneto) [reply lib/luadata.c]: > I checked lunatik.h and saw that lunatik_checknull raises the standard 'not enough memory' error if the pointer is NULL.
> 
> So I can just simplify it to this, correct? `data->ptr = lunatik_checknull(L, lunatik_realloc(L, data->ptr, new_size));`

precisely =)

Comment (lneto) [reply lib/luadata.c]: > Also for tailroom and external memory this are fine right ??
> 
> tailroom ->
> if (skb_tailroom(data->skb) < needed)
>       return luaL_error(L, "insufficient tailroom for resize");
> 
> external memory ->
> else
> 	return luaL_error(L, "cannot resize external memory");

LGTM =). Please notice that, though it's a common patter in Lua, we don't use `return luaL_error(...)` on Lunatik because it raises warns on kbuild

Comment (abhijeetw035) [reply lib/luadata.c]: > > Also for tailroom and external memory this are fine right ??
> > tailroom ->
> > if (skb_tailroom(data->skb) < needed)
> > return luaL_error(L, "insufficient tailroom for resize");
> > external memory ->
> > else
> > return luaL_error(L, "cannot resize external memory");
> 
> LGTM =). Please notice that, though it's a common patter in Lua, we don't use `return luaL_error(...)` on Lunatik because it raises warns on kbuild

sure, i removed the return keyword before luaL_error() calls to avoid kbuild warnings. Thanks!
made required changes please review

File: lib/luadata.c (hunk 8 above)
Comment (abhijeetw035) [reply lib/luadata.c:409]: thanks for the clarification. I've updated luadata_create and luadata_reset to explicitly check the OPT_SKB flag.

File: lib/luadata.c [hunk 14]
```diff
@@ -56,7 +62,8 @@ static inline void *luadata_checkbounds(lua_State *L, int ix, luadata_t *data, l
 {
 	int bounds = offset >= 0 && length > 0 && offset + length <= data->size;
 	luaL_argcheck(L, bounds, ix, "out of bounds");
-	return (void *)(data->ptr + offset);
+
```
lneto: please, don't change cosmetics along with logic changes.. it makes it harder to review patches

lib/luanetfilter.c
lneto: we use `--fixup`s while development, we shouldn't remove than re-add code chunks on a PR.. otherwise, it gets harder to keep track of the history.. we use git log/blame during development to assess how code was introduced while handling issues or new features.. could you please sanitize the commit history of this branch.. it's OK to rebase and push -f it

lib/luadata.c
```diff
@@ -244,6 +251,41 @@ static int luadata_setstring(lua_State *L)
...
+    if (data->opt & LUADATA_OPT_SKB) {
+		if (new_size > data->size) {
+			size_t needed = new_size - data->size;
+			if (skb_tailroom(data->skb) < needed)
+				luaL_error(L, "insufficient tailroom for resize");
+
+			skb_put(data->skb, needed);
+			
+		} else if (new_size < data->size)
```
lneto: I think I have reported this already.. could you review your entire patch regarding this?
```suggestion
		}
		else if (new_size < data->size)
```

lib/luadata.c
```diff
@@ -244,6 +251,41 @@ static int luadata_setstring(lua_State *L)
...
+    luadata_checkwritable(L, data);
+
+    if (data->opt & LUADATA_OPT_SKB) {
+		if (new_size > data->size) {
+			size_t needed = new_size - data->size;
+			if (skb_tailroom(data->skb) < needed)
+				luaL_error(L, "insufficient tailroom for resize");
+
+			skb_put(data->skb, needed);
+
```
```suggestion
```

lib/luadata.c
```diff
@@ -244,6 +251,41 @@ static int luadata_setstring(lua_State *L)
...
+    size_t new_size = (size_t)luaL_checkinteger(L, 2);
+    
+    luadata_checkwritable(L, data);
+
+    if (data->opt & LUADATA_OPT_SKB) {
+		if (new_size > data->size) {
+			size_t needed = new_size - data->size;
+			if (skb_tailroom(data->skb) < needed)
+				luaL_error(L, "insufficient tailroom for resize");
+
```
```suggestion
```

lib/luadata.c
```diff
@@ -56,7 +62,8 @@ static inline void *luadata_checkbounds(lua_State *L, int ix, luadata_t *data, l
 {
 	int bounds = offset >= 0 && length > 0 && offset + length <= data->size;
 	luaL_argcheck(L, bounds, ix, "out of bounds");
-	return (void *)(data->ptr + offset);
+
+	return (void *)(LUADATA_PTR(data) + offset);
```
lneto: wondering if this function shouldn't return `char *` or if we shouldn't store the pointer as `void *`. Perhaps having a `void *ptr` would be better than the union.. it would simplify a bit the assignment block and we could have two similar macros.. TO/FROMPTR using `?` operator.. what do you think?

@Physic69 can you share your thoughts as well?

abhijeetw035: yes i think using a generic void *ptr instead of the union is better way to simplify the code. It would clean up the assignment blocks in `luadata_create` and `luadata_reset` by removing tthat `if...else...` blocks
its likes access is managed entirely by macros
also union doesn't add value since only one pointer is valid at a time based on the opt flags.

also since `void *` is universal pointer type so it converts implicitly to any pointer type without casting
so if we changed it to char * then every caller would need an explicit cast

Physic69: +1 for void* . Socket iov are native void*. char* added casts everywhere.

Physic69: +1 for void* . Socket iov are native void*. char* added casts everywhere.

Physic69: Simple FROM macro & TO macro can do for strings.
Why to use '?' operator. 
Or I didnt get your intent

Physic69: Simple FROM macro & TO macro can do for strings.
Why to use '?' operator. 
Or I didnt get your intent

lneto: > Simple FROM macro & TO macro can do for strings. Why to use '?' operator. Or I didnt get your intent

Because of SKB ;-)

lib/luanetfilter.c
abhijeetw035: Thanks for the info on using fixups! I've sanitized the history by rebasing the branch into a single

abhijeetw035:
Please review the current implementation, once everything done the logic and the refactor, let me know and I'll move this from Draft to Ready for Review.

lib/luadata.c
```diff
@@ -31,8 +32,11 @@
 
 #include "luadata.h"
 
+#define LUADATA_TOPTR(d)  ((d)->opt & LUADATA_OPT_SKB ? (char *)(((struct sk_buff *)(d)->ptr)->data) : (char *)(d)->ptr)
```
```suggestion
#define LUADATA_TOPTR(d)  ((d)->opt & LUADATA_OPT_SKB ? (((struct sk_buff *)(d)->ptr)->data) : (d)->ptr)
```

lib/luadata.c
```diff
@@ -56,7 +60,7 @@ static inline void *luadata_checkbounds(lua_State *L, int ix, luadata_t *data, l
 {
 	int bounds = offset >= 0 && length > 0 && offset + length <= data->size;
 	luaL_argcheck(L, bounds, ix, "out of bounds");
-	return (void *)(data->ptr + offset);
+	return (void *)(LUADATA_TOPTR(data) + offset);
```
```suggestion
	return (LUADATA_TOPTR(data) + offset);
```

lib/luadata.c
```diff
@@ -266,7 +306,7 @@ static int luadata_length(lua_State *L)
 static int luadata_tostring(lua_State *L)
 {
 	luadata_t *data = luadata_check(L, 1);
-	lua_pushlstring(L, data->ptr, data->size);
+	lua_pushlstring(L, LUADATA_TOPTR(data), data->size);
```
```suggestion
	lua_pushlstring(L, (char *)LUADATA_TOPTR(data), data->size);
```

lib/luadata.c
```diff
@@ -244,6 +248,42 @@ static int luadata_setstring(lua_State *L)
...
+    if (data->opt & LUADATA_OPT_SKB) {
+		struct sk_buff *skb = LUADATA_TOSKB(data);
+		if (new_size > data->size) {
+			size_t needed = new_size - data->size;
+			if (skb_tailroom(skb) < needed)
+				luaL_error(L, "insufficient tailroom for resize");
+			skb_put(skb, needed);
+		}
+		else if (new_size < data->size)
+			skb_trim(skb, new_size);
```
lneto: wondering if it would be better to move this block to an aux func. What do you think? feel free to choose =). No strong feelings on this..

abhijeetw035: yes better , because of code readability of that if else blocks i was adding space before
using aux function is also good way for it. thanks for the suggestion. will add the aux function

lneto (approved):

abhijeetw035:
Thanks for the review! really enjoyed working on this PR and learned lot of things like getting deeper into kernel memory management and SKB handling stuff, and feel more comfortable with lunatik
Looking forward to more contributions 🙂

lneto (changes requested):
sorry, I was about to merge it and just notice that you missed this line https://github.com/luainkernel/lunatik/blob/master/lib/luaxtable.c#L143, it should be OPT_SKB as well, right? also, can you remove the `feat:` prefix from your commit message? then, I can merge it. thanks!

lneto:
> Thanks for the review! really enjoyed working on this PR and learned lot of things like getting deeper into kernel memory management and SKB handling stuff, and feel more comfortable with lunatik Looking forward to more contributions 🙂

thank you for your contrib! btw, if you are interested in memory allocation, there's something that on my plate that you might like to implement.. a new optional allocator based on pools.. let me know if you want to tackle that..

abhijeetw035:
> sorry, I was about to merge it and just notice that you missed this line https://github.com/luainkernel/lunatik/blob/master/lib/luaxtable.c#L143, it should be OPT_SKB as well, right? also, can you remove the `feat:` prefix from your commit message? then, I can merge it. thanks!

hii , for consistency we can also do the same in the luaxtable_domatch function right ?

image

abhijeetw035:
> > Thanks for the review! really enjoyed working on this PR and learned lot of things like getting deeper into kernel memory management and SKB handling stuff, and feel more comfortable with lunatik Looking forward to more contributions 🙂
> 
> thank you for your contrib! btw, if you are interested in memory allocation, there's something that on my plate that you might like to implement.. a new optional allocator based on pools.. let me know if you want to tackle that..

yes suree ! will like to tackle the optional pool-based allocator... please let me know the details or if there's an issue I should follow to get started. ready to take it on

lneto:
> > sorry, I was about to merge it and just notice that you missed this line https://github.com/luainkernel/lunatik/blob/master/lib/luaxtable.c#L143, it should be OPT_SKB as well, right? also, can you remove the `feat:` prefix from your commit message? then, I can merge it. thanks!
> 
> hii , for consistency we can also do the same in the luaxtable_domatch function right ?

sure, we can do it for consistency; but resize() cannot be applied to READONLY data objects.

lneto:
> > Did you test this patch? Is it running properly?
> 
> Yes, I have verified the patch. I tested both expansion (skb_put) and shrinking (skb_trim) using Netfilter hooks in an atomic context (sleep=false).
> 
> the dmesg logs shows that the packet buffer sizes are being correctly updated in the kernel:

perhaps it would be good to have such tests in the repo.. if it's clean enough, please consider to create a new PR with them.. if it will give you much work, I'd prioritize that.. up to you actually ;-)

PR #336, opened 2025-12-20
add NOSHARE flag

Physic69:


lib/luacompletion.c
```diff
@@ -119,7 +119,7 @@ static const luaL_Reg luacompletion_mt[] = {
 static const lunatik_class_t luacompletion_class = {
 	.name = "completion",
 	.methods = luacompletion_mt,
-	.sleep = false,
+	.flags = false,
```
lneto: it shouldn't be false anymore, right? but a proper flag value

lib/luadata.c
```diff
@@ -361,7 +361,7 @@ static const lunatik_class_t luadata_class = {
 	.name = "data",
 	.methods = luadata_mt,
 	.release = luadata_release,
-	.sleep = false,
+	.flags = LUNATIK_CLASS_NOSHARE,
```
lneto: it's not acceptable to force all data objects to be "noshare", right? what happens if we need to share such objects (as we currently do)?

lunatik_obj.c
```diff
@@ -58,9 +58,13 @@ EXPORT_SYMBOL(lunatik_checkpobject);
 
 void lunatik_cloneobject(lua_State *L, lunatik_object_t *object)
 {
+	const lunatik_class_t *class = object->class;
+
+	if (class->flags & LUNATIK_CLASS_NOSHARE)
+		luaL_error(L, "%s objects cannot be shared across runtimes",class->name);
```
lneto: please mind our code style, e.g., `, class->...`

lunatik.h
```diff
@@ -17,9 +17,15 @@
 
 #define LUNATIK_VERSION	"Lunatik 3.7"
 
+#define LUNATIK_CLASS_SLEEPABLE  0x01
+#define LUNATIK_CLASS_NOSHARE 0x02
```
lneto: I think "SINGLE" would be a better name than "NOSHARE", what do you think?

lneto: how do you handle allocations? we could have "single" mode objects that might or might not sleep on allocation, besides the fact they aren't shared. Perhaps, a better idea is to have two bools instead of `flags`, `sleep` and `shared`. What do you think?

Physic69: Im bad with naming :|
Single sounds better, (single runtime)

lunatik.h
```diff
@@ -17,9 +17,12 @@
 
 #define LUNATIK_VERSION	"Lunatik 3.7"
 
+#define lunatik_object_issleepable(obj) ((obj)->sleep)
+#define lunatik_class_issleepable(cls)  ((cls)->sleep)
```
```suggestion
#define lunatik_issleepable(o) ((o)->sleep)
```

lneto: btw, I wouldn't change this on the same PR.. it seems only cosmetic and not require to the main change your are adding.. let's keep the patches as self-contained as possible..

lunatik.h
```diff
@@ -77,6 +80,7 @@ typedef struct lunatik_class_s {
 	const luaL_Reg *methods;
 	void (*release)(void *);
 	bool sleep;
+	bool shared;
```
lneto: I think we should invert this flag.. so we don't need to change all modules and just the places we really need (such as pointer)..

```suggestion
	bool single;
```

lneto: actually.. I wouldn't add this to class now, we don't need classes that are exclusively "single", right? I would just have it on object

lunatik_obj.c
```diff
@@ -58,6 +58,9 @@ EXPORT_SYMBOL(lunatik_checkpobject);
 
 void lunatik_cloneobject(lua_State *L, lunatik_object_t *object)
 {
+	 if (!object->shared)
+        luaL_error(L, "%s objects cannot be shared across runtimes", object->class->name);
```
lneto: it's not enough, right? we probably need to add `single` flag to the API for creating new objects..

lneto: also, avoiding creating mutex/spinlock and refcount unnecessarily..

lunatik.h
```diff
@@ -89,14 +93,15 @@ typedef struct lunatik_object_s {
 		spinlock_t spin;
 	};
```
lneto, line 93: maybe we should have a substruct allocated dynamically for kref and lock.. but let's worry about this after having the first version settled..

lib/luadata.h
```diff
@@ -13,6 +13,7 @@ LUNATIK_LIB(data);
 #define	LUADATA_OPT_NONE	0x00
 #define	LUADATA_OPT_READONLY	0x01
 #define	LUADATA_OPT_FREE	0x02
+#define LUADATA_OPT_SINGLE  0x04
```
lneto: you can just use the obj flag, right?

lib/luadata.c
```diff
@@ -362,17 +362,23 @@ static const lunatik_class_t luadata_class = {
 	.methods = luadata_mt,
 	.release = luadata_release,
 	.sleep = false,
+	.shared = true,
 };
 
 static int luadata_lnew(lua_State *L)
 {
 	size_t size = (size_t)luaL_checkinteger(L, 1);
+	bool single = lua_toboolean(L, 2); /* false -> sharable */
```
```suggestion
	bool single = lua_toboolean(L, 2);
```

lunatik.h
```diff
@@ -17,9 +17,12 @@
 
 #define LUNATIK_VERSION	"Lunatik 3.7"
 
+#define lunatik_object_issleepable(obj) ((obj)->sleep)
+#define lunatik_class_issleepable(cls)  ((cls)->sleep)
+
 #define lunatik_locker(o, mutex_op, spin_op)	\
 do {						\
```
lneto, line 21: it should check if it's single, right? it should lock in that case.. also, refcount shouldn't be applied if single

Physic69:
This version lunatik driver is working properly...
I didnt finish comments, yet.
Ill ping you once done

lunatik.h
```diff
@@ -104,24 +92,20 @@ typedef struct lunatik_object_s {
...
+	bool single;
 	gfp_t gfp;
 } lunatik_object_t;
 
 extern lunatik_object_t *lunatik_env;
 
 static inline int lunatik_trylock(lunatik_object_t *object)
 {
-	return object->sleep ? mutex_trylock(&object->mutex) : spin_trylock_bh(&object->spin);
+	return object->single ? 1 : lunatik_object_issleepable(object) ? mutex_trylock(&object->mutex) : spin_trylock_bh(&object->spin);
```
lneto: why should we return 1 if single? what's the use case?

lunatik.h
```diff
@@ -193,18 +177,20 @@ static inline void lunatik_setclass(lua_State *L, const lunatik_class_t *class)
...
+static inline void lunatik_setobject(lunatik_object_t *object, const lunatik_class_t *class, bool sleep, bool single)
 {
 	kref_init(&object->kref);
 	object->private = NULL;
 	object->class = class;
 	object->sleep = sleep;
+    object->single = single;
 	object->gfp = sleep ? GFP_KERNEL : GFP_ATOMIC;
-	lunatik_newlock(object);
+	if(!single)
```
lneto, line 189: please mind our code style

lunatik.h
```diff
@@ -193,18 +177,20 @@ static inline void lunatik_setclass(lua_State *L, const lunatik_class_t *class)
...
+		lunatik_newlock(object);
```
lneto, line 190: we don't need refcount as well, right?

lunatik.h
```diff
@@ -193,18 +177,20 @@ static inline void lunatik_setclass(lua_State *L, const lunatik_class_t *class)
 	lua_setiuservalue(L, -2, 1); /* pop class */
 }
 
-static inline void lunatik_setobject(lunatik_object_t *object, const lunatik_class_t *class, bool sleep)
+static inline void lunatik_setobject(lunatik_object_t *object, const lunatik_class_t *class, bool sleep, bool single)
```
lneto, line 180: I'm rethinking this; this a good reason for having a single `flags` parameter.. otherwise, we will need a new argument for every new decision we need.. it doesn't scale.. what do you think?

lib/luadata.c
```diff
@@ -420,9 +367,12 @@ static const lunatik_class_t luadata_class = {
 static int luadata_lnew(lua_State *L)
 {
 	size_t size = (size_t)luaL_checkinteger(L, 1);
+	bool single = lua_toboolean(L, 2);
+
 	lunatik_object_t *object = lunatik_newobject(L, &luadata_class, sizeof(luadata_t));
 	luadata_t *data = (luadata_t *)object->private;
 
+	object->single = single;
```
lneto, line 375: doing this after the object creation won't avoid the lock init

lunatik.h
Physic69: Yup, I agree. There are many places where i need to change this. Lets settle on a working version first, then ill move to a single parameter. Sounds good?

lib/luadata.c
Physic69: Wanted to know if im going in correct direction for creating objects:[here](https://github.com/luainkernel/lunatik/pull/336/changes#diff-6dddb85ed3c5b1111e4d90a3fe6d8fad2c346c0a388035fcf59f94011b9688b8R14)
I didnt update the function here yet... Will do update this.. (the line is just redundancy as of now, will remove)

Physic69:
Hi
Need some thoughts if i should continue this / use the latest externalstring function ?
IMO this might be better... externelstring is just r only. Here its luadata & its r/w.

lneto:
> Hi Need some thoughts if i should continue this / use the latest externalstring function ? IMO this might be better... externelstring is just r only. Here its luadata & its r/w.

Agreed; also, this will be useful for other objects as well, I think

lneto:
hi @Physic69, I'm thinking that it would be better for me to tackle this issue if it's cool for you. I think I will need diff exec modes for bottom halves and irqs, I could bundle single mode on this change or at least create the flag (for implementing the decision later). Is it alright for you? Then, you could tackle another issue ;-). Thanks.

Physic69:
> hi @Physic69, I'm thinking that it would be better for me to tackle this issue if it's cool for you. I think I will need diff exec modes for bottom halves and irqs, I could bundle single mode on this change or at least create the flag (for implementing the decision later). Is it alright for you? Then, you could tackle another issue ;-). Thanks.

Sure, that’s totally fine 👍...
Thanks for taking this up.

PR #338, opened 2025-12-21
provide automatic nop fallback for missing callbacks

Zapper9982:
### Changes:
Added:  
- `lunatik_nop1 : A generic no-op lua_CFunction in `lunatik.h`
- `lunatik_defaultfield` : reusable helper to default table fields to a function 
- Modified device.new to automatically apply `lunatik_nop` when callbacks are missing

### Testing : 

- Verified with devices using default callbacks
- Verified custom callbacks are preserved
- Verified mixed scenarios (custom + default)
`All tests pass with kernel module loaded successfully`

### Screenshots : 

image

### DIFF
--- a/examples/systrack.lua
+++ b/examples/systrack.lua
@@ -10,10 +10,8 @@ local systab = require("syscall.table")
 
 local syscalls = {"openat", "read", "write", "readv", "writev", "close"}
 
-local function nop() end -- do nothing
-
 local s = linux.stat
-local driver = {name = "systrack", open = nop, release = nop, mode = s.IRUGO}
+local driver = {name = "systrack", mode = s.IRUGO}
 
 local track = {}
 local toggle = true
@@ -36,7 +34,7 @@ for _, symbol in ipairs(syscalls) do
 		track[symbol] = track[symbol] + 1
 	end
 
-	probe.new(address, {pre = handler, post = nop})
+	probe.new(address, {pre = handler})
 end
 
 device.new(driver)
--- a/lib/luadevice.c
+++ b/lib/luadevice.c
@@ -96,11 +96,7 @@ static int luadevice_fop(lua_State *L, luadevice_t *luadev, const char *fop, int
 		goto err;
 	}
 
-	if (lua_getfield(L, -1, fop) != LUA_TFUNCTION) {
-		lua_getfield(L, -2, "name");
-		pr_err("%s: operation isn't defined for /dev/%s\n", fop, lua_tostring(L, -1));
-		goto err;
-	}
+	lunatik_optcfunction(L, -1, fop, lunatik_nop);
 
 	lua_insert(L, base + 1); /* fop */
 	lua_insert(L, base + 2); /* driver */
--- a/lunatik.h
+++ b/lunatik.h
@@ -102,6 +102,11 @@ static inline int lunatik_trylock(lunatik_object_t *object)
 int lunatik_runtime(lunatik_object_t **pruntime, const char *script, bool sleep);
 int lunatik_stop(lunatik_object_t *runtime);
 
+static inline int lunatik_nop(lua_State *L)
+{
+	return 0;
+}
+
 static inline void *lunatik_realloc(lua_State *L, void *ptr, size_t size)
 {
 	void *ud = NULL;
@@ -326,6 +331,14 @@ do {									\
 	priv->field = lua_isnil(L, -1) ? opt : lua_tointeger(L, -1);	\
 	lua_pop(L, 1);							\
 } while (0)
+
+static inline void lunatik_optcfunction(lua_State *L, int idx, const char *field, lua_CFunction default_func)
+{
+	if (lua_getfield(L, idx, field) != LUA_TFUNCTION) {
+		lua_pop(L, 1);
+		lua_pushcfunction(L, default_func);
+	}
+}
 
 #define lunatik_checkbounds(L, idx, val, min, max)	luaL_argcheck(L, val >= min && val <= max, idx, "out of bounds")
 
--- a/examples/spyglass.lua
+++ b/examples/spyglass.lua
@@ -24,9 +24,7 @@ local function printable(keysym)
 	return keysym >= 32 and keysym <= 126
 end
 
-local function nop() end
-
-local spyglass = {name = "spyglass", open = nop, release = nop, log = ""}
+local spyglass = {name = "spyglass", log = ""}
 
 function spyglass:read()
 	local log = self.log
--- a/examples/tap.lua
+++ b/examples/tap.lua
@@ -12,10 +12,8 @@ local RAW       = socket.sock.RAW
 local ETH_P_ALL = 0x0003
 local MTU       = 1500
 
-local function nop() end
-
 local s = linux.stat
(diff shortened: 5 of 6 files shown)


lib/luadevice.c
```diff
@@ -363,6 +363,9 @@ static int luadevice_new(lua_State *L)
 
 	luaL_checktype(L, 1, LUA_TTABLE); /* driver */
 
+	lunatik_defaultfield(L, 1, "open", lunatik_nop);
+	lunatik_defaultfield(L, 1, "release", lunatik_nop);
```
lneto: this wasn't what I meant.. I meant to just failover on the call.. not to check the field and store it.. look for the place we are actually calling and failing when such callbacks aren't defined and instead of failing (if type(field) is not function, use luantik_nop.

Zapper9982: okay yeah makes sense , i misinterpreted the requirement , but got it now

examples/systrack.lua
```diff
@@ -13,7 +13,7 @@ local syscalls = {"openat", "read", "write", "readv", "writev", "close"}
 local function nop() end -- do nothing
```
lneto: Unused, right?

Zapper9982:
> Please, squash your commits into a single one. Could you also review Luadevice documentation to see if it needs to be updated? Thanks!

Yes sure will do that

examples/systrack.lua
Zapper9982: okay , so the runtime fallback is added for `device.new` and not `probe.new` so ideally if i remove that, it will break . 
so i am wondering if i should add a similar handler for probe as well ?

lneto: ah.. it's used on `probe.new` below, right? yeah, if it's apply, I think we should do this change as well.. and perhaps add a new macro that could be used for both.. take a look at [optinteger()](https://github.com/luainkernel/lunatik/blob/master/lunatik.h#L323).

lunatik.h
```diff
@@ -102,6 +102,17 @@ static inline int lunatik_trylock(lunatik_object_t *object)
...
+{
+	return 0;
+}
+
+#define lunatik_optfunction(L, idx, field, default_func) \
+	((lua_getfield((L), (idx), (field)) == LUA_TFUNCTION) ? \
+		(void)0 : \
+		(lua_pop((L), 1), lua_pushcfunction((L), (default_func))))
+
+
```
lneto: please, mind our style..

```suggestion

```

lunatik.h
```diff
@@ -102,6 +102,17 @@ static inline int lunatik_trylock(lunatik_object_t *object)
...
 
+static inline int lunatik_nop(lua_State *L)
+{
+	return 0;
+}
+
+#define lunatik_optfunction(L, idx, field, default_func) \
+	((lua_getfield((L), (idx), (field)) == LUA_TFUNCTION) ? \
+		(void)0 : \
+		(lua_pop((L), 1), lua_pushcfunction((L), (default_func))))
```
lneto: any reason for not following the `do { } while (0)` approach? btw, I would create a template, e.g., `NEWOPT` and create optcfunction and optinteger using it.

lunatik.h
```diff
@@ -102,6 +102,17 @@ static inline int lunatik_trylock(lunatik_object_t *object)
 int lunatik_runtime(lunatik_object_t **pruntime, const char *script, bool sleep);
 int lunatik_stop(lunatik_object_t *runtime);
 
+static inline int lunatik_nop(lua_State *L)
+{
+	return 0;
+}
+
+#define lunatik_optfunction(L, idx, field, default_func) \
```
lneto: please move it next to optinteger, it will make easier to find this API in the header later =)

lunatik.h
Zapper9982: > any reason for not following the `do { } while (0)` approach? btw, I would create a template, e.g., `NEWOPT` and create optcfunction and optinteger using it.

I thought the ternary operator would do the trick. At the end, both achieve the same result functionally, but you're right that do-while-0 is more idiomatic for statement-like macros. 

I'll go ahead and make a `NEWOPT`  template .

Zapper9982: @lneto  but , isnt optinteger being used to access struct fields , while in optcfunction the ideal requirement , is accessing Lua Table fields , so i think its better  just refactor optcfunction .

lunatik.h
```diff
@@ -327,6 +332,13 @@ do {									\
 	lua_pop(L, 1);							\
 } while (0)
 
+#define lunatik_optcfunction(L, idx, field, default_func)		\
+do {									\
+	(lua_getfield(L, idx, field) == LUA_TFUNCTION) ?		\
+		(void)0 :						\
+		(lua_pop(L, 1), lua_pushcfunction(L, default_func));	\
+} while (0)
```
```suggestion
static inline void lunatik_optcfunction(L, idx, field, opt)
{
	if (lua_getfield(L, idx, field) != LUA_TFUNCTION) {
		lua_pop(L, 1);
		lua_pushcfunction(L, opt));
	}
}
```

I'm not fond of using the same name pattern as optinteger as we are not actually setting a struct field as pointed out, but I couldn't think on something better.. anyway, in this case, we don't need a macro.. an inline will do =). Please, move it to next optinteger anyway, it will be easier to maintain. Thanks.

lunatik.h
lneto: @Zapper9982 you are right. Perhaps, we should change optinteger instead to just get the field and have another macro calling it to set the struct field, that's breaking it into two macros. What do you think? (of course, we can tackle it as a separate PR).

Zapper9982: Breaking lunatik_optinteger into a getter macro (that just pushes the value or the default onto the stack) and an assignment wrapper would make it much more composable for sure

lunatik.h
Zapper9982: Done :))

lneto (commented):
It's looking good to me.. the only two things are missing is 1) to move the function declaration to the region I mentioned and 2) squash your commits. Thanks!

Zapper9982:
> It's looking good to me.. the only two things are missing is 1) to move the function declaration to the region I mentioned and 2) squash your commits. Thanks!
 
 Its already there :)) 
image


but yes ill squash my commits now , was waiting on getting the final green flag , so ill squash them at once

lib/luadevice.c
```diff
@@ -96,11 +96,8 @@ static int luadevice_fop(lua_State *L, luadevice_t *luadev, const char *fop, int
 		goto err;
 	}
 
-	if (lua_getfield(L, -1, fop) != LUA_TFUNCTION) {
-		lua_getfield(L, -2, "name");
-		pr_err("%s: operation isn't defined for /dev/%s\n", fop, lua_tostring(L, -1));
-		goto err;
-	}
+
```
lneto:
```suggestion
```

lneto: plase mind that, we don't do double \n

lneto (changes requested):
sorry.. I was about to merge and found that `\n`, can you fix it? then I can merge.. also, could you remove this "feat(core)" from your commit message? we don't use this kind of prefix.. please, check our history to get a better sense of our commit messages..

PR #343, opened 2025-12-24
replace static errno table with errname() wrapper

abhijeetw035:
This PR replaces the manually maintained lualinux_errno table with a dynamic lookup using the kernel's native errname() function.

key changes : 
- Implemented a lunatik_errname() wrapper in lunatik.h that is safely guarded by CONFIG_SYMBOLIC_ERRNAME.
- Updated the lunatik_tryret macro to use the new guarded wrapper.
- Removed the static lualinux_errno array and its namespace entry from lualinux.c to eliminate code bloat.
- Exposed the dynamic lookup via a new linux.errname(err) function in the linux module.

### DIFF
--- a/lib/lualinux.c
+++ b/lib/lualinux.c
@@ -25,6 +25,9 @@
 #include <linux/pid.h> 
 #include <linux/signal.h>
 #include <linux/errno.h>
+#include <linux/errname.h>
+#include <linux/version.h>
+#include <linux/err.h>
 
 #include <lua.h>
 #include <lauxlib.h>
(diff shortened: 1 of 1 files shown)


lunatik.h
```diff
@@ -142,10 +127,19 @@ static inline void *lunatik_checknull(lua_State *L, void *ptr)
 
 #define lunatik_checkalloc(L, s)	(lunatik_checknull((L), lunatik_malloc((L), (s))))
 
+static inline const char *lunatik_errname(int err)
+{
+#ifdef CONFIG_SYMBOLIC_ERRNAME
+	return errname(err);
+#else
+	return NULL;
+#endif
```
lneto: I think it's already covered by the kernel header.. https://elixir.bootlin.com/linux/v6.6/source/include/linux/errname.h#L10

as @marcelstanley pointed out, the real problem is that it's not being exported on kernel < 6.7

lib/lualinux.c
```diff
@@ -569,10 +422,31 @@ static const lunatik_reg_t lualinux_signal[] = {
...
+* @usage
+* local name = linux.errname(2)
+* if name then print("Error name:", name) end -- "ENOENT"
+*/
+static int lualinux_errname(lua_State *L)
+{
+	int e = (int)luaL_checkinteger(L, 1);
+	const char *n = lunatik_errname(e);
+	if (n)
+		lua_pushstring(L, n);
```
lneto: what happens if we just push a NULL here and avoid the code bellow?

lunatik.h
marcelstanley: Indeed, I've just stumbled upon it while building lunatik on openwrt-24.10. 

So, for kernel < 6.7, I believe we could go with their suggestion on https://github.com/torvalds/linux/commit/21db931445d85a14c0a64a9d54c7617a8c6e5762:
> this functionality is
otherwise only available for error pointers via %pE - bcachefs uses this
for better error messages.

abhijeetw035: Hii , so should I make the check version specific to handle the export issue in older kernels? 
something like: `#if defined(CONFIG_SYMBOLIC_ERRNAME) && LINUX_VERSION_CODE >= KERNEL_VERSION(6, 7, 0)`

This would ensure compatibility for users on kernels < 6.7 where errname isn't exported

lib/lualinux.c
abhijeetw035: oh means we will simply return n ? 1 : 0; at the end right ?

lneto: nop, I meant to check what is pushed on pushstring(L, NULL).. is it an empty string? nil? does it explode? Also, what's the best error treatment here? is it returning nil? should we raise an error? perhaps it's better to return "unknown". Think about the use case case; the scripts that use such API.

lunatik.h
lneto: I don't think we should care about `CONFIG_SYMBOLIC_ERRNAME` as the kernel already has a fall back for it. I think we should use the string modifier @marcelstanley mentioned in the case of < 6.7. Did you check the reference he shared?

lib/lualinux.c
abhijeetw035: you're right, `lua_pushstring(L, NULL)` would cause a crash , since the Lua C API expects a valid pointer. we can return the original error number as a fallback if the symbolic name is 'unknown' or cannot be resolved.

lib/lualinux.c
```diff
@@ -569,10 +422,31 @@ static const lunatik_reg_t lualinux_signal[] = {
...
+* if name then print("Error name:", name) end -- "ENOENT"
+*/
+static int lualinux_errname(lua_State *L)
+{
+	int e = (int)luaL_checkinteger(L, 1);
+	const char *n = lunatik_errname(e);
+	if (n && *n != '\0' && strcmp(n, "unknown") != 0)
+		lua_pushstring(L, n);
+	else
+		lua_pushinteger(L, e);
```
lneto: it wasn't what I meant.. I meant something like this..
```suggestion
	const char *n = lunatik_errname(e);
	lua_pushstring(L, n != NULL ? n : "unknown");
```

my point is to think about its usage.. we would typically use such API to print or take decision based on the error name corresponding to an error code, right? why would we would like to get the error code returned back? IMO, we would like just a clean indication that the error code isn't defined. What's the better way to do it? nil? ""? "unknown"? error()? I'm not sure. IMO, we need to look to the other side (the API client) instead of rush it.

lunatik.h
```diff
@@ -147,10 +128,21 @@ static inline void *lunatik_checknull(lua_State *L, void *ptr)
 
 #define lunatik_checkalloc(L, s)	(lunatik_checknull((L), lunatik_malloc((L), (s))))
 
+static inline const char *lunatik_errname(int err)
+{
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 7, 0)
+	return errname(err);
+#else
+	static char buf[32];
```
lneto: why 32?

lunatik.h
```diff
@@ -147,10 +128,21 @@ static inline void *lunatik_checknull(lua_State *L, void *ptr)
...
+	snprintf(buf, sizeof(buf), "%pE", ERR_PTR(-err));
+	return buf;
+#endif
```
```suggestion
#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 7, 0)
#define lunatik_errname(e)	errname(e)
#else
static inline const char *lunatik_errname(int err)
{
	static char buf[32];
	snprintf(buf, sizeof(buf), "%pE", ERR_PTR(-err));
	return buf;
}
#endif
```

lunatik.h
```diff
@@ -147,10 +128,21 @@ static inline void *lunatik_checknull(lua_State *L, void *ptr)
...
 #define lunatik_checkalloc(L, s)	(lunatik_checknull((L), lunatik_malloc((L), (s))))
 
+static inline const char *lunatik_errname(int err)
+{
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 7, 0)
+	return errname(err);
+#else
+	static char buf[32];
+	snprintf(buf, sizeof(buf), "%pE", ERR_PTR(-err));
+	return buf;
```
lneto: you shouldn't do that; if we have different threads calling this concurrently it will corrupt the memory.

lunatik.h
```diff
@@ -147,10 +128,21 @@ static inline void *lunatik_checknull(lua_State *L, void *ptr)
 
 #define lunatik_checkalloc(L, s)	(lunatik_checknull((L), lunatik_malloc((L), (s))))
 
+static inline const char *lunatik_errname(int err)
+{
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 7, 0)
+	return errname(err);
+#else
+	static char buf[32];
+	snprintf(buf, sizeof(buf), "%pE", ERR_PTR(-err));
```
lneto: I would do this directly in the lualinux_errname binding.. pushing it to the lua stack, where it will be copied instead of using a static (Oo) var for this.

lunatik.h
abhijeetw035: i chose 32  as safe upper bound since standard error names are generally shorter e.g `EADDRNOTAVAIL` is 13 chars and `ENOTRECOVERABLE` is of 15chars so....

lib/lualinux.c
abhijeetw035: Got it.  make sense return type consistent makes things much safer for the Lua side. I’ll add the "unknown" fallback.

abhijeetw035: so i will update the header in lunatik.h to simply return NULL for older kernels (instead of using the static buffer there). this forces the logic to fall through to the binding below.

since I am moving this block into lualinux.c to use a thread safe local buffer, I am thinking implement the fallback logic like this in the else block ( please correct me if i am wrong):
lualinux.c
```c 
char buf[32];
snprintf(buf, sizeof(buf), "%pE", ERR_PTR(-e));

/* %pE writes the error number e.g "-22" if the name is not found
  We detect this case to ensure we return "unknown" instead of the number string,
  keeping the API consistent.
 */
if (buf[0] == '-' || (buf[0] >= '0' && buf[0] <= '9'))  
    lua_pushstring(L, "unknown");
else
    lua_pushstring(L, buf);

Comment (abhijeetw035) [reply lib/lualinux.c]: lunatik.h something like this

```c
#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 7, 0)
#define lunatik_errname(e)	errname(e)
#else
static inline const char *lunatik_errname(int err)
{
	return NULL;
}
#endif

File: lunatik.h (hunk 1 above)
Comment (carloslack) [reply lunatik.h]: Notice that `errname` was only introduced in kernel `v5.5` , prior to that maybe you'll have to stick with static errno.

File: lunatik.h (hunk 4 above)
Comment (lneto) [reply lunatik.h]: I would add a define (e.g., MAXERR) or use something as BUFSIZE

File: lunatik.h (hunk 1 above)
Comment (lneto) [reply lunatik.h]: actually, even on [5.5, 6.7) we cannot use it because it's exported to kmods..

File: lib/lualinux.c (hunk 3 above)
Comment (lneto) [reply lib/lualinux.c]: > lunatik.h something like this
> 
> #if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 7, 0)
> #define lunatik_errname(e)	errname(e)
> #else
> static inline const char *lunatik_errname(int err)
> {
> 	return NULL;
> }
> #endif

why not just adding the ifdef to the binding implementation? IMO, it would be simpler.

Comment (lneto) [reply lib/lualinux.c]: > so i will update the header in lunatik.h to simply return NULL for older kernels (instead of using the static buffer there). this forces the logic to fall through to the binding below.
> 
> since I am moving this block into lualinux.c to use a thread safe local buffer, I am thinking implement the fallback logic like this in the else block ( please correct me if i am wrong): lualinux.c
> 
> char buf[32];
> snprintf(buf, sizeof(buf), "%pE", ERR_PTR(-e));
> 
> /* %pE writes the error number e.g "-22" if the name is not found
>   We detect this case to ensure we return "unknown" instead of the number string,
>   keeping the API consistent.
>  */
> if (buf[0] == '-' || (buf[0] >= '0' && buf[0] <= '9'))  
>     lua_pushstring(L, "unknown");
> else
>     lua_pushstring(L, buf);

I wouldn't add such complex logic here.. if we want to handle both negative and positive, I would just use mod operation. I definitely wouldn't parse the numeral format here.

Comment (abhijeetw035) [reply lib/lualinux.c]: > > lunatik.h something like this
> > #if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 7, 0)
> > #define lunatik_errname(e)	errname(e)
> > #else
> > static inline const char *lunatik_errname(int err)
> > {
> > return NULL;
> > }
> > #endif
> 
> why not just adding the ifdef to the binding implementation? IMO, it would be simpler.

Yeah, keeping it in one place is cleaner 
I’ll implement it in lualinux.c like this: (please correct me if i am wrong)
```c
#define MAXERR 32

static int lualinux_errname(lua_State *L)
{
    int e = (int)luaL_checkinteger(L, 1);
    if (e < 0)  // handle both positive/negative inputs 
        e = -e;

#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 7, 0)
    const char *n = errname(e);
    lua_pushstring(L, n ? n : "unknown");
#else
    char buf[MAXERR];
    snprintf(buf, sizeof(buf), "%pE", ERR_PTR(-e));
    lua_pushstring(L, buf);
#endif

    return 1;
}

Comment (lneto) [reply lib/lualinux.c]: yeah.. I'm thinking on something like this.. just prefix MAXERR with LUALINUX_, please.. also, we probably have a module macro somewhere.. just use it instead of recreating it ;-).. one more thing.. please, avoid comments like this.. if our code for module isn't clear enough, we should rewrite it (instead of adding too much natural language to the code base ;-).. thanks!

Comment (abhijeetw035) [reply lib/lualinux.c]: so I checked lunatik_conf.h for an existing macro and found LUAL_BUFFERSIZE, but it is defined as 256 bytes. since kernel error names are very short  allocating 256 bytes on the stack seems excessive for this specific fallback. if you prefer consistency over stack usage we can go with LUAL_BUFFERSIZE. 

https://github.com/luainkernel/lunatik/blob/81a178547ff7faa816d0548e2dbf9f1a11db79c6/lunatik_conf.h#L51C9-L51C24

Comment (lneto) [reply lib/lualinux.c]: what prejudice you see by allocating 256 bytes on the stack? I would just reuse this definition; I don't see any constrain on using 256 bytes on the activation frame of this function. Let's always keep the first rule of optimization in mind (while contributing to Lunatik ;-).

btw, did you find the module macro?

Comment (abhijeetw035) [reply lib/lualinux.c]: sure will do the required. 

yes i found it , `LUAL_BUFFERSIZE` was the macro I found in lunatik_conf.h this was the one you were referring to right ? 
also attached permalink in previous message 
-> https://github.com/luainkernel/lunatik/blob/81a178547ff7faa816d0548e2dbf9f1a11db79c6/lunatik_conf.h#L51C9-L51C24

Comment (lneto) [reply lib/lualinux.c]: Sorry, I meant `abs` instead of `modulo`, my brain is a bit off with this port of v5.5.. heh.. I meant to replace that `if e < 0 ...` logic with `abs(e)`.

Comment (abhijeetw035) [reply lib/lualinux.c]: haha :)) yes will add it in the commit. thanks

**Comment by abhijeetw035** [comment]
> Please change the file mode back to 644. Other than this, LGTM.

hii !! I checked my local git index (git ls-files -s) and it shows 100644. i tried to force push but it shows everything is up to date. please let me know if you still see the wrong mode on your end.

image

**Comment by marcelstanley** [comment]
> > Please change the file mode back to 644. Other than this, LGTM.
> 
> hii !! I checked my local git index (git ls-files -s) and it shows 100644. i tried to force push but it shows everything is up to date. please let me know if you still see the wrong mode on your end.

Sorry, perhaps it was something on my end. You're right. Nevermind.

**Comment by marcelstanley** [comment]
On the other hand, @abhijeetw035, I'm having trouble compiling lunatik with your patch on kernel 6.6.110 both on OpenWrt and Ubuntu. Here's the error:
```
  MODPOST /home/marcel/lunatik/Module.symvers
ERROR: modpost: "errname" [/home/marcel/lunatik/lib/luasocket.ko] undefined!
ERROR: modpost: "errname" [/home/marcel/lunatik/lib/luacrypto_shash.ko] undefined!
ERROR: modpost: "errname" [/home/marcel/lunatik/lib/luacrypto_skcipher.ko] undefined!
ERROR: modpost: "errname" [/home/marcel/lunatik/lib/luacrypto_aead.ko] undefined!
ERROR: modpost: "errname" [/home/marcel/lunatik/lib/luacrypto_rng.ko] undefined!
ERROR: modpost: "errname" [/home/marcel/lunatik/lib/luacrypto_comp.ko] undefined!
make[3]: *** [scripts/Makefile.modpost:145: /home/marcel/lunatik/Module.symvers] Error 1
make[2]: *** [/usr/src/linux-headers-6.6.110-0606110-generic/Makefile:1883: modpost] Error 2
make[1]: *** [Makefile:234: __sub-make] Error 2
make[1]: Leaving directory '/usr/src/linux-headers-6.6.110-0606110-generic'
```

See [full logs](https://github.com/user-attachments/files/24396935/lunatik_failure_kernel_6.6.110.txt).

lneto:
> On the other hand, @abhijeetw035, I'm having trouble compiling lunatik with your patch on kernel 6.6.110 both on OpenWrt and Ubuntu. Here's the error:
> 
> ```
>   MODPOST /home/marcel/lunatik/Module.symvers
> ERROR: modpost: "errname" [/home/marcel/lunatik/lib/luasocket.ko] undefined!
> ERROR: modpost: "errname" [/home/marcel/lunatik/lib/luacrypto_shash.ko] undefined!
> ERROR: modpost: "errname" [/home/marcel/lunatik/lib/luacrypto_skcipher.ko] undefined!
> ERROR: modpost: "errname" [/home/marcel/lunatik/lib/luacrypto_aead.ko] undefined!
> ERROR: modpost: "errname" [/home/marcel/lunatik/lib/luacrypto_rng.ko] undefined!
> ERROR: modpost: "errname" [/home/marcel/lunatik/lib/luacrypto_comp.ko] undefined!
> make[3]: *** [scripts/Makefile.modpost:145: /home/marcel/lunatik/Module.symvers] Error 1
> make[2]: *** [/usr/src/linux-headers-6.6.110-0606110-generic/Makefile:1883: modpost] Error 2
> make[1]: *** [Makefile:234: __sub-make] Error 2
> make[1]: Leaving directory '/usr/src/linux-headers-6.6.110-0606110-generic'
> ```
> 
> See [full logs](https://github.com/user-attachments/files/24396935/lunatik_failure_kernel_6.6.110.txt).

it's not related to this patch, right.. though, it can be fixed on this PR as well.. it was actually the original proposal of @abhijeetw035.. I somehow mislead him on this.. sorry about that.. but the code quality is better now anyway, IMO.. and we can just move it as is to a `luanatik_pusherrname` API and use it both on `lunatik_tryret` and on `lualinux_errname`. What do you both think?

marcelstanley:
> it's not related to this patch, right.. though, it can be fixed on this PR as well.. it was actually the original proposal of @abhijeetw035.. I somehow mislead him on this.. sorry about that.. but the code quality is better now anyway, IMO.. and we can just move it as is to a `luanatik_pusherrname` API and use it both on `lunatik_tryret` and on `lualinux_errname`. What do you both think?

In fact, it's not related. My bad. And I agree with your suggestion, @lneto, we can do it here.

lneto:
@marcelstanley @abhijeetw035 I'll merge this then I'll provide the fix for tryret on another PR

abhijeetw035:
> @marcelstanley @abhijeetw035 I'll merge this then I'll provide the fix for tryret on another PR

Sure! I agree with both handling it in this PR is and fixing in a separate PR as well... I’d be happy to work on that. Thanks for the guidance!

lneto:
> > @marcelstanley @abhijeetw035 I'll merge this then I'll provide the fix for tryret on another PR
> 
> Sure! I agree with both handling it in this PR is and fixing in a separate PR as well... I’d be happy to work on that. Thanks for the guidance!

ack.. so, let's do it on another PR.. and get things going ;-)

PR #347, opened 2025-12-25
clean up luahid

lneto:
* add lunatik_monitor
* remove unecessary branch on error handling
* simplify error levels
* fix style

### DIFF
--- a/lib/luahid.c
+++ b/lib/luahid.c
@@ -170,7 +170,10 @@ static inline lunatik_object_t *luahid_getdescriptor(lua_State *L, luahid_t *hid
 	return data;
 }
 
-typedef struct hid_device_id luahid_probe_arg_t;
+typedef struct {
+	const struct hid_device_id *id;
+	int ret;
+} luahid_probe_arg_t;
 
 typedef struct {
 	__u8 *rdesc;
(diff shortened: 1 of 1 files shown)


lib/luahid.c
```diff
@@ -226,48 +230,47 @@ static void luahid_postreport_fixup(lua_State *L, struct hid_device *hdev, luahi
 static void luahid_postraw_event(lua_State *L, struct hid_device *hdev, luahid_raw_event_arg_t *arg)
 {
 	lunatik_object_t *data = luahid_getdescriptor(L, luahid_gethid(hdev));
+
 	if (data)
 		luadata_clear(data);
-	if (!lua_isboolean(L, -1))
-		lua_pushinteger(L, -EINVAL);
-	else
-		lua_pushinteger(L, 0);
```
lneto, line 234: @qrsikno2 why should we return 0 if it's not error?

lib/luahid.c
```diff
@@ -226,48 +230,47 @@ static void luahid_postreport_fixup(lua_State *L, struct hid_device *hdev, luahi
...
+		lua_pushinteger(L, -ECANCELED);					\
+		return 1;		                   			\
+	}									\
+										\
+	luahid_post##NAME(L, hdev, arg);					\
+	return NRET;                                                           	\
 }
 
-LUAHID_CALLBACK(probe, 2, 1, err);
-LUAHID_CALLBACK(report_fixup, 4, 0, warn);
```
lneto, line 269: @qrsikno2 there was any reason to use warn here?

lib/luahid.c
```diff
@@ -56,6 +56,7 @@ static const luaL_Reg luahid_lib[] = {
 };
 
 static const luaL_Reg luahid_mt[] = {
+	{"__index", lunatik_monitorobject},
```
lneto, line 59: @qrsikno2 there was any reason to not use monitor here?

lib/luahid.c
```diff
@@ -226,48 +230,47 @@ static void luahid_postreport_fixup(lua_State *L, struct hid_device *hdev, luahi
...
+static int luahid_do##NAME(lua_State *L)					\
+{										\
+	struct hid_device *hdev = (struct hid_device *)lua_touserdata(L, 1);	\
+	luahid_##NAME##_arg_t *arg = LUAHID_ARG(L, NAME, 2);			\
+	luahid_t *hid = luahid_gethid(hdev);					\
+										\
+	if (luahid_checkdriver(L, hid, -1)) {					\
+		pr_err(#NAME ": invaild driver\n");				\
+		lua_pushinteger(L, -ENXIO);					\
+		return 1;		                   			\
```
lneto, line 286: @qrsikno2 please check if it will make any difference.. from Lua perspective, it will just discard the return value if there's no assignment

lneto:
@qrsikno2 I noticed that we use both lunatik_run and lunatik_runirq to run different callbacks.. is that guaranteed that the driver will not call them concurrently?

lib/luahid.c
```diff
@@ -226,48 +230,47 @@ static void luahid_postreport_fixup(lua_State *L, struct hid_device *hdev, luahi
...
 	lunatik_object_t *data = luahid_getdescriptor(L, luahid_gethid(hdev));
+
 	if (data)
 		luadata_clear(data);
-	if (!lua_isboolean(L, -1))
-		lua_pushinteger(L, -EINVAL);
-	else
-		lua_pushinteger(L, 0);
+
+	lua_pushinteger(L, !lua_isboolean(L, -1) ? EINVAL : 0);
```
carloslack: Maybe you want `-EINVAL` here

lneto: good catch! thanks!

glk0 (commented):
Nothing looks bad to me.

lneto:
#366

PR #349, opened 2025-12-26
Refactor luahid callbacks with function pointers

### LINKED ISSUE #344: modularize properly luahid callbacks
instead of base this on the callback name, suffixing it with [pre](https://github.com/luainkernel/lunatik/blob/master/lib/luahid.c#L256) and [post](https://github.com/luainkernel/lunatik/blob/master/lib/luahid.c#L264), use a proper data structure with function pointers. It might be a good idea to split these callbacks in separate files depending on their complexity.

sneaky-potato:
Fixes #344

### DIFF
--- a/lib/luahid.c
+++ b/lib/luahid.c
@@ -170,7 +170,10 @@ static inline lunatik_object_t *luahid_getdescriptor(lua_State *L, luahid_t *hid
 	return data;
 }
 
-typedef struct hid_device_id luahid_probe_arg_t;
+typedef struct {
+	const struct hid_device_id *id;
+	int ret;
+} luahid_probe_arg_t;
 
 typedef struct {
 	__u8 *rdesc;
@@ -181,61 +184,94 @@ typedef struct {
 	struct hid_report *report;
 	u8 *data;
 	int size;
+	int ret;
 } luahid_raw_event_arg_t;
 
 #define LUAHID_ARG(L, name, ix)	((luahid_##name##_arg_t *)lua_touserdata(L, ix))
 
-static void luahid_preprobe(lua_State *L, struct hid_device *hdev, luahid_probe_arg_t *arg)
+static void luahid_preprobe(lua_State *L, struct hid_device *hdev, void *arg)
 {
-	luahid_newtable(L, arg, driver_data);
+	luahid_probe_arg_t *probe_arg = arg;
+	luahid_newtable(L, probe_arg->id, driver_data);
 }
 
-static void luahid_prereport_fixup(lua_State *L, struct hid_device *hdev, luahid_report_fixup_arg_t *arg)
+static void luahid_prereport_fixup(lua_State *L, struct hid_device *hdev, void *arg)
 {
 	luahid_pushhdev(L, hdev);
 	luahid_pushinfo(L, -4, hdev);
 	lunatik_object_t *data = luahid_getdescriptor(L, luahid_gethid(hdev));
+	luahid_report_fixup_arg_t *report_fixup_arg = arg;
 	if (data)
-		luadata_reset(data, arg->rdesc, arg->rsize, LUADATA_OPT_NONE);
+		luadata_reset(data, report_fixup_arg->rdesc, report_fixup_arg->rsize, LUADATA_OPT_NONE);
 }
 
-static void luahid_preraw_event(lua_State *L, struct hid_device *hdev, luahid_raw_event_arg_t *arg)
+static void luahid_preraw_event(lua_State *L, struct hid_device *hdev, void *arg)
 {
 	luahid_pushhdev(L, hdev);
 	luahid_pushinfo(L, -4, hdev);
-	luahid_pushreport(L, arg->report);
+	luahid_raw_event_arg_t *raw_event_arg = arg;
+	luahid_pushreport(L, raw_event_arg->report);
 	lunatik_object_t *data = luahid_getdescriptor(L, luahid_gethid(hdev));
 	if (data)
-		luadata_reset(data, arg->data, arg->size, LUADATA_OPT_NONE);
+		luadata_reset(data, raw_event_arg->data, raw_event_arg->size, LUADATA_OPT_NONE);
 }
 
-static void luahid_postprobe(lua_State *L, struct hid_device *hdev, luahid_probe_arg_t *arg)
+static void luahid_postprobe(lua_State *L, struct hid_device *hdev, void *arg)
 {
+	luahid_probe_arg_t *probe_arg = arg;
 	if (lua_istable(L, -1)) {
 		lua_pushlightuserdata(L, hdev);
 		lua_pushvalue(L, -2);
 		lua_settable(L, -4);
 		hid_set_drvdata(hdev, luahid_gethid(hdev));
+		probe_arg->ret = 0;
 	}
-	lua_pushinteger(L, 0);
 }
 
-static void luahid_postreport_fixup(lua_State *L, struct hid_device *hdev, luahid_report_fixup_arg_t *arg)
+static void luahid_postreport_fixup(lua_State *L, struct hid_device *hdev, void *arg)
 {
 	lunatik_object_t *data = luahid_getdescriptor(L, luahid_gethid(hdev));
 	if (data)
 		luadata_clear(data);
 }
 
-static void luahid_postraw_event(lua_State *L, struct hid_device *hdev, luahid_raw_event_arg_t *arg)
+static void luahid_postraw_event(lua_State *L, struct hid_device *hdev, void *arg)
 {
+	luahid_raw_event_arg_t *raw_event_arg = arg;
 	lunatik_object_t *data = luahid_getdescriptor(L, luahid_gethid(hdev));
 
 	if (data)
 		luadata_clear(data);
 
-	lua_pushinteger(L, !lua_isboolean(L, -1) ? -EINVAL : 0);
+	if (!lua_isboolean(L, -1)) {
+		raw_event_arg->ret = -EINVAL;
+		return;
+	}
+
+	raw_event_arg->ret = lua_toboolean(L, -1) ? 1 : 0;
 }
+
+typedef void (*luahid_cb_t)(lua_State *L, struct hid_device *hdev, void *arg);
+
+struct luahid_cb_ops {
+	luahid_cb_t pre;
+	luahid_cb_t post;
+};
+
+static const struct luahid_cb_ops luahid_probe_ops = {
+    .pre  = luahid_preprobe,
+    .post = luahid_postprobe,
+};
+
+static const struct luahid_cb_ops luahid_report_fixup_ops = {
+    .pre  = luahid_prereport_fixup,
+    .post = luahid_postreport_fixup,
+};
+
+static const struct luahid_cb_ops luahid_raw_event_ops = {
+    .pre  = luahid_preraw_event,
+    .post = luahid_postraw_event,
+};
 
 #define LUAHID_CALLBACK(NAME, NARG, NRET)					\
 static int luahid_do##NAME(lua_State *L)					\
@@ -245,7 +281,7 @@ static int luahid_do##NAME(lua_State *L)					\
 	luahid_t *hid = luahid_gethid(hdev);					\
 										\
 	if (luahid_checkdriver(L, hid, -1)) {					\
-		pr_err(#NAME ": invaild driver\n");				\
+		pr_err(#NAME ": invalid driver\n");				\
 		lua_pushinteger(L, -ENXIO);					\
 		return 1;		                   			\
 	}									\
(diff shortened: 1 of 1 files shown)


lneto:
thanks, @sneaky-potato! can you rebase your PR on lneto_hid? we are using it as a integration branch.. than we can merge this there before going to master

lib/luahid.c
```diff
@@ -234,7 +234,27 @@ static void luahid_postraw_event(lua_State *L, struct hid_device *hdev, luahid_r
 		lua_pushinteger(L, 0);
 }
 
-#define LUAHID_CALLBACK(NAME, NARG, NRET, LEVEL)							\
+struct luahid_cb_ops {
+	void (*pre)(lua_State *L, struct hid_device *hdev, void *arg);
```
lneto: we can use a typedef for this prototype.

lib/luahid.c
```diff
@@ -253,21 +273,23 @@ static int luahid_do##NAME(lua_State *L)								\
 	}                                                                      				\
 													\
 	lua_pushvalue(L, -3);                                                  				\
-	luahid_pre##NAME(L, hdev, arg);                               					\
+	if ((OPS) && (OPS)->pre)                               					\
```
lneto: I would avoid this verification, as we require both callbacks to be defined for all cases, right?

lib/luahid.c
lneto: btw, you also need to change the signature of the actual callbacks to receive `void *` and then typecast it in the proper type

lneto: btw, arg doesn't seem to be used on any post function, we can get rid of it, right? and have two diff callback types.. in this case, I think we don't need to bother with the typedef

lneto: hm.. looking again at the code.. perhaps it's a good opportunity to use the arg structure to pass back the return value instead of the lua stack, getting rid of those "pushinteger"s on post*.. in that case, I would keep the typedef.. I think that makes more sense

lib/luahid.c
sneaky-potato: Resolved in latest commit

lib/luahid.c
sneaky-potato: Good point. I agree that using arg as in/out state would clean up the return handling and avoid stack-based return values. I’ll rework the callbacks to store return values in the arg structs and keep a single callback typedef.

lib/luahid.c
```diff
@@ -181,63 +184,103 @@ typedef struct {
...
 		lua_pushlightuserdata(L, hdev);
 		lua_pushvalue(L, -2);
 		lua_settable(L, -4);
 		hid_set_drvdata(hdev, luahid_gethid(hdev));
+		probe_arg->ret = 0;
+	} else if (lua_isinteger(L, -1)) {
+		probe_arg->ret = lua_tointeger(L, -1);
+	} else {
+		probe_arg->ret = 0;
 	}
```
lneto, line 228: I'm lost here.. why this? isn't the case of just `probe_arg->ret = 0;`?

lib/luahid.c
```diff
@@ -181,63 +184,103 @@ typedef struct {
...
 
 	if (data)
 		luadata_clear(data);
-
-	lua_pushinteger(L, !lua_isboolean(L, -1) ? EINVAL : 0);
+	if (lua_isboolean(L, -1)) {
+		raw->handled = lua_toboolean(L, -1);
+		raw_event_arg->ret = 0;
+	} else {
+		raw->handled = false;
```
lneto: why? isn't it enough to `raw_event_arg->ret = !lua_isboolean(L, -1) ? EINVAL : 0`?

sneaky-potato: I noticed that a boolean is used for returning in `luahid_runraw_event` [here](https://github.com/luainkernel/lunatik/blob/bc91085c2cb833c3784dddb8854860b735778d0b/lib/luahid.c#L317C1-L323C2). For setting this variable, I added this additional bool member `handled` (whether the event was handled is communicated via this boolean). Can this be done without this bool?

lneto: Yes, this code is a bit odd.. I would take a look how the examples are using this contract and just simplify it.. it seems to me that an int (not sure if a bool) would do the job..

lneto: Btw, it would be good if you could try to run the gestures example..

lib/luahid.c
sneaky-potato: Refactored the control flow.

lib/luahid.c
lneto: it seems that `ret` is not used here. Why can't we get rid of this `ret_bool`, having only `ret`?

sneaky-potato: Pushed a change which removes `ret_bool` and used `ret` in post*, and run* methods for raw_event.

lneto: Btw, did you manage to run gestures.lua? I'm having some trouble with it..

sneaky-potato: I noticed the code was not compiling, pushed another change, it should compile now. I tried writing a very simple hid script (based on gesture.lua) to test on my host machine because I am having some trouble with qemu. I will revert back once I manage to run my script or example itself.

lib/luahid.c
```diff
@@ -181,63 +184,101 @@ typedef struct {
...
 } luahid_raw_event_arg_t;
 
 #define LUAHID_ARG(L, name, ix)	((luahid_##name##_arg_t *)lua_touserdata(L, ix))
 
-static void luahid_preprobe(lua_State *L, struct hid_device *hdev, luahid_probe_arg_t *arg)
+static void luahid_preprobe(lua_State *L, struct hid_device *hdev, void *arg)
 {
-	luahid_newtable(L, arg, driver_data);
+	luahid_probe_arg_t *probe_arg = arg;
+	probe_arg->ret = 0;
```
lneto: why setting this to 0 here and setting it again to 0 on post, unconditionally? It looks like it's not used at all.

lib/luahid.c
```diff
@@ -181,63 +184,101 @@ typedef struct {
...
 	if (lua_istable(L, -1)) {
 		lua_pushlightuserdata(L, hdev);
 		lua_pushvalue(L, -2);
 		lua_settable(L, -4);
 		hid_set_drvdata(hdev, luahid_gethid(hdev));
+		return;
 	}
-	lua_pushinteger(L, 0);
+	if (lua_isinteger(L, -1))
+		probe_arg->ret = lua_tointeger(L, -1);
```
lneto: I don't follow this.. the original code was passing 0 through the Lua stack.. why are you fetching an integer from the top of the stack? what's supposed to be there?

lib/luahid.c
sneaky-potato: I was able to run a simple script detecting left and right swipes using a USB mouse on the host machine. The example given does not unlock the mouse for some reason (lunatik using my branch), will debug further.

lib/luahid.c
sneaky-potato: I assume Lua returns values, C interprets them once and stores them in `arg`? Isn't the return value supposed to be on top of stack at this point?

sneaky-potato: Should `probe()` in Lua be allowed to return an error code?

I believe this refactor has changed the contract a bit.
With this change (checking top of stack) interpreting error code is now possible
```lua
function driver:probe(devid)
    if devid.vendor ~= 0x1234 then
        return -ENODEV
    end
    return {}
end
```

But i guess for the old contract to remain as is, we need to remove this check part (probe will unconditionally succeed in that case, right?)

lneto: > I assume Lua returns values, C interprets them once and stores them in `arg`?

I don't follow. What you mean as "stores them in `arg`"? Which `arg`?

> Isn't the return value supposed to be on top of stack at this point?

Yes, Lua calls returns values on the top of the stack. However, what's the function call you are referring here? The probe callback? isn't it returning a table?

lneto: > Should `probe()` in Lua be allowed to return an error code?
> 
> I believe this refactor has changed the contract a bit. With this change (checking top of stack) interpreting error code is now possible
> 
> function driver:probe(devid)
>     if devid.vendor ~= 0x1234 then
>         return -ENODEV
>     end
>     return {}
> end
> But i guess for the old contract to remain as is, we need to remove this check part (probe will unconditionally succeed in that case, right?)

It's a good question, actually. Can you take a look at other Linux HID's drivers and check what's the common pattern?

sneaky-potato: Is [this](https://elixir.bootlin.com/linux/v6.13.5/source/drivers/hid/hid-logitech-hidpp.c#L4363) a valid example? I can see `-ENOMEM` being returned from the probe in this case.

lneto: It's but ENOMEM hasn't much usage from the Lua scripts themselves as they don't manage memory by their own; it could be returned by the binding though. However, I think you are right anyway. There're usage for returning error codes here. See:

https://github.com/torvalds/linux/blob/c8ebd433459bcbf068682b09544e830acd7ed222/drivers/base/dd.c#L574C12-L574C29

and 

https://elixir.bootlin.com/linux/v6.18.2/source/drivers/base/core.c#L5030

I'm just not sure there's real usage for us to return `EPROBE_DEFER`. Perhaps, it would be better to just return `-ENODEV` in the binding if the script doesn't return a table. What do you think?

sneaky-potato: I agree. Allowing arbitrary errno values from Lua is probably unnecessary at this stage, and `-EPROBE_DEFER` doesn’t seem meaningful for HID scripts. Let’s define the contract such that if the Lua `probe()` does not return a table, the binding returns `-ENODEV` and the driver does **not** bind.

Essentially, Lua returns data (table) or no data, C code decides the kernel-visible error (`-ENODEV` in this case).

lib/luahid.c
sneaky-potato: Resolved in latest commit, now setting `ret` member only in post methods.

lib/luahid.c
```diff
@@ -256,34 +294,40 @@ static int luahid_do##NAME(lua_State *L)					\
...
+	return arg->ret;
 }
 
 static int luahid_probe(struct hid_device *hdev, const struct hid_device_id *id)
 {
 	luahid_t *hid = luahid_gethid(hdev);
 	int ret;
 
-	lunatik_run(hid->runtime, luahid_runprobe, ret, hdev, (luahid_probe_arg_t *)id);
+	luahid_probe_arg_t probe_arg;
```
```suggestion
	luahid_probe_arg_t probe_arg = {.id = id, .ret = -ENODEV};
```

lneto: then, we just need to flip it zero if we get a table.. simplifying the flow..

sneaky-potato: Resolved in latest commit

lib/luahid.c
```diff
@@ -256,34 +293,38 @@ static int luahid_do##NAME(lua_State *L)					\
 	}                                                              		\
 										\
 	lua_pushvalue(L, -3);                                          		\
-	luahid_pre##NAME(L, hdev, arg);                         		\
+	if ((OPS)->pre)                                         		\
```
lneto: I would just assume by contract that _ops must be there and use ##NAME to retrieve it here..

lib/luahid.c
```diff
@@ -409,7 +448,7 @@ static int luahid_register(lua_State *L)
 
 	lunatik_checkfield(L, 1, "id_table", LUA_TTABLE);
 	luaL_argcheck(L, (user_driver->id_table = luahid_setidtable(L, -1)) != NULL,
-		      2, "invaild id_table");
+		      2, "invalid id_table");
```
lneto, line 448: it's not clear what you are changing here

lib/luahid.c
```diff
@@ -181,71 +184,105 @@ typedef struct {
...
+static void luahid_postprobe(lua_State *L, struct hid_device *hdev, void *arg)
 {
+	luahid_probe_arg_t *probe_arg = arg;
 	if (lua_istable(L, -1)) {
 		lua_pushlightuserdata(L, hdev);
 		lua_pushvalue(L, -2);
 		lua_settable(L, -4);
 		hid_set_drvdata(hdev, luahid_gethid(hdev));
+		probe_arg->ret = 0;
+		return;
```
lneto: ?

sneaky-potato: We flip the ret value to 0 in case a table is found, this was discussed [here](https://github.com/luainkernel/lunatik/pull/349#discussion_r2655463371). 
`probe_arg->ret` is initialized to `-ENODEV`, and success is defined as the Lua probe() returning a table.
Am I missing something?

lib/luahid.c
sneaky-potato: there’s no functional change here. This hunk only fixes a typo in the error message string ("invaild" → "invalid").
I can split this into a separate commit if you prefer.

lib/luahid.c
lneto: but why do we still need this return statement?

sneaky-potato:
Hi @lneto, sure I am fine with this idea, I'll focus on my other ongoing PR

lib/luahid.c
sneaky-potato: Yeah, that return is not requires, resolved in latest commit

lib/luahid.c
lneto: Interesting.. it's calling raw_event?

sneaky-potato: Yes I was able to generate logs on swipes. But I used my own USB mouse, updated the vender and product ids accordingly.

PR #351, opened 2025-12-27
Add signal functions (modify, pending)

Physic69:


lib/lualinux.c
```diff
@@ -117,6 +117,71 @@ static int lualinux_schedule(lua_State *L)
...
 }
 
+/***
+* Blocks/unblocks signal for current task.
+*
+* @function block
+* @tparam integer sig Signal number to send ('linux.signal.KILL`).
+* @within linux
+*/
+static int lualinux_block(lua_State *L)
```
Physic69: Should I merge both the functions block and allow, where 
function having 2 params, one telling signal number & another for telling allow/block..

lneto: Yeah, I think it's the best approach =)

Physic69: Just one thing:
if we want the SETMASK in our code, i just added for now just in case we need that..
else ill just revert back to [No sigmask](https://github.com/luainkernel/lunatik/pull/351/changes/6f55358383ea15e4d7b76c74737013edbf218014#diff-ebaea58d0c04910938024880228d50c76d071c363b631259cef65a6c4cbdf5feR135)

lneto: I think it's overly complex..I would rather have simpler bindings and handle more advanced logic in Lua itself. I would just export lower-level functions to Lua and then writing this complex logic in signal.lua.

lib/lualinux.c
```diff
@@ -117,6 +117,53 @@ static int lualinux_schedule(lua_State *L)
...
+* @tparam integer sig Signal number
+* @tparam[opt] integer cmd 0=BLOCK (default), 1=UNBLOCK
+* @treturn boolean `true` on success
+* @treturn[error] boolean `false` followed by negated errno
+* @within linux
+* @usage
+*   linux.sigmodify(15, true)          -- Unblock SIGTERM
+*   linux.sigmodify(linux.signal.KILL) -- Block SIGKILL
+*/
+static int lualinux_sigmodify(lua_State *L)
```
lneto: I would call this `sigmask`, I think it would be clearer.

lib/lualinux.c
```diff
@@ -117,6 +117,53 @@ static int lualinux_schedule(lua_State *L)
...
+    
+    int signum = luaL_checkinteger(L, 1);
+    int cmd = luaL_optinteger(L, 2, 0);
+    
+    sigaddset(&newmask, signum);
+    
+    int ret = sigprocmask(cmd, &newmask, NULL);
+    if (ret) {
+        lua_pushboolean(L, false);
+        lua_pushinteger(L, -ret);
```
lneto: I would use the same macros we are using on luasocket to return errname

lib/lualinux.c
```diff
@@ -117,6 +117,53 @@ static int lualinux_schedule(lua_State *L)
...
+}
+
+/***
+* Checks current task pending signals.
+*
+* @function pending
+* @treturn boolean
+* @within linux
+*/
+static int lualinux_pending(lua_State *L)
```
lneto: I would call this sigpending (trying to "standardize" at the bare minimum at least the signal namespace ;-))

lib/lualinux.c
```diff
@@ -158,6 +205,19 @@ static int lualinux_kill(lua_State *L)
 	return 2;
 }
 
+/***
+* Checks if signal is allowed (unblocked) for current task.
+* @function isallowed
+* @tparam integer sig Signal number to send (`linux.signal.KILL`).
+* @treturn boolean `true` if signal can be delivered (not blocked)
+*/
+static int lualinux_isallowed(lua_State *L)
```
lneto: I would rather have `sigismemeber` here.. you can have an optional string to describe "blocked", ["pending"](https://elixir.bootlin.com/linux/v6.6/source/fs/autofs/waitq.c#L53), etc.

lib/lualinux.c
Physic69: like this: ` lunatik_tryret(L, ret, sigprocmask...)` ?

lneto: as we don't need the actual ret, I think just `lunatik_try`, right? so it will raise the error accordingly..

lib/lualinux.c
```diff
@@ -158,6 +199,20 @@ static int lualinux_kill(lua_State *L)
...
+{
+    int signum = luaL_checkinteger(L, 1);
+    const char *state = luaL_optstring(L, 2, "blocked");
+    
+    if (strcmp(state, "blocked") == 0)
+        lua_pushboolean(L, sigismember(&current->blocked, signum));
+    else if (strcmp(state, "pending") == 0) 
+        lua_pushboolean(L, sigismember(&current->pending.signal, signum));
+    else 
+        lua_pushboolean(L, !sigismember(&current->blocked, signum));
```
```suggestion
    const char *state = luaL_tostring(L, 2);
    
    if (strcmp(state, "pending") == 0) 
        lua_pushboolean(L, sigismember(&current->pending.signal, signum));
    else 
        lua_pushboolean(L, sigismember(&current->blocked, signum));
```

lib/lualinux.c
```diff
@@ -158,6 +199,20 @@ static int lualinux_kill(lua_State *L)
 	return 2;
 }
 
+static int lualinux_sigstate(lua_State *L)
```
lneto, line 173: please, document this API as well

lib/lualinux.c
lneto: actually, I think it will be clearer to use [luaL_checkoption](https://www.lua.org/manual/5.5/manual.html#luaL_checkoption) here. Assuming "blocked" as default seems a bit confusing to me. See https://github.com/luainkernel/lua/blob/v5.5.0-kernel/lbaselib.c#L208

lneto: btw, it would be good to check the kernel source for other useful options =).

Physic69: > actually, I think it will be clearer to use [luaL_checkoption](https://www.lua.org/manual/5.5/manual.html#luaL_checkoption) here. Assuming "blocked" as default seems a bit confusing to me. See https://github.com/luainkernel/lua/blob/v5.5.0-kernel/lbaselib.c#L208

like [this](https://github.com/luainkernel/lunatik/pull/351/changes#diff-ebaea58d0c04910938024880228d50c76d071c363b631259cef65a6c4cbdf5feR228) right?
Will look more into kernel..

lib/lualinux.c
```diff
@@ -161,6 +199,49 @@ static int lualinux_kill(lua_State *L)
...
+static const char *const sigstate_opts[] = {
+    "blocked", 
+    "pending",   
+    "allowed", 
+    "realtime",
+    NULL
+};
+
+static const int sigstate_cmds[] = {
+    0,      1,     2,      3
```
lneto: please, mind our code style. also, I would use enum here.

lib/lualinux.c
```diff
@@ -161,6 +199,49 @@ static int lualinux_kill(lua_State *L)
 	return 2;
 }
 
+static const char *const sigstate_opts[] = {
```
lneto: I think it would be better to move both to the scope of the function; it won't be used anywhere else

lib/lualinux.c
```diff
@@ -161,6 +199,49 @@ static int lualinux_kill(lua_State *L)
...
+* @within linux
+*/
+static int lualinux_sigstate(lua_State *L)
+{
+    int signum = luaL_checkinteger(L, 1);
+    int opt = sigstate_cmds[luaL_checkoption(L, 2, "pending", sigstate_opts)];
+    
+    switch (opt) {
+        case 0: 
+            lua_pushboolean(L, sigismember(&current->blocked, signum));
```
lneto: you can move the common pattern to just after the switch block, and only assign the pointer to a var on the cases..

lib/lualinux.c
```diff
@@ -120,6 +117,47 @@ static int lualinux_schedule(lua_State *L)
...
+    sigset_t newmask;
+    sigemptyset(&newmask);
+    
+    int signum = luaL_checkinteger(L, 1);
+    int cmd = luaL_optinteger(L, 2, 0);
+    
+    sigaddset(&newmask, signum);
+
+    lunatik_try(L, sigprocmask, cmd, &newmask, NULL);
+    lua_pushboolean(L, true);
```
lneto: why pushboolean here?

lib/lualinux.c
```diff
@@ -161,6 +199,50 @@ static int lualinux_kill(lua_State *L)
...
+*/
+static int lualinux_sigstate(lua_State *L)
+{
+    enum sigstate_cmd {
+        SIGSTATE_BLOCKED,
+        SIGSTATE_PENDING,
+        SIGSTATE_ALLOWED,
+    };
+
+    static const char *const sigstate_opts[] = {
```
lneto: it shouldn't be static, right?

lib/lualinux.c
```diff
@@ -161,6 +199,50 @@ static int lualinux_kill(lua_State *L)
...
+        [SIGSTATE_BLOCKED] = "blocked",
+        [SIGSTATE_PENDING] = "pending",
+        [SIGSTATE_ALLOWED] = "allowed",
+        NULL
```
```suggestion
        NULL,
```

lneto: actually, we don't need this NULL here, right?

lib/lualinux.c
```diff
@@ -161,6 +199,50 @@ static int lualinux_kill(lua_State *L)
...
+    };
+
+    int signum = luaL_checkinteger(L, 1);
+    enum sigstate_cmd cmd = (enum sigstate_cmd)luaL_checkoption(L, 2, "blocked", sigstate_opts);
```
lneto, line 188: I'm not sure we should have a default = blocked here.. but I will leave this decision to you ;-).. what would be the better ergonomics for this API..

lib/lualinux.c
```diff
@@ -161,6 +199,50 @@ static int lualinux_kill(lua_State *L)
 	return 2;
 }
 
+/***
+* Checks signal state for current task.
+*
+* @function sigstate
+* @tparam integer sig Signal number
```
lneto, line 165: actually a string, right?

lib/lualinux.c
```diff
@@ -591,13 +795,15 @@ static const lunatik_namespace_t lualinux_flags[] = {
 static const luaL_Reg lualinux_lib[] = {
 	{"random", lualinux_random},
 	{"schedule", lualinux_schedule},
+	{"sigmask", lualinux_sigmask},        
+	{"sigpending", lualinux_sigpending},
 	{"kill", lualinux_kill},
+	{"sigstate", lualinux_sigstate},
```
lneto, line 802: I would move this function to next to the two others

lneto (commented):
it's looking good to me.. just small comments..

lib/lualinux.c
Physic69: i feel like its better than having pending at the moment

lib/lualinux.c
Physic69: as well as a signal num

lib/lualinux.c
lneto: thing is.. we don't need to have any default.. but it's alright to me.. =)

PR #352, opened 2025-12-27
Fix lunatik starter unexpected behaviour

### LINKED ISSUE #350: lunatik starter script: unexpected behaviour
An expected behaviour is seen with `bin/lunatik` where if an invalid argument is given, let's say `foobar` , the script will proceed loading the modules and `then` show the help message with the expected parameters.

> $ sudo lunatik foobar ; lsmod |grep lua
> usage: lunatik [load|unload|reload|status|list] [run|spawn|stop <script>]
> luacpu                 12288  0
> luacrypto_comp         12288  0
> ...
> lunatik               344064  22 luaprobe,luacrypto_rng,luacrypto_skcipher,luaxtable,luafib,luacpu,luaxdp,luasyscall,luarcu,luathread,lualinux,luacrypto_shash,luasocket,lunatik_run,luadata,luacompletion,luacrypto_comp,luafifo,luadevice,luanetfilter,luacrypto_aead,luanotifier
> x_tables               65536  2 luaxtable,ip_tables

This can be confusing for new users, correct behaviour, IMO, would be for the script to just show the help message and exit cleanly.

carloslack:
Fix #350

Bahaviour with this patch:

$ sudo lunatik foobar
usage: lunatik [load|unload|reload|status|list] [run|spawn|stop <script>] $ lsmod |grep lua
$ sudo lunatik load
$ lsmod |grep lunatik
lunatik_run            12288  0
luarcu                 12288  4 luaxtable,luaxdp,lunatik_run
lunatik               344064  22 luaprobe,luacrypto_rng,luacrypto_skcipher,...
$ sudo lunatik unload
$ lsmod |grep lunatik
$

### DIFF
--- a/bin/lunatik
+++ b/bin/lunatik
@@ -98,13 +98,6 @@ if command then
 	os.exit()
 end
 
-if not lunatik.probe() then
-	lunatik.commands.load()
-end
-
-local version = lunatik.dostring("return _LUNATIK_VERSION")
-lunatik.copyright = version .. "  " .. lunatik.copyright
-
 local function set(t)
 	local s = {}
 	for _, key in ipairs(t) do s[key] = true end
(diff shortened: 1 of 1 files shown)


bin/lunatik
lneto, line 119: actually, we are already handling this condition here, right? wouldn't be the case of just moving `if not lunatik.probe() ... lunatik.copyright` to after this block?

carloslack: yes! I've applied the change and it is working fine with my tests.

lneto (approved):

Issue #309, opened 2025-07-20
CI or Automated Workflow for testing builds


sheharyaar:
We need a CI or Github Workflow to test Lunatik pushes in the `master` against kernel versions `5.x` onwards. This will help catch us issues like: #307.

PR #355, opened 2025-12-28
Add GitHub Actions workflow for building modules

Physic69:
See:
https://github.com/Physic69/lunatik/pulls
for working

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,53 @@
+# Based on https://gist.github.com/domenic/ec8b0fc8ab45f39403dd
+name: Build Modules
+on:
+  pull_request:
+    branches:
+    - master
+  push:
+    branches:
+    - master
+jobs:
+  build:
+    name: Build ${{ matrix.kernel-version }}
+    strategy:
+      matrix:
+        kernel-version: [5.15, 6.11]
+      fail-fast: false
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+        submodules: recursive
+    - name: Install dependencies
+      run: |
+        sudo apt-get update -qq
+        sudo apt-get install -yq build-essential bc bison flex libssl-dev dwarves libelf-dev libncurses-dev clang llvm pkg-config libpcap-dev lua5.4-dev m4
+    - name: Setup kernel ${{ matrix.kernel-version }}
+      run: |
+        KERNEL_VERSION="${{ matrix.kernel-version }}"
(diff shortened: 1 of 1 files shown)


sheharyaar:
Should the workflow also try to load the kernel module after building just for a better check ?

carloslack:
> Should the workflow also try to load the kernel module after building just for a better check ?

This would be bordering tests domain IMO

.github/workflows/build.yaml
```diff
@@ -0,0 +1,46 @@
...
+        make modules_prepare -j$(nproc)
+        echo "KERNEL_DIR=$(pwd)" >> $GITHUB_ENV
+    - name: Build Lunatik
+      run: |
+        export KERNEL_DIR="${{ env.KERNEL_DIR }}"
+        make clean || true
+        make -j$(nproc)
+    - name: Verify modules built
+      run: |
+        test -f lunatik.ko || test -f lib/lunatik.ko || (echo "Build failed" && exit 1)
```
carloslack: Is it really needed to test for `lib/lunatik.ko` ?

Physic69: Nope, removed it

Physic69:
> Should the workflow also try to load the kernel module after building just for a better check ?

I feel like, they are not required. This catches most of missed cases (which is rare), XD..
Also added 5.x kernel version as well, as per issue stated 5.x onwards..

.github/workflows/build.yaml
```diff
@@ -0,0 +1,53 @@
...
+        make modules_prepare -j$(nproc)
+        echo "KERNEL_DIR=$(pwd)" >> $GITHUB_ENV
+    - name: Build Lunatik
+      run: |
+        export KERNEL_DIR="${{ env.KERNEL_DIR }}"
+        make clean || true
+        make -j$(nproc)
+    - name: Verify modules built
+      run: |
+        test -f lunatik.ko || (echo "Build failed" && exit 1)
```
carloslack: Should use `&&` here? I think a `;` would work, this echo cannot fail.

Physic69: yup, missed that..
Thanks for pointing that

lneto:
hey @sheharyaar @carloslack, do you think we can merge this or do you still have open concerns?

sheharyaar:
> hey @sheharyaar @carloslack, do you think we can merge this or do you still have open concerns?

LGTM, we can merge this.

lneto:
> > hey @sheharyaar @carloslack, do you think we can merge this or do you still have open concerns?
> 
> LGTM, we can merge this.

Ack; will merge it after releasing 4.0

sheharyaar:
I guess this fixes #309.

PR #357, opened 2025-12-29
update lua/ to v5.5.0

lneto:


jperon:
I just tried (checkout this branch, compiled and rebooted), but when launching `lunatik`, I get the trace joined here. Otherwise, simple tests (like `print"test"` or `lunatik run rng` and so on) work as expected.
[log.txt](https://github.com/user-attachments/files/24379804/log.txt)

lneto:
> I just tried (checkout this branch, compiled and rebooted), but when launching `lunatik`, I get the trace joined here. Otherwise, simple tests (like `print"test"` or `lunatik run rng` and so on) work as expected. [log.txt](https://github.com/user-attachments/files/24379804/log.txt)

Thank you very much! ~Can you check if these configs are on? `CONFIG_UBSAN_BOUNDS` and `CONFIG_FORTIFY_SOURCE`? It seems that [this line](https://github.com/lua/lua/compare/v5.5.0...luainkernel:lua:v5.5.0-kernel#diff-a71aa75c20a11677951d73e7d6836e4333e163aa8e4db23976965aa6feeb4945R445) is not evaluating to true. Can you also try to force LUNATIK_FLEXARRAY.~

ouch, me culpa, I forgot to push `lunatik_conf.h`; just pushed -f it now.

lneto:
> (...) simple tests (like `print"test"` or `lunatik run rng` and so on) work as expected.

btw, can you also run moontastik? I'm not sure if Lua 5.5 will break MoonScript transpilation..

jperon:
So, some tests later:

- "force-pulling" fixed the kernel trace;
- I get some errors about error codes in my crypto tests; I guess it comes from the recent changes in another merged PR? Some details like "EINVAL" instead of 22, which I’ll have to fix;
- when running `lunatik unload`, i get those warnings:
```
déc. 30 18:16:09 shbox kernel: Lua warning: 
déc. 30 18:16:09 shbox kernel: error in 
déc. 30 18:16:09 shbox kernel: __gc
déc. 30 18:16:09 shbox kernel:  (
déc. 30 18:16:09 shbox kernel: /lib/modules/lua/crypto/hkdf.lua:24: calling '__close' on bad self (null-pointer dereference)
déc. 30 18:16:09 shbox kernel: )
déc. 30 18:16:09 shbox kernel: 
déc. 30 18:16:09 shbox kernel: Lua warning: 
déc. 30 18:16:09 shbox kernel: error in 
déc. 30 18:16:09 shbox kernel: __gc
déc. 30 18:16:09 shbox kernel:  (
déc. 30 18:16:09 shbox kernel: /lib/modules/lua/crypto/hkdf.lua:24: calling '__close' on bad self (null-pointer dereference)
```
- otherwise, only errors that got raised were about read-only for-loop variables, especially this one that will need to be fixed in net.lua (and I’ll have other similar fixes for ipparse):
```
/lib/modules/lua/net.lua:27: attempt to assign to const variable 'n'
```

lneto:
> So, some tests later:
> 
> * "force-pulling" fixed the kernel trace;
> * I get some errors about error codes in my crypto tests; I guess it comes from the recent changes in another merged PR? Some details like "EINVAL" instead of 22, which I’ll have to fix;

Yeah, this was due to the change on the error handling. Sorry about that =(.

> * when running `lunatik unload`, i get those warnings:
> 
> ```
> déc. 30 18:16:09 shbox kernel: Lua warning: 
> déc. 30 18:16:09 shbox kernel: error in 
> déc. 30 18:16:09 shbox kernel: __gc
> déc. 30 18:16:09 shbox kernel:  (
> déc. 30 18:16:09 shbox kernel: /lib/modules/lua/crypto/hkdf.lua:24: calling '__close' on bad self (null-pointer dereference)
> déc. 30 18:16:09 shbox kernel: )
> déc. 30 18:16:09 shbox kernel: 
> déc. 30 18:16:09 shbox kernel: Lua warning: 
> déc. 30 18:16:09 shbox kernel: error in 
> déc. 30 18:16:09 shbox kernel: __gc
> déc. 30 18:16:09 shbox kernel:  (
> déc. 30 18:16:09 shbox kernel: /lib/modules/lua/crypto/hkdf.lua:24: calling '__close' on bad self (null-pointer dereference)
> ```

here I have no clue; but searching for the error string, this should be raised by luaL_argerror on shash. Probably called from closeobject -> checkpobject -> argchecknull. Hm.. perhaps we are double closing it as you are defining both __close and __gc as close. However, you have a verification there to call it only once. Can you print(self.tfm) on close? To understand if we are calling it twice for the same tfm userdata somehow.

> * otherwise, only errors that got raised were about read-only for-loop variables, especially this one that will need to be fixed in net.lua (and I’ll have other similar fixes for ipparse):
> 
> ```
> /lib/modules/lua/net.lua:27: attempt to assign to const variable 'n'
> ```

I will add a fix for net.lua on this commit. Now, according to the [manual](https://www.lua.org/manual/5.5/manual.html#8):

"The control variable in for loops is read only. If you need to change it, declare a local variable with the same name in the loop body."

lneto:
> > So, some tests later:
> > 
> > * "force-pulling" fixed the kernel trace;
> > * I get some errors about error codes in my crypto tests; I guess it comes from the recent changes in another merged PR? Some details like "EINVAL" instead of 22, which I’ll have to fix;
> 
> Yeah, this was due to the change on the error handling. Sorry about that =(.
> 
> > * when running `lunatik unload`, i get those warnings:
> > 
> > ```
> > déc. 30 18:16:09 shbox kernel: Lua warning: 
> > déc. 30 18:16:09 shbox kernel: error in 
> > déc. 30 18:16:09 shbox kernel: __gc
> > déc. 30 18:16:09 shbox kernel:  (
> > déc. 30 18:16:09 shbox kernel: /lib/modules/lua/crypto/hkdf.lua:24: calling '__close' on bad self (null-pointer dereference)
> > déc. 30 18:16:09 shbox kernel: )
> > déc. 30 18:16:09 shbox kernel: 
> > déc. 30 18:16:09 shbox kernel: Lua warning: 
> > déc. 30 18:16:09 shbox kernel: error in 
> > déc. 30 18:16:09 shbox kernel: __gc
> > déc. 30 18:16:09 shbox kernel:  (
> > déc. 30 18:16:09 shbox kernel: /lib/modules/lua/crypto/hkdf.lua:24: calling '__close' on bad self (null-pointer dereference)
> > ```
> 
> here I have no clue; but searching for the error string, this should be raised by luaL_argerror on shash. Probably called from closeobject -> checkpobject -> argchecknull. Hm.. perhaps we are double closing it as you are defining both __close and __gc as close. However, you have a verification there to call it only once. Can you print(self.tfm) on close? To understand if we are calling it twice for the same tfm userdata somehow.

update: my guess is that it's [this line](https://github.com/luainkernel/lunatik/blob/master/lunatik_obj.c#L91) raising the error. I'm not sure it's related, but shouldn't we set `obj->tfm = NULL` [here](https://github.com/luainkernel/lunatik/blob/master/lib/luacrypto_shash.c#L36)?

lneto:
> > > * I get some errors about error codes in my crypto tests; I guess it comes from the recent changes in another merged PR? Some details like "EINVAL" instead of 22, which I’ll have to fix;
> > 
> > 
> > Yeah, this was due to the change on the error handling. Sorry about that =(.
> > > * when running `lunatik unload`, i get those warnings:
> >  (...)
> > here I have no clue; but searching for the error string, this should be raised by luaL_argerror on shash. Probably called from closeobject -> checkpobject -> argchecknull. Hm.. perhaps we are double closing it as you are defining both __close and __gc as close. However, you have a verification there to call it only once. Can you print(self.tfm) on close? To understand if we are calling it twice for the same tfm userdata somehow.
> 
> update: my guess is that it's [this line](https://github.com/luainkernel/lunatik/blob/master/lunatik_obj.c#L91) raising the error. I'm not sure it's related, but shouldn't we set `obj->tfm = NULL` [here](https://github.com/luainkernel/lunatik/blob/master/lib/luacrypto_shash.c#L36)?

maybe we should change [these lines](https://github.com/luainkernel/lunatik/blob/master/lunatik_obj.c#L91-L92) to just avoid to explode.. to something like this..

```
	if (private != NULL)
		lunatik_releaseprivate(object->class, private);
```

I really think the problem is [here](https://github.com/luainkernel/lunatik/blob/master/lib/crypto/hkdf.lua#L29-L30).

lneto:
> > > > * I get some errors about error codes in my crypto tests; I guess it comes from the recent changes in another merged PR? Some details like "EINVAL" instead of 22, which I’ll have to fix;
> > > 
> > > 
> > > Yeah, this was due to the change on the error handling. Sorry about that =(.
> > > > * when running `lunatik unload`, i get those warnings:
> > > >   (...)
> > > >   here I have no clue; but searching for the error string, this should be raised by luaL_argerror on shash. Probably called from closeobject -> checkpobject -> argchecknull. Hm.. perhaps we are double closing it as you are defining both __close and __gc as close. However, you have a verification there to call it only once. Can you print(self.tfm) on close? To understand if we are calling it twice for the same tfm userdata somehow.
> > 
> > 
> > update: my guess is that it's [this line](https://github.com/luainkernel/lunatik/blob/master/lunatik_obj.c#L91) raising the error. I'm not sure it's related, but shouldn't we set `obj->tfm = NULL` [here](https://github.com/luainkernel/lunatik/blob/master/lib/luacrypto_shash.c#L36)?
> 
> maybe we should change [these lines](https://github.com/luainkernel/lunatik/blob/master/lunatik_obj.c#L91-L92) to just avoid to explode.. to something like this..
> 
> ```
> 	if (private != NULL)
> 		lunatik_releaseprivate(object->class, private);
> ```
> 
> I really think the problem is [here](https://github.com/luainkernel/lunatik/blob/master/lib/crypto/hkdf.lua#L29-L30).

update.. that was hard, but I think now I understand what's happening.. I could reproduce it simply by changing this [line](https://github.com/luainkernel/lunatik/blob/master/lib/socket/inet.lua#L35) to:

```
	self.__gc = self.close
```

and running this script:

```
inet = require("socket/inet") s = inet.udp() s = nil collectgarbage()
```

what seems to happen is that the finalizer is called firstly for the userdata (s.socket) then to the table (s), what I think it's the expected behavior; see:

"When a marked object becomes dead, it is not collected immediately by the garbage collector. Instead, Lua puts it in a list. After the collection, Lua goes through that list. For each object in the list, it checks the object's __gc metamethod: If it is present, Lua calls it with the object as its single argument.

At the end of each garbage-collection cycle, the finalizers are called in the *reverse order* that the objects were marked for finalization, among those collected in that cycle; that is, the first finalizer to be called is the one associated with the object marked last in the program. The execution of each finalizer may occur at any point during the execution of the regular code."[[1]](https://www.lua.org/manual/5.5/manual.html#2.5.3)

thus, if I got it correctly, `s` is marked to be collected (`s = nil`), then `s.socket` is marked to be collected. When GC runs (`collectgarbage()`), it will call `s.socket.__gc` firstly, which it will call [lunatik_deleteobject](https://github.com/luainkernel/lunatik/blob/master/lunatik_obj.c#L110). This call will release the object pointed by the userdata but not release the userdata itself. Then, Lua will call `s.__gc`, which will call `s.socket.__close` passing the userdata that was finalized (in the sense it called its `__gc` metamethod), but not collected yet. Thus, [lunatik_closeobject](https://github.com/luainkernel/lunatik/blob/master/lunatik_obj.c#L83) will call [lunatik_checkobject](https://github.com/luainkernel/lunatik/blob/master/lunatik_obj.c#L81), which will call luaL_checkudata and [acquires it's pointer](https://github.com/luainkernel/lunatik/blob/master/lunatik_obj.c#L53), which was already nullified by `lunatik_deleteobject` call. Then, it will raise the warning we are observing on [lunatik_argchecknull](https://github.com/luainkernel/lunatik/blob/master/lunatik_obj.c#L54).

That all said =), I think we just need to change hkdf to not call the userdata's __close on the table's __gc.

lneto:
pushed -f it again..

* fixed lib/net.lua
* improved short string patch (now using `unsafe_memcpy`)

lneto:
@jperon I'm merging this, then we can fix hkdf on another PR

PR #360, opened 2025-12-31
Factor common RAW AF_PACKET socket pattern

### LINKED ISSUE #358: Factor common AF_PACKET RAW socket pattern
While working on LLDP support, I noticed a recurring pattern [here](https://github.com/luainkernel/lunatik/pull/341#discussion_r2648424943) when creating AF_PACKET raw sockets, for example:

```lua
-- receiving socket 
local rx <close> = socket.new(socket.af.PACKET, socket.sock.RAW, ETH_P_ALL)
rx:bind(string.pack(">H", ETH_P_ALL), ifindex)
-- ...
-- transmitting socket
local tx <close> = socket.new(socket.af.PACKET, socket.sock.RAW, ETH_P_LLDP)
tx:bind(ifindex)
```

This pattern appears whenever we want to send/receive raw frames. This could be factored in a common place like `socket/raw.lua` for ready future use.

sneaky-potato:
Move recurring raw socket receive and send pattern in a common place.

Fixes #358

### DIFF
--- a/lib/socket/raw.lua
+++ b/lib/socket/raw.lua
@@ -1,0 +1,39 @@
+--
+-- SPDX-FileCopyrightText: (c) 2025 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+---
+-- RAW AF_PACKET socket operations.
+-- This module provides a higher-level abstraction over the `socket` module.
+--
+-- @module socket.raw
+-- @see socket
+--
+local socket = require("socket")
+
+local af   = socket.af
+local sock = socket.sock
+
+local ETH_P_ALL  = 0x0003
+
+local raw = {}
+
+---
+-- Creates and binds a raw packet socket for receiving and sending frames.
+-- @param proto (number) EtherType (defaults to ETH_P_ALL).
+-- @param ifindex (number) Interface index.
+-- @return A new raw packet socket bound for RX.
+-- @raise Error if socket.new() or socket.bind() fail.
+-- @see socket.new
+-- @see socket.bind
+function raw.new(proto, ifindex)
+	local proto = proto or ETH_P_ALL
+	local ifindex = ifindex or 0
+	local s = socket.new(af.PACKET, sock.RAW, proto)
+	s:bind(string.pack(">H", proto), ifindex)
+	return s
+end
+
+return raw
+
--- a/examples/lldpd.lua
+++ b/examples/lldpd.lua
@@ -4,7 +4,7 @@
 --
 
 -- Daemon to send LLDP frames on given interface
-local socket = require("socket")
+local raw    = require("socket.raw")
 local linux  = require("linux")
 local thread = require("thread")
 
@@ -36,8 +36,7 @@ local config = {
 local ethertype = string.pack(">H", ETH_P_LLDP)
 
 local function get_src_mac(ifindex)
-	local rx <close> = socket.new(socket.af.PACKET, socket.sock.RAW, ETH_P_ALL)
-	rx:bind(string.pack(">H", ETH_P_ALL), ifindex)
+	local rx <close> = raw.bind(ETH_P_ALL, ifindex)
 	local frame = rx:receive(2048)
 	return frame:sub(7, 12)
 end
@@ -80,8 +79,7 @@ local src_mac = get_src_mac(ifindex)
 local lldp_frame = build_lldp_frame(src_mac)
 
 local function worker()
-	local tx <close> = socket.new(socket.af.PACKET, socket.sock.RAW, ETH_P_LLDP)
-	tx:bind(ifindex)
+	local tx <close> = raw.bind(ETH_P_LLDP, ifindex)
 
 	while (not shouldstop()) do
 		tx:send(lldp_frame)
--- a/examples/tap.lua
+++ b/examples/tap.lua
@@ -4,19 +4,15 @@
 --
 
 local device = require("device")
-local socket = require("socket")
+local raw    = require("socket.raw")
 local linux  = require("linux")
 
-local PACKET    = socket.af.PACKET
-local RAW       = socket.sock.RAW
-local ETH_P_ALL = 0x0003
 local MTU       = 1500
 
 local s = linux.stat
 local tap = {name = "tap", mode = s.IRUGO}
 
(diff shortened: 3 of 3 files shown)


lib/socket/raw.lua
```diff
@@ -0,0 +1,79 @@
...
+-- @param ifindex (number) Interface index.
+-- @param proto (number) EtherType.
+-- @param fn (function) procedure to execute when transmitting frames
+-- @return Outout of raw packet socket bound for TX.
+-- @raise Error if socket.raw.connect() fail.
+-- @see socket.raw.connect
+function raw.with_tx(ifindex, proto, fn)
+    local s <close> = raw.connect(ifindex, proto)
+    return fn(s)
+end
```
lneto: I don't think we need this.. I'd rather keep it simple.. having only bind and connect

lib/socket/raw.lua
```diff
@@ -0,0 +1,51 @@
...
+---
+-- Creates and binds a raw packet socket for transmitting frames.
+-- @param ifindex (number) Interface index.
+-- @param proto (number) EtherType.
+-- @return A new raw packet socket bound for TX.
+-- @raise Error if socket.new() or socket.bind() fail.
+-- @see socket.new
+-- @see socket.bind
+function raw.connect(ifindex, proto)
+	assert(proto, "raw.connect: protocol required")
```
lneto: why do we need this? what's happen if we pass proto = nil to socket.new?

lneto: btw, why not having ETH_P_ALL as default?

lneto (commented):
btw, please use `--fixup` on your commits or squash it.. let's keep our history as clean as possible

lib/socket/raw.lua
sneaky-potato: I read some code from [luasocket_bind](https://github.com/luainkernel/lunatik/blob/3e4a3d65773243fa71eab3da07e42d5a5f1c80ad/lib/luasocket.c#L249) and [luasocket_checkaddr](https://github.com/luainkernel/lunatik/blob/3e4a3d65773243fa71eab3da07e42d5a5f1c80ad/lib/luasocket.c#L54) and have some questions before finalizing abstraction
- will a socket like this receive `ETH_P_ALL` protocol frames on `ifindex` interface? 
```lua
local rx <close> = socket.new(socket.af.PACKET, socket.sock.RAW, ETH_P_ALL)
rx:bind(ifindex) -- notice no protocol is passed here 
```
- Also, for sending raw frames, is the recommended approach to use `bind(ifindex)` with protocol from `socket.new()`, or to use `connect()` instead?

lneto: > I read some code from [luasocket_bind](https://github.com/luainkernel/lunatik/blob/3e4a3d65773243fa71eab3da07e42d5a5f1c80ad/lib/luasocket.c#L249) and [luasocket_checkaddr](https://github.com/luainkernel/lunatik/blob/3e4a3d65773243fa71eab3da07e42d5a5f1c80ad/lib/luasocket.c#L54) and have some questions before finalizing abstraction
> 
> * will a socket like this receive `ETH_P_ALL` protocol frames on `ifindex` interface?
> 
> local rx <close> = socket.new(socket.af.PACKET, socket.sock.RAW, ETH_P_ALL)
> rx:bind(ifindex) -- notice no protocol is passed here 

It's supposed to work, did you try it?

> * Also, for sending raw frames, is the recommended approach to use `bind(ifindex)` with protocol from `socket.new()`, or to use `connect()` instead?

though it's weird to connect() a raw socket, I would say we should use it (such we do for unix and udp, right?). What do you think?

sneaky-potato: > It's supposed to work, did you try it?

I am not getting any traffic using this socket, however it starts capturing traffic if `s:bind(string.pack(">H", ETH_P_ALL), ifindex)` is used. (I am using `s:receive(1500)` to get the frame)

sneaky-potato: I believe this is true
- for receiving raw socket, `ifindex` is optional (listen on all interfaces like `tap.lua` if `ifindex` not given), but `protocol` is required
- for transmitting raw socket, `protocol` is optional, but `ifindex` is required

please correct me if I am wrong here.

sneaky-potato: > though it's weird to connect() a raw socket, I would say we should use it (such we do for unix and udp, right?). What do you think?

I got `[38409.252883] luathread: [0000000046b5e27a] EOPNOTSUPP` on invoking `s:connect(proto, ifindex)`, is connect not supported yet for AF_PACKET raw sockets?

lneto: Hm.. I'm not sure.. I think you should take a look at the kernel sources ;-).. but it's looking the case

lneto: Perhaps we should have only the bind method here?

lib/socket/raw.lua
```diff
@@ -0,0 +1,43 @@
...
+
+---
+-- Creates and binds a raw packet socket for receiving and sending frames.
+-- @param proto (number) EtherType (defaults to ETH_P_ALL).
+-- @param ifindex (number) Interface index.
+-- @return A new raw packet socket bound for RX.
+-- @raise Error if socket.new() or socket.bind() fail.
+-- @see socket.new
+-- @see socket.bind
+new("bind")
```
lneto:
```suggestion
---
-- Creates and binds a raw packet socket for receiving and sending frames.
-- @param proto (number) EtherType (defaults to ETH_P_ALL).
-- @param ifindex (number) Interface index.
-- @return A new raw packet socket bound for RX.
-- @raise Error if socket.new() or socket.bind() fail.
-- @see socket.new
-- @see socket.bind
function bind.new(proto, ifindex)
	local proto = proto or ETH_P_ALL
	local ifindex = ifindex or 0
	local s = socket.new(af.PACKET, sock.RAW, proto)
	s:bind(string.pack(">H", proto), ifindex)
	return s
end
```

if we will have only one method for creating a new raw socket, I'd keep that simpler and avoid the factory..

sneaky-potato: resolved in latest push

sneaky-potato:
Hi @lneto 
Sorry I missed the calling signature, examples lldpd.lua and tap.lua do not work because of that

Also I believe we need to call `bind()` differently when using both proto, ifindex vs only ifindex. I will create a quick pull request with working examples.

PR #364, opened 2026-01-03
Fix bind signature in raw.lua pattern

sneaky-potato:
We need separate call signatures for 
- receiving socket (protocol is required, passed in the form of packed string)
- transmitting socket (only numeric interface index is passed to bind)

Related PR
#360

### DIFF
--- a/examples/lldpd.lua
+++ b/examples/lldpd.lua
@@ -1,5 +1,5 @@
 --
--- SPDX-FileCopyrightText: (c) 2025 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+-- SPDX-FileCopyrightText: (c) 2025-2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
 -- SPDX-License-Identifier: MIT OR GPL-2.0-only
 --
 
@@ -33,10 +33,10 @@ local config = {
 	},
 }
 
-local ethertype = string.pack(">H", ETH_P_LLDP)
+local ethertype = string.pack(">I2", ETH_P_LLDP)
 
-local function get_src_mac(ifindex)
-	local rx <close> = raw.bind(ETH_P_ALL, ifindex)
+local function get_src_mac()
+	local rx <close> = raw.bind(ETH_P_ALL)
 	local frame = rx:receive(2048)
 	return frame:sub(7, 12)
 end
@@ -45,13 +45,13 @@ local function tlv(t, payload, subtype)
 	if subtype then
 		payload = string.char(subtype) .. payload
 	end
-	return string.pack(">H", (t << 9) | #payload) .. payload
+	return string.pack(">I2", (t << 9) | #payload) .. payload
 end
 
 local function build_lldp_frame(chassis_id)
 	local port_id = config.interface
-	local ttl = string.pack(">H", config.system.ttl)
-	local capabilities = string.pack(">HH", config.system.capabilities, config.system.capabilities_enabled)
+	local ttl = string.pack(">I2", config.system.ttl)
+	local capabilities = string.pack(">I2I2", config.system.capabilities, config.system.capabilities_enabled)
 
 	local pdu = {
 		-- Ethernet header
@@ -75,7 +75,7 @@ end
 
 local ifindex = linux.ifindex(config.interface)
 
-local src_mac = get_src_mac(ifindex)
+local src_mac = get_src_mac()
 local lldp_frame = build_lldp_frame(src_mac)
 
 local function worker()
--- a/lib/socket/raw.lua
+++ b/lib/socket/raw.lua
@@ -1,5 +1,5 @@
 --
--- SPDX-FileCopyrightText: (c) 2025 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+-- SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
 -- SPDX-License-Identifier: MIT OR GPL-2.0-only
 --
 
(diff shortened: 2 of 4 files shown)


sneaky-potato:
If we want to do this in a single new method, we can pass whatever arguments we get to the `bind` call 
something like this
```lua
-- Arguments are forwarded verbatim to socket:bind().
--
-- Examples:
--   raw.new(string.pack(">H", ETH_P_ALL), ifindex) RX-style bind
--   raw.new(ifindex) TX-style bind
--   raw.new() unbound socket
--
function raw.new(...)
    local s = socket.new(af.PACKET, sock.RAW)

    local argc = select("#", ...)
    if argc > 0 then
        s:bind(...)
    end

    return s
end

return raw
```

lib/socket/raw.lua
```diff
@@ -20,19 +20,33 @@ local ETH_P_ALL  = 0x0003
...
 -- @see socket.bind
-function raw.new(proto, ifindex)
+function raw.rx(proto, ifindex)
 	local proto = proto or ETH_P_ALL
 	local ifindex = ifindex or 0
-	local s = socket.new(af.PACKET, sock.RAW, proto)
-	s:bind(string.pack(">H", proto), ifindex)
-	return s
+    local s = socket.new(af.PACKET, sock.RAW, proto)
+    s:bind(string.pack(">H", proto), ifindex)
```
lneto: Wouldn't this work for sending as well? Are you sure?

lib/socket/raw.lua
```diff
@@ -20,19 +20,33 @@ local ETH_P_ALL  = 0x0003
...
--- @param proto (number) EtherType (defaults to ETH_P_ALL).
--- @param ifindex (number) Interface index.
+-- Creates and binds a raw packet socket for receiving frames.
+-- @param proto (number) EtherType.
+-- @param ifindex (number) Interface index (defaults to listen all interfaces).
 -- @return A new raw packet socket bound for RX.
 -- @raise Error if socket.new() or socket.bind() fail.
 -- @see socket.new
 -- @see socket.bind
-function raw.new(proto, ifindex)
```
lneto, line 30: I misread the original patch; it was supposed to be called "bind", not "new"

lib/socket/raw.lua
sneaky-potato: I tried to use this form to send frames over raw socket
```lua
local function worker()
	local tx <close> = socket.new(socket.af.PACKET, socket.sock.RAW, ETH_P_LLDP)
	tx:bind(string.pack(">H", ETH_P_LLDP), ifindex)

	while (not shouldstop()) do
		tx:send(lldp_frame)
		print(string.format("[lldpd] frame sent on ifindex=%d (%d bytes)", ifindex, #lldp_frame))
		linux.schedule(config.tx_interval_ms)
	end

	print("[lldpd] worker stopped")
end

return worker
```

got this error
[  495.463722] luathread: [000000004887f053] ENXIO

The socket starts working when called with `bind(ifindex)`

sneaky-potato: I believe [this](https://github.com/luainkernel/lunatik/blob/72cc7be7a6ec1f6723ab96e3e7db7f3c0a531293/lib/luasocket.c#L79C2-L85C3) branch executes when bind is called with string parameter.
So it blindly copies 2 bytes (ETH_P_LLDP) into sockaddr_storage.__data
Everything else is uninitialized garbage. (This explains ENXIO no such device error)

Summary
- `bind(ifindex)` binds on a specific interface (this cannot be used to capture traffic because protocol is never set in this case)
- `bind(string.pack(">H", proto))` binds on all interfaces (0) (should be used when receiving frames)
- anything else does not make sense

lneto: We were using ">I2" instead ">H" on tap.. why did you change it?

sneaky-potato: I read [this](https://github.com/luainkernel/lunatik/blob/72cc7be7a6ec1f6723ab96e3e7db7f3c0a531293/lib/luasocket.c#L235) line in the documentation.

lneto: I don't think it's correct as short might vary depending on arch.. I would fix both the code and the doc

lib/socket/raw.lua
```diff
@@ -20,19 +20,33 @@ local ETH_P_ALL  = 0x0003
...
+---
+-- Creates and binds a raw packet socket for sending frames.
+-- @param proto (number) EtherType.
+-- @param ifindex (number) Interface index.
+-- @return A new raw packet socket bound for RX.
+-- @raise Error if socket.new() or socket.bind() fail.
+-- @see socket.new
+-- @see socket.bind
+function raw.tx(proto, ifindex)
+    local s = socket.new(af.PACKET, sock.RAW, proto)
```
lneto: I would create a "new" function just to avoid repeat this line.. not fond of tx/Rx but couldn't figure better names.. it's up to you.. I can merge it

lneto (commented):
Ah.. you need to add this to config.ld right?

lib/socket/raw.lua
sneaky-potato: What about `bind` and `connect` as the names for the methods?

lneto: I think it's better from the API client's perspective.. though it sounds a bit misleading to bind() on connect(), I will leave this decision to you ;-)

sneaky-potato: I think a single bind call will suffice, we can call `bind(ifindex)` if ifindex is given else the usual listening socket `bind(string.pack(">I2", proto))`

sneaky-potato:
@lneto from the `luasocket_bind` and more specifically [luasocket_checkaddr](https://github.com/luainkernel/lunatik/blob/72cc7be7a6ec1f6723ab96e3e7db7f3c0a531293/lib/luasocket.c#L74C2-L85C3), it seems that binding a listening AF_PACKET socket to a specific interface is not currently supported.

For AF_PACKET
- when the argument is a number, it is always interpreted as `sll_ifindex`, and no path sets `sll_protocol`. 
- string-based path copies raw bytes into `sockaddr_storage` but does not populate struct `sockaddr_ll` fields in a structured way.

For a listening raw socket bound to an interface, we would need something equivalent to: 
```c
sll_protocol = htons(ETH_P_ALL);
sll_ifindex  = ifindex;
```
Currently there’s no way to express this combination from Lua. Is this by design? If not, would it make sense to open an issue to track proper AF_PACKET bind support for listening sockets?

lib/socket/raw.lua
```diff
@@ -20,19 +20,25 @@ local ETH_P_ALL  = 0x0003
...
-	local ifindex = ifindex or 0
-	local s = socket.new(af.PACKET, sock.RAW, proto)
-	s:bind(string.pack(">H", proto), ifindex)
-	return s
+    local s = socket.new(af.PACKET, sock.RAW, proto)
+	if ifindex then
+		s:bind(ifindex)
+	else
+		s:bind(string.pack(">I2", proto))
+	end
```
lneto:
```suggestion
	local parm = ifindex and ifindex or string.pack(">I2", proto)
	s:bind(parm)
```

lneto (approved):
it's looking good to me, just a small comment.. btw, can you take the opportunity and bump the copyright year of those files as well? thanks

lib/socket/raw.lua
sneaky-potato: Resolved and tested, both lldpd and tap are working as expected.

examples/lldpd.lua
```diff
@@ -1,5 +1,5 @@
 --
--- SPDX-FileCopyrightText: (c) 2025 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+-- SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
```
lneto:
```suggestion
-- SPDX-FileCopyrightText: (c) 2025-2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
```

right?

PR #365, opened 2026-01-04
Auto generate network constants for lua

### LINKED ISSUE #354: Consider auto-generate network constants for Lua use
As per discussion with @lneto [here](https://github.com/luainkernel/lunatik/pull/341#discussion_r2649758805) , consider auto-generating some constants for Lua use.

For example the `ETH_P_*` from `/usr/src/linux-headers-$(uname -r)/include/uapi/linux/if_ether.h` 

Maybe something like 

```
local CONSTANTS = {
   ETH_P_ALL = 0x0003,
   ETH_P_LLDP = 0x88CC,
   ...
```

We already do something like that during build-time using `gensymbols.sh`

sneaky-potato:
Generate linux constants from `$(MODULES_PATH)/$(KERNEL_RELEASE)/build/include/uapi/linux/*.h` for lua usage

Fixes #354

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,8 @@ all: lunatik_sym.h defines
 	CONFIG_LUNATIK_NETFILTER=m CONFIG_LUNATIK_COMPLETION=m \
 	CONFIG_LUNATIK_CRYPTO_SHASH=m CONFIG_LUNATIK_CRYPTO_SKCIPHER=m \
 	CONFIG_LUNATIK_CRYPTO_AEAD=m CONFIG_LUNATIK_CRYPTO_RNG=m \
-	CONFIG_LUNATIK_CRYPTO_COMP=m CONFIG_LUNATIK_CPU=m CONFIG_LUNATIK_HID=m
+	CONFIG_LUNATIK_CRYPTO_COMP=m CONFIG_LUNATIK_CPU=m CONFIG_LUNATIK_HID=m \
+    CONFIG_LUNATIK_SIGNAL=m CONFIG_LUNATIK_BYTEORDER=m
 
 clean:
 	${MAKE} -C ${MODULES_BUILD_PATH} M=${PWD} clean
@@ -53,7 +54,8 @@ scripts_install:
 	${INSTALL} -m 0644 lib/syscall/*.lua ${SCRIPTS_INSTALL_PATH}/syscall
 	${INSTALL} -m 0644 lib/crypto/*.lua ${SCRIPTS_INSTALL_PATH}/crypto
 	${INSTALL} -m 0644 ${AUTOGEN_DIR}/linux/*.lua ${SCRIPTS_INSTALL_PATH}/linux
-	${INSTALL} -D -m 0755 bin/lunatik ${LUNATIK_INSTALL_PATH}/bin/lunatik
+	${INSTALL} -D -m 0755 bin/lunatik ${LUNATIK_INSTALL_PATH}/lunatik
+
 
 scripts_uninstall:
 	${RM} ${SCRIPTS_INSTALL_PATH}/driver.lua
@@ -124,7 +126,7 @@ uninstall: scripts_uninstall modules_uninstall
 	depmod -a
 
 lunatik_sym.h: $(LUA_API) gensymbols.sh
-	${shell ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
+	${shell CC='$(CC)' ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
 
 ${AUTOGEN_DIR}/linux:
 	${MKDIR} ${AUTOGEN_DIR}/linux
--- a/examples/lldpd.lua
+++ b/examples/lldpd.lua
@@ -33,8 +33,9 @@ local config = {
 
 local ethertype = string.pack(">I2", eth.LLDP)
 
-local function get_src_mac()
-	local rx <close> = raw.bind(eth.ALL)
+local function get_src_mac(ifindex)
+	local rx <close> = raw.bind(eth.ALL, ifindex)
+
 	local frame = rx:receive(2048)
 	return frame:sub(7, 12)
 end
@@ -73,7 +74,7 @@ end
 
 local ifindex = linux.ifindex(config.interface)
 
-local src_mac = get_src_mac()
+local src_mac = get_src_mac(ifindex)
 local lldp_frame = build_lldp_frame(src_mac)
 
 local function worker()
--- a/Kbuild
+++ b/Kbuild
@@ -64,4 +64,6 @@ obj-$(CONFIG_LUNATIK_CRYPTO_RNG) += lib/luacrypto_rng.o
 obj-$(CONFIG_LUNATIK_CRYPTO_COMP) += lib/luacrypto_comp.o
 obj-$(CONFIG_LUNATIK_CPU) += lib/luacpu.o
 obj-$(CONFIG_LUNATIK_HID) += lib/luahid.o
+obj-$(CONFIG_LUNATIK_SIGNAL) += lib/luasignal.o
+obj-$(CONFIG_LUNATIK_BYTEORDER) += lib/luabyteorder.o
 
--- a/examples/dnsblock/common.lua
+++ b/examples/dnsblock/common.lua
@@ -6,6 +6,7 @@
 -- Common code for new netfilter framework and legacy iptables dnsblock example
 
 local linux = require("linux")
+local byteorder = require("byteorder")
 local string = require("string")
 
 local common = {}
@@ -34,7 +35,7 @@ end
 
 function common.hook(skb, thoff, proto)
 	if proto == udp then
-		local dstport = linux.ntoh16(skb:getuint16(thoff + 2))
+		local dstport = byteorder.ntoh16(skb:getuint16(thoff + 2))
 		if dstport == dns then
 			local qoff = thoff + 20
 			local name = get_domain(skb, qoff)
--- a/examples/dnsdoctor/common.lua
+++ b/examples/dnsdoctor/common.lua
@@ -7,6 +7,7 @@
 
 local nf = require("netfilter")
 local linux = require("linux")
+local byteorder = require("byteorder")
 local action = nf.action
 local dns = 0x35
 
@@ -18,14 +19,14 @@ local function get_domain(skb, off)
 end
 
 function common.hook(skb, thoff, target_dns, target_ip, dst_ip, packet_dst)
-	if packet_dst ~= linux.hton32(dst_ip) then
+	if packet_dst ~= byteorder.hton32(dst_ip) then
 		return action.ACCEPT
 	end
 
-	local srcport = linux.ntoh16(skb:getuint16(thoff))
+	local srcport = byteorder.ntoh16(skb:getuint16(thoff))
 	if srcport == dns then
 		local dnsoff = thoff + 8
-		local nanswers = linux.ntoh16(skb:getuint16(dnsoff + 6))
+		local nanswers = byteorder.ntoh16(skb:getuint16(dnsoff + 6))
 
 		-- check the domain name
 		dnsoff = dnsoff + 12
@@ -34,9 +35,9 @@ function common.hook(skb, thoff, target_dns, target_ip, dst_ip, packet_dst)
 			dnsoff = dnsoff + nameoff + 4 -- skip over type, label fields
 			-- iterate over answers
 			for i = 1, nanswers do
-				local atype = linux.hton16(skb:getuint16(dnsoff + 2))
+				local atype = byteorder.hton16(skb:getuint16(dnsoff + 2))
 				if atype == 1 then
-					skb:setuint32(dnsoff + 12, linux.hton32(target_ip))
+					skb:setuint32(dnsoff + 12, byteorder.hton32(target_ip))
 				end
 				dnsoff = dnsoff + 16
 			end
--- a/gensymbols.sh
+++ b/gensymbols.sh
@@ -2,6 +2,7 @@
 # SPDX-FileCopyrightText: (c) 2023-2026 Ring Zero Desenvolvimento de Software LTDA
 # SPDX-License-Identifier: MIT OR GPL-2.0-only
 
-cat "$@" | sed "s/.*luaconf.h.*//g" | cpp -Ilua/ -D_KERNEL -DLUNATIK_GENSYMBOLS -E -pipe | \
+sed "s/.*luaconf.h.*//g" "$@" | \
+        ${CC} -Ilua/ -D_KERNEL -DLUNATIK_GENSYMBOLS -E - | \
 	grep API | sed "s/.*(\(\w*\))\s*(.*/EXPORT_SYMBOL(\1);/g"
 
--- a/lib/luabyteorder.c
+++ b/lib/luabyteorder.c
@@ -1,0 +1,191 @@
+/*
+* SPDX-FileCopyrightText: (c) 2023-2026 Ring Zero Desenvolvimento de Software LTDA
+* SPDX-License-Identifier: MIT OR GPL-2.0-only
+*/
+
+#include <linux/module.h>
+#include <linux/byteorder/generic.h>
+
+#include <lunatik.h>
+
+/***
+* Byte Order Conversion
+* @section byte_order
+*/
+#define LUABYTEORDER_BYTESWAPPER(swapper, T)		\
+static int luabyteorder_##swapper(lua_State *L)		\
+{							\
+	T x = (T)luaL_checkinteger(L, 1);		\
+	lua_pushinteger(L, (lua_Integer)swapper(x));	\
+	return 1;					\
+}
+
+/***
+* Converts a 16-bit integer from host byte order to big-endian byte order.
+* @function htobe16
+* @tparam integer num The 16-bit integer in host byte order.
+* @treturn integer The integer in big-endian byte order.
+*/
+LUABYTEORDER_BYTESWAPPER(cpu_to_be16, u16);
+
+/***
+* Converts a 32-bit integer from host byte order to big-endian byte order.
+* @function htobe32
+* @tparam integer num The 32-bit integer in host byte order.
+* @treturn integer The integer in big-endian byte order.
+*/
+LUABYTEORDER_BYTESWAPPER(cpu_to_be32, u32);
+
+/***
+* Converts a 16-bit integer from host byte order to little-endian byte order.
+* @function htole16
+* @tparam integer num The 16-bit integer in host byte order.
+* @treturn integer The integer in little-endian byte order.
+*/
+LUABYTEORDER_BYTESWAPPER(cpu_to_le16, u16);
+
+/***
+* Converts a 32-bit integer from host byte order to little-endian byte order.
+* @function htole32
+* @tparam integer num The 32-bit integer in host byte order.
+* @treturn integer The integer in little-endian byte order.
+*/
+LUABYTEORDER_BYTESWAPPER(cpu_to_le32, u32);
+
+/***
+* Converts a 16-bit integer from big-endian byte order to host byte order.
+* @function be16toh
+* @tparam integer num The 16-bit integer in big-endian byte order.
+* @treturn integer The integer in host byte order.
+*/
+LUABYTEORDER_BYTESWAPPER(be16_to_cpu, u16);
+
+/***
+* Converts a 32-bit integer from big-endian byte order to host byte order.
+* @function be32toh
+* @tparam integer num The 32-bit integer in big-endian byte order.
+* @treturn integer The integer in host byte order.
+*/
+LUABYTEORDER_BYTESWAPPER(be32_to_cpu, u32);
+
+/***
+* Converts a 16-bit integer from little-endian byte order to host byte order.
+* @function le16toh
+* @tparam integer num The 16-bit integer in little-endian byte order.
+* @treturn integer The integer in host byte order.
+*/
+LUABYTEORDER_BYTESWAPPER(le16_to_cpu, u16);
+
+/***
+* Converts a 32-bit integer from little-endian byte order to host byte order.
+* @function le32toh
+* @tparam integer num The 32-bit integer in little-endian byte order.
+* @treturn integer The integer in host byte order.
+*/
+LUABYTEORDER_BYTESWAPPER(le32_to_cpu, u32);
+
+/***
+* Converts a 64-bit integer from host byte order to big-endian byte order.
+* @function htobe64
+* @tparam integer num The 64-bit integer in host byte order.
+* @treturn integer The integer in big-endian byte order.
+*/
+LUABYTEORDER_BYTESWAPPER(cpu_to_be64, u64);
+
+/***
+* Converts a 64-bit integer from host byte order to little-endian byte order.
+* @function htole64
+* @tparam integer num The 64-bit integer in host byte order.
+* @treturn integer The integer in little-endian byte order.
+*/
+LUABYTEORDER_BYTESWAPPER(cpu_to_le64, u64);
+
+/***
+* Converts a 64-bit integer from big-endian byte order to host byte order.
+* @function be64toh
+* @tparam integer num The 64-bit integer in big-endian byte order.
+* @treturn integer The integer in host byte order.
+*/
+LUABYTEORDER_BYTESWAPPER(be64_to_cpu, u64);
+
+/***
+* Converts a 64-bit integer from little-endian byte order to host byte order.
+* @function le64toh
+* @tparam integer num The 64-bit integer in little-endian byte order.
+* @treturn integer The integer in host byte order.
+*/
+LUABYTEORDER_BYTESWAPPER(le64_to_cpu, u64);
+
+static const luaL_Reg luabyteorder_lib[] = {
+/***
+* Converts a 16-bit integer from network (big-endian) byte order to host byte order.
+* @function ntoh16
+* @tparam integer num The 16-bit integer in network byte order.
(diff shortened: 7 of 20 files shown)


lib/luaconstants.h
```diff
@@ -0,0 +1,19 @@
...
+#include <linux/if_ether.h>
+
+#define CONST(x) CONST_##x = x,
+
+enum {
+CONST(ETH_P_ALL)
+CONST(ETH_P_IP)
+CONST(ETH_P_ARP)
+CONST(ETH_P_LLDP)
+CONST(ETH_P_IPV6)
```
lneto: actually, the idea is to autogen these fields, instead of adding them manually.. see gensymbols

lneto: btw, I think it might be easier to do it in Lua instead of using sed + cpp as we do on autogen.. but, it's up to who gonna implement it ;-)

sneaky-potato: > actually, the idea is to autogen these fields, instead of adding them manually.. see gensymbols

My idea was to whitelist some constants and then get them inside a Lua table. On what criteria do we autogen these fields from? I could regex match for constants matching `ETH_P_*`, do you have something else in mind?

lneto: Yes, I would apply some pattern matching to autogenerate our constants.

Makefile
```diff
@@ -10,17 +10,20 @@ BTF_INSTALL_PATH = ${MODULES_RELEASE_PATH}/build
...
 SCRIPTS_INSTALL_PATH := ${MODULES_PATH}/lua
+ETH_HEADER_PATH := ${MODULES_RELEASE_PATH}/build/include/uapi/linux/if_ether.h
+LUA_CONSTANTS := lib/constants.lua
 
 
 LUNATIK_INSTALL_PATH = /usr/local/sbin
 LUNATIK_EBPF_INSTALL_PATH = /usr/local/lib/bpf/lunatik
 MOONTASTIK_RELEASE ?= v0.1c
 LUA_API = lua/lua.h lua/lauxlib.h lua/lualib.h
+
```
lneto:
```suggestion
```

Makefile
```diff
@@ -10,17 +10,20 @@ BTF_INSTALL_PATH = ${MODULES_RELEASE_PATH}/build
 MODULES_BUILD_PATH ?= ${BTF_INSTALL_PATH}
 MODULES_INSTALL_PATH := ${MODULES_RELEASE_PATH}/kernel
 SCRIPTS_INSTALL_PATH := ${MODULES_PATH}/lua
+ETH_HEADER_PATH := ${MODULES_RELEASE_PATH}/build/include/uapi/linux/if_ether.h
```
lneto: perhaps it would be better to config this on the lua script itself.. I think it would be easier to evolve it for parsing other hearders..

genconstants.lua
```diff
@@ -0,0 +1,22 @@
...
+print("-- auto-generated, do not edit")
+print("-- source: " .. header)
+print("")
+print("return {")
+
+for line in io.lines(header) do
+	-- Match only literal ETH_P_* defines
+	local name, value = line:match(pattern)
+	if name and value then
+		print(string.format("  %s = %s,", name, value))
```
lneto: we aren't handling defines conditioned to particular kernel versions anywhere, right?

sneaky-potato: this is a point which came to my mind as well
So conditional defines and indirect defines will not be handled if we continue with this approach
I suppose using cpp will eliminate these issue, but for network constants (if_ether.h) there are no conditional defines and indirect defines

lneto: we can do this in two steps, right? as we do with cpp + sed on gensymbols, but with cpp + lua here

examples/lldpd.lua
```diff
@@ -33,10 +31,10 @@ local config = {
 	},
 }
 
-local ethertype = string.pack(">I2", ETH_P_LLDP)
+local ethertype = string.pack(">I2", constants.ETH_P_LLDP)
 
 local function get_src_mac()
-	local rx <close> = raw.bind(ETH_P_ALL)
+	local rx <close> = raw.bind(constants.ETH_P_ALL)
```
lneto: perhaps it would be good to have namespaces.. also I'm not fond of constants.. perhaps, we can add it to different files under `linux` umbrella.. e.g.,

```
eth = require("linux/eth")
...
	local rx <close> = raw.bind(eth.ALL)
```

perhaps not `defines` instead of `linux`? not sure which is better.. `uapi`, `userspace`, `user`? not sure ;-)

Makefile
sneaky-potato: resolved in latest push

examples/lldpd.lua
sneaky-potato: I agree, pushed a script which regenerate constants under namespaces. We can add more headers in future.

lib/linux/eth.lua
```diff
@@ -0,0 +1,90 @@
+-- auto-generated, do not edit
```
lneto: this file shouldn't be added to this commit, right?

lib/linux/eth.lua
```diff
@@ -0,0 +1,90 @@
...
+-- kernel: 6.12.63-1-lts
+
+local eth = {}
+
+eth._1588 = 0x88F7
```
lneto: I wouldn't prefix numbers like this.. it's more natural to lua to have
```suggestion
eth["1588"] = 0x88F7
```

lib/linux/eth.lua
```diff
@@ -0,0 +1,90 @@
...
+eth.TDLS = 0x890D
+eth.TEB = 0x6558
+eth.TIPC = 0x88CA
+eth.TRAILER = 0x001C
+eth.TSN = 0x22F0
+eth.WCCP = 0x883E
+eth.X25 = 0x0805
+eth.XDSA = 0x00F8
+
+return eth
```
lneto: let's include a `\n` on autogen, please

lib/linux/eth.lua
```diff
@@ -0,0 +1,90 @@
+-- auto-generated, do not edit
+-- kernel: 6.12.63-1-lts
```
lneto: that's a good approach to have this metadata here.. =)

genconstants.lua
```diff
@@ -0,0 +1,106 @@
...
+	local defs = {}
+	for name, value in text:gmatch("#define%s+(" .. prefix .. "%w+)%s+(%S+)") do
+		defs[name] = value
+	end
+	return defs
+end
+
+local function sanitize(name)
+	-- add a single underscore if name starts with digit
+	-- example: ETH_P_1588 = 0x88F7
```
lneto: we can use keys as strings for all..

lneto:
hey @sneaky-potato, can you do an exercise on adding [this constants](https://github.com/luainkernel/lunatik/blob/master/lib/lualinux.c#L416) as well? I just received a report that it's breaking MIPS build due to the lack of some defines.. also, it's a good a way to make sure your patch is not tied to a single set of defines..

lib/linux/eth.lua
lneto: Or we can keep "P_" prefix.. not sure what's better.. what do you think?

genconstants.lua
sneaky-potato: Yeah, that will make things straight forward

sneaky-potato:
I was able to generate constants mentioned [here](https://github.com/luainkernel/lunatik/blob/e32eb67b68a4d472835996e5c86ea689eec5a22c/lib/lualinux.c#L416) with my script. I have included the spec for generating these constants. Just try `make` in my branch and you can verify

lib/linux/eth.lua
sneaky-potato: I think string would be best.

marcelstanley:
> hey @sneaky-potato, can you do an exercise on adding [this constants](https://github.com/luainkernel/lunatik/blob/master/lib/lualinux.c#L416) as well? I just received a report that it's breaking MIPS build due to the lack of some defines.. also, it's a good a way to make sure your patch is not tied to a single set of defines..

@lneto Actually the issue I found on MIPS is related to the signal API. As the [man page says](https://man7.org/linux/man-pages/man7/signal.7.html), some signals are absent in some architectures. On MIPS, for example, [SIGSTKFLT](https://github.com/luainkernel/lunatik/blob/e32eb67b68a4d472835996e5c86ea689eec5a22c/lib/lualinux.c#L495) is not found and the build fails.

If we can tackle the generation of these contants as well, it would be great, @sneaky-potato. Thanks!

sneaky-potato:
> If we can tackle the generation of these constants as well, it would be great

@marcelstanley @lneto Thanks, this helps clarify the issue.
I could extend the constants generator to parse `uapi/linux/signal.h` and emit a `lib/linux/signal.lua` module containing only the signal constants that are actually defined for the running kernel.

If that direction sounds reasonable, we could then consider dropping the hardcoded signal table in `lualinux.c` in a separate PR.

genconstants.lua
```diff
@@ -0,0 +1,107 @@
...
+}
+
+local function exit(msg)
+	io.stderr:write("genconstants.lua: " .. msg .. "\n")
+	os.exit(1)
+end
+
+local function preprocess(path, base)
+	local cmd = table.concat({
+		"cpp",
```
lneto: please, mind this PR https://github.com/luainkernel/lunatik/pull/372, we should grab CC from the build environment to enable crosscompiling

genconstants.lua
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+local function exit(msg)
+	io.stderr:write("genconstants.lua: " .. msg .. "\n")
+	os.exit(1)
+end
+
+local function preprocess(path, base)
+	local cmd = table.concat({
```
lneto: I think it would be clearer to have just a command line here instead of this table..

MarcFontaine:
I would like to add a comment here.
If you find it relevant, I'm happy to discuss it further.
`Generate linux constants from /lib/modules/$(uname -r)/build/include/uapi/linux/*.h for lua usage`
is not a good idea.
It is highly problematic, when a build pulls in data from a filesystem path that is not managed by the build system.
This gives up on reproducibility, The build will work on one machine and not on an other.
It will make it impossible to cross compile the project ($(uname -r) is only valid for the host machine not the target machine)
What one wants it a `hermetic build`.

The good news is, I have successfully build lunatik on NixOS (with some minor changes).
The buildsystem of NixOS runs builds in a sandbox and the fact that it was easy to build lunatik on NixOS is shows,
that the project at the moment has the property of an `hermetic` build.

To access `uapi/linux/*.h` one should call the symbol generator from within `Kbuild` and respect the
include paths that are set by the kernel build infrastructure.

lneto:
> I would like to add a comment here. If you find it relevant, I'm happy to discuss it further. `Generate linux constants from /lib/modules/$(uname -r)/build/include/uapi/linux/*.h for lua usage` is not a good idea. It is highly problematic, when a build pulls in data from a filesystem path that is not managed by the build system. This gives up on reproducibility, The build will work on one machine and not on an other. It will make it impossible to cross compile the project ($(uname -r) is only valid for the host machine not the target machine) What one wants it a `hermetic build`.

You are absolutely right. Thanks for bringing this up =). We should retrieve the path from the build environment.

> The good news is, I have successfully build lunatik on NixOS (with some minor changes). The buildsystem of NixOS runs builds in a sandbox and the fact that it was easy to build lunatik on NixOS is shows, that the project at the moment has the property of an `hermetic` build.

It's good to hear. BTW, you are welcome to update our README with instructions to NixOS.

> To access `uapi/linux/*.h` one should call the symbol generator from within `Kbuild` and respect the include paths that are set by the kernel build infrastructure.

You're welcome to add inline suggestions on this (or any other PR).

Thanks!

sneaky-potato:
> You are absolutely right. Thanks for bringing this up =). We should retrieve the path from the build environment.

Right now the approach is: top level Makefile provides value of CC, kernel release and the header path and lua script generates the constants (by invoking $CC) based on a configurable spec.

Does this align with the goal of hermetic builds? Or do we want to invoke symbol generation from the build system itself?
In the latter case  we could have a dedicated Makefile inside `lib/linux/Makefile` invoked via top level Makefile.

lneto:
> > You are absolutely right. Thanks for bringing this up =). We should retrieve the path from the build environment.
> 
> Right now the approach is: top level Makefile provides value of CC, kernel release and the header path and lua script generates the constants (by invoking $CC) based on a configurable spec.
> 
> Does this align with the goal of hermetic builds? Or do we want to invoke symbol generation from the build system itself? In the latter case we could have a dedicated Makefile inside `lib/linux/Makefile` invoked via top level Makefile.

I think it's safe to rely on the Makefile; one way to validate this is through our OpenWRT build..

sneaky-potato:
> I think it's safe to rely on the Makefile; 

Just to clarify: do you mean that symbol generation should be fully driven by the build system?
If so, it might be cleaner to follow the existing gensymbols.sh pattern and move this logic into a small shell script invoked by the Makefile, with the configuration (which headers/prefixes) living there, rather than in Lua.

.gitignore
```diff
@@ -12,4 +12,5 @@ lunatik_sym.h
 moontastik*
 dnstest
 doc/
+lib/linux/*.lua
```
lneto: we shouldn't do this.. perhaps we should have a separate dir for autogen files.. we might want to have things like linux/foo.lua

lneto:
> > I think it's safe to rely on the Makefile;
> 
> Just to clarify: do you mean that symbol generation should be fully driven by the build system? 

I meant that we should rely on the build system paths and defines for the target platform. For instance, this script should be integrated on https://github.com/luainkernel/openwrt_feed/blob/openwrt-23.05/kernel/lunatik/Makefile#L112.

> If so, it might be cleaner to follow the existing gensymbols.sh pattern and move this logic into a small shell script invoked by the Makefile, with the configuration (which headers/prefixes) living there, rather than in Lua.

What would be the difference between shell and Lua in this sense? One advantage I see regarding shell is portability, though I think this logic would be clearer in Lua. I would say if we can easily integrate such a script in our openwrt feed, it would be good enough. What do you think? I'm ok on using shells script as well, I just think the code is getting a bit complex and Lua gives us more flexibility for parsing. For instance, we might want to integrate enums as well.

sneaky-potato:
I also noticed that relying on paths like `$(MODULES_PATH)/$(KERNEL_RELEASE)/build/include/uapi/linux/*.h` is not distribution-independent.
Instead, we could extract UAPI constants via the compiler’s default include search path, for example:
```bash
echo '#include <linux/signal.h>' | cpp -E -dM - | grep '#define SIG'
```

This way we avoid hard coded paths. Not sure if all distros will have kernel headers symlinked at `/lib/module/$(uname -r)/build/include/uapi/linux`, kernel build trees are not guaranteed to be present on all systems…”
@lneto what do you think?

lneto:
@sneaky-potato I took a better look on this, it turns out that our Makefile is already host-oriented, we use `uname -r`. I don't think this PR is the right place to change it. I would just pass the kernel headers path from the makefile to your script. Thus, on openwrt feed we can do the same using $KDIR as argument.

genconstants.lua
```diff
@@ -0,0 +1,93 @@
...
+		local out, err = io.open(filename, "w")
+		if not out then
+			exit("cannot open " .. filename .. ": " .. err)
+		end
+
+		out:write("-- auto-generated, do not edit\n")
+		out:write("-- kernel: " .. spec.header .. "\n\n")
+		out:write("local " .. spec.module_name .. " = {}\n\n")
+
+		local keys = {}
```
lneto: let's use a more descriptive name here.. what's `key` in this context? a `define`?

genconstants.lua
```diff
@@ -0,0 +1,93 @@
...
+				defs[k]
+			))
+		end
+
+		out:write("\nreturn " .. spec.module_name .. "\n\n")
+		out:close()
+	end
+end
+
+main()
```
lneto: we don't use much this pattern in Lua.. I would modularize your "main" function a bit more and just have your main logic inline here..

genconstants.lua
```diff
@@ -0,0 +1,93 @@
+--
+-- SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+local MODULE_PATH = "./lib/linux"
```
lneto: this should be the headers path and come from a argument..

lneto:
> @sneaky-potato I took a better look on this, it turns out that our Makefile is already host-oriented, we use `uname -r`. I don't think this PR is the right place to change it. I would just pass the kernel headers path from the makefile to your script. Thus, on openwrt feed we can do the same using $KDIR as argument.

we could do something like [this](https://github.com/luainkernel/openwrt_feed/blob/openwrt-23.05/kernel/lunatik/Makefile#L114).

genconstants.lua
sneaky-potato: One clarification question:

Instead of passing an explicit UAPI header path, would something like the following be considered problematic in this script?
```shell
echo '#include <linux/signal.h>' | cpp -E -dM - | grep '#define SIG'
```

This avoids hard-coding kernel header paths and relies on the compiler’s default include resolution.
I’m not sure if this is acceptable from a build-system / portability standpoint (e.g. host vs OpenWRT, glibc headers vs kernel UAPI headers), so I wanted to ask before going further.

lneto: I think it could work if you use CFLAGS from the build system, however, I think it would be simpler just to have a var with kernel headers path.. it's already defined in our makefile as you are using.. we just need to pass the right one on openwrt build.. what changes from your script perspective is that you get this from a argument.. just it..

lneto: actually, I think we would need to call cpp anyway to evaluate the ifdefs of the platform..

genconstants.lua
sneaky-potato: Yes, specifically the macro name (e.g., `ETH_P_ALL` in case of `#define ETH_P_ALL 0x0003`)

genconstants.lua
sneaky-potato: There is one more issue
for signal constants, we have some indirect defines like [SIGPOLL](https://elixir.bootlin.com/linux/v6.18.4/source/arch/x86/include/uapi/asm/signal.h#L52) and [SIGRTMAX](https://elixir.bootlin.com/linux/v6.18.4/source/arch/x86/include/uapi/asm/signal.h#L62)
Right now the script produces constants like 
```lua
signal["POLL"] = SIGIO
signal["RTMAX"] = _NSIG
```
I checked whether this can be solved purely with `cpp -dM`, but that mode only dumps macro definitions and does not resolve aliases.
One approach could be to use a small compiled helper, but I am not sure about how it will grow with complexity as more constants are added. I am still pondering how to solve this.

lneto: `signal["POLL"] = SIGIO` could be "easily" handled generating `signal["POLL"] = signal["SIGIO"]`, right? `RTMAX`, I would ignore for now and create a issue for handling this later as an improvement.. we could just raise a warning if it's not number or another prefixed define.. what do you think?

genconstants.lua
```diff
@@ -0,0 +1,102 @@
...
+	local cmd = string.format("%s -E -dM -I%s %s/%s", cc, uapi_linux_path, uapi_linux_path, header)
+
+	local p = io.popen(cmd)
+
+	if not p then
+		exit("failed to run cpp on " .. header)
+	end
+
+	local out = p:read("*a")
+	p:close()
```
lneto: `<close>`

genconstants.lua
```diff
@@ -0,0 +1,102 @@
...
+	io.stderr:write("genconstants.lua: " .. msg .. "\n")
+	os.exit(1)
+end
+
+local function preprocess(header, uapi_linux_path)
+	local cc = os.getenv("CC")
+	local cmd = string.format("%s -E -dM -I%s %s/%s", cc, uapi_linux_path, uapi_linux_path, header)
+
+	local p = io.popen(cmd)
+
```
lneto:
```suggestion
	local p = io.popen(cmd)
```

genconstants.lua
```diff
@@ -0,0 +1,102 @@
...
+}
+
+local function exit(msg)
+	io.stderr:write("genconstants.lua: " .. msg .. "\n")
+	os.exit(1)
+end
+
+local function preprocess(header, uapi_linux_path)
+	local cc = os.getenv("CC")
+	local cmd = string.format("%s -E -dM -I%s %s/%s", cc, uapi_linux_path, uapi_linux_path, header)
```
lneto: we should move this to an aux to avoid repeat this parm..

Makefile
```diff
@@ -10,6 +10,7 @@ BTF_INSTALL_PATH = ${MODULES_RELEASE_PATH}/build
 MODULES_BUILD_PATH ?= ${BTF_INSTALL_PATH}
 MODULES_INSTALL_PATH := ${MODULES_RELEASE_PATH}/kernel
 SCRIPTS_INSTALL_PATH := ${MODULES_PATH}/lua
+UAPI_LINUX_PATH := ${MODULES_BUILD_PATH}/include/uapi/linux
```
lneto: I think it's better to pass the include path, not uapi/linux.. so we can parse other headers later..

Makefile
```diff
@@ -123,6 +125,10 @@ uninstall: scripts_uninstall modules_uninstall
 lunatik_sym.h: $(LUA_API) gensymbols.sh
 	${shell ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
 
+constants:
```
lneto: I think "defines" is clearer than "constants" for what we are doing here.. but I have no strong feelings on this..

genconstants.lua
```diff
@@ -0,0 +1,102 @@
...
+
+	table.sort(constant_keys)
+	return constants, constant_keys
+end
+
+local function resolve_indirect_macros(constants, constant_keys, prefix, module_name)
+	for _, macro in ipairs(constant_keys) do
+		local value = constants[macro]
+		if value and not value:match("^%-?%d") and not value:match("^0x%x+") then
+			if constants[value] then
```
lneto: this condition could be added to the if above.. `and ...`, we don't need to branch again here.. perhaps it's worth to move it to an aux predicate as well..

genconstants.lua
```diff
@@ -0,0 +1,102 @@
...
+	for macro, literal in cpp_output:gmatch("#define%s+(" .. prefix .. "%w+)%s+(%S+)") do
+		constants[macro] = literal
+		table.insert(constant_keys, macro)
+	end
+
+	table.sort(constant_keys)
+	return constants, constant_keys
+end
+
+local function resolve_indirect_macros(constants, constant_keys, prefix, module_name)
```
lneto: let's avoid long names like this.. both `resolve` and `macros` don't say much in this context..

genconstants.lua
```diff
@@ -0,0 +1,102 @@
...
+	for _, macro in ipairs(constant_keys) do
+		local value = constants[macro]
```
lneto: let's use more meaningful names.. avoid things like `keys` and `value`

genconstants.lua
```diff
@@ -0,0 +1,102 @@
...
+	out:write("-- kernel: " .. KERNEL .. "\n\n")
+	out:write("local " .. spec.module_name .. " = {}\n\n")
+
+	for _, macro in ipairs(constant_keys) do
+		local lua_macro_name = macro:gsub("^" .. spec.prefix, "")
+		out:write(string.format("%s[\"%s\"] = %s\n", spec.module_name, lua_macro_name, constants[macro]))
+	end
+
+	out:write("\nreturn " .. spec.module_name .. "\n\n")
+	out:close()
```
lneto: `<close>`

genconstants.lua
lneto: `resolve_aliases`, `fix_aliases`.. couldn't think on anything better..

genconstants.lua
```diff
@@ -0,0 +1,102 @@
...
+	}
+}
+
+local function exit(msg)
+	io.stderr:write("genconstants.lua: " .. msg .. "\n")
+	os.exit(1)
+end
+
+local function preprocess(header, uapi_linux_path)
+	local cc = os.getenv("CC")
```
lneto: we don't need to consult env all the time, right? you can init this along with the arguments.. e.g.,
```
local CC = os.getenv("CC")
```

lneto: actually you can initiate the whole command bellow as `CC` previously.. it doesn't change..

sneaky-potato: Resolved  in latest push

genconstants.lua
sneaky-potato: Fixed in last push

genconstants.lua
sneaky-potato: Refactored and made this logic simpler

Makefile
sneaky-potato: Right, fixed in latest push

genconstants.lua
sneaky-potato: Fixed in last push

lneto:
Hey @sneaky-potato, do you mind if I merge this to an integration branch so I can make some fixups on my own and then we can merge it to master? I think we can work together on this. Let me know.

sneaky-potato:
Hello @lneto 
That sounds good to me, thanks :D

If possible, I'd love to take a look at the changes before it lands on master so I can learn from them, and it'd be great to keep authorship credit (or add a Co-authored-by) when it is merged.

Thanks again.

lneto:
Of course, there's no doubt on the authorship =). I just want to add some fixups on your original commit and, of course, have your review on them. Thanks.

lneto:
moved to https://github.com/luainkernel/lunatik/pull/395

PR #366, opened 2026-01-05
revert support for hardirq and refactor luahid

lneto:
tried to simplify the most; it's still have room for improvement but it was the best I could do during the holidays..

@sneaky-potato @qrsikno2 would be of great help if you can test it.. @glk0 I would really appreciate if you can review as well..

update: while refactoring luahid, we noticed that we actually don't need to run on a hard irq.

### DIFF
--- a/examples/gesture.lua
+++ b/examples/gesture.lua
@@ -1,67 +1,69 @@
 --
--- SPDX-FileCopyrightText: (c) 2025 Jieming Zhou <qrsikno@gmail.com>
+-- SPDX-FileCopyrightText: (c) 2025-2026 Jieming Zhou <qrsikno@gmail.com>
 -- SPDX-License-Identifier: MIT OR GPL-2.0-only
 --
 
--- Driver for Qemu's USB mouse with gesture(dragging),
--- swipeing up to lock the mouse and swipeing down to unlock it.
+-- Driver for QEMU's USB mouse with gesture (dragging),
+-- swiping right to lock the mouse and swiping left to unlock it.
 
 local hid = require("hid")
 
-local driver = {
-	name = "gesture",
-	id_table = {
-		{ vendor = 0x0627, product = 0x0001 }
-	}
-}
+local driver = {name = "gesture", id_table = {{vendor = 0x0627, product = 0x0001}}}
 
-function driver:probe(devid)
-	return { x = 0, y = 0, drag = false, lock = false }
+local threshold = 2
+
+local function debug(fmt, ...)
+	print(string.format(fmt, ...))
 end
 
-function driver:raw_event(hdev, state, report, raw_data)
-	local btn = raw_data:getbyte(0)
-	local dx_byte = raw_data:getbyte(1)
-	local dy_byte = raw_data:getbyte(2)
-	-- complement conversion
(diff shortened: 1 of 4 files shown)


lib/luahid.c
```diff
@@ -350,75 +295,36 @@ static int luahid_raw_event(struct hid_device *hdev, struct hid_report *report,
...
+	driver->report_fixup = luahid_report_fixup;
+	driver->raw_event = luahid_raw_event;
 
 	lunatik_setruntime(L, hid, hid);
 	lunatik_getobject(hid->runtime);
-	lunatik_registerobject(L, 2, object);
+	lunatik_registerobject(L, 1, object);
 
-	if (__hid_register_driver(user_driver, THIS_MODULE, KBUILD_MODNAME) != 0) {
+	if (__hid_register_driver(driver, THIS_MODULE, KBUILD_MODNAME) != 0) {
```
sheharyaar: Why are we using `__hid_register_driver` instead of just `hid_register_driver` ?

lib/luanetfilter.c
```diff
@@ -104,7 +104,7 @@ static inline unsigned int luanetfilter_docall(luanetfilter_t *luanf, struct sk_
 	if (likely(luanf->mark != skb->mark))
 		goto out;
 
-	lunatik_runbh(luanf->runtime, luanetfilter_hook_cb, ret, luanf, skb);
+	lunatik_run(luanf->runtime, luanetfilter_hook_cb, ret, luanf, skb);
```
sheharyaar, line 107: I think we should touch these files in a different PR, since the PR name will be misleading, or we can change the PR name.

lneto: the thing is that we need to bundle those changes together as we are removing/changing a core API, otherwise we will break the base merging them separately. I've just update the PR title and description; please take a look =).

lib/luahid.c
lneto: that's a good question, actually.. I recall @qrsikno2 mentioning that it was needed for having multiple drivers in the same module, but it doesn't seem to be the case. I've just updated the PR with this change. Thanks!

examples/gesture.lua
```diff
@@ -1,67 +1,65 @@
...
+local function count(state, direction)
+	if direction ~= "neutral" then
+		if state.direction == direction then
+			state.count = state.count + 1
+		else
+			state.count = 0
+			state.direction = direction
+		end
+	end
+	return state.count
```
qrsikno2, line 36: It's necessary to have `state.count` initialized to prevent `nil` "exception" between comparison with `threshold`.

examples/gesture.lua
```diff
@@ -1,67 +1,65 @@
...
-function driver:probe(devid)
-	return { x = 0, y = 0, drag = false, lock = false }
+local threshold = 2
+
+function driver:probe(id)
+	self.state = {x = 0, lock = false}
+end
+
+local function forward(x0, x1)
+	return (x0 >= 255 and x1 <= 25) or (x1 > x0)
```
qrsikno2: Note that some devices (like the QEMU USB Tablet) use 16-bit absolute coordinates. 

We might need to adjust this depending on the target device

lib/luahid.c
qrsikno2: Correct.

PR #371, opened 2026-01-08
Create install path for lunatik binary

MarcFontaine:
Create the install path for the lunatik binary.
This makes it possible to install to new paths.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ scripts_install:
 	${INSTALL} -m 0644 lib/socket/*.lua ${SCRIPTS_INSTALL_PATH}/socket
 	${INSTALL} -m 0644 lib/syscall/*.lua ${SCRIPTS_INSTALL_PATH}/syscall
 	${INSTALL} -m 0644 lib/crypto/*.lua ${SCRIPTS_INSTALL_PATH}/crypto
-	${INSTALL} -m 0755 bin/lunatik ${LUNATIK_INSTALL_PATH}
+	${INSTALL} -D -m 0755 bin/lunatik ${LUNATIK_INSTALL_PATH}/bin/lunatik
 
 scripts_uninstall:
 	${RM} ${SCRIPTS_INSTALL_PATH}/driver.lua


Makefile
```diff
@@ -50,6 +50,7 @@ scripts_install:
 	${INSTALL} -m 0644 lib/socket/*.lua ${SCRIPTS_INSTALL_PATH}/socket
 	${INSTALL} -m 0644 lib/syscall/*.lua ${SCRIPTS_INSTALL_PATH}/syscall
 	${INSTALL} -m 0644 lib/crypto/*.lua ${SCRIPTS_INSTALL_PATH}/crypto
+	${MKDIR} ${LUNATIK_INSTALL_PATH}
```
lneto: perhaps a better approach is to `LUNATIK_INSTALL_PATH ?= /usr/local/sbin` allowing us to redefine this variable on `make install`; what do you think? I'm good with both approaches..

carloslack: Also, you should be able to use `${INSTALL} -d` and avoid this mkdir call

MarcFontaine: `LUNATIK_INSTALL_PATH := /usr/local/sbin` allows overwriting as well.
`${MKDIR}` followed by `${INSTALL}` is the pattern that is used everywhere in the Makefile.
I did not want to change the pattern that was used before.

MarcFontaine: > perhaps a better approach is to `LUNATIK_INSTALL_PATH ?= /usr/local/sbin` allowing us to redefine this variable on `make install`; what do you think? I'm good with both approaches..

This is not enough. In the NixOS case the path will be something like.
`/nix/store/brari1r0vbx4m3znd4rirpmh39nxslj5-lunatik-linux-module-nix-unstable/bin/lunatik`.
`/nix/store/brari1r0vbx4m3znd4rirpmh39nxslj5-lunatik-linux-module-nix-unstable` is created by the NixOS build system, but `bin` needs to be created in some way.

lneto: > Also, you should be able to use `${INSTALL} -d` and avoid this mkdir call

that makes sense.. @MarcFontaine can you try this?

MarcFontaine: ${INSTALL} -d  does not work for me

lneto (approved):

Makefile
carloslack: @MarcFontaine my bad, I think it is `-D`, as stated in the manpage:

`-D     create all leading components of DEST except the last, or all components of --target-directory, then copy SOURCE to DEST`

I could run it nicely from command line, like:

`install -D file.txt ./test/file.txt`

MarcFontaine: You have `./test/file.txt` as argument to install
while the the Makefile uses `$(LUNATIK_INSTALL_PATH = /usr/local/bin`.
That is the PATH without the filename itself.
`install -D -m 0755 bin/lunatik /usr/local/bin` will create directory /usr/local/ and write a file named `bin`.

carloslack: @MarcFontaine I think it is OK to leave like this for now, I'm fine with it.

lneto: wouldn't be the case of simply `install -D -m 0755 bin/lunatik /usr/local/bin/lunatik`?

carloslack: Exactly, my eyes missed that, can you give a quick try @MarcFontaine ? It should work this way

MarcFontaine: > wouldn't be the case of simply `install -D -m 0755 bin/lunatik /usr/local/bin/lunatik`? 

That would break the convention that is used for the other installs in the Makefile. The path variables dont contain the filename.

Also a hardcoded filepath does not work

lneto: > > wouldn't be the case of simply `install -D -m 0755 bin/lunatik /usr/local/bin/lunatik`?
> 
> That would break the convention that is used for the other installs in the Makefile. The path variables dont contain the filename.

Sorry, I don't see why it's a problem.

> Also a hardcoded filepath does not work

I didn't mean to hardcode it, but to have something like this:

```
${INSTALL} -D -m 0755 bin/lunatik ${LUNATIK_INSTALL_PATH}/lunatik
```

doesn't it work for you?

MarcFontaine: Yes that works very well !

Makefile
```diff
@@ -30,7 +29,8 @@ all: lunatik_sym.h
 	CONFIG_LUNATIK_NETFILTER=m CONFIG_LUNATIK_COMPLETION=m \
 	CONFIG_LUNATIK_CRYPTO_SHASH=m CONFIG_LUNATIK_CRYPTO_SKCIPHER=m \
 	CONFIG_LUNATIK_CRYPTO_AEAD=m CONFIG_LUNATIK_CRYPTO_RNG=m \
-	CONFIG_LUNATIK_CRYPTO_COMP=m CONFIG_LUNATIK_CPU=m CONFIG_LUNATIK_HID=m
+	CONFIG_LUNATIK_CRYPTO_COMP=m CONFIG_LUNATIK_CPU=m CONFIG_LUNATIK_HID=m \
+	CONFIG_LUNATIK_ROOT=/tmp/
```
lneto: why do you need this?

Makefile
```diff
@@ -121,7 +121,7 @@ uninstall: scripts_uninstall modules_uninstall
 	depmod -a
 
 lunatik_sym.h: $(LUA_API) gensymbols.sh
-	${shell ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
+	${shell CC='$(CC)' ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
```
lneto: this belongs to your other PR, right?

Makefile
MarcFontaine: Sorry, I mixed up some commits..

PR #372, opened 2026-01-08
Use ${CC} instead of cpp in gensymbols.sh

MarcFontaine:
This fixes a build problem on NixOS.
cpp is not predefined in the kernel build environment on NixOS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,7 +121,7 @@ uninstall: scripts_uninstall modules_uninstall
 	depmod -a
 
 lunatik_sym.h: $(LUA_API) gensymbols.sh
-	${shell ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
+	${shell CC='$(CC)' ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
 
 moontastik_install_%:
 	[ $* ] || (echo "usage: make moontastik_install_TARGET" ; exit 1)
--- a/gensymbols.sh
+++ b/gensymbols.sh
@@ -2,6 +2,7 @@
 # SPDX-FileCopyrightText: (c) 2023-2026 Ring Zero Desenvolvimento de Software LTDA
 # SPDX-License-Identifier: MIT OR GPL-2.0-only
 
-cat "$@" | sed "s/.*luaconf.h.*//g" | cpp -Ilua/ -D_KERNEL -DLUNATIK_GENSYMBOLS -E -pipe | \
+sed "s/.*luaconf.h.*//g" "$@" | \
+        ${CC} -Ilua/ -D_KERNEL -DLUNATIK_GENSYMBOLS -E - | \
 	grep API | sed "s/.*(\(\w*\))\s*(.*/EXPORT_SYMBOL(\1);/g"
 


gensymbols.sh
```diff
@@ -2,6 +2,8 @@
 # SPDX-FileCopyrightText: (c) 2023-2026 Ring Zero Desenvolvimento de Software LTDA
 # SPDX-License-Identifier: MIT OR GPL-2.0-only
 
-cat "$@" | sed "s/.*luaconf.h.*//g" | cpp -Ilua/ -D_KERNEL -DLUNATIK_GENSYMBOLS -E -pipe | \
+set -euo pipefail
```
lneto: sorry, why do we need this?

Makefile
```diff
@@ -121,7 +121,7 @@ uninstall: scripts_uninstall modules_uninstall
 	depmod -a
 
 lunatik_sym.h: $(LUA_API) gensymbols.sh
-	${shell ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
+	${shell CC='$(CC)' ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
```
lneto, line 124: @sneaky-potato this might affect your PR as well..

gensymbols.sh
lneto: I think it's bashism, right? do you really need this to build?

gensymbols.sh
```diff
@@ -2,6 +2,8 @@
...
+cat "$@" | sed "s/.*luaconf.h.*//g" | \
+        ${CC} -Ilua/ -D_KERNEL -DLUNATIK_GENSYMBOLS -E -pipe -x c - | \
```
lneto: this didn't work for me; CC usually defines C compiler, please notice we are using C preprocessor here

gensymbols.sh
carloslack: there is some detail [here](https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425?permalink_comment_id=3799230#set--o-pipefail) how the `pipefail` will behave

I think this would mark the error code properly, in the case the fail happens previous to the pipe, like `cat "$@" | sed "s/.*luaconf.h.*//g"` when used with `pipefail` should guarantee a non-zero status code.

It depends on how you want that to behave

lneto: yeah.. it's bash, I think we shouldn't tie the build to a particular shell

MarcFontaine: `set -euo pipefail` makes the script fail if any command in the pipeline fails.
This makes the build fail early in case of an error.

Makefile
MarcFontaine: AFAIK `CC` is preddefined in make. This makes the definition from make available in gensymbols.sh.

gensymbols.sh
MarcFontaine: This surprises me. cpp does more or less the same as ${CC}  -E.
This calls the compiler-pipeline and -E means stop after running the preprocessor.

gensymbols.sh
carloslack: No need to `cat` command, you can pass `$@` as 2nd `sed` argument

Makefile
lneto: @MarcFontaine can you try this and see if it works for you?
```
cat "$@" | sed "s/.*luaconf.h.*//g" | cc -E - -Ilua/ -D_KERNEL -DLUNATIK_GENSYMBOLS -E -pipe | \
        grep API | sed "s/.*(\(\w*\))\s*(.*/EXPORT_SYMBOL(\1);/g"
```

MarcFontaine: The basic idea is that instead of using just the cpp that is installed. use the compiler that is defined
by make, that is $(CC). At the same time this makes is possible to overwrite the compiler in a standard way.
Changing the compiler is needed for example for cross-compiling lunatik for arm64 on an x64 system.
(I did the crosscompiling and it worked well)

gensymbols.sh
MarcFontaine: Would you tell me what error you see ?

Makefile
lneto: > The basic idea is that instead of using just the cpp that is installed. use the compiler that is defined by make, that is $(CC). At the same time this makes is possible to overwrite the compiler in a standard way. Changing the compiler is needed for example for cross-compiling lunatik for arm64 on an x64 system. (I did the crosscompiling and it worked well)

thing is CC is the compiler itself, not the preprocessor; we need `-E` to call the preprocessor. We don't need any definition from the target arch for gensymbols, we just want to figure it out exported Lua APIs, but it will indeed impact the @sneaky-potato's PR on autogen constants. I agree it's a good practice anyway =). Just notice, we cannot call C compiler direct here, we need CPP (C preprocessor) instead, for this, you can pass `-E`.

gensymbols.sh
lneto: > This surprises me. cpp does more or less the same as ${CC} -E. This calls the compiler-pipeline and -E means stop after running the preprocessor.

this exactly what we want to do here; running the cpp ;-).

lneto: > Would you tell me what error you see ?

it's not the issue I'm raising here; my point is that this script explicitly needs cpp, it doesn't need to compile anything, just evaluate the macros. We don't have any particular error with this particular version, but we have had depending on ifdefs and includes.

Anyways, this should address your problem and use the cross compiler; what's not needed here, but I agree it's a better practice.

```suggestion
        ${CC} -E - -Ilua/ -D_KERNEL -DLUNATIK_GENSYMBOLS -E -pipe -x c - | \
```

gensymbols.sh
lneto: > `set -euo pipefail` makes the script fail if any command in the pipeline fails. This makes the build fail early in case of an error.

yes, but it's bash-centric; for instance, it breaks the build on Ubuntu + dash

MarcFontaine: OK,I see `set -euo pipefail` is not enough to a actually stop the build.
I just noticed that the build as it is now just continues to compile if
` ./gensymbols.sh $(LUA_API) > lunatik_sym.h` fails and produces an empty `lunatik_sym.h`.
It will include an empty list of symbols and fail much later.
When cross building on NixOS neither `cc` nor `cpp` are available (by default).
By default, the build system will set ${CC} properly when calling `make`.

MarcFontaine:
Here is an updated version.

Makefile
MarcFontaine: > @MarcFontaine can you try this and see if it works for you?
> 
> ```
> cat "$@" | sed "s/.*luaconf.h.*//g" | cc -E - -Ilua/ -D_KERNEL -DLUNATIK_GENSYMBOLS -E -pipe | \
>         grep API | sed "s/.*(\(\w*\))\s*(.*/EXPORT_SYMBOL(\1);/g"
> ```

It works have way but not for the cross build.

lneto: > > [@MarcFontaine](https://github.com/MarcFontaine) can you try this and see if it works for you?
> > ```
> > cat "$@" | sed "s/.*luaconf.h.*//g" | cc -E - -Ilua/ -D_KERNEL -DLUNATIK_GENSYMBOLS -E -pipe | \
> >         grep API | sed "s/.*(\(\w*\))\s*(.*/EXPORT_SYMBOL(\1);/g"
> > ```
> 
> It works have way but not for the cross build.

But this https://github.com/luainkernel/lunatik/pull/372/changes#r2672227528 should, right?

lneto (commented):
if you change it to  `${CC} -E -` and squash your commits, I can merge it. Thanks.

lneto (approved):

Issue #369, opened 2026-01-06
implement errname() fallback when it's not present


lneto:
see https://github.com/luainkernel/lunatik/pull/343#issuecomment-3702611386

lneto:
@abhijeetw035 are you still willing to tackle this? please, let us know; I would like to have this fix on the next release.

abhijeetw035:
Yes @lneto happy to tackle this. will get started on implementing the lunatik_pusherrname helper . I'll open a PR soon.

PR #375, opened 2026-01-09
introduce lunatik_pusherrname helper for error handling

abhijeetw035:
This PR implements a central error handling helper to fix build failures on Linux kernels older than 6.7 (where `errname()` is not yet available).

It introduces a fallback mechanism that ensures `lunatik` compiles and runs correctly across different kernel versions, consolidating the error reporting logic into a single API.

## Changes

* **Introduced `lunatik_pusherrname(L, err)` in `lunatik_aux.c`:**
    * Implements a version check using `LINUX_VERSION_CODE`.
    * Uses the native `errname(err)` on Linux 6.7+.
    * Provides a fallback using `snprintf(..., "%pE", ...)` for older kernels.
* **Updated `lunatik_tryret` macro in `lunatik.h`:**
    * Replaced the direct `errname()` call with the new `lunatik_pusherrname()` helper.
    * This fixes compilation errors for all modules (Socket, Crypto, etc.) that use `lunatik_tryret`.
* **Refactored `lib/lualinux.c`:**
    * Updated `lualinux_errname` to use the new shared helper.
    * Removed unused headers (`<linux/errname.h>`, `<linux/version.h>`) and redundant logic.

## Related Issue
Fixes [#369](https://github.com/luainkernel/lunatik/issues/369)

### DIFF
--- a/lunatik_aux.c
+++ b/lunatik_aux.c
@@ -5,6 +5,8 @@
 
 #include <linux/slab.h>
 #include <linux/fs.h>
+#include <linux/version.h>
+#include <linux/errname.h>
 
 #include <lua.h>
 #include <lauxlib.h>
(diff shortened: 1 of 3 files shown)


lunatik_aux.c
```diff
@@ -63,6 +66,22 @@ int lunatik_loadfile(lua_State *L, const char *filename, const char *mode)
 }
 EXPORT_SYMBOL(lunatik_loadfile);
 
+void lunatik_pusherrname(lua_State *L, int err)
+{
+    if (err < 0)
+        err = -err;
```
lneto:
```suggestion
    err = abs(err);
```

lunatik_aux.c
```diff
@@ -5,6 +5,9 @@
 
 #include <linux/slab.h>
 #include <linux/fs.h>
+#include <linux/version.h>
+#include <linux/err.h>
+#include <linux/errname.h>
```
lneto: do we really need both? I had the impression one includes the other

lneto (approved):
LGTM; just small comments

lunatik_aux.c
abhijeetw035: I checked the headers again. you’re right that we don’t need to include it explicitly, but it’s actually `linux/fs.h` that ends up pulling in `linux/err.h`, not `linux/errname.h`. Since lunatik_aux already includes `fs.h` anyway (like we need it for `IS_ERR` and `ERR_PTR`), I’ll just drop the extra include.

PR #378, opened 2026-01-11
Require explicit (protocol, ifindex) pair in raw sockets bind()

### LINKED ISSUE #367: AF_PACKET listening socket cannot be bound on interface
From the documentation of `luasocket_bind` and more specifically [luasocket_checkaddr](https://github.com/luainkernel/lunatik/blob/72cc7be7a6ec1f6723ab96e3e7db7f3c0a531293/lib/luasocket.c#L74C2-L85C3), it seems that binding a listening AF_PACKET socket to a specific interface is not currently supported.

For AF_PACKET
- when the argument is a number, it is always interpreted as `sll_ifindex`, and no path sets `sll_protocol`. 
- string-based path copies raw bytes into `sockaddr_storage` but does not populate struct `sockaddr_ll` fields in a structured way.

For a listening raw socket bound to an interface, we would need something equivalent to: 
```c
sll_protocol = htons(ETH_P_ALL);
sll_ifindex  = ifindex;
```
Currently there’s no way to express this combination from Lua. Is this by design? If not, can we add proper AF_PACKET bind support for listening sockets?

sneaky-potato:
AF_PACKET sockets previously allowed ambiguous bind() arguments, leaving `sockaddr_ll.sll_protocol` uninitialized in some cases.
This change enforces an explicit (protocol, ifindex) pair when binding AF_PACKET sockets, ensuring `sll_protocol` is always initialized

Fixes #367

### DIFF
--- a/lib/luasocket.c
+++ b/lib/luasocket.c
@@ -1,5 +1,5 @@
 /*
-* SPDX-FileCopyrightText: (c) 2023-2025 Ring Zero Desenvolvimento de Software LTDA
+* SPDX-FileCopyrightText: (c) 2023-2026 Ring Zero Desenvolvimento de Software LTDA
 * SPDX-License-Identifier: MIT OR GPL-2.0-only
 */
 
@@ -57,7 +57,7 @@ static size_t luasocket_checkaddr(lua_State *L, struct socket *socket, struct so
 	if (addr->ss_family == AF_INET) {
 		struct sockaddr_in *addr_in = (struct sockaddr_in *)addr;
 		addr_in->sin_addr.s_addr = htonl((u32)luaL_checkinteger(L, ix));
-		addr_in->sin_port = htons((u16)luaL_checkinteger(L, ix + 1));
+		addr_in->sin_port = htons((u16)lunatik_checkinteger(L, ix + 1, 0, U16_MAX));
 		return sizeof(struct sockaddr_in);
 	}
 #ifdef CONFIG_UNIX
@@ -71,9 +71,10 @@ static size_t luasocket_checkaddr(lua_State *L, struct socket *socket, struct so
 		return sizeof(struct sockaddr_un);
 	}
 #endif
-	else if (addr->ss_family == AF_PACKET && lua_type(L, ix) == LUA_TNUMBER) {
+	else if (addr->ss_family == AF_PACKET) {
 		struct sockaddr_ll *addr_ll = (struct sockaddr_ll *)addr;
-		addr_ll->sll_ifindex = (int)lua_tointeger(L, ix);
+		addr_ll->sll_protocol = htons((u16)lunatik_checkinteger(L, ix, 0, U16_MAX));
+		addr_ll->sll_ifindex = (int)lunatik_checkinteger(L, ix + 1, 0, INT_MAX);;
 		return sizeof(struct sockaddr_ll);
 	}
 	else {
@@ -226,23 +227,23 @@ static int luasocket_receive(lua_State *L)
 *   - `AF_INET` (IPv4): An integer representing the IPv4 address (e.g., from `net.aton()`).
 *     Use `0` (or `net.aton("0.0.0.0")`) to bind to all available interfaces.
 *     The `port` argument is also required.
-*   - `AF_PACKET`:
-*     - If `addr` is an integer: It's treated as the network interface index (e.g., from `linux.ifindex("eth0")`).
-*       The `sll_protocol` field of the underlying `sockaddr_ll` structure would need to be set to `htons(ETH_P_ALL)` or similar
-*       if not binding to a specific protocol via other means (e.g. `socket.new`'s protocol argument for `AF_PACKET` might set this).
-*     - If `addr` is a string: It's a packed string directly representing parts of the `sockaddr_ll` structure
-*       (specifically, the fields after `sll_family`, like `sll_protocol`).
-*       Example from `tap.lua` for binding to `ETH_P_ALL` (0x0003): `sock:bind(string.pack(">I2", 0x0003))`.
-*   - Other families: A packed string representing the family-specific address structure.
-* @tparam[opt] integer port The local port number (required and used only if the family is `AF_INET`).
+*   - `AF_PACKET`: An integer representing the ethernet protocol in host byte order
+*     (e.g., `0x0003` for `ETH_P_ALL`, `0x88CC` for `ETH_P_LLDP`)
+*     The `port` argument is also required.
+*   - Other families: A packed string directly representing parts of the family-specific address structure.
+*
+* @tparam[opt] integer port The local port or interface index.
+*   - `AF_INET`: TCP/UDP port number.
+*   - `AF_PACKET`: Network interface index (e.g., from `linux.ifindex("eth0")`).
+*
 * @treturn nil
 * @raise Error if the bind operation fails (e.g., address already in use, invalid address).
 * @usage
 *   -- Bind TCP/IPv4 socket to localhost, port 8080
 *   tcp_server_sock:bind(net.aton("127.0.0.1"), 8080)
 *
-*   -- Bind AF_PACKET socket to receive all protocols (ETH_P_ALL = 0x0003)
-*   af_packet_sock:bind(string.pack(">H", 0x0003))
+*   -- Bind AF_PACKET socket to protocol `ETH_P_LLDP` on a specific interface
+*   af_packet_sock:bind(0x88CC, linux.ifindex("eth0"))
 * @see net.aton
 * @see linux.ifindex
 */
--- a/lunatik.h
+++ b/lunatik.h
@@ -340,6 +340,13 @@ static inline void lunatik_optcfunction(lua_State *L, int idx, const char *field
 #define lunatik_checkbounds(L, idx, val, min, max)	\
 	luaL_argcheck(L, val >= min && val <= max, idx, "out of bounds")
 
+static inline lua_Integer lunatik_checkinteger(lua_State *L, int idx, lua_Integer min, lua_Integer max)
+{
+	lua_Integer v = luaL_checkinteger(L, idx);
+	lunatik_checkbounds(L, idx, v, min, max);
+	return v;
+}
+
 static inline unsigned int lunatik_checkuint(lua_State *L, int idx)
 {
 	lua_Integer val = luaL_checkinteger(L, idx);
--- a/examples/lldpd.lua
+++ b/examples/lldpd.lua
@@ -35,8 +35,8 @@ local config = {
 
 local ethertype = string.pack(">I2", ETH_P_LLDP)
 
-local function get_src_mac()
-	local rx <close> = raw.bind(ETH_P_ALL)
+local function get_src_mac(ifindex)
+	local rx <close> = raw.bind(ETH_P_ALL, ifindex)
 	local frame = rx:receive(2048)
 	return frame:sub(7, 12)
 end
@@ -75,7 +75,7 @@ end
 
 local ifindex = linux.ifindex(config.interface)
 
-local src_mac = get_src_mac()
+local src_mac = get_src_mac(ifindex)
 local lldp_frame = build_lldp_frame(src_mac)
 
 local function worker()
--- a/lib/socket/raw.lua
+++ b/lib/socket/raw.lua
@@ -32,9 +32,9 @@ local raw = {}
 -- @see socket.bind
 function raw.bind(proto, ifindex)
 	local proto = proto or ETH_P_ALL
+	local ifindex = ifindex or 0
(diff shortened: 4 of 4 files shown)


lib/luasocket.c
```diff
@@ -71,6 +71,17 @@ static size_t luasocket_checkaddr(lua_State *L, struct socket *socket, struct so
 		return sizeof(struct sockaddr_un);
 	}
 #endif
+	else if (addr->ss_family == AF_PACKET && lua_type(L, ix) == LUA_TSTRING && lua_type(L, ix + 1) == LUA_TNUMBER) {
```
lneto: we should always avoid duplicate code while developing our code base.. we should do something like this..

```suggestion
	else if (addr->ss_family == AF_PACKET) {
	     /* common code */
		 int type = lua_type(L, ix);
		 if (type == LUA_TNUMBER) {
		 	 	/* specific code */
		 }
	 	 else if (type == LUA_TSTRING && lua_type(L, ix + 1) == LUA_TNUMBER) {
	 	 	     /* specific code */
	 	 }
	 	 /* common code */
	 }
```

lib/luasocket.c
```diff
@@ -71,17 +71,30 @@ static size_t luasocket_checkaddr(lua_State *L, struct socket *socket, struct so
...
 #endif
-	else if (addr->ss_family == AF_PACKET && lua_type(L, ix) == LUA_TNUMBER) {
+	else if (addr->ss_family == AF_PACKET) {
 		struct sockaddr_ll *addr_ll = (struct sockaddr_ll *)addr;
-		addr_ll->sll_ifindex = (int)lua_tointeger(L, ix);
+		int type = lua_type(L, ix);
+
+		if (type == LUA_TNUMBER) {
+			addr_ll->sll_ifindex = (int)lua_tointeger(L, ix);
+		} else if (type == LUA_TSTRING && lua_type(L, ix + 1) == LUA_TNUMBER) {
```
lneto: please, mind the code style with this file.. we use this form..
```
if (x) {
...
}
else {
...
}
```

please try to stick with it on your PRs.

lib/luasocket.c
```diff
@@ -71,17 +71,30 @@ static size_t luasocket_checkaddr(lua_State *L, struct socket *socket, struct so
...
+			size_t len;
+			const char *proto = luaL_checklstring(L, ix, &len);
+			__be16 proto_be;
+			luaL_argcheck(L, len == sizeof(proto_be), ix, "protocol must be 2 bytes");
+			memcpy(&proto_be, proto, sizeof(proto_be));
```
lneto: it seems weird to me; why not using an integer here?

lneto: you could just check the range..

lib/luasocket.c
```diff
@@ -71,17 +71,30 @@ static size_t luasocket_checkaddr(lua_State *L, struct socket *socket, struct so
...
+			addr_ll->sll_protocol = proto_be;
+			addr_ll->sll_ifindex  = (int)lua_tointeger(L, ix + 1);
+		} else {
+			size_t len;
+			const char *addr_data = luaL_checklstring(L, ix, &len);
+			luaL_argcheck(L, len <= LUASOCKET_ADDRMAX, ix, "out of bounds");
+			memcpy(addr->__data, addr_data, len);
+		}
```
lneto: why changing our logic here?

sneaky-potato: oh right, I just realized this is wrong. For non-special families, we should allow passing a raw packed sockaddr payload. Will fix this in next push.

lib/luasocket.c
sneaky-potato: Are you suggesting to convert into byte order inside the block and accept an integer protocol?
something like this
```c
u16 proto = luaL_checkinteger(L, ix);
luaL_argcheck(L, proto <= 0xffff, ix, "protocol out of range");
addr_ll->sll_protocol = htons(proto);
```

this makes binding raw socket on lua side easier as string packing will not be required. But will need to document this behavior as right now bind accepts a packed string as protocol

lneto: I think we have a macro to check range, right? and it should be [0, UINT16_MAX].. but yeah, it was what I meant =)

lneto: if we don't have such a macro, we should create one..

sneaky-potato:
I realized that bind(ifindex) does not make much sense once we extend the API to support bind(protocol, ifindex).
Allowing the single-argument form leaves sll_protocol uninitialized.
I’ve refactored the API to require the explicit (protocol, ifindex) pair for AF_PACKET binds so that sockaddr_ll is always fully initialized.

Made relevant changes in raw.lua and lldpd.lua

lib/luasocket.c
```diff
@@ -71,9 +71,12 @@ static size_t luasocket_checkaddr(lua_State *L, struct socket *socket, struct so
 		return sizeof(struct sockaddr_un);
 	}
 #endif
-	else if (addr->ss_family == AF_PACKET && lua_type(L, ix) == LUA_TNUMBER) {
+	else if (addr->ss_family == AF_PACKET) {
 		struct sockaddr_ll *addr_ll = (struct sockaddr_ll *)addr;
-		addr_ll->sll_ifindex = (int)lua_tointeger(L, ix);
+		lua_Integer proto = luaL_checkinteger(L, ix);
+		luaL_argcheck(L, proto >= 0 && proto <= U16_MAX, ix, "protocol out of bounds");
```
lneto: did you check if you have a macro for this?

lneto:
> I realized that bind(ifindex) does not make much sense once we extend the API to support bind(protocol, ifindex). Allowing the single-argument form leaves sll_protocol uninitialized. I’ve refactored the API to require the explicit (protocol, ifindex) pair for AF_PACKET binds so that sockaddr_ll is always fully initialized.
> 
> Made relevant changes in raw.lua and lldpd.lua

btw, can you check if tap.lua is still working? thanks

lib/luasocket.c
sneaky-potato: [Here](https://elixir.bootlin.com/linux/v4.0/source/include/linux/kernel.h#L35) is the macro defined.
I was able to build and test the changes. One question
I tried running the following code
```lua
local rx <close> = raw.bind(70000000, ifindex)
```

running this produces an error
```shell
bad argument #2 to '?' (protocol out of bounds)
```

Is this some known bug in `luaL_argcheck` ?

sneaky-potato:
> btw, can you check if tap.lua is still working? thanks

Yes `tap.lua` is working.

lib/luasocket.c
lneto: it's a known issue on our monitor.. if you comment out lunatik_monitorobject on luasocket, it should disappear I think.. if you want to tackle this improvement, it's a good one.. I think we just need to handle the error properly [here](https://github.com/luainkernel/lunatik/blob/master/lunatik_obj.c#L135).

lneto: > [Here](https://elixir.bootlin.com/linux/v4.0/source/include/linux/kernel.h#L35) is the macro defined. I was able to build and test the changes. One question I tried running the following code

I was referring to our macros for bounds.. not kernel defines.. https://github.com/luainkernel/lunatik/blob/master/lunatik.h#L360

lunatik.h
```diff
@@ -347,6 +347,13 @@ static inline unsigned int lunatik_checkuint(lua_State *L, int idx)
 	return (unsigned int)val;
 }
 
+static inline u16 lunatik_checku16(lua_State *L, int idx)
```
lneto: let's generalize this in a macro.. e.g., `lunatik_checkint(L, ix, t, T)` and used like `lunatik_checkint(L, 1, u16, U16)`.. I would also search for other places where this pattern is being used..

lunatik.h
```diff
@@ -340,6 +340,13 @@ static inline void lunatik_optcfunction(lua_State *L, int idx, const char *field
 #define lunatik_checkbounds(L, idx, val, min, max)	\
 	luaL_argcheck(L, val >= min && val <= max, idx, "out of bounds")
 
+#define lunatik_checkint(L, idx, type, TYPE) \
+({                                                     \
+	lua_Integer val = luaL_checkinteger(L, idx);     \
+	lunatik_checkbounds(L, idx, val, 0, TYPE##_MAX); \
+	(type)val;                                       \
+})
```
lneto: please, mind our standard.. we use `do while (0)` in such cases..

lib/luasocket.c
```diff
@@ -57,7 +57,7 @@ static size_t luasocket_checkaddr(lua_State *L, struct socket *socket, struct so
 	if (addr->ss_family == AF_INET) {
 		struct sockaddr_in *addr_in = (struct sockaddr_in *)addr;
 		addr_in->sin_addr.s_addr = htonl((u32)luaL_checkinteger(L, ix));
-		addr_in->sin_port = htons((u16)luaL_checkinteger(L, ix + 1));
+		addr_in->sin_port = htons(lunatik_checkint(L, ix + 1, u16, U16));
```
lneto: we should apply this macro to other places that are using the same pattern in the code base (e.g., checkinteger + checkbounds).. but we can do it as a separate PR. Could you open an issue for this?

lneto: btw, I just checked we have already a checkuint API, ideally, its users should be migrated to this new one.. however, we need to check why they are expecting `1` as the lower bound..

lunatik.h
```diff
@@ -340,6 +340,13 @@ static inline void lunatik_optcfunction(lua_State *L, int idx, const char *field
 #define lunatik_checkbounds(L, idx, val, min, max)	\
 	luaL_argcheck(L, val >= min && val <= max, idx, "out of bounds")
 
+#define lunatik_checkint(L, idx, type, TYPE) \
+({                                                     \
+	lua_Integer val = luaL_checkinteger(L, idx);     \
+	lunatik_checkbounds(L, idx, val, 0, TYPE##_MAX); \
```
```suggestion
	lunatik_checkbounds(L, idx, val, TYPE##_MIN, TYPE##_MAX); \
```
?

lib/luasocket.c
lneto: perhaps, a better idea is to have two APIs,  one that receives the ranges and other that uses `TYPE##_{MIN, MAX}`

lunatik.h
sneaky-potato: There is no standard or kernel-defined `U16_MIN`. I believe this is because unsigned minimum is always 0.
One option could be to make the macro explicit about bounds as well
```c
#define lunatik_checkint(L, idx, out, type, min, max) \
do {                                                  \
	lua_Integer val = luaL_checkinteger(L, idx); \
	lunatik_checkbounds(L, idx, val, min, max);  \
	(out) = (type)val;                           \
} while (0)
```

This can be called like so
```c
u16 port;
lunatik_checkint(L, ix, port, u16, 0, U16_MAX);
```

lneto: C standard specifies UINT16_MIN, defined by stdint; if it's not available in the kernel header, we add it to ours on include/stdint.h. However, I think it could be a bit overkill. I think an API like this should be enough, but I would use a inline for returning the value and then use typecast on calling.. e.g.,
```
u16 port = (u16)lunatik_checkinteger(L, ...);
```

I would call it checkinteger, instead of checking to make it clearer that's lua_Integer, but no strong feelings on this name..

lib/luasocket.c
```diff
@@ -71,9 +71,10 @@ static size_t luasocket_checkaddr(lua_State *L, struct socket *socket, struct so
 		return sizeof(struct sockaddr_un);
 	}
 #endif
-	else if (addr->ss_family == AF_PACKET && lua_type(L, ix) == LUA_TNUMBER) {
+	else if (addr->ss_family == AF_PACKET) {
 		struct sockaddr_ll *addr_ll = (struct sockaddr_ll *)addr;
-		addr_ll->sll_ifindex = (int)lua_tointeger(L, ix);
+		addr_ll->sll_protocol = htons((u16)lunatik_checkinteger(L, ix, 0, U16_MAX));
+		addr_ll->sll_ifindex  = (int)luaL_checkinteger(L, ix + 1);
```
lneto: perhaps ifindex should be limited to [0, INT_MAX], right?

lunatik.h
```diff
@@ -340,6 +340,13 @@ static inline void lunatik_optcfunction(lua_State *L, int idx, const char *field
 #define lunatik_checkbounds(L, idx, val, min, max)	\
 	luaL_argcheck(L, val >= min && val <= max, idx, "out of bounds")
 
+static inline lua_Integer lunatik_checkinteger(lua_State *L, int idx, lua_Integer min, lua_Integer max)
```
lneto, line 343: can you create another issue to apply this new API to other modules and remove lunatik_checkuint?

lib/luasocket.c
sneaky-potato: indeed, that should be limited to [0, INT_MAX]

PR #379, opened 2026-01-12
fix SIGSTKFLT and add support for 64-bit integer on 32-bit arch

lneto:


lneto:
Changed my mind on a new 3.7 minor because of other fixes not related to the Lua update to 5.5; let's embrace 4.0 ;-)

PR #384, opened 2026-01-13
Fix install path for lunatik script

MarcFontaine:
This fixes a bug that was introduced in #371.


MarcFontaine:
Sorry, I lost concentration in #384 and the final version that got merged had the install path wrong.
Here is a fix

PR #385, opened 2026-01-13
split signal and byteorder into top-level modules

### LINKED ISSUE #381: break lib/lualinux on many lib/lualinux_<sub> sub libs
lualinux has grown significantly on complexity.. we might want to split it on many sublibs

abhijeetw035:
This PR addresses #381  by refactoring `lib/lualinux.c` and extracting the signal handling and byte-order conversion logic into their own top-level modules.

### Changes:
- Created `lib/luasignal.c` to handle signal masking, pending signals, signal states, and the `kill` function.
- Created `lib/luabyteorder.c` to handle all CPU/Big-Endian/Little-Endian conversions.
- Cleaned up `lib/lualinux.c` by removing the extracted logic.
- Updated `Kbuild` and `Makefile` to include the new modules in the build process.

The new modules can now be loaded independently:
- `require("signal")`
- `require("byteorder")`

Verified the build on my environment and all modules compile and link correctly.

Closes #381


lib/luabyteorder.c
```diff
@@ -0,0 +1,195 @@
...
+}
+
+static void __exit luabyteorder_exit(void)
+{
+}
+
+module_init(luabyteorder_init);
+module_exit(luabyteorder_exit);
+MODULE_LICENSE("Dual MIT/GPL");
+MODULE_AUTHOR("Lourival Vieira Neto <lourival.neto@ringzero.com.br>");
```
lneto:
```suggestion
MODULE_AUTHOR("Lourival Vieira Neto <lourival.neto@ringzero.com.br>");

```

lib/lualinux.c
```diff
@@ -52,34 +46,34 @@
 /* based on math_random() @ lua/lmathlib.c */
 static int lualinux_random(lua_State *L)
 {
-	lua_Integer low, up, rand;
+    lua_Integer low, up, rand;
```
lneto: why changing this? please, review your patch; we use tabs.

lneto (changes requested):
please review your patch format; \n at EOF, tabs, avoid cosmetic changes, etc

Makefile
```diff
@@ -30,7 +30,8 @@ all: lunatik_sym.h
 	CONFIG_LUNATIK_NETFILTER=m CONFIG_LUNATIK_COMPLETION=m \
 	CONFIG_LUNATIK_CRYPTO_SHASH=m CONFIG_LUNATIK_CRYPTO_SKCIPHER=m \
 	CONFIG_LUNATIK_CRYPTO_AEAD=m CONFIG_LUNATIK_CRYPTO_RNG=m \
-	CONFIG_LUNATIK_CRYPTO_COMP=m CONFIG_LUNATIK_CPU=m CONFIG_LUNATIK_HID=m
+	CONFIG_LUNATIK_CRYPTO_COMP=m CONFIG_LUNATIK_CPU=m CONFIG_LUNATIK_HID=m \
+    CONFIG_LUNATIK_SIGNAL=m CONFIG_LUNATIK_BYTEORDER=m
```
lneto, line 34: btw, you should check examples and tests..

lib/luabyteorder.c
```diff
@@ -0,0 +1,195 @@
...
+*/
+
+#include <linux/module.h>
+#include <linux/byteorder/generic.h>
+#include <linux/errno.h>
+
+#include <lua.h>
+#include <lauxlib.h>
+
+#include <lunatik.h>
```
lneto, line 9: can you take the opportunity and check what headers are really necessary here? I've the impression that we just need..
```suggestion
#include <linux/module.h>
#include <linux/byteorder/generic.h>

#include <lunatik.h>
```

lneto (approved):
it's looking good to me.. can you double check if nothing else (e.g., examples and tests) needs to be changed? thanks!

abhijeetw035:
> it's looking good to me.. can you double check if nothing else (e.g., examples and tests) needs to be changed? thanks!

did  grep sweep of the tests/ and examples/ directories. Aside from dnsblock and dnsdoctor (which I updated in the previous commit), there are no other references to the old linux.signal or byte-order functions. Thanks.

lib/luasignal.c
```diff
@@ -0,0 +1,212 @@
...
+
+	if (ret)
+		goto err;
+
+	lua_pushboolean(L, true);
+	return 1;
+err:
+	lua_pushboolean(L, false);
+	lua_pushinteger(L, ret);
+	return 2;
```
lneto, line 140: I think we should just throw an error here.. but it should be handled as a new PR..

lneto (approved):

lib/luasignal.c
abhijeetw035: Thanks ! . will make  new PR soon to clean up the error handling.

PR #388, opened 2026-01-15
Add ifindex to nf hook callback

carloslack:
Also add luasocket side MAC getter

- Add `ifindex` on stack for `netfilter` hook
- `ifindex` can be used as parameter for luasocket's `getmacaddress`

There are corner cases I don't handle yet, like in netfilter's `skb->dev` can be `NULL` if , for example nf hook is `LOCAL_OUT` 

Also, is `luasocket` the best placehold for `getmacaddr` ?
I've not added any documentation yet, as we'll discuss the best approach here.

### DIFF
--- a/lib/luanetfilter.c
+++ b/lib/luanetfilter.c
@@ -81,8 +81,15 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
 	else
 		luadata_reset(data, skb, skb_headlen(skb), LUADATA_OPT_SKB);
 
-	if (lua_pcall(L, 1, 2, 0) != LUA_OK) {
+	struct net_device *dev = skb->dev;
+	if (dev)
+		lua_pushinteger(L, dev->ifindex);
+	else
+		lua_pushnil(L); /* dev may be NULL if hook is LOCAL_OUT */
+
+	if (lua_pcall(L, 2, 2, 0) != LUA_OK) {
 		pr_err("%s\n", lua_tostring(L, -1));
+		lua_pop(L, 1);
 		return -1;
 	}
 


lib/luanetfilter.c
```diff
@@ -81,13 +82,19 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
 	else
 		luadata_reset(data, skb, skb_headlen(skb), LUADATA_OPT_SKB);
 
-	if (lua_pcall(L, 1, 2, 0) != LUA_OK) {
+	if (skb && skb->dev)
```
lneto: skb is always available here, right? if it isn't we would have other problems above..

lib/luanetfilter.c
```diff
@@ -81,13 +82,19 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
...
+	lua_pushinteger(L, ifindex);
+	if (lua_pcall(L, 2, 2, 0) != LUA_OK) {
 		pr_err("%s\n", lua_tostring(L, -1));
+		lua_pop(L, 1);
 		return -1;
 	}
 
 	if (!lua_isnil(L, -1))
 		skb->mark = (u32)lua_tointeger(L, -1);
+
```
lneto: let's avoid mix cosmetic changes along with code new features of fixes..

```suggestion
```

lib/luanetfilter.c
```diff
@@ -81,13 +82,19 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
 	else
 		luadata_reset(data, skb, skb_headlen(skb), LUADATA_OPT_SKB);
 
-	if (lua_pcall(L, 1, 2, 0) != LUA_OK) {
+	if (skb && skb->dev)
+		ifindex = skb->dev->ifindex;
+
+	lua_pushinteger(L, ifindex);
```
lneto:
```suggestion
	struct net_device *dev = skb->dev;
	lua_pushinteger(L, dev ? dev->ifindex : 0);
```

btw, is it alright to assume `0` if dev is not available?

lib/luasocket.c
```diff
@@ -163,6 +163,26 @@ static int luasocket_send(lua_State *L)
...
 }
 
+static int luasocket_getmacaddress(lua_State *L)
+{
+	struct socket *socket = luasocket_check(L, 1);
+	struct sock *sk = socket->sk;
+	unsigned char mac[ETH_ALEN] = {0};
+	size_t len = 0;
+
+	if (likely(sk)) {
```
lneto: I don't think there's a possible flow where sk is NULL here..

lib/luasocket.c
```diff
@@ -163,6 +163,26 @@ static int luasocket_send(lua_State *L)
 	return 1;
 }
 
+static int luasocket_getmacaddress(lua_State *L)
+{
+	struct socket *socket = luasocket_check(L, 1);
+	struct sock *sk = socket->sk;
+	unsigned char mac[ETH_ALEN] = {0};
```
lneto: we can use a luaL_Buffer here to avoid copying twice the same data (e.g., memcpy + pushlstring); take a look at the code code base we have some examples there..

lib/luasocket.c
```diff
@@ -163,6 +163,26 @@ static int luasocket_send(lua_State *L)
...
+	if (likely(sk)) {
+		struct net_device *dev = dev_get_by_index(sock_net(sk), luaL_checkinteger(L, 2));
+		if (likely(dev)) {
+			len = min_t(size_t, dev->addr_len, ETH_ALEN);
+			memcpy(mac, dev->dev_addr, len);
+			dev_put(dev);
+		}
+	}
+
+	lua_pushlstring(L, (const char*)mac, len);
```
lneto: we will push an empty string in case we don't find a device, right? I think we should follow this module standard and throw an error.. perhaps ENODEV? btw, I've introduced a new lunatik_throw macro in my PR regarding luacrypto.. it might be useful for this here if there's no direct way to use lunatik_try here (it's probably the case because dev_put, right?).

lib/luasocket.c
```diff
@@ -163,6 +163,26 @@ static int luasocket_send(lua_State *L)
...
 
+static int luasocket_getmacaddress(lua_State *L)
+{
+	struct socket *socket = luasocket_check(L, 1);
+	struct sock *sk = socket->sk;
+	unsigned char mac[ETH_ALEN] = {0};
+	size_t len = 0;
+
+	if (likely(sk)) {
+		struct net_device *dev = dev_get_by_index(sock_net(sk), luaL_checkinteger(L, 2));
```
lneto: what's the second argument? ifindex? how does the API client know this?

lneto:
> Also add luasocket side MAC getter

perhaps we could break this PR and two and merge netfiler change beforehand..

> * Add `ifindex` on stack for `netfilter` hook
> * `ifindex` can be used as parameter for luasocket's `getmacaddress`
> 
> There are corner cases I don't handle yet, like in netfilter's `skb->dev` can be `NULL` if , for example nf hook is `LOCAL_OUT`

It's alright, but in this case, it would be misleading to push(0) as it's a valid ifindex, in this case, I think we should pushnil or none. Also, we should document this behavior and open an issue for improvement.

> Also, is `luasocket` the best placehold for `getmacaddr` ? I've not added any documentation yet, as we'll discuss the best approach here.

yes, this API on socket seems a bit odd to me.. perhaps a better approach would be to have such API on lualinux and a use the default network namespace, afterwords we can improve it with other APIs to handle namespaces..

lib/luasocket.c
carloslack: good points! Let me address these.

lib/luanetfilter.c
carloslack: I give preference to returning `-ENODEV` in this case, because device is not set
We can still can do a `pushnil` instead, but we'd have 2 distinct calls depending on the case, `pushnil` or `pushinteger`

carloslack:
> > Also add luasocket side MAC getter
> 
> perhaps we could break this PR and two and merge netfiler change beforehand..
> 
> > * Add `ifindex` on stack for `netfilter` hook
> > * `ifindex` can be used as parameter for luasocket's `getmacaddress`
> > 
> > There are corner cases I don't handle yet, like in netfilter's `skb->dev` can be `NULL` if , for example nf hook is `LOCAL_OUT`
> 
> It's alright, but in this case, it would be misleading to push(0) as it's a valid ifindex, in this case, I think we should pushnil or none. Also, we should document this behavior and open an issue for improvement.
> 
> > Also, is `luasocket` the best placehold for `getmacaddr` ? I've not added any documentation yet, as we'll discuss the best approach here.
> 
> yes, this API on socket seems a bit odd to me.. perhaps a better approach would be to have such API on lualinux and a use the default network namespace, afterwords we can improve it with other APIs to handle namespaces..

Let's go for 2 PR's then? I agree with you because we can now concentrate only about `ifindex` , and it is a more simple PR that is unlikely to affect old users.
I can open a improvement PR as soon as we settle this one.

lib/luanetfilter.c
```diff
@@ -81,8 +82,11 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
 	else
 		luadata_reset(data, skb, skb_headlen(skb), LUADATA_OPT_SKB);
 
-	if (lua_pcall(L, 1, 2, 0) != LUA_OK) {
+	dev = skb->dev;
+	lua_pushinteger(L, dev ? dev->ifindex : -ENODEV); /* dev may be NULL if using LOCAL_OUT */
```
lneto: the problem here is that the Lua script has no way to use this info (`ENODEV`), we are currently using a string for errno (using `errname()`). Also, in that case, the script has only two possibilities for this parm, a valid or an undefined index, I don't see any decision the script should perform based on errno here. Moreover, having an undefined device here is not a proper error, it's somehow an "expected" behavior for UDP on LOCAL (both IN and OUT, right?). Thus, I think the more adequate is to use nil or none. What do you think?

lneto: e.g.,

```suggestion
	if ((dev = skb->dev) == NULL) /* dev may be NULL if using LOCAL_OUT */
		lua_pushnil(L);
	else
		lua_pushinteger(L, dev->ifindex);
```

lib/luanetfilter.c
```diff
@@ -72,6 +72,7 @@ static inline lunatik_object_t *luanetfilter_pushskb(lua_State *L, luanetfilter_
 static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_buff *skb)
 {
 	lunatik_object_t *data;
+	struct net_device *dev;
```
lneto: we can init dev here, right?

lneto: or actually, we can move the declaration below.. we are using C99 on Lunatik, so we don't need to declare vars on the beginning of blocks only..

lneto (approved):

lib/luanetfilter.c
```diff
@@ -81,8 +82,15 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
 	else
 		luadata_reset(data, skb, skb_headlen(skb), LUADATA_OPT_SKB);
 
-	if (lua_pcall(L, 1, 2, 0) != LUA_OK) {
+	dev = skb->dev;
+	if (dev)
+		lua_pushinteger(L, dev->ifindex);
+	else
+		lua_pushnil(L); /* dev may be NULL if hook is LOCAL_OUT */
```
lneto, line 88: can you confirm if it's only OUT or both OUT and IN?

lib/luanetfilter.c
lneto: it's up to you.. it's approved, let me know and I can merge it.. btw, you can turn this into a normal PR instead of a draft ;-)

lib/luanetfilter.c
carloslack: It is only OUT.
I did few quick tests and only case `dev` is not available is out

lib/luanetfilter.c
carloslack: Related to `ifndex` being zero, for example in in kernel 5.4 [here](https://elixir.bootlin.com/linux/v5.4/source/net/core/dev.c#L9035) and [here](https://elixir.bootlin.com/linux/v5.4/source/net/core/dev.c#L8454) conflict with the way kernel 6.8 for example  [here](https://elixir.bootlin.com/linux/v6.8/source/net/core/dev.c#L10188) 

So perhaps, for  now at least we should do what you suggested, return the value of `ifindex`, or nil for error, makes more sense.

lib/luanetfilter.c
carloslack: ah, we can of course, I just did't because there is the `luanetfilter_pushcb` call, that if fail, will return early.
But I'm fine either way, what do you think?

carloslack: At least in `luanetfilter`, I don't see other examples of variables not at the beginning of scope, so I didn't want to break that flow :)

lib/luanetfilter.c
carloslack: Thanks!

lib/luanetfilter.c
lneto: I prefer it nearest to its usage (that is, bellow); but I have no strong feelings.. it's up to you =). I would just try to do the assignment on declaration. Thanks!

carloslack: Fair enough 👍

lneto (approved):

PR #391, opened 2026-01-16
raise errno-name errors in signal.kill

abhijeetw035:
This PR refactors `signal.kill` to raise Lua errors on failure instead of returning `false, errno`.

### Changes
* Removed the `goto err` pattern .
* Updated the function to raise a Lua error when `pid` lookup fails or `kill_pid` returns a non-zero status.

### Implementation Details
implemented the error raising using a combination of `lunatik_pusherrname` and `lua_error(L)` .

1.  `lunatik_pusherrname(L, -ret)` converts the raw kernel errno (e.g., `-3`) into its standard string representation (e.g., `"ESRCH"`).
2.  `lua_error(L)` then takes this string from the stack and raises it as the error message.

This approach ensures that Lua scripts receive standard, clean error strings (e.g., `lunatik: ESRCH`) which are easier to handle programmatically via `pcall`, compared to formatted strings from `luaL_error`.

### DIFF
--- a/lib/luasignal.c
+++ b/lib/luasignal.c
@@ -108,10 +108,7 @@ static int luasignal_sigstate(lua_State *L)
 * @tparam integer pid Process ID to kill.
 * @tparam[opt] integer sig Signal number to send (default: `signal.flags.KILL`).
 * @treturn boolean `true` if the signal was sent successfully.
-* @treturn[error] boolean `false` followed by an error number if the operation fails.
-* @raise Errors:
-* - (3): The specified PID doesn't exist
-* - other errno values depending on the failure cause (e.g., `EPERM`, `EINVAL`, etc.)
+* @raise Error string (e.g., "ESRCH", "EPERM") if the operation fails.
 * @usage
 * signal.kill(1234)  -- Kill process 1234 with SIGKILL (default)
 * signal.kill(1234, signal.flags.TERM)  -- Kill process 1234 with SIGTERM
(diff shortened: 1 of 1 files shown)


lib/luasignal.c
```diff
@@ -121,23 +118,22 @@ static int luasignal_kill(lua_State *L)
...
 	struct pid *pid = find_get_pid(nr);
-
-	int ret = ESRCH;
-	if (pid == NULL)
-		goto err;
+	int ret;
+	
+	if (pid == NULL) {
+		lunatik_pusherrname(L, ESRCH);
+		return lua_error(L);
```
lneto: I have introduced `lunatik_throw` macro, you can use it here..

lneto:
Please, don't add merge commits; use rebase, instead.

lib/luasignal.c
```diff
@@ -121,23 +118,19 @@ static int luasignal_kill(lua_State *L)
...
 	struct pid *pid = find_get_pid(nr);
-
-	int ret = ESRCH;
+	int ret;
+	
 	if (pid == NULL)
-		goto err;
+		lunatik_throw(L, ESRCH);
 
 	ret = kill_pid(pid, sig, 1);
```
lneto: You declare ret in here..

abhijeetw035:
> Please, don't add merge commits; use rebase, instead.

Sure. Did reset -> fetch -> rebase.

lib/luasignal.c
lneto: I think you reverted throw

abhijeetw035: oh , ig that was reverted when i did reset for the rebase . made required changes . please check

lib/luasignal.c
```diff
@@ -121,23 +118,18 @@ static int luasignal_kill(lua_State *L)
 	pid_t nr = (pid_t)luaL_checkinteger(L, 1);
 	int sig = luaL_optinteger(L, 2, SIGKILL);
 	struct pid *pid = find_get_pid(nr);
-
-	int ret = ESRCH;
+
```
lneto: I think there's a trailing whitespace here.. please, mind that

abhijeetw035: yes removed it.

PR #397, opened 2026-01-17
Add lualinux mac retriever from interface index

carloslack:
Continuing from #388 ,  I add mac retrieval from `lualinux` (similar to what is done for `ifindex` there) 

Use `luaL_Buffer` as well, as discussed [here](https://github.com/luainkernel/lunatik/pull/388#discussion_r2694051872)

This simple test works for well for me:

```
local ifindex = linux.ifindex("enp1s0")
local mac = linux.mac(ifindex)
print(string.byte(mac,1,6))
```
I use `mac` as fn name, but could be `macaddress` or `getmac` or something similar, I am open to suggestions.

### DIFF
--- a/lib/lualinux.c
+++ b/lib/lualinux.c
@@ -209,6 +209,32 @@ static int lualinux_ifindex(lua_State *L)
 }
 
 /***
+* Gets the HW address for the network interface index.
+*
+* @function ifaddr
+* @tparam integer The interface index number.
+* @treturn string The interface HW address.
+* @raise Error if the device is not found.
+* @usage
+*   local addr = linux.ifaddr(index)
+*   print(string.byte(addr,1,6))
+*/
+static int lualinux_ifaddr(lua_State *L)
+{
+	int ifindex = luaL_checkinteger(L, 1);
+	luaL_Buffer B;
+	char *addr = luaL_buffinitsize(L, &B, MAX_ADDR_LEN);
+	struct net_device *dev = dev_get_by_index(&init_net, ifindex);
+
+	luaL_argcheck(L, dev != NULL, 1, "device not found");
+	size_t len = min(dev->addr_len, MAX_ADDR_LEN);
+	memcpy(addr, dev->dev_addr, len);
+	dev_put(dev);
+	luaL_pushresultsize(&B, len);
+	return 1;
+}
+
+/***
 * Table of task state constants.
 * Exports task state flags from `<linux/sched.h>`. These are used with
 * `linux.schedule()`.
@@ -292,6 +318,7 @@ static const luaL_Reg lualinux_lib[] = {
 	{"difftime", lualinux_difftime},
 	{"lookup", lualinux_lookup},
 	{"ifindex", lualinux_ifindex},
+	{"ifaddr", lualinux_ifaddr},
 	{"errname", lualinux_errname},
 	{NULL, NULL}
 };


lib/lualinux.c
```diff
@@ -208,6 +208,35 @@ static int lualinux_ifindex(lua_State *L)
...
+*   print(string.byte(mac,1,6))
+*/
+static int lualinux_mac(lua_State *L)
+{
+	int ifindex = luaL_checkinteger(L, 1);
+	struct net_device *dev = dev_get_by_index(&init_net, ifindex);
+
+	luaL_argcheck(L, dev != NULL, 1, "device not found");
+
+	luaL_Buffer b;
```
lneto: I don't think we need luaL_Buffer here; it's advantage is to avoid copying me multiple times. In this case, I think simply pushlstring.

lib/lualinux.c
```diff
@@ -208,6 +208,35 @@ static int lualinux_ifindex(lua_State *L)
...
+	luaL_buffinit(L, &b);
+
+	int len = min_t(size_t, dev->addr_len, ETH_ALEN);
+	luaL_addlstring(&b, (const char*)dev->dev_addr, len);
+	luaL_pushresult(&b);
+	dev_put(dev);
```
lneto, line 232: Actually, perhaps we would need buffer because of dev_put, as lua APIs can raise errors. We should prepbuf; memcpy; dev_put and then pushresult

lib/lualinux.c
```diff
@@ -208,6 +208,35 @@ static int lualinux_ifindex(lua_State *L)
...
+{
+	int ifindex = luaL_checkinteger(L, 1);
+	struct net_device *dev = dev_get_by_index(&init_net, ifindex);
+
+	luaL_argcheck(L, dev != NULL, 1, "device not found");
+
+	luaL_Buffer b;
+	luaL_buffinit(L, &b);
+
+	int len = min_t(size_t, dev->addr_len, ETH_ALEN);
```
lneto: btw, why `min_t` instead of `min`? and why we need to check this? I mean, it's safe to push the if address whatever the protocol it uses, right?

lib/lualinux.c
carloslack: What if I did something like

```
int len = dev->addr_len;
luaL_Buffer b;
char *addr = luaL_buffinitsize(L, &b, len);
memcpy(addr, dev->dev_addr, len);
dev_put(dev);
luaL_pushresultsize(&b, len);
```

But I am not sure if buffinitsize can longjmp and leak dev ref

lneto: Actually, I was thinking on something like this, but indeed it can raise an error.. see that [.., m] in the man https://www.lua.org/manual/5.5/manual.html#luaL_buffinitsize, it means that it can fail alloc'ing mem. Another idea is to allocate ETH_ALEN before getting device.. so we fail safely

lneto: Btw, this is useful to understand which APIs can raise errors.. https://www.lua.org/manual/5.5/manual.html#4.6

lneto: e.g.,

```
...
luaL_Buffer B;
char *addr = luaL_buffinitsize(L, &B, ETH_ALEN);
struct net_device *dev = dev_get_by_index(&init_net, ifindex);
size_t len = min(dev->addr_len, ETH_ALEN);

memcpy(addr, dev->dev_addr, len);
dev_put(dev);
luaL_pushresultsize(&B, len);
...
```

carloslack: I like this solution, we can use a maximum fixed length, but maybe instead of `ETH_ALEN`, [this](https://elixir.bootlin.com/linux/v5.4/source/include/uapi/linux/netdevice.h#L35) could cover different layouts?

lneto: Indeed; we just need a better name for this API.. `ifaddr`, `netdevaddr`? not sure.. btw, it might evolve to a `netdev` module on the future.. `ifaddress` might be alright for now, as we already have `ifindex`. What do you think?

carloslack: Maybe `ifaddr` is a bit more standard, in userspace we have `getifaddrs`, for example.

lib/lualinux.c
```diff
@@ -208,6 +208,33 @@ static int lualinux_ifindex(lua_State *L)
...
+* @treturn string The interface HW address.
+* @raise Error if the device is not found.
+* @usage
+*   local addr = linux.ifaddr(index)
+*   print(string.byte(addr,1,6))
+*/
+static int lualinux_ifaddr(lua_State *L)
+{
+	int ifindex = luaL_checkinteger(L, 1);
+	luaL_Buffer b;
```
lneto: quite nitpick, but the pattern is to
```suggestion
	luaL_Buffer B;
```

lib/lualinux.c
```diff
@@ -208,6 +208,33 @@ static int lualinux_ifindex(lua_State *L)
...
+
+	luaL_argcheck(L, dev != NULL, 1, "device not found");
+	size_t len = min(dev->addr_len, MAX_ADDR_LEN);
+	memcpy(addr, dev->dev_addr, len);
+	dev_put(dev);
+	luaL_pushresultsize(&b, len);
+	return 1;
+}
+
+
```
```suggestion

```

lneto: we try to avoid double \n as wel.. (nitpick again, sorry ;-)

lib/lualinux.c
carloslack: ah ok, changing that :) 
`B` is more fit indeed.

BTW we still have some un-matching `b` pattern across `lib/` , would like me to open a separate PR adjusting those?

lneto: > ah ok, changing that :) `B` is more fit indeed.

actually, I don't know why Roberto uses this pattern, but it ends up being a Lua "idiom".. hehe..

> BTW we still have some un-matching `b` pattern across `lib/` , would like me to open a separate PR adjusting those?

that would be much appreciated, thanks! =)

PR #398, opened 2026-01-17
fix opaque error messages in lunatik_monitor

### LINKED ISSUE #382: Bug with luaL_argcheck
As discussed with @lneto [here](https://github.com/luainkernel/lunatik/pull/378#discussion_r2680278791) it seems we need to properly handle the error in [lunatik_monitor](https://github.com/luainkernel/lunatik/blob/48c9bfc1b0bf718d684e698d58f6f76cd382d44c/lunatik_obj.c#L135)

Calling `socket.bind()` with certain argument patterns triggers the fallback branch in [luasocket_checkaddr()](https://github.com/luainkernel/lunatik/blob/48c9bfc1b0bf718d684e698d58f6f76cd382d44c/lib/luasocket.c#L54C1-L54C110), which uses [luaL_argcheck](https://github.com/luainkernel/lunatik/blob/48c9bfc1b0bf718d684e698d58f6f76cd382d44c/lib/luasocket.c#L82C1-L82C67).
Instead of a clear Lua error, this results in opaque messages.

The error can be reproduced with this minimal lua script
```lua

local af   = socket.af
local sock = socket.sock

local AF_NETLINK = 16
local NETLINK_ROUTE = 0

local tx = socket.new(AF_NETLINK, sock.RAW, NETLINK_ROUTE)
tx:bind(string.rep("A", 200))
```
Observed output
```shell
bad argument #2 to '?' (out of bounds)
```

Expected output
```shell
bad argument #2 to 'bind' (out of bounds)
```

And in some cases causes stack overflow due to some recursive error handling inside Lunatik.

Another example using syscalls, below script should hit [luaL_argcheck](https://github.com/luainkernel/lunatik/blob/48c9bfc1b0bf718d684e698d58f6f76cd382d44c/lib/luasyscall.c#L43C1-L43C70)
```lua
local syscall = require("syscall")
syscall.address(1000)
```
Observed output
```shell
stack overflow
```

This makes the errors unreliable and difficult to debug.

sneaky-potato:
This patch introduces better error handling and fixes opaque error messages by passing method name as second upvalue to `lunatik_monitor`.

Tested with this script
```lua

local af   = socket.af
local sock = socket.sock

local AF_NETLINK = 16
local NETLINK_ROUTE = 0

local tx = socket.new(AF_NETLINK, sock.RAW, NETLINK_ROUTE)
tx:bind(string.rep("A", 200))
```

Output
```shell
bad argument #2 to 'bind' (out of bounds)
stack traceback:
	/lib/modules/lua/bug.lua:10: in main chunk
	[C]: in ?
```

Fixes #382

### DIFF
--- a/lunatik_obj.c
+++ b/lunatik_obj.c
@@ -124,6 +124,18 @@ int lunatik_deleteobject(lua_State *L)
 }
 EXPORT_SYMBOL(lunatik_deleteobject);
 
+inline static void lunatik_fixerror(lua_State *L, const char *method)
+{
+	if (method) {
+		const char *error = lua_tostring(L, -1);
+		luaL_gsub(L, error, "?", method);
+		lua_remove(L, -2); /* error */
+	}
+	luaL_traceback(L, L, lua_tostring(L, -1), 1);
+	lua_remove(L, -2); /* fixed error */
+	lua_error(L);
+}
+
 static int lunatik_monitor(lua_State *L)
 {
 	int ret, n = lua_gettop(L);
@@ -136,8 +148,10 @@ static int lunatik_monitor(lua_State *L)
 	ret = lua_pcall(L, n, LUA_MULTRET, 0);
 	lunatik_unlock(object);
 
-	if (ret != LUA_OK)
-		lua_error(L);
+	if (ret != LUA_OK) {
+		const char *method = lua_tostring(L, lua_upvalueindex(2));
+		lunatik_fixerror(L, method);
+	}
 	return lua_gettop(L);
 }
 
@@ -147,7 +161,8 @@ void lunatik_monitorobject(lua_State *L, const lunatik_class_t *class)
 	for (reg = class->methods; reg->name != NULL; reg++) {
 		if (!lunatik_ismetamethod(reg)) {
 			lua_getfield(L, -1, reg->name);
-			lua_pushcclosure(L, lunatik_monitor, 1); /* stack: mt, method */
+			lua_pushstring(L, reg->name);
+			lua_pushcclosure(L, lunatik_monitor, 2); /* stack: mt, method, method name*/
 			lua_setfield(L, -2, reg->name);
 		}
 	}


lunatik_obj.c
```diff
@@ -123,6 +123,7 @@ static int lunatik_monitor(lua_State *L)
 {
 	int ret, n = lua_gettop(L);
 	lunatik_object_t *object = lunatik_checkobject(L, 1);
+	const char *method_name = lua_tostring(L, lua_upvalueindex(2));
```
lneto: It's better to fetch this only on the error case..

lunatik_obj.c
```diff
@@ -132,19 +133,43 @@ static int lunatik_monitor(lua_State *L)
 	lunatik_unlock(object);
 
 	if (ret != LUA_OK)
+	{
```
lneto: I think it's better to create an aux function for this block..

lunatik_obj.c
```diff
@@ -132,19 +133,43 @@ static int lunatik_monitor(lua_State *L)
...
+				int arg_num = 0;
+				if (sscanf(msg, "bad argument #%d", &arg_num) == 1) {
+					const char *suffix = found + strlen(placeholder);
+					luaL_where(L, 1);
+					lua_pushfstring(L, "bad argument #%d to '%s'%s", arg_num, method_name, suffix);
+					lua_concat(L, 2);
+					lua_remove(L, -2); /* remove original error msg */
+				}
+			}
+		}
```
lneto: it seems to me that it would be easier to do this in a lua script.. we can just pcall it from here..

lunatik_obj.c
```diff
@@ -132,19 +133,43 @@ static int lunatik_monitor(lua_State *L)
...
 		lua_error(L);
+	}
 	return lua_gettop(L);
 }
 
 int lunatik_monitorobject(lua_State *L)
 {
 	lua_getmetatable(L, 1);
 	lua_insert(L, 2); /* stack: object, metatable, key */
+	lua_pushvalue(L, 3); /* Copy method name to top of stack */
```
lneto: please, use the format above

lunatik_obj.c
sneaky-potato: I realized that patching a generated error message after the main pcall is not a good idea. Instead, I have moved this logic into a dedicated error handler passed to `lua_pcall`.

lunatik_obj.c
sneaky-potato: I have simplified this block of code a lot in the latest push.

lunatik_obj.c
lneto: > I realized that patching a generated error message after the main pcall is not a good idea. Instead, I have moved this logic into a dedicated error handler passed to `lua_pcall`.

yeah.. it's looking way better now =)..

lneto: btw, food for thought https://github.com/luainkernel/lua/blob/lunatik/lua.c#L136

lunatik_obj.c
```diff
@@ -119,18 +119,47 @@ int lunatik_deleteobject(lua_State *L)
...
 
+static int lunatik_error_handler(lua_State *L)
+{
+	/* stack: err_msg */
+	const char *method_name = lua_tostring(L, lua_upvalueindex(1));
+	if (lua_isstring(L, -1) && method_name)
+	{
+		const char *error_msg = lua_tostring(L, -1);
+		const char *placeholder = "'?'";
+		const char *found = strstr(error_msg, placeholder);
```
lneto: you can use https://www.lua.org/manual/5.5/manual.html#luaL_gsub right?

lunatik_obj.c
```diff
@@ -119,18 +119,47 @@ int lunatik_deleteobject(lua_State *L)
 }
 EXPORT_SYMBOL(lunatik_deleteobject);
 
+static int lunatik_error_handler(lua_State *L)
+{
+	/* stack: err_msg */
+	const char *method_name = lua_tostring(L, lua_upvalueindex(1));
+	if (lua_isstring(L, -1) && method_name)
+	{
```
lneto:
```suggestion
	if (lua_isstring(L, -1) && method_name) {
```

please mind our style on the overall patch

lunatik_obj.c
```diff
@@ -119,18 +119,47 @@ int lunatik_deleteobject(lua_State *L)
 }
 EXPORT_SYMBOL(lunatik_deleteobject);
 
+static int lunatik_error_handler(lua_State *L)
+{
+	/* stack: err_msg */
```
lneto: I don't think we need this comment here.. it's expected for the error handler, right?

sneaky-potato:
Testing summary:
I tested my changes for the bind example mentioned in the issue and was able to patch the error message correctly
```shell
$ sudo lunatik run bind_bug
bad argument #2 to 'bind' (out of bounds)
stack traceback:
	[C]: in ?
	[C]: in method 'bind'
	/lib/modules/lua/bind_bug.lua:10: in main chunk
	[C]: in ?

$ sudo lunatik run syscall_bug
stack overflow
```
EDIT: I realized `syscall.address` is a module-level function, NOT a method on a userdata object. I suggest we investigate the stack overflow issue separately?

lunatik_obj.c
sneaky-potato: Yes, that's is much cleaner

lneto:
> EDIT: I realized `syscall.address` is a module-level function, NOT a method on a userdata object. I suggest we investigate the stack overflow issue separately?

Yup, exactly. Also, it shouldn't be related to your patch, right? Did you try it on master? One thing we could do though to improve the error message in this case is to show stack trace on the [driver](https://github.com/luainkernel/lunatik/blob/master/driver.lua#L25).

sneaky-potato:
> Also, it shouldn't be related to your patch, right? Did you try it on master?

Yup this issue is unrelated to this patch. I got confused because of `luaL_argcheck` (still learning the embeddable nature of Lua :D)

lunatik_obj.c
```diff
@@ -119,18 +119,36 @@ int lunatik_deleteobject(lua_State *L)
...
+	return 1;
+}
+
 static int lunatik_monitor(lua_State *L)
 {
 	int ret, n = lua_gettop(L);
 	lunatik_object_t *object = lunatik_checkobject(L, 1);
 
+	lua_pushvalue(L, lua_upvalueindex(2)); /* method name */
+	lua_pushcclosure(L, lunatik_error_handler, 1); /* stack: object, args..., errhandler */
```
lneto: one concern I have is creating this other closure on every call.. perhaps it's a bit of premature optimization on my side, but it would be good if you can run some benchmarks for this change.. overall, I found it quite an elegant solution.. I'm most concerned on this on hotpaths..

btw, one thing that I've left for later and never got back is to apply memoization to monitor; perhaps it's a good opportunity to do it now..

sneaky-potato: > btw, one thing that I've left for later and never got back is to apply memoization to monitor;

Could you give me some pointers on how to memoize it? I'm thinking of caching the wrapped monitor closures (monitorobject)  in the registry to avoid recreating them on every call, but I want to make sure I'm following the right pattern for Lunatik.

I thought of doing something like this:
```c
int lunatik_monitorobject(lua_State *L)
{
    lua_pushlightuserdata(L, (void*)lunatik_monitorobject);
    lua_rawget(L, LUA_REGISTRYINDEX);
    if (!lua_isnil(L, -1))
        // found closure cached
        return 1;
    }
    lua_pop(L, 1);
    // function body
    lua_pushlightuserdata(L, (void *)lunatik_monitorobject);
    lua_pushvalue(L, -2);
    lua_rawset(L, LUA_REGISTRYINDEX);
    return 1;
}
```

lneto: perhaps we should just update the object metatable, right?

lneto: but I would measure the effect of this change before going to memoization; it's a good exercise anyway ;-).. you can mesure the time of n-calls on a luadata method (e.g., getbyte).. and assess the difference.. it might be negligible.. then we can merge this as is and leave memoization for later.. (or not ;-)

sneaky-potato:
Hello @lneto 
I've completed the benchmark comparing the current master against the new closure-based monitor implementation.
Results (1M iterations of getbyte):
    Master: ~266 ns/call (avg)
    This branch: ~370 ns/call (avg)
The closure creation adds roughly 104ns of overhead per call. The monitoring logic introduces a ~35 - 45% performance hit
this is the script I used
```lua
local data = require("data")

local d = data.new(1024)
local iterations = 1000000

local start_ns = linux.time()
for i = 1, iterations do
    d:getbyte(0)
end
local end_ns = linux.time()
local total_ns = end_ns - start_ns

local ns_per_call = total_ns / iterations

print("Total ns: " .. total_ns)
print("Ns per call: " .. ns_per_call)
```

lneto:
> Hello @lneto I've completed the benchmark comparing the current master against the new closure-based monitor implementation. Results (1M iterations of getbyte): Master: ~266 ns/call (avg) This branch: ~370 ns/call (avg) The closure creation adds roughly 104ns of overhead per call. The monitoring logic introduces a ~35 - 45% performance hit this is the script I used
> 
> ```lua
> local linux = require("linux")
> local data = require("data")
> 
> local d = data.new(1024)
> local iterations = 1000000
> 
> local start_ns = linux.time()
> for i = 1, iterations do
>     d:getbyte(0)
> end
> local end_ns = linux.time()
> local total_ns = end_ns - start_ns
> 
> local ns_per_call = total_ns / iterations
> 
> print("Total ns: " .. total_ns)
> print("Ns per call: " .. ns_per_call)
> ```

Nice job! Perhaps we can add this to our tests? 35% is looking to much for enriching error messages; what do you think? I think we should try memoization. Btw, can you disable lunatik_monitorobject on luadata (by commenting out it on its metatable) and run the same benchmark? Perhaps we are already imposing an unnecessary overhead by creating the monitor closure all the time.. memoization could help us there as well.. also the new "single" option that has been discussed for not sharing objects.. good work! Thanks!

sneaky-potato:
> what do you think? I think we should try memoization.

Yes I agree, we should try memoization. 

> Btw, can you disable lunatik_monitorobject on luadata (by commenting out it on its metatable) and run the same benchmark? Perhaps we are already imposing an unnecessary overhead by creating the monitor closure all the time.. 

disabling the monitorobject, I can see latency drops to just ~62 ns/call (avg).
How should we go about memoizing the monitor wrapper? 
I tried caching the wrapped monitor closures lazily via `__index` by updating the object metatable (inside `lunatik_monitorobject`), but that turned out to be unsafe (kernel panic).

lneto:
> > what do you think? I think we should try memoization.
> 
> Yes I agree, we should try memoization.

perhaps, a better idea is to create closures on the object creation instead of memoizing it. what do you think? We could even leverage uservalues.

> > Btw, can you disable lunatik_monitorobject on luadata (by commenting out it on its metatable) and run the same benchmark? Perhaps we are already imposing an unnecessary overhead by creating the monitor closure all the time..
> 
> disabling the monitorobject, I can see latency drops to just ~62 ns/call (avg). 

Wow; that would be a great optimization!

> How should we go about memoizing the monitor wrapper? I tried caching the wrapped monitor closures lazily via `__index` by updating the object metatable (inside `lunatik_monitorobject`), but that turned out to be unsafe (kernel panic).

can you share your draft and the stack trace?

sneaky-potato:
> perhaps, a better idea is to create closures on the object creation instead of memoizing it. what do you think? We could even leverage uservalues.

That will be better yes. But where in the lifecycle will we create these closures? Won't this require too many changes in the current implementation, given that __index and `lunatik_monitorobject` handle dynamic lookup during `lua_newobject` ?

> Wow; that would be a great optimization!

I have a pushed a change which memoizes the monitor, the latency for same script for latest change is ~330 ns/cal

> can you share your draft and the stack trace?

I was trying to mess with metatable and __index, got a kernel panic related to some reentry issue. With the latest push, I’ve been able to safely memoize it to some extent. (330 ns/call, so 21% performance hit)

sneaky-potato:
This is what I could think about creating closures instead of memoization, (taken example of luadata class)

C side:
- luadata_lnew
- lunatik_newobject
    - userdata allocated
    - metatable 
    - create uservalue table
    - create monitored closures ONCE and store in uservalue
- return object

Lua side:
- obj:method()
    - __index
    - return uservalue[method]
    
and thus we get rid of `lunatik_monitorobject`

what do you think @lneto ?

lneto:
> This is what I could think about creating closures instead of memoization, (taken example of luadata class)
> 
> C side:
> 
> * luadata_lnew
> * lunatik_newobject
>   
>   * userdata allocated
>   * metatable
>   * create uservalue table
>   * create monitored closures ONCE and store in uservalue
> * return object
> 
> Lua side:
> 
> * obj:method()
>   
>   * __index
>   * return uservalue[method]
> 
> and thus we get rid of `lunatik_monitorobject`
> 
> what do you think @lneto ?

I was thinking on something like this, but on lunatik_newclass (or setclass) itself, we could create our modified metatable there..

lneto:
> Hello @lneto I've completed the benchmark comparing the current master against the new closure-based monitor implementation. Results (1M iterations of getbyte): Master: ~266 ns/call (avg) This branch: ~370 ns/call (avg) The closure creation adds roughly 104ns of overhead per call. The monitoring logic introduces a ~35 - 45% performance hit this is the script I used
> 
> ```lua
> local linux = require("linux")
> local data = require("data")
> 
> local d = data.new(1024)
> local iterations = 1000000
> 
> local start_ns = linux.time()
> for i = 1, iterations do
>     d:getbyte(0)
> end
> local end_ns = linux.time()
> local total_ns = end_ns - start_ns
> 
> local ns_per_call = total_ns / iterations
> 
> print("Total ns: " .. total_ns)
> print("Ns per call: " .. ns_per_call)
> ```

I was thinking here.. we could also measure the impact of memoization by doing `local m_getbyte = d.getbyte` before the loop and the initialization of `start_ns`. So, we can estimate what's the impact of the closure creation and the impact of synchronization itself. What do you think? Can you run it as well?

sneaky-potato:
> So, we can estimate what's the impact of the closure creation and the impact of synchronization itself. 

Some metrics for the following script
```lua
local data = require("data")
local d = data.new(1024)
local iterations = 1000000

local m_getbyte = d.getbyte 

local start_ns = linux.time()
for i = 1, iterations do
    m_getbyte(d, 0)
end
local end_ns = linux.time()

local total_ns = end_ns - start_ns
print("Localized Method Total ns: " .. total_ns)
print("Localized Method Ns per call: " .. (total_ns / iterations))
```

master: Localized Method Ns per call: 117ns
memoization approach (specifically [this commit](https://github.com/luainkernel/lunatik/pull/398/changes/155e3126530c0435f57c24913e959879363642a4): Localized Method Ns per call: 227ns
latest changes (tried something for creating closures on new_class): Localized Method Ns per call: 220ns

and for the following script
```lua
local data = require("data")

local d = data.new(1024)
local iterations = 1000000

local start_ns = linux.time()
for i = 1, iterations do
    d:getbyte(0)
end
local end_ns = linux.time()
local total_ns = end_ns - start_ns

local ns_per_call = total_ns / iterations

print("Total ns: " .. total_ns)
print("Ns per call: " .. ns_per_call)
```

latest changes in this branch report ~240ns/call (master reports 266 ns/call)
It means the original Lunatik method lookup was actually quite expensive. By pre-creating the closures in lunatik_newclass, we can optimize the lookup phase.

lunatik.h
```diff
@@ -236,14 +237,47 @@ static inline bool lunatik_hasindex(lua_State *L, int index)
...
+	const luaL_Reg *reg;
+	for (reg = class->methods; reg->name != NULL; reg++) {
+		lua_getfield(L, -1, reg->name);  // stack: mt, method
+
+		if (!lua_isfunction(L, -1)) {
+			lua_pop(L, 1);
+			continue;
+		}
+
+		// Skip metamethods (starting with __)
```
lneto: please, mind our style; we don't do C++ style comments

lneto:
> > So, we can estimate what's the impact of the closure creation and the impact of synchronization itself.
> 
> Some metrics for the following script
> 
> ```lua
> local linux = require("linux")
> local data = require("data")
> local d = data.new(1024)
> local iterations = 1000000
> 
> local m_getbyte = d.getbyte 
> 
> local start_ns = linux.time()
> for i = 1, iterations do
>     m_getbyte(d, 0)
> end
> local end_ns = linux.time()
> 
> local total_ns = end_ns - start_ns
> print("Localized Method Total ns: " .. total_ns)
> print("Localized Method Ns per call: " .. (total_ns / iterations))
> ```
> 
> master: Localized Method Ns per call: 117ns memoization approach (specifically [this commit](https://github.com/luainkernel/lunatik/pull/398/changes/155e3126530c0435f57c24913e959879363642a4): Localized Method Ns per call: 227ns latest changes (tried something for creating closures on new_class): Localized Method Ns per call: 220ns
> 
> and for the following script
> 
> ```lua
> local linux = require("linux")
> local data = require("data")
> 
> local d = data.new(1024)
> local iterations = 1000000
> 
> local start_ns = linux.time()
> for i = 1, iterations do
>     d:getbyte(0)
> end
> local end_ns = linux.time()
> local total_ns = end_ns - start_ns
> 
> local ns_per_call = total_ns / iterations
> 
> print("Total ns: " .. total_ns)
> print("Ns per call: " .. ns_per_call)
> ```
> 
> latest changes in this branch report ~240ns/call (master reports 266 ns/call) It means the original Lunatik method lookup was actually quite expensive. By pre-creating the closures in lunatik_newclass, we can optimize the lookup phase.

I think it's getting very complex. I would give a step back and break up this problem in two. Firstly, I would try to implement the wrapper, without the improved error handling; then, after it's round, I would introduce the error hadling improvement. What do you think?

sneaky-potato:
> I think it's getting very complex. I would give a step back and break up this problem in two.

Yes, I have created a [PR](https://github.com/luainkernel/lunatik/pull/410). Once that’s reviewed, we can continue iterating here on the remaining pieces.

lneto:
Hey @sneaky-potato, did you rerun your benchmarks? How is the effect of this patch now?

lunatik_obj.c
```diff
@@ -124,18 +124,36 @@ int lunatik_deleteobject(lua_State *L)
 }
 EXPORT_SYMBOL(lunatik_deleteobject);
 
+static int lunatik_error_handler(lua_State *L)
+{
+	const char *method_name = lua_tostring(L, lua_upvalueindex(1));
```
lneto: I would also measure if pushing the method name on the stack and using a cfunction would be more efficient than creating a closure

sneaky-potato: Presently there is no noticeable difference in metrics from latest master. However let me apply this approach and document the metrics again.

lneto: you mean that even pushing the error handling closure, you kept the same gain you introduced on the other PR?

sneaky-potato: I made a mistake while checking, these are the metrics

Master (No descriptive errors):131ns
Closure Approach (current):      263ns

> I would also measure if pushing the method name on the stack and using a cfunction would be more efficient than creating a closure

How will this work? Lua will call error handler with a new stack frame containing only the error message. Any "extra" values we push onto the stack before the call are inaccessible to the handler. Right?

lneto: > > I would also measure if pushing the method name on the stack and using a cfunction would be more efficient than creating a closure
> 
> How will this work? Lua will call error handler with a new stack frame containing only the error message. Any "extra" values we push onto the stack before the call are inaccessible to the handler. Right?

it shouldn't, right? if no one has pop'ed it, it should be there, before the function frame.. pcall doesn't clean the stack from the caller, only the callee..

lunatik_obj.c
```diff
@@ -124,18 +124,36 @@ int lunatik_deleteobject(lua_State *L)
...
+	return 1;
+}
+
 static int lunatik_monitor(lua_State *L)
 {
 	int ret, n = lua_gettop(L);
 	lunatik_object_t *object = lunatik_checkobject(L, 1);
 
+	lua_pushvalue(L, lua_upvalueindex(2)); /* method name */
+	lua_pushcclosure(L, lunatik_error_handler, 1);
```
lneto: perhaps, before trying to use the method name in the stack, you should just try to change this to a pushcfunction, that's ignoring method name for now, and measure the effect.. to check if it will really make difference here..

sneaky-potato: Latest metrics
Master: 136 ns/call
This PR: 152 ns/call

Used `lua_getstack` to inspect the monitor's stack and get method name.

lunatik_obj.c
```diff
@@ -147,7 +173,8 @@ void lunatik_monitorobject(lua_State *L, const lunatik_class_t *class)
 	for (reg = class->methods; reg->name != NULL; reg++) {
 		if (!lunatik_ismetamethod(reg)) {
 			lua_getfield(L, -1, reg->name);
-			lua_pushcclosure(L, lunatik_monitor, 1); /* stack: mt, method */
+			lua_pushstring(L, reg->name);
+			lua_pushcclosure(L, lunatik_monitor, 2); /* stack: mt, method, method name*/
```
lneto, line 165: I wasn't expecting such overhead.. I'm not sure it's due to pushcfunction or to the stack operations (insert).. can you try an alternative approach? just occurred to me that you can push another closure here for error handling as the second upvalue instead of the string (method name).. then, push the 2nd upvalue as the error handler.. let's see what's perform better.. I think this alternative would be more elegant.. if it's perform better, we can go with it.. ;-)

lunatik_obj.c
lneto: > Latest metrics Master:  ns/call This PR: 152 ns/call
> 
> Used `lua_getstack` to inspect the monitor's stack and get method name.

I'm wondering if we couldn't just "prettify" the traceback string, the method name is already there.. actually, for me, the traceback string is already good enough..  I'm not sure ~11% of overhead on every call would be a good tradeoff to have a prettier error message.. what do you think?

lunatik_obj.c
```diff
@@ -124,18 +124,44 @@ int lunatik_deleteobject(lua_State *L)
...
+  	lua_pushvalue(L, lua_upvalueindex(1));
+    lua_insert(L, 2); /* stack: errhandler, method, object, args */
 
 	lunatik_lock(object);
-	ret = lua_pcall(L, n, LUA_MULTRET, 0);
+	ret = lua_pcall(L, n, LUA_MULTRET, 1);
 	lunatik_unlock(object);
 
+	lua_remove(L, 1); /* remove error handler */
 	if (ret != LUA_OK)
```
lneto: we could also try to move the error handling logic to here.. only when the error really occurs.. it should speed up the hot path.. what do you think? the error message will be on the top.. we can just fix it in here and throw the error afterwards

sneaky-potato: This is what I thought initially too but this is like patching the error after it had occurred. This will bring down the latency though.

lunatik_obj.c
sneaky-potato: That is indeed a clever approach but I think we will still incur some latency there. I'll try this though.

sneaky-potato: Tried this approach, this is actually cleaner
final stats:
master: 134 ns/call
this patch: 154 ns/call

lunatik_obj.c
lneto: What's the drawback you see on patching the error?

lunatik_obj.c
lneto: Yeah.. it looks cleaner.. however, it adds 15% of overhead, right.. mixed feelings.. what do you think?

lneto: Peharps we could run a more realistic benchmark instead of just fetching 1 byte.. parsing a TCP header, perhaps.. not sure.. maybe adapting the LLDP and DNS examples..

lunatik_obj.c
sneaky-potato: Building on what discussion we had on matrix
> Perhaps on lunatik_run? It's the one that will start the driver. I think lunatik_runtime is the one that pcalls
> Perhaps this is the right place

I think this should be the place to patch the error message (before re-throwing the error in monitor) because:
- The method name is available as an upvalue
- We fix it before it bubbles up through the stack

lunatik_obj.c
```diff
@@ -136,8 +136,17 @@ static int lunatik_monitor(lua_State *L)
 	ret = lua_pcall(L, n, LUA_MULTRET, 0);
 	lunatik_unlock(object);
 
-	if (ret != LUA_OK)
+	if (ret != LUA_OK) {
```
lneto, line 151: LGTM, but I would move this block to an aux.. I think it would be clearer..

lunatik_obj.c
```diff
@@ -136,8 +136,17 @@ static int lunatik_monitor(lua_State *L)
...
+		const char *method_name = lua_tostring(L, lua_upvalueindex(2));
+		if (method_name) {
+			const char *error_msg = lua_tostring(L, -1);
+			luaL_gsub(L, error_msg, "?", method_name);
+			lua_remove(L, -2);
```
lneto: Can you add a small comment on the side of these two remove calls? Just to indicate what value you are removing from the stack..

sneaky-potato: Resolved in latest push

lunatik_obj.c
```diff
@@ -124,6 +124,18 @@ int lunatik_deleteobject(lua_State *L)
 }
 EXPORT_SYMBOL(lunatik_deleteobject);
 
+static void lunatik_report_error(lua_State *L, const char *method_name)
```
lneto:
```suggestion
inline static void lunatik_fixerror(lua_State *L, const char *method)
```

let's try to use shorter names while maintaining clarity =)

lunatik_obj.c
```diff
@@ -124,6 +124,18 @@ int lunatik_deleteobject(lua_State *L)
...
+{
+	if (method_name) {
+		const char *error_msg = lua_tostring(L, -1);
```
lneto:
```suggestion
		const char *error = lua_tostring(L, -1);
```

lunatik_obj.c
```diff
@@ -124,6 +124,18 @@ int lunatik_deleteobject(lua_State *L)
...
+		luaL_gsub(L, error_msg, "?", method_name);
+		lua_remove(L, -2); /* remove the original error message, keeping the gsub'd string */
```
lneto:
```suggestion
		lua_remove(L, -2); /* error */
```

lunatik_obj.c
```diff
@@ -124,6 +124,18 @@ int lunatik_deleteobject(lua_State *L)
...
+	}
+	luaL_traceback(L, L, lua_tostring(L, -1), 1);
+	lua_remove(L, -2); /* remove the message that lacked the traceback */
```
lneto:
```suggestion
	lua_remove(L, -2); /* fixed error */
```

lunatik_obj.c
```diff
@@ -136,8 +148,10 @@ static int lunatik_monitor(lua_State *L)
 	ret = lua_pcall(L, n, LUA_MULTRET, 0);
 	lunatik_unlock(object);
 
-	if (ret != LUA_OK)
-		lua_error(L);
+	if (ret != LUA_OK) {
+		const char *method_name = lua_tostring(L, lua_upvalueindex(2));
+		lunatik_report_error(L, method_name);
```
lneto:
```suggestion
		const char *method = lua_tostring(L, lua_upvalueindex(2));
		lunatik_fixerror(L, method);
```

lneto (approved):
LGTM; thank for this! sorry re nitpicking, but I think it will integrate better in the code base reducing friction while reading it =)

PR #387, opened 2026-01-15
feat Adding Kconfig to support modularization on openwrt

Zapper9982:
This PR adds a root `Kconfig` file to the Lunatik repository to support granular module selection. 
 
Changes

- Added Kconfig file.
- Defined CONFIG_LUNATIK as the main trigger.
- Created individual config entries for all sub-modules mirroring the existing Kbuild definitions.
- Defaults set to m (module) to maintain standard behavior.

Testing : 

- Confirmed enabling/disabling specific configs correctly aligns with lunatik-objs and obj-* lists in Kbuild

Fixes #373

PR #395, opened 2026-01-16
Auto generate network constants for lua (cont)

lneto:
@sneaky-potato can you take a look? I'm trying to brush it up.. it's still generating `signal["RTMAX"] = _NSIG`; it should be skipped.. please, feel free to add fixups to this branch.

PR #400, opened 2026-01-18
autogenerate lunatik config from makefile flags

### LINKED ISSUE #396: autogenerate modules table on the build based on CONFIG_* from the Makefile
https://github.com/luainkernel/lunatik/blob/master/bin/lunatik#L10, we can have a autogen/lunatik/modules.lua, for instance.

see: https://github.com/luainkernel/lunatik/pull/395

abhijeetw035:
Closes #396

This PR removes the hardcoded module list in `bin/lunatik` and replaces it with an auto-generated configuration file created during the build process.

### Changes
* **New Script (`genconfig.lua`):** Parses `CONFIG_LUNATIK_*` flags to generate `autogen/lunatik/config.lua` containing the `modules` table and kernel release version.
* **Makefile:** Added a `config` target to execute the generator. The configuration flags are passed explicitly to the script to avoid modifying the `all` target variables
* Updated `bin/lunatik` to `require("lunatik.config")` instead of using a static list.
* **Git Handling:** Added `autogen/lunatik/.gitignore` to ensure the directory structure exists on fresh clones while ignoring generated artifacts.

I included an ignore list in `genconfig.lua` to filter out non-module flags (e.g., `INSTALL_EXAMPLES`, `INSTALL_TESTS`) introduced in pending PR #387.

image


Makefile
```diff
@@ -33,10 +33,25 @@ all: lunatik_sym.h
...
+	"CONFIG_LUNATIK=m CONFIG_LUNATIK_RUN=m CONFIG_LUNATIK_RUNTIME=y CONFIG_LUNATIK_DEVICE=m \
+	CONFIG_LUNATIK_LINUX=m CONFIG_LUNATIK_NOTIFIER=m CONFIG_LUNATIK_SOCKET=m \
+	CONFIG_LUNATIK_RCU=m CONFIG_LUNATIK_THREAD=m CONFIG_LUNATIK_FIB=m \
+	CONFIG_LUNATIK_DATA=m CONFIG_LUNATIK_PROBE=m CONFIG_LUNATIK_SYSCALL=m \
+	CONFIG_LUNATIK_XDP=m CONFIG_LUNATIK_FIFO=m CONFIG_LUNATIK_XTABLE=m \
+	CONFIG_LUNATIK_NETFILTER=m CONFIG_LUNATIK_COMPLETION=m \
+	CONFIG_LUNATIK_CRYPTO_SHASH=m CONFIG_LUNATIK_CRYPTO_SKCIPHER=m \
+	CONFIG_LUNATIK_CRYPTO_AEAD=m CONFIG_LUNATIK_CRYPTO_RNG=m \
+	CONFIG_LUNATIK_CRYPTO_COMP=m CONFIG_LUNATIK_CPU=m CONFIG_LUNATIK_HID=m \
+	CONFIG_LUNATIK_SIGNAL=m CONFIG_LUNATIK_BYTEORDER=m"
```
lneto: you shouldn't redefine them here, but access this flags from your lua script.. perhaps we could use CPP for this..

genconfig.lua
```diff
@@ -0,0 +1,61 @@
...
+
+file:write("-- auto-generated, do not edit\n")
+file:write("local config = {}\n\n")
+file:write(string.format('config.kernel_release = "%s"\n', KERNEL_RELEASE))
+file:write("config.modules = {\n")
+for _, mod in ipairs(modules) do
+    file:write(string.format('\t"%s",\n', mod))
+end
+file:write("}\n\n")
+file:write("return config\n")
```
lneto: please.. mind \n

autogen/lunatik/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
```
lneto, line 4: \n

genconfig.lua
```diff
@@ -0,0 +1,61 @@
+-- SPDX-FileCopyrightText: (c) 2023-2026 Ring Zero Desenvolvimento de Software LTDA
```
lneto: it should be your copywrite.. if you are borrowing code from the other script, I recommend you properly modularize it, moving common code to another file and just require it..

abhijeetw035:
Thanks for the feedback @lneto !

1.  I will update the header to use my name in genconfig.lua.

2. Modularization: I checked PR #395. Since gendefines.lua doesn't exist in my branch yet, I can't easily extract common code (like the exit function) into a shared utility file right now. I suggest we merge one of these PRs first, and then do a quick follow-up PR to refactor/deduplicate the common logic?

3. CPP / Duplication: I see that gendefines.lua uses cpp -dM to parse C headers. However, for this task, the input data is the list of Makefile variables (CONFIG_LUNATIK_*), which don't exist in any C header file until the compiler command runs.

Since we can't move these flags to a Makefile variable (to avoid Kconfig breakage), and the Preprocessor can't read the Makefile directly ..... like both ways not possible ... I'm unsure how to use CPP here to avoid the duplication.

so.... would you prefer I stick to the current approach (explicitly passing the flags) as a temporary solution until the Kconfig PR merges (which will give us a .config file to parse)? or do you have a specific way in mind to pipe these Makefile vars to CPP?

lneto:
The problem here is not only to avoid duplication, but redefinition of the flags values.. think about the case you disable some module on `make menuconfig`, it should reflect the autogen. For using CPP, you could have an "artificial" header using such flags.. however, ideally, we should simply export this flags from the build system to your script.. I'm not sure, but perhaps we can access a .config file or something.. or access such values as a environment var.. it needs more investigation, I think

abhijeetw035:
> The problem here is not only to avoid duplication, but redefinition of the flags values.. think about the case you disable some module on `make menuconfig`, it should reflect the autogen. For using CPP, you could have an "artificial" header using such flags.. however, ideally, we should simply export this flags from the build system to your script.. I'm not sure, but perhaps we can access a .config file or something.. or access such values as a environment var.. it needs more investigation, I think

I dug into this a bit. I checked for artifacts, but no `.config` or `auto.conf` files are generated in the directory. It seems the flags only exist transiently on the command line while `make` is running and aren't saved anywhere.
image
So right now, there isn't really a persistent file for the script to read from.

lneto:
hey @abhijeetw035 I took a look at this.. I think we should have two separate approaches.. one for the off-tree build, that is more or less what you already doing.. but assigning the value of the Makefile variable (instead of redefining it).. e.g.,

`CONFIG_LUNATIK_LINUX=$(CONFIG_LUNATIK_LINUX)`, instead of `CONFIG_LUNATIK_LINUX=m`. I hope that the difference is clear..

Another think is the Kconfig, for this, we should assume it will be build on-tree and leverage `/lib/modules/$(uname -r)/build/include/generated/autoconf.h`, which should take precedence over the Makefile flags. I would split this issue in two and left Kconfig/autoconf as a next step. What do you think?

abhijeetw035:
That sounds good !  will handle the off-tree logic now,  and we can look at the Kconfig / autoconf side in a follow-up once there’s some support for that.

I will refactor the Makefile to use variable references as u suggested so it correctly respects user overrides

Pushing those changes shortly!

abhijeetw035:
I've refactored the Makefile 

Moved all flags to top-level variables using ?= m and confirmed that running make `CONFIG_LUNATIK_SIGNAL=n` correctly removes the module from the generated config 

image

Makefile
```diff
@@ -20,23 +20,87 @@ RM = rm -f
...
+	CONFIG_LUNATIK_COMPLETION=$(CONFIG_LUNATIK_COMPLETION) \
+	CONFIG_LUNATIK_CRYPTO_SHASH=$(CONFIG_LUNATIK_CRYPTO_SHASH) \
+	CONFIG_LUNATIK_CRYPTO_SKCIPHER=$(CONFIG_LUNATIK_CRYPTO_SKCIPHER) \
+	CONFIG_LUNATIK_CRYPTO_AEAD=$(CONFIG_LUNATIK_CRYPTO_AEAD) \
+	CONFIG_LUNATIK_CRYPTO_RNG=$(CONFIG_LUNATIK_CRYPTO_RNG) \
+	CONFIG_LUNATIK_CRYPTO_COMP=$(CONFIG_LUNATIK_CRYPTO_COMP) \
+	CONFIG_LUNATIK_CPU=$(CONFIG_LUNATIK_CPU) \
+	CONFIG_LUNATIK_HID=$(CONFIG_LUNATIK_HID) \
+	CONFIG_LUNATIK_SIGNAL=$(CONFIG_LUNATIK_SIGNAL) \
+	CONFIG_LUNATIK_BYTEORDER=$(CONFIG_LUNATIK_BYTEORDER)
```
lneto, line 79: isn't it possible to put this in a var to avoid duplication? it would be easier for maintaining the Makefile

lneto:
hey @abhijeetw035 can you test this along with our OpenWRT feed? My main concern now is that these autogen script don't break it. Thanks! Let me know if you need help.

lneto:
> hey @abhijeetw035 can you test this along with our OpenWRT feed? My main concern now is that these autogen script don't break it. Thanks! Let me know if you need help.

most specifically, you should call your autogen script here https://github.com/luainkernel/openwrt_feed/blob/openwrt-23.05/kernel/lunatik/Makefile#L127.

abhijeetw035:
image

I just simulated the OpenWRT build locally by feeding the exact `EXTRA_KCONFIG` string from the feed into genconfig.lua.
It works perfectly
modules set to =n (like `PROBE` and `SYSCALL` in your example) were correctly excluded from the generated list.

So, once this is merged, you can safely update the OpenWRT feed to call this script in Build/Prepare (line 127) passing `$(EXTRA_KCONFIG)`, and it will generate the correct config.lua.

lneto:
> image
> I just simulated the OpenWRT build locally by feeding the exact `EXTRA_KCONFIG` string from the feed into genconfig.lua. It works perfectly modules set to =n (like `PROBE` and `SYSCALL` in your example) were correctly excluded from the generated list.
> 
> So, once this is merged, you can safely update the OpenWRT feed to call this script in Build/Prepare (line 127) passing `$(EXTRA_KCONFIG)`, and it will generate the correct config.lua.

it's not enough, we should run it using the OpenWRT build

lneto:
>> ...
> it's not enough, we should run it using the OpenWRT build

see this discussion, for instance, https://github.com/luainkernel/lunatik/pull/387

abhijeetw035:
> > > ...
> > > it's not enough, we should run it using the OpenWRT build
> 
> see this discussion, for instance, #387

Suree will go through it and update you.

lneto:
> > > > ...
> > > > it's not enough, we should run it using the OpenWRT build
> > 
> > 
> > see this discussion, for instance, #387
> 
> Suree will go through it and update you.

Well, I will need to work on this for https://github.com/luainkernel/lunatik/pull/395. Perhaps you want to put this issue on hold until we merge the other and work on another issue for a while. It might be better usage of both of our time. What do you think?

abhijeetw035:
> > > > > ...
> > > > > it's not enough, we should run it using the OpenWRT build
> > > 
> > > 
> > > see this discussion, for instance, #387
> > 
> > 
> > Suree will go through it and update you.
> 
> Well, I will need to work on this for #395. Perhaps you want to put this issue on hold until we merge the other and work on another issue for a while. It might be better usage of both of our time. What do you think?

yes suree ....it makes sense to synchronize the OpenWRT work since both PRs touch the same integration points.

genconfig.lua
```diff
@@ -0,0 +1,62 @@
...
+	os.exit(1)
+end
+
+local KERNEL_RELEASE = arg[1]
+local OUTPUT_DIR = arg[2]
+local CONFIGS = arg[3]
+
+if not KERNEL_RELEASE or not OUTPUT_DIR or not CONFIGS then
+	exit("usage: lua genconfig.lua <KERNEL_RELEASE> <OUTPUT_DIR> <CONFIGS>")
+end
```
lneto: the other PR is merged.. let's avoid replicate code and sharing the common code

lneto: mabe a better approach is just to insert your logic on gendefines.lua, se we will have a single entry point on the Makefile

lneto: @abhijeetw035 it's weird, it looks like you are re-adding the changes that are already in the base.. did you rebase your branch on master?

gendefines.lua
```diff
@@ -1,5 +1,6 @@
 --
 -- SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+-- SPDX-FileCopyrightText: (c) 2026 Abhijeet Waghmare <abhijeetw035@gmail.com>
```
lneto: We usually don't do this, we keep the original copyright unless there's very significant change in the original, such as major refactor. Which is not the case of adding new features. If you want to modularize it and move the config portion to a sublib, it's alright to add your copyright to the new file.

lneto: We could have autogen/{init,defines,config}.lua for instance

genconfig.lua
abhijeetw035: > @abhijeetw035 it's weird, it looks like you are re-adding the changes that are already in the base.. did you rebase your branch on master?

that was an artifact from squashing the merge commit. I have rebased onto origin/master to filter out the upstream changes. The diff should be clean now!

abhijeetw035: > Please, see this [#395 (comment)](https://github.com/luainkernel/lunatik/pull/395#issuecomment-3794542741).
> 
> We don't use merge commits.. `--ff-only`



> @abhijeetw035 it's weird, it looks like you are re-adding the changes that are already in the base.. did you rebase your branch on master?

fixing it ...

gendefines.lua
abhijeetw035: I've rebased onto the latest master and squashed everything into a single commit to keep the history linear.
Changes:
copyright: removed my copyright header from `gendefines.lua` .
runtime: updated `bin/lunatik` to load the generated config dynamically (added `/lib/modules/lua` to package path).

Regarding the modularization (`autogen/{defines,config}.lua`): I agree that is a cleaner architecture. however, to keep this PR focused and safe, i think we should merge this functional version first.  I can then submit a follow-up PR dedicated to splitting the scripts and organizing the autogen/ structure properly. what do you think?

abhijeetw035:
I noticed the CI build is failing with make: /bin/sh: 1: lua5.4: not found.

It seems the GitHub Actions runner doesn't have lua5.4 installed. what can we do for now ?

bin/lunatik
```diff
@@ -1,16 +1,17 @@
-#!/usr/bin/lua5.4
+#!/usr/bin/lua
```
lneto: Please, don't.

Makefile
```diff
@@ -130,9 +184,6 @@ uninstall: scripts_uninstall modules_uninstall
 lunatik_sym.h: $(LUA_API) gensymbols.sh
 	${shell CC='$(CC)' ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
 
-defines:
```
lneto, line 133: ?

lneto:
> I noticed the CI build is failing with make: /bin/sh: 1: lua5.4: not found.
> 
> It seems the GitHub Actions runner doesn't have lua5.4 installed. what can we do for now ?

I don't think it's the problem as master (and other branches are alright with CI build)..

Makefile
```diff
@@ -21,23 +21,76 @@ RM = rm -f
...
-	CONFIG_LUNATIK_CRYPTO_SHASH=m CONFIG_LUNATIK_CRYPTO_SKCIPHER=m \
-	CONFIG_LUNATIK_CRYPTO_AEAD=m CONFIG_LUNATIK_CRYPTO_RNG=m \
-	CONFIG_LUNATIK_CRYPTO_COMP=m CONFIG_LUNATIK_CPU=m CONFIG_LUNATIK_HID=m \
-	CONFIG_LUNATIK_SIGNAL=m CONFIG_LUNATIK_BYTEORDER=m
+	${MAKE} -C ${MODULES_BUILD_PATH} M=${PWD} $(LUNATIK_CONFIG_FLAGS)
+
+defines:
+	@mkdir -p autogen/lunatik autogen/linux
+	CC='${CC}' lua gendefines.lua "$(KERNEL_RELEASE)" "$(INCLUDE_PATH)" "autogen/linux" "$(LUNATIK_CONFIG_FLAGS)"
+	@mv autogen/linux/config.lua autogen/lunatik/
```
lneto: why this?

Makefile
```diff
@@ -21,23 +21,76 @@ RM = rm -f
...
-	CONFIG_LUNATIK_NETFILTER=m CONFIG_LUNATIK_COMPLETION=m \
-	CONFIG_LUNATIK_CRYPTO_SHASH=m CONFIG_LUNATIK_CRYPTO_SKCIPHER=m \
-	CONFIG_LUNATIK_CRYPTO_AEAD=m CONFIG_LUNATIK_CRYPTO_RNG=m \
-	CONFIG_LUNATIK_CRYPTO_COMP=m CONFIG_LUNATIK_CPU=m CONFIG_LUNATIK_HID=m \
-	CONFIG_LUNATIK_SIGNAL=m CONFIG_LUNATIK_BYTEORDER=m
+	${MAKE} -C ${MODULES_BUILD_PATH} M=${PWD} $(LUNATIK_CONFIG_FLAGS)
+
+defines:
+	@mkdir -p autogen/lunatik autogen/linux
+	CC='${CC}' lua gendefines.lua "$(KERNEL_RELEASE)" "$(INCLUDE_PATH)" "autogen/linux" "$(LUNATIK_CONFIG_FLAGS)"
```
lneto: this should be `lua5.4`

Makefile
```diff
@@ -21,23 +21,76 @@ RM = rm -f
...
-	CONFIG_LUNATIK_DATA=m CONFIG_LUNATIK_PROBE=m CONFIG_LUNATIK_SYSCALL=m \
-	CONFIG_LUNATIK_XDP=m CONFIG_LUNATIK_FIFO=m CONFIG_LUNATIK_XTABLE=m \
-	CONFIG_LUNATIK_NETFILTER=m CONFIG_LUNATIK_COMPLETION=m \
-	CONFIG_LUNATIK_CRYPTO_SHASH=m CONFIG_LUNATIK_CRYPTO_SKCIPHER=m \
-	CONFIG_LUNATIK_CRYPTO_AEAD=m CONFIG_LUNATIK_CRYPTO_RNG=m \
-	CONFIG_LUNATIK_CRYPTO_COMP=m CONFIG_LUNATIK_CPU=m CONFIG_LUNATIK_HID=m \
-	CONFIG_LUNATIK_SIGNAL=m CONFIG_LUNATIK_BYTEORDER=m
+	${MAKE} -C ${MODULES_BUILD_PATH} M=${PWD} $(LUNATIK_CONFIG_FLAGS)
+
+defines:
```
lneto: why did you move this block around?

bin/lunatik
```diff
@@ -1,16 +1,17 @@
...
 --
 -- SPDX-FileCopyrightText: (c) 2023-2026 Ring Zero Desenvolvimento de Software LTDA
 -- SPDX-License-Identifier: MIT OR GPL-2.0-only
 --
 
+package.path = "/lib/modules/lua/?.lua;" .. package.path
```
lneto: why do we need this? I don't think we should add it.

gendefines.lua
```diff
@@ -102,9 +103,62 @@ local function write_module(spec, constants, macro_names)
 	file:write("\nreturn " .. spec.module_name .. "\n")
 end
 
+
+local function generate_build_config(kernel_release, output_dir, configs)
+    local ignore_list = {
```
lneto: can you confirm you are using tabs?

gendefines.lua
```diff
@@ -102,9 +103,62 @@ local function write_module(spec, constants, macro_names)
...
+        end
+    end
+
+    local filepath = output_dir .. "/config.lua"
+    local file <close>, err = io.open(filepath, "w")
+    if not file then
+        exit("cannot open " .. filepath .. ": " .. err)
+    end
+
+    file:write("-- auto-generated, do not edit\n")
```
lneto: we should avoid repeat this, instead we should proper modularize the code base.

abhijeetw035: yes.... I will submit a follow-up PR immediately after this merges to split gendefines.lua into proper modules. I kept it inline here to keep this PR focused on the feature.

Makefile
abhijeetw035: Fixed. That was an accidental move during editing.

Makefile
abhijeetw035: yes i was checking if the `lua` works .like before ci build was failing due to that 
/bin/sh: 1: lua5.4: not found

Makefile
abhijeetw035: the script was generating the file in the autogen/linux directory (based on the output argument), so I added this to move it to the correct lunatik folder.

 now updated gendefines.lua to write the config file directly to autogen/lunatik/config.lua, so this mv command is removed.

abhijeetw035:
> > I noticed the CI build is failing with make: /bin/sh: 1: lua5.4: not found.
> > It seems the GitHub Actions runner doesn't have lua5.4 installed. what can we do for now ?
> 
> I don't think it's the problem as master (and other branches are alright with CI build)..

I checked the build.yaml file and I see the issue.

The workflow installs `lua5.4-dev`, which provides the headers for compiling the C module (which is why master works). However, it does not explicitly install the lua5.4 package, which provides the CLI interpreter.

Since my PR requires running a Lua script during the build, the runner fails to find the binary.

Makefile
```diff
@@ -131,7 +180,8 @@ lunatik_sym.h: $(LUA_API) gensymbols.sh
 	${shell CC='$(CC)' ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
 
 defines:
-	${shell CC='$(CC)' lua5.4 gendefines.lua $(KERNEL_RELEASE) $(INCLUDE_PATH) autogen/linux}
+	@mkdir -p autogen/lunatik autogen/linux
```
lneto: we shouldn't do this.. it's handled by the .gitignore

Makefile
```diff
@@ -131,7 +180,8 @@ lunatik_sym.h: $(LUA_API) gensymbols.sh
...
+	CC='${CC}' lua5.4 gendefines.lua "$(KERNEL_RELEASE)" "$(INCLUDE_PATH)" "autogen/linux" "$(LUNATIK_CONFIG_FLAGS)"
```
lneto: why did you remove `${shell ...}`? this seems to be CI issue to me.

gendefines.lua
```diff
@@ -102,9 +103,62 @@ local function write_module(spec, constants, macro_names)
...
+	local modules = {}
+
+	for token in configs:gmatch("%S+") do
+		local key = token:match("(CONFIG_[%w_]+)=[my]")
+		if key then
+			local name
+			if key == "CONFIG_LUNATIK" then
+				name = "lunatik"
+			elseif key == "CONFIG_LUNATIK_RUN" then
+				name = "lunatik_run"
```
lneto: you can use the same approach of `ignore_list` here.. having an `overlay_list` (couldn't figure a better name)..

gendefines.lua
```diff
@@ -102,9 +103,62 @@ local function write_module(spec, constants, macro_names)
...
+	local ignore_list = {
+		["INSTALL_EXAMPLES"] = true,
+		["INSTALL_TESTS"] = true,
+		["RUNTIME"] = true
+	}
+
+	local modules = {}
+
+	for token in configs:gmatch("%S+") do
+		local key = token:match("(CONFIG_[%w_]+)=[my]")
```
lneto: let's avoid use `key` as a variable name and have more meaningful names instead..

lneto:
> > > I noticed the CI build is failing with make: /bin/sh: 1: lua5.4: not found.
> > > It seems the GitHub Actions runner doesn't have lua5.4 installed. what can we do for now ?
> > 
> > 
> > I don't think it's the problem as master (and other branches are alright with CI build)..
> 
> I checked the build.yaml file and I see the issue.
> 
> The workflow installs `lua5.4-dev`, which provides the headers for compiling the C module (which is why master works). However, it does not explicitly install the lua5.4 package, which provides the CLI interpreter.
> 
> Since my PR requires running a Lua script during the build, the runner fails to find the binary.

the defines

> > > I noticed the CI build is failing with make: /bin/sh: 1: lua5.4: not found.
> > > It seems the GitHub Actions runner doesn't have lua5.4 installed. what can we do for now ?
> > 
> > 
> > I don't think it's the problem as master (and other branches are alright with CI build)..
> 
> I checked the build.yaml file and I see the issue.
> 
> The workflow installs `lua5.4-dev`, which provides the headers for compiling the C module (which is why master works). However, it does not explicitly install the lua5.4 package, which provides the CLI interpreter.
> 
> Since my PR requires running a Lua script during the build, the runner fails to find the binary.

master is already running gendefines.lua on build.. I think the problem is to suppress `${shell ...}`; please, be more careful when removing or moving code around.

gendefines.lua
lneto: my point is that we should at least avoid to repeat strings that can be added to a local var or very simple patterns that can be added to a local function..

gendefines.lua
```diff
@@ -102,9 +103,62 @@ local function write_module(spec, constants, macro_names)
...
+				end
+			end
+
+			if name then
+				table.insert(modules, name)
+			end
+		end
+	end
+
+	local filepath = "autogen/lunatik/config.lua"
```
lneto: This should come from the args, like on defines. You coul pass `autogen` dir and concat subdir on both cases.

gendefines.lua
```diff
@@ -29,6 +29,7 @@ end
 local KERNEL = arg[1]
 local INCLUDE_PATH = arg[2]
 local OUTPUT_DIR = arg[3]
+local CONFIG_FLAGS = arg[4]
 
 if not KERNEL or not INCLUDE_PATH or not OUTPUT_DIR then
 	exit("usage: lua gendefines.lua <KERNEL> <INCLUDE_PATH> <OUTPUT_DIR>")
```
lneto: It should be updated with CONFIG_FLAGS

gendefines.lua
```diff
@@ -102,9 +103,62 @@ local function write_module(spec, constants, macro_names)
...
+	file:write("return config\n")
+end
+
 for _, spec in ipairs(specs) do
 	local cpp_output = preprocess(spec.header)
 	local constants, macro_names = collect_constants(cpp_output, spec.prefix)
 	write_module(spec, constants, macro_names)
 end
 
+if CONFIG_FLAGS then
```
lneto: it should be checked along with the other args.. it should be a requirement..

lneto:
@sneaky-potato can you take a look at this as well? thanks!

gendefines.lua
```diff
@@ -102,9 +103,62 @@ local function write_module(spec, constants, macro_names)
...
+	file:write(string.format('config.kernel_release = "%s"\n', kernel_release))
+	file:write("config.modules = {\n")
+	for _, mod in ipairs(modules) do
+		file:write(string.format('\t"%s",\n', mod))
+	end
+	file:write("}\n\n")
+	file:write("return config\n")
+end
+
+local linux_dir = OUTPUT_DIR .. "/linux"
```
sneaky-potato:
```suggestion

```

gendefines.lua
```diff
@@ -102,9 +103,62 @@ local function write_module(spec, constants, macro_names)
...
+
 for _, spec in ipairs(specs) do
 	local cpp_output = preprocess(spec.header)
 	local constants, macro_names = collect_constants(cpp_output, spec.prefix)
-	write_module(spec, constants, macro_names)
+	write_module(spec, constants, macro_names, linux_dir)
```
sneaky-potato:
```suggestion
	write_module(spec, constants, macro_names, OUTPUT_DIR)
```

gendefines.lua
```diff
@@ -83,8 +84,8 @@ local function resolve_value(value, constants, prefix, module_name, seen)
     return resolve_value(constants[value], constants, prefix, module_name, seen)
 end
 
-local function write_module(spec, constants, macro_names)
-	local filepath = string.format("%s/%s.lua", OUTPUT_DIR, spec.module_name)
+local function write_module(spec, constants, macro_names, dir)
+	local filepath = string.format("%s/%s.lua", dir, spec.module_name)
```
sneaky-potato:
```suggestion
local function write_module(spec, constants, macro_names, output_dir)
	local filepath = string.format("%s/linux/%s.lua", output_dir, spec.module_name)
```

gendefines.lua
sneaky-potato: To keep the script uniform, let's pass the same base directory to both `write_module` and `generate_build_config`

lneto:
@sneaky-potato, it seems @abhijeetw035 has committed your suggestions. Please, let me know if we can merge it. Thanks!

PR #407, opened 2026-01-21
fix SKB resize for MAC header on luanetfilter

lneto:
introduced on https://github.com/luainkernel/lunatik/commit/c0bcb81d88d06b95692c2b223c488c786aa318c1

### DIFF
--- a/lib/luadata.c
+++ b/lib/luadata.c
@@ -32,11 +32,13 @@
 
 #include "luadata.h"
 
-#define LUADATA_TOPTR(d)  ((d)->opt & LUADATA_OPT_SKB ? (((struct sk_buff *)(d)->ptr)->data) : (d)->ptr)
 #define LUADATA_TOSKB(d)  ((struct sk_buff *)(d)->ptr)
+#define LUADATA_TOPTR(d)	\
+	(((d)->opt & LUADATA_OPT_SKB ? (LUADATA_TOSKB(d)->data) : (d)->ptr) + (d)->offset)
 
 typedef struct luadata_s {
 	void *ptr;
+	ptrdiff_t offset;
 	size_t size;
 	uint8_t opt;
 } luadata_t;
@@ -177,7 +179,7 @@ static int luadata_resize(lua_State *L)
 	if (data->opt & LUADATA_OPT_SKB)
 		luadata_skb_resize(L, data, new_size); 
 	else if (data->opt & LUADATA_OPT_FREE)
-		data->ptr = lunatik_checknull(L, lunatik_realloc(L, data->ptr, new_size)); 
+		data->ptr = lunatik_checknull(L, lunatik_realloc(L, data->ptr, new_size));
 	else
 		luaL_error(L, "cannot resize external memory");
 
@@ -409,15 +411,21 @@ static const lunatik_class_t luadata_class = {
 	.sleep = false,
 };
 
+static inline void luadata_set(luadata_t *data, void *ptr, ptrdiff_t offset, size_t size, uint8_t opt)
+{
+	data->ptr = ptr;
+	data->offset = offset;
+	data->size = size;
+	data->opt = opt;
+}
+
 static int luadata_lnew(lua_State *L)
 {
 	size_t size = (size_t)luaL_checkinteger(L, 1);
 	lunatik_object_t *object = lunatik_newobject(L, &luadata_class, sizeof(luadata_t));
 	luadata_t *data = (luadata_t *)object->private;
 
-	data->ptr = lunatik_checkalloc(L, size);
-	data->size = size;
-	data->opt = LUADATA_OPT_FREE;
+	luadata_set(data, lunatik_checkalloc(L, size), 0, size, LUADATA_OPT_FREE);
 	return 1; /* object */
 }
 
@@ -429,9 +437,7 @@ static inline lunatik_object_t *luadata_create(void *ptr, size_t size, bool slee
 
 	if (object != NULL) {
 		luadata_t *data = (luadata_t *)object->private;
-		data->ptr = ptr;
-		data->size = size;
-		data->opt = opt;
+		luadata_set(data, ptr, 0, size, opt);
 	}
 	return object;
 }
@@ -444,7 +450,7 @@ lunatik_object_t *luadata_new(lua_State *L)
 }
 EXPORT_SYMBOL(luadata_new);
 
-int luadata_reset(lunatik_object_t *object, void *ptr, size_t size, uint8_t opt)
+int luadata_reset(lunatik_object_t *object, void *ptr, ptrdiff_t offset, size_t size, uint8_t opt)
 {
 	luadata_t *data;
 
@@ -456,9 +462,8 @@ int luadata_reset(lunatik_object_t *object, void *ptr, size_t size, uint8_t opt)
 		return -1;
 	}
 
-	data->ptr = ptr;
-	data->size = size;
-	data->opt = opt & LUADATA_OPT_KEEP ? data->opt : opt;
+	opt = opt & LUADATA_OPT_KEEP ? data->opt : opt;
+	luadata_set(data, ptr, offset, size, opt);
 
 	lunatik_unlock(object);
 	return 0;
@@ -477,5 +482,5 @@ static void __exit luadata_exit(void)
 module_init(luadata_init);
 module_exit(luadata_exit);
 MODULE_LICENSE("Dual MIT/GPL");
-MODULE_AUTHOR("Lourival Vieira Neto <lourival.neto@ring-0.io>");
+MODULE_AUTHOR("Lourival Vieira Neto <lourival.neto@ringzero.com.br>");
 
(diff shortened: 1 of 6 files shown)


lneto:
@abhijeetw035 can you take a look at this and try to run your resize tests?

sheharyaar:
Sorry if I missed the previous commit's context, why do we need `offset` here ?

lneto:
> Sorry if I missed the previous commit's context, why do we need `offset` here ?

Because of the MAC header, see the commit that introduced the resize. It has discarded the MAC header offset; accessing skb->data on both cases.

sheharyaar:
Oh yeah, I see it here: https://github.com/luainkernel/lunatik/commit/c0bcb81d88d06b95692c2b223c488c786aa318c1#diff-7af6ccd3b938d3efe6965a52e59f84c7f34aae6f825864536661cf83f544a277L59

carloslack:
oh the previous change is from December, how did you see the first time, did it crash or only messed with the data?

carloslack:
I did a test before and after your fix, with my `local function dnsblock_hook(skb, ifindex)`  it is at least behaving the same way.

lneto:
> Oh yeah, I see it here: [c0bcb81#diff-7af6ccd3b938d3efe6965a52e59f84c7f34aae6f825864536661cf83f544a277L59](https://github.com/luainkernel/lunatik/commit/c0bcb81d88d06b95692c2b223c488c786aa318c1#diff-7af6ccd3b938d3efe6965a52e59f84c7f34aae6f825864536661cf83f544a277L59)

https://github.com/luainkernel/lunatik/commit/c0bcb81d88d06b95692c2b223c488c786aa318c1#diff-001cb862a4cd47321d93ddf2043955a2a438442c500673db65e5399279a4195dL80

See above, we were passing the pointer to MAC header when it was available; but on this change we started passing skb->data for both cases, what prevent us to inspect L2.

lneto:
> oh the previous change is from December, how did you see the first time, did it crash or only messed with the data?

It was because I wasn't running master on daily bases, I was using the latest release on my router. When I upgraded it to v4.0, it wasn't able to parse L2 anymore. But it doesn't crash; it's just the case that it has no access to MAC header anymore.

abhijeetw035:
> @abhijeetw035 can you take a look at this and try to run your resize tests?

Hii @lneto ,

ihave tested the resize functionality with the fix from commit `05fc16e`. 

### Test Setup
```bash
sudo lunatik run examples/test_netfilter_resize false
ping -c 3 127.0.0.1
```

### Test Results
The resize functionality is working correctly across different packet sizes:
**Successful SKB Resize Operations:**
- 52 → 62 → 52 bytes 
- 118 → 128 → 118 bytes 
- 152 → 162 → 152 bytes 
- 163 → 173 → 163 bytes 

**Expected Error Handling:**
- 110, 111, 112, 114 byte packets: `luanetfilter: insufficient tailroom for resize` 
  (This is correct behavior when there isn't enough tailroom to expand)

no crashes were there.

image

### Test script used :
```c
local netfilter = require("netfilter")
local data = require("data")

local test_count = 0

local hook = netfilter.register({
    pf = netfilter.family.INET,
    hooknum = netfilter.inet_hooks.LOCAL_OUT,
    priority = netfilter.ip_priority.FIRST,
    hook = function(skb, ifindex)
        -- Only run tests on first packet
        if test_count > 0 then
            return netfilter.action.ACCEPT
        end
        test_count = test_count + 1
        
        print("--- Test 1: Raw Buffer Resize ---")
        local d = data.new(10)
        print("Initial size: " .. tostring(#d))
        
        d:setstring(0, "hello")
        
        d:resize(20)
        print("After resize(20): " .. tostring(#d))
        
        d:resize(5)
        print("After resize(5): " .. tostring(#d))
        
        print("--- Test 2: SKB Resize ---")
        local orig_size = #skb
        print("Original SKB size: " .. tostring(orig_size))
        
        -- Test expansion
        local ok, err = pcall(function()
            skb:resize(orig_size + 10)
        end)
        
        if ok then
            print("After resize(+10): " .. tostring(#skb))
            
            -- Test shrinking back
            skb:resize(orig_size)
            print("After resize(back): " .. tostring(#skb))
        else
            print("SKB resize failed: " .. tostring(err))
        end
        
        print("=== Tests Complete ===")
        
        return netfilter.action.ACCEPT
    end
})

print("Hook registered") ```

File: lib/luaxdp.c [hunk 1]
```diff
@@ -51,7 +51,7 @@ static inline lunatik_object_t *luaxdp_pushdata(lua_State *L, int upvalue, void
 
 	lua_pushvalue(L, lua_upvalueindex(upvalue));
 	data = (lunatik_object_t *)lunatik_toobject(L, -1);
-	luadata_reset(data, ptr, size, LUADATA_OPT_KEEP);
+	luadata_reset(data, ptr, 0, size, LUADATA_OPT_KEEP);
```
carloslack, line 54: should you use a constant for `0` here? for readability

lneto:
> > @abhijeetw035 can you take a look at this and try to run your resize tests?
> 
> Hii @lneto ,
> 
> ihave tested the resize functionality with the fix from commit `05fc16e`.
> (...)

Thank you very much!

lib/luaxdp.c
lneto: I'm not sure, what do you have in mind? it's an `offset`, so I would expect a number here not a definition like a option/flag.

lneto: and we should do the same when size = 0, right? like here https://github.com/luainkernel/lunatik/pull/407/changes#diff-d1ecc792151b5ec5af5df97a5be9a66e2648ab2547a6ec49cf9beb20a93a9275R19

lib/luadata.c
```diff
@@ -177,7 +179,7 @@ static int luadata_resize(lua_State *L)
     if (data->opt & LUADATA_OPT_SKB)
 		luadata_skb_resize(L, data, new_size); 
 	else if (data->opt & LUADATA_OPT_FREE)
-		data->ptr = lunatik_checknull(L, lunatik_realloc(L, data->ptr, new_size)); 
+		data->ptr = lunatik_checknull(L, lunatik_realloc(L, data->ptr, new_size + data->offset));
```
lneto: as we cannot reset LUADATA_OPT_FREE, data->offset will never be diff than 0, here.. so we shouldn't take this in consideration here..

lib/luaxdp.c
carloslack: It is nit actually, someone looking at first time may wonder the `0` is, as we have `data`, `ptr`, etc
but one can always look at the function anyway and see, not a big deal.

lib/luaxtable.c
```diff
@@ -138,7 +138,7 @@ static int luaxtable_domatch(lua_State *L, luaxtable_t *xtable, const struct sk_
 
 static int luaxtable_dotarget(lua_State *L, luaxtable_t *xtable, struct sk_buff *skb, const struct xt_action_param *par, int fallback)
 {
-	if (luaxtable_call(L, "target", xtable, skb, par, (luaxtable_info_t *)par->targinfo, LUADATA_OPT_SKB) != 0)
+	if (luaxtable_call(L, "target", xtable, skb, par, (luaxtable_info_t *)par->targinfo, LUADATA_OPT_NONE) != 0)
```
carloslack, line 141: `LUADATA_OPT_NONE` makes sense here , right? My question is why was it assuming skb, was it an issue by itself?

lneto: Yeah, it was the original value. `LUADATA_OPT_SKB` was introduced on the commit we are fixing here. I think this was an attempt to add support for SKB resize on stables, but it's not enough. For this, we would need to pass the SKB struct pointer on `reset` and the right offset. Perhaps we should open a new issue to do this. However, I'm not sure we should keep support for new things on luaxtable as iptables is legacy. Maybe we should just drop support for iptables on 4.x; not sure. What do you think?

@sheharyaar it would be good to have your view on this as well ;-).

lneto:
@carloslack @sheharyaar @abhijeetw035 if you don't oppose, I would like to merge this and proceed with 4.0 release ;-)

lib/luaxtable.c
sheharyaar: Since we have already decided to drop support for kernel 4.x, we can drop iptables support for 4.x as well.

sheharyaar:
> @carloslack @sheharyaar @abhijeetw035 if you don't oppose, I would like to merge this and proceed with 4.0 release ;-)

Sure we can go ahead if the tests are working fine.

lib/luaxtable.c
carloslack: The fix can be done, but I second @sheharyaar because kernel `4.x` support is no more.

abhijeetw035:
> > @carloslack @sheharyaar @abhijeetw035 if you don't oppose, I would like to merge this and proceed with 4.0 release ;-)
> 
> Sure we can go ahead if the tests are working fine.

Yes i also checked from my side 
Tests were working fine without crashing.
image

PR #410, opened 2026-01-24
wrap class methods eagerly instead of via __index

sneaky-potato:
This patch eliminates the need for the `__index` metamethod to handle method dispatching for shared modules. By eagerly pre-wrapping methods in the `lunatik_monitor` closure at class initialization, we avoid repeated closure allocations and metatable churn in performance-critical paths.

Tested performance with the following script
```lua
local data = require("data")

local d = data.new(1024)
local iterations = 1000000

local start_ns = linux.time()
for i = 1, iterations do
    d:getbyte(0)
end
local end_ns = linux.time()
local total_ns = end_ns - start_ns

local ns_per_call = total_ns / iterations

print("Total ns: " .. total_ns)
print("Ns per call: " .. ns_per_call)


local m_getbyte = d.getbyte 

start_ns = linux.time()
for i = 1, iterations do
    m_getbyte(d, 0)
end
end_ns = linux.time()
total_ns = end_ns - start_ns

ns_per_call = total_ns / iterations

print("Localized method total ns: " .. total_ns)
print("Localized method Ns per call: " .. ns_per_call)
```

master:
```shell
Total ns: 275151687
Ns per call: 275
Localized method total ns: 137531899
Localized method Ns per call: 137
```

this branch:
```shell
Total ns: 128897733
Ns per call: 128
Localized method total ns: 130847719
Localized method Ns per call: 130
```

### DIFF
--- a/lib/luacrypto_aead.c
+++ b/lib/luacrypto_aead.c
@@ -224,7 +224,6 @@ static const luaL_Reg luacrypto_aead_mt[] = {
 	{"decrypt", luacrypto_aead_decrypt},
 	{"__gc", lunatik_deleteobject},
 	{"__close", lunatik_closeobject},
-	{"__index", lunatik_monitorobject},
 	{NULL, NULL}
 };
 
@@ -237,6 +236,7 @@ static const lunatik_class_t luacrypto_aead_class = {
 	.methods = luacrypto_aead_mt,
 	.release = luacrypto_aead_release,
 	.sleep = true,
+	.shared = true,
 	.pointer = true,
 };
 
--- a/lib/luafifo.c
+++ b/lib/luafifo.c
@@ -116,7 +116,6 @@ static const luaL_Reg luafifo_lib[] = {
 * @treturn nil
 */
 static const luaL_Reg luafifo_mt[] = {
-	{"__index", lunatik_monitorobject},
 	{"__gc", lunatik_deleteobject},
 	{"__close", lunatik_closeobject},
 	{"close", lunatik_closeobject},
@@ -130,6 +129,7 @@ static const lunatik_class_t luafifo_class = {
 	.methods = luafifo_mt,
 	.release = luafifo_release,
 	.sleep = false,
+	.shared = true,
 };
 
 static int luafifo_new(lua_State *L)
--- a/lunatik.h
+++ b/lunatik.h
@@ -76,6 +76,7 @@ typedef struct lunatik_class_s {
 	const luaL_Reg *methods;
 	void (*release)(void *);
 	bool sleep;
+	bool shared;
 	bool pointer;
 } lunatik_class_t;
 
@@ -205,7 +206,7 @@ void lunatik_cloneobject(lua_State *L, lunatik_object_t *object);
 void lunatik_releaseobject(struct kref *kref);
 int lunatik_closeobject(lua_State *L);
 int lunatik_deleteobject(lua_State *L);
-int lunatik_monitorobject(lua_State *L);
+void lunatik_monitorobject(lua_State *L, const lunatik_class_t *class);
 
 #define LUNATIK_ERR_NULLPTR	"null-pointer dereference"
 
@@ -241,6 +242,8 @@ static inline void lunatik_newclass(lua_State *L, const lunatik_class_t *class)
 	luaL_newmetatable(L, class->name); /* mt = {} */
 	luaL_setfuncs(L, class->methods, 0);
 	if (!lunatik_hasindex(L, -1)) {
+		if (class->shared)
+			lunatik_monitorobject(L, class);
 		lua_pushvalue(L, -1);  /* push mt */
 		lua_setfield(L, -2, "__index");  /* mt.__index = mt */
 	}
--- a/lunatik_obj.c
+++ b/lunatik_obj.c
@@ -1,5 +1,5 @@
 /*
-* SPDX-FileCopyrightText: (c) 2023-2024 Ring Zero Desenvolvimento de Software LTDA
+* SPDX-FileCopyrightText: (c) 2023-2026 Ring Zero Desenvolvimento de Software LTDA
 * SPDX-License-Identifier: MIT OR GPL-2.0-only
 */
 
@@ -10,6 +10,11 @@
 #include "lunatik.h"
 
 #ifdef LUNATIK_RUNTIME
+
+#define lunatik_ismetamethod(reg)          \
+	((!strncmp(reg->name, "__", 2)) ||     \
+	(reg)->func == lunatik_deleteobject || \
+	(reg)->func == lunatik_closeobject)
 
 lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class, size_t size)
 {
@@ -136,17 +141,16 @@ static int lunatik_monitor(lua_State *L)
 	return lua_gettop(L);
 }
 
-int lunatik_monitorobject(lua_State *L)
+void lunatik_monitorobject(lua_State *L, const lunatik_class_t *class)
 {
-	lua_getmetatable(L, 1);
-	lua_insert(L, 2); /* stack: object, metatable, key */
-	if (lua_rawget(L, 2) == LUA_TFUNCTION) {
-		lua_CFunction method = lua_tocfunction(L, -1);
-
-		if (likely(method != lunatik_deleteobject && method != lunatik_closeobject))
-			lua_pushcclosure(L, lunatik_monitor, 1);
+	const luaL_Reg *reg;
+	for (reg = class->methods; reg->name != NULL; reg++) {
+		if (!lunatik_ismetamethod(reg)) {
+			lua_getfield(L, -1, reg->name);
+			lua_pushcclosure(L, lunatik_monitor, 1); /* stack: mt, method */
+			lua_setfield(L, -2, reg->name);
+		}
 	}
-	return 1;
 }
 EXPORT_SYMBOL(lunatik_monitorobject);
 
--- a/lib/luacrypto_comp.c
+++ b/lib/luacrypto_comp.c
@@ -92,7 +92,6 @@ static const luaL_Reg luacrypto_comp_mt[] = {
 	{"decompress", luacrypto_comp_decompress},
 	{"__gc", lunatik_deleteobject},
 	{"__close", lunatik_closeobject},
-	{"__index", lunatik_monitorobject},
 	{NULL, NULL}
 };
 
@@ -101,6 +100,7 @@ static const lunatik_class_t luacrypto_comp_class = {
 	.methods = luacrypto_comp_mt,
 	.release = luacrypto_comp_release,
 	.sleep = true,
+	.shared = true,
 	.pointer = true,
 };
 
--- a/lib/luacrypto_rng.c
+++ b/lib/luacrypto_rng.c
@@ -134,7 +134,6 @@ static const luaL_Reg luacrypto_rng_mt[] = {
 	{"info", luacrypto_rng_info},
 	{"__gc", lunatik_deleteobject},
 	{"__close", lunatik_closeobject},
-	{"__index", lunatik_monitorobject},
 	{NULL, NULL}
 };
 
@@ -147,6 +146,7 @@ static const lunatik_class_t luacrypto_rng_class = {
 	.methods = luacrypto_rng_mt,
 	.release = luacrypto_rng_release,
 	.sleep = true,
+	.shared = true,
 	.pointer = true,
 };
 
--- a/lib/luacrypto_shash.c
+++ b/lib/luacrypto_shash.c
@@ -209,7 +209,6 @@ static const luaL_Reg luacrypto_shash_mt[] = {
 	{"import", luacrypto_shash_import},
 	{"__gc", lunatik_deleteobject},
 	{"__close", lunatik_closeobject},
-	{"__index", lunatik_monitorobject},
 	{NULL, NULL}
 };
 
@@ -223,6 +222,7 @@ static const lunatik_class_t luacrypto_shash_class = {
 	.methods = luacrypto_shash_mt,
 	.release = luacrypto_shash_release,
 	.sleep = true,
+	.shared = true,
 	.pointer = true,
 };
 
--- a/lib/luacrypto_skcipher.c
+++ b/lib/luacrypto_skcipher.c
@@ -190,7 +190,6 @@ static const luaL_Reg luacrypto_skcipher_mt[] = {
 	{"decrypt", luacrypto_skcipher_decrypt},
 	{"__gc", lunatik_deleteobject},
 	{"__close", lunatik_closeobject},
-	{"__index", lunatik_monitorobject},
 	{NULL, NULL}
 };
 
@@ -205,6 +204,7 @@ static const lunatik_class_t luacrypto_skcipher_class = {
 	.methods = luacrypto_skcipher_mt,
 	.release = luacrypto_skcipher_release,
 	.sleep = true,
+	.shared = true,
 	.pointer = true,
 };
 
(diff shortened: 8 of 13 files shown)


lunatik.h
```diff
@@ -236,11 +236,36 @@ static inline bool lunatik_hasindex(lua_State *L, int index)
...
+			lua_pop(L, 1);
+			continue;
+		}
+		if (reg->func == lunatik_deleteobject || reg->func == lunatik_closeobject) {
+			lua_pop(L, 1);
+			continue;
+		}
+
+		lua_pushcclosure(L, lunatik_monitor, 1); /* stack: class, method */
+		lua_setfield(L, -2, reg->name);
```
lneto: if I got it right, we are rewriting the actual metatable here, right? I'd indeed suggesting this instead of another table nested on the metatable. Good job ;-).

lib/luacrypto_aead.c
```diff
@@ -224,7 +224,6 @@ static const luaL_Reg luacrypto_aead_mt[] = {
 	{"decrypt", luacrypto_aead_decrypt},
 	{"__gc", lunatik_deleteobject},
 	{"__close", lunatik_closeobject},
-	{"__index", lunatik_monitorobject},
```
lneto, line 227: we will need another way to signalize where a class (actually, it would better to have it per object) need a monitor or not. Please notice, that on this patch all classes will use monitor, and we cannot do this. For instance, it probably will break RCU. I thought on having a bitfield carrying the object flags, including "shared"  ("monitored") and "sleep". What do you think?

lunatik.h
```diff
@@ -236,11 +236,36 @@ static inline bool lunatik_hasindex(lua_State *L, int index)
 	return hasindex;
 }
 
+static inline void lunatik_wrap_methods(lua_State *L, const lunatik_class_t *class)
+{
+	const luaL_Reg *reg;
+	for (reg = class->methods; reg->name != NULL; reg++) {
+		lua_getfield(L, -1, reg->name);
+
+		if (!lua_isfunction(L, -1)) {
```
lneto: is this a possible case?

lneto: btw, you can just OR all these cases..

lunatik.h
```diff
@@ -236,11 +236,36 @@ static inline bool lunatik_hasindex(lua_State *L, int index)
...
+			lua_pop(L, 1);
+			continue;
+		}
+		if (reg->name[0] == '_' && reg->name[1] == '_') {
```
lneto: I think strcmp would be clearer here, and even clearer to have a macro defined with it.. something like `lunatik_ismetamethod`, perhaps carrying the whole OR'ed cases.

lneto:
> Methods are now wrapped once in lunatik_newclass(). This reduces metatable churn and avoids per-access closure allocation in performance-sensitive paths.
> 
> Tested performance with the following script
> 
> ```lua
> local linux = require("linux")
> local data = require("data")
> 
> local d = data.new(1024)
> local iterations = 1000000
> 
> local start_ns = linux.time()
> for i = 1, iterations do
>     d:getbyte(0)
> end
> local end_ns = linux.time()
> local total_ns = end_ns - start_ns
> 
> local ns_per_call = total_ns / iterations
> 
> print("Total ns: " .. total_ns)
> print("Ns per call: " .. ns_per_call)
> ```
> 
> master: ~300 ns/call this branch: ~166 ns/call

can you also add to your tests that case of `local d_getbyte = d.getbyte`? so we can always compare the three cases? thanks!

lib/luacrypto_aead.c
sneaky-potato: Right, I agree here. Currently I am pre wrapping all the objects which do not have index in the metatable. It would be better to check if an object really needs monitoring.
Couple of questions regarding structs [lunatik_class_t](https://github.com/luainkernel/lunatik/blob/35520a5aba5a9bf48b72ac345f61a7e3af9e8861/lunatik.h#L74) and lunatik_object_t:
- Why is there `bool sleep` in both `lunatik_class_t` and `lunatik_object_t`. Is it safe to assume that `sleep` in `lunatik_class_t` is redundant?
- Lock type is unclear [here](https://github.com/luainkernel/lunatik/blob/35520a5aba5a9bf48b72ac345f61a7e3af9e8861/lunatik.h#L86) (mutex vs spinlock), for instance, what happens if a wrong lock is taken?
> I thought on having a bitfield carrying the object flags, including "shared" ("monitored") and "sleep". What do you think?

Yes, a flag will be helpful here. What fields (other than monitoring) do you think we should track?

lunatik.h
sneaky-potato: Resolved in latest push

lib/luacrypto_aead.c
lneto: > (...)
> * Why is there `bool sleep` in both `lunatik_class_t` and `lunatik_object_t`. Is it safe to assume that `sleep` in `lunatik_class_t` is redundant?

No, it isn't. Object has precedence over class. For classes, `sleep = true` means that their objects are allowed to sleep, that is, their objects can either be used on contexts that might sleep or not. Thus, classes with `sleep = true` can have objects with both `sleep = false` and `sleep = true`. However, classes with `sleep = false` can only have objects with `sleep = false`. A good example is luadata and luanetfilter, the latter cannot sleep because it's called from softirqs; the former can be used on both contexts. Please, let me know if you need more clarification on this.

> * Lock type is unclear [here](https://github.com/luainkernel/lunatik/blob/35520a5aba5a9bf48b72ac345f61a7e3af9e8861/lunatik.h#L86) (mutex vs spinlock), for instance, what happens if a wrong lock is taken?

How do you see the wrong lock being taken? It shouldn't, but perhaps we have a bug there.

> > I thought on having a bitfield carrying the object flags, including "shared" ("monitored") and "sleep". What do you think?
> 
> Yes, a flag will be helpful here. What fields (other than monitoring) do you think we should track?

I was thinking on having a RCU mode in the future where we lock only for writings, otherwise we use rcu_read_lock. Maybe some allocation flags. I didn't anticipate much on this, but I think it would be good to have some capabilities.

lib/luacrypto_aead.c
lneto, line 238: we could add `.monitor` here (perhaps `.shared` is a better name).

lunatik.h
```diff
@@ -236,11 +241,38 @@ static inline bool lunatik_hasindex(lua_State *L, int index)
 	return hasindex;
 }
 
+static int lunatik_dispatch(lua_State *L)
```
lneto: you missed `inline` here.. however, I think this should be moved to lunatik_obj.c, we don't need to expose these internals to the bindings..

lunatik.h
```diff
@@ -236,11 +241,38 @@ static inline bool lunatik_hasindex(lua_State *L, int index)
...
+{
+	lunatik_object_t *obj = lunatik_checkobject(L, 1);
+	lua_CFunction fn = lua_tocfunction(L, lua_upvalueindex(1));
+
+	/* fast path, direct function call */
```
lneto: I don't think we need comments like this and the one bellow.. you code is quite clear in those points ;-).. let's use comments as exceptions, not the norm..

lunatik.h
```diff
@@ -236,11 +241,38 @@ static inline bool lunatik_hasindex(lua_State *L, int index)
...
+
+	/* slow path */
+	return lunatik_monitor(L);
+}
+static inline void lunatik_wrap_methods(lua_State *L, const lunatik_class_t *class)
+{
+	const luaL_Reg *reg;
+	for (reg = class->methods; reg->name != NULL; reg++) {
+		lua_getfield(L, -1, reg->name);
+		if (lunatik_ismetamethod(reg)) {
```
lneto: we can avoid wrapping entirely if class is not monitored..

lneto: this could speedup even more the "fast path"

lunatik_obj.c
```diff
@@ -20,6 +20,7 @@ lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class,
 	lunatik_setobject(object, class, class->sleep);
 	lunatik_setclass(L, class);
 
+	object->monitored = true;
```
lneto: for setting this, we could change our C APIs to pass opt instead sleep, then we bundle all options we need as this argument.. for now, it can be only sleep and shared/monitored, then we can evolve.. does it make sense to you?

lunatik.h
sneaky-potato: We don't need this dispatch, I think, I will push another fixup since we dont need to wrap anything in case class has `.shared = false`

lunatik.h
sneaky-potato: Yes, I realized this. Pushing another fixup to speed up the fast path.

lib/luacrypto_aead.c
sneaky-potato: yes, indeed. I have pushed a fix which checks closure creation on class level. We need to effectively test the modules having `.shared = false` also. Perhaps luarcu could be used for that.

lunatik.h
sneaky-potato: Updated the description with latest results. Looks like we have some performance gains :D

lib/luafib.c
```diff
@@ -142,6 +142,7 @@ static const luaL_Reg luafib_lib[] = {
 
 static const lunatik_class_t luafib_class = {
 	.sleep = true,
+	.shared = true,
```
lneto: we should set it only for the classes that has monitor already

lib/luafifo.c
```diff
@@ -116,7 +116,6 @@ static const luaL_Reg luafifo_lib[] = {
 * @treturn nil
 */
 static const luaL_Reg luafifo_mt[] = {
-	{"__index", lunatik_monitorobject},
```
lneto, line 119: you missed shared for fifo, right?

sneaky-potato: Oh yeah, I think I added the shared member in luafib instead of luafifo

lunatik.h
```diff
@@ -33,6 +33,10 @@ do {						\
 
 #define lunatik_cannotsleep(L, s)	((s) && !lunatik_toruntime(L)->sleep)
 #define lunatik_getstate(runtime)	((lua_State *)runtime->private)
+#define lunatik_ismetamethod(reg)                         \
+	 (((reg)->name[0] == '_' && (reg)->name[1] == '_') || \
```
lneto: I think strncmp would be clearer here..

lunatik.h
```diff
@@ -236,11 +241,26 @@ static inline bool lunatik_hasindex(lua_State *L, int index)
 	return hasindex;
 }
 
+static inline void lunatik_wrap_methods(lua_State *L, const lunatik_class_t *class)
```
lneto: I would call this lunatik_metamethod and move it to lunatik_obj (so we export it instead of lunatik_monitor). What do you think?

sneaky-potato: Yes, that is a much cleaner abstraction, taking care of this in next push

lunatik.h
```diff
@@ -33,6 +33,10 @@ do {						\
 
 #define lunatik_cannotsleep(L, s)	((s) && !lunatik_toruntime(L)->sleep)
 #define lunatik_getstate(runtime)	((lua_State *)runtime->private)
+#define lunatik_ismetamethod(reg)           \
```
lneto: This can be moved to lunatik_obj as well, right? It's only used there.

sneaky-potato: Resolved in latest push

lunatik_obj.c
```diff
@@ -11,6 +11,11 @@
 
 #ifdef LUNATIK_RUNTIME
 
+#define lunatik_ismetamethod(reg)           \
+     ((!strncmp(reg->name, "__", 2)) ||     \
+     (reg)->func == lunatik_deleteobject || \
+     (reg)->func == lunatik_closeobject)
```
lneto: it seems to me that you are not using tabs here

lneto: can you review your patch?

lunatik.h
```diff
@@ -241,6 +242,7 @@ static inline void lunatik_newclass(lua_State *L, const lunatik_class_t *class)
 	luaL_newmetatable(L, class->name); /* mt = {} */
 	luaL_setfuncs(L, class->methods, 0);
 	if (!lunatik_hasindex(L, -1)) {
+        lunatik_metamethod(L, class);
```
lneto: here too.. it seems like spaces..

lunatik_obj.c
```diff
@@ -136,19 +141,20 @@ static int lunatik_monitor(lua_State *L)
...
-			lua_pushcclosure(L, lunatik_monitor, 1);
+	const luaL_Reg *reg;
+	for (reg = class->methods; reg->name != NULL; reg++) {
+		lua_getfield(L, -1, reg->name);
+		if (lunatik_ismetamethod(reg) || !class->shared) {
+			lua_pop(L, 1);
+			continue;
+		}
+		lua_pushcclosure(L, lunatik_monitor, 1); /* stack: mt, method */
+		lua_setfield(L, -2, reg->name);
```
lneto:
```suggestion
		if (!lunatik_ismetamethod(reg) && class->shared) {
			lua_getfield(L, -1, reg->name);
			lua_pushcclosure(L, lunatik_monitor, 1); /* stack: mt, method */
			lua_setfield(L, -2, reg->name);
		}
```
it's equivalent, right?

lunatik_obj.c
```diff
@@ -136,19 +141,20 @@ static int lunatik_monitor(lua_State *L)
 	return lua_gettop(L);
 }
 
-int lunatik_monitorobject(lua_State *L)
+void lunatik_metamethod(lua_State *L, const lunatik_class_t *class)
```
lneto: sorry.. seeing from this prisma here.. I think `lunatik_monitorobject` is a better name..

lneto (commented):
@sneaky-potato thank your for this work.. I think, after this round we can merge it.. sorry for nitpicking =)

lunatik_obj.c
sneaky-potato: It is. One more observation: if `class->shared == false` we don't need to iterate the loop at all, something like this:
```suggestion
	if (!class->shared)
		return;
	for (reg = class->methods; reg->name != NULL; reg++) {
		if (!lunatik_ismetamethod(reg)) {
			lua_getfield(L, -1, reg->name);
			lua_pushcclosure(L, lunatik_monitor, 1); /* stack: mt, method */
			lua_setfield(L, -2, reg->name);
		}
	}
```

lunatik_obj.c
sneaky-potato: Resolved in latest push

lunatik.h
sneaky-potato: Yes, fixed in latest push

lunatik_obj.c
lneto: I think it would be clear to not call this function at first..
```
if (!class->shared)
	lunatik_monitorobject(L, class);
```

again.. reading it like this.. we should have a better name for this function.. setmonitor?

sneaky-potato: > I think it would be clear to not call this function at first..

Fair point

> we should have a better name for this function.. setmonitor?

agreed, `lunatik_setmonitor` is a much clearer name. I will push an updated version.

lunatik.h
```diff
@@ -241,6 +242,7 @@ static inline void lunatik_newclass(lua_State *L, const lunatik_class_t *class)
 	luaL_newmetatable(L, class->name); /* mt = {} */
 	luaL_setfuncs(L, class->methods, 0);
 	if (!lunatik_hasindex(L, -1)) {
+		lunatik_monitorobject(L, class);
```
lneto: I've made a comment while you were updating the code.. not sure you read it.. but I think it would be clearer to move the decision to here..

```suggestion
		if (class->shared)
				lunatik_monitorobject(L, class);
```

lneto: and I rest my case.. let's keep it named monitorobject.. we can evolve from that on another PR ;-)

lneto (approved):

PR #411, opened 2026-01-24
add unregister methods to luanetfilter, luaxdp, luahid

### LINKED ISSUE #403: add unregister methods to luanetfilter, luaxdp, luaxtables, luahid
these methods can be based on stop methods present on luadevice, luanotifier, et al; using lunatik_unregisterobject().

Eshaan-byte:
Adds explicit unregister() methods following the pattern from luadevice_stop() and luanotifier_stop().

- luanetfilter: adds unregister() and refactors cleanup into helper
- luaxdp: adds unregister as alias for detach()
- luahid: adds unregister() and refactors cleanup into helper

All use lunatik_unregisterobject() for explicit cleanup.

Fixes #403

### DIFF
--- a/lib/luahid.c
+++ b/lib/luahid.c
@@ -64,6 +64,7 @@ static void luahid_release(void *private)
 }
 
 static int luahid_register(lua_State *L);
+static int luahid_unregister(lua_State *L);
 
 static const luaL_Reg luahid_lib[] = {
 	{"register", luahid_register},
@@ -73,6 +74,7 @@ static const luaL_Reg luahid_lib[] = {
 static const luaL_Reg luahid_mt[] = {
 	{"__index", lunatik_monitorobject},
 	{"__gc", lunatik_deleteobject},
+	{"unregister", luahid_unregister},
 	{NULL, NULL},
 };
 
@@ -278,6 +280,23 @@ static int luahid_raw_event(struct hid_device *hdev, struct hid_report *report, 
 
 	luahid_run(raw_event, &ctx, hid, hdev, ret);
 	return ret;
+}
+
+static int luahid_unregister(lua_State *L)
+{
+	lunatik_object_t *object = lunatik_checkobject(L, 1);
+	luahid_t *hid = (luahid_t *)object->private;
+
+	lunatik_lock(object);
+	if (hid->registered) {
+		hid_unregister_driver(&hid->driver);
+		hid->registered = false;
+	}
+	lunatik_unlock(object);
+
+	if (lunatik_toruntime(L) == hid->runtime)
+		lunatik_unregisterobject(L, object);
+	return 0;
 }
 
 /***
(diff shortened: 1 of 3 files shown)


lib/luahid.c
```diff
@@ -46,24 +46,32 @@ typedef struct luahid_ctx_s {
...
+
 static void luahid_release(void *private)
 {
 	luahid_t *hid = (luahid_t *)private;
-	if (hid->registered)
-		hid_unregister_driver(&hid->driver);
+	luahid_delete(hid);
 
 	lunatik_object_t *runtime = hid->runtime;
-	if (runtime != NULL) {
```
lneto, line 56: please, don't mix cosmetic changes on new feat and fixes.. if you want to suggest changes on our code style, you are welcome to do so, but do it as a separate PR.

lib/luanetfilter.c
```diff
@@ -203,6 +205,25 @@ static void luanetfilter_release(void *private)
 	nf->runtime = NULL;
 }
 
+static void luanetfilter_release(void *private)
```
lneto: why having this indirection?

lib/luahid.c
Eshaan-byte: You're right, sorry about that. I will revert it. I'll keep style changes separate in the future.

lib/luanetfilter.c
Eshaan-byte: The helper avoids duplicating the cleanup logic in both release() and unregister(). I followed the same pattern from luadevice_delete() which is called from both luadevice_stop() and luadevice_release().

Happy to change it if you have a different approach in mind!

lib/luahid.c
lneto: please, let me know then I can give it another review round.. thanks..

lib/luanetfilter.c
lneto: it's not the same, right? we don't simply call one in the other on luadevice.. in this case here, I think we should just call the nested function directly instead of having a indirect call.

lib/luahid.c
Eshaan-byte: Done! Reverted the cosmetic changes and kept the original `!= NULL` style. Ready for another review round.

lib/luanetfilter.c
Eshaan-byte: You're right! I removed the helper function and inlined the cleanup code directly in both release() and unregister(). Changes pushed.

lib/luahid.c
```diff
@@ -280,6 +282,23 @@ static int luahid_raw_event(struct hid_device *hdev, struct hid_report *report,
 	return ret;
 }
 
+static int luahid_unregister(lua_State *L)
+{
+	lunatik_object_t *object = lunatik_checkobject(L, 1);
+	luahid_t *hid = (luahid_t *)object->private;
+
+	lunatik_lock(object);
```
lneto, line 290: Did you test it? It looks like a deadlock.

lneto:
Hi @Eshaan-byte, are you still working on this? Please, let me know, otherwise we should close this PR.

PR #417, opened 2026-01-29
Retrieve vlan tag id from skb, if present

carloslack:
Added a way of retrieving the vlan ID from `skb` 
Tried to make it support non-linear skb and it is idiomatic for kernel 6.x ATM

However I wasn't able to fully test in a vlan network (working on that) but I think it is safe to PR this and get some reviews in the meantime.

### DIFF
--- a/lib/luanetfilter.c
+++ b/lib/luanetfilter.c
@@ -90,7 +90,13 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
 	else
 		lua_pushnil(L); /* dev may be NULL if hook is LOCAL_OUT */
 
-	if (lua_pcall(L, 2, 2, 0) != LUA_OK) {
+	int narg = 2;
+	if (skb_vlan_tag_present(skb)) {
+		lua_pushinteger(L, skb_vlan_tag_get_id(skb));
+		narg++;
+	}
+
+	if (lua_pcall(L, narg, 2, 0) != LUA_OK) {
 		pr_err("%s\n", lua_tostring(L, -1));
 		lua_pop(L, 1);
 		goto clear;


lib/luanetfilter.c
```diff
@@ -79,7 +79,8 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
 	if (!luanetfilter_pushcb(L, luanf) || (data = luanetfilter_pushskb(L, luanf, skb)) == NULL)
 		goto out;
 
-	if (skb_mac_header_was_set(skb))
+	bool mac_header_was_set = skb_mac_header_was_set(skb);
```
lneto: we try to not have very long names..
```suggestion
	bool has_mac_header = skb_mac_header_was_set(skb);
```
couldn't figure anything better =)

lib/luanetfilter.c
```diff
@@ -90,7 +91,21 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
 	else
 		lua_pushnil(L); /* dev may be NULL if hook is LOCAL_OUT */
 
-	if (lua_pcall(L, 2, 2, 0) != LUA_OK) {
+	/* work out vlan TAG id, if present */
```
lneto: I would move the block below to an aux.. e.g., luanetfilter_pushvlanid

lib/luanetfilter.c
```diff
@@ -90,7 +91,21 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
...
+	if (skb_vlan_tag_present(skb)) {
+		lua_pushinteger(L, skb_vlan_tag_get_id(skb));
+	} else if (mac_header_was_set && (skb->protocol == htons(ETH_P_8021Q) || skb->protocol == htons(ETH_P_8021AD))) {
```
lneto: we use `\n` after `}`, on if-then-else

lneto: btw, if tag is not present, wouldn't be safe to give up? why should we extract it by ourselves? if needed, we can do this logic directly in lua, right?

lib/luanetfilter.c
```diff
@@ -90,7 +91,21 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
...
+			struct vlan_ethhdr _veh;
+			struct vlan_ethhdr *veh = skb_header_pointer(skb, skb_mac_offset(skb), sizeof(_veh), &_veh);
+			if (veh)
+				lua_pushinteger(L, ntohs(veh->h_vlan_TCI) & VLAN_VID_MASK);
+			else
+				lua_pushnil(L);
```
lneto: we don't need to pass nil as the third arg.. we can pass none.. it's both more luaish and clearer, I think.. for this, we might use a `narg` var and increment it in the cases we push it..

lib/luanetfilter.c
carloslack: That makes it more simple, I think it should be possible to retrieve from lua, similar to what we had in lldp example, for mac address.
I will push a change

lib/luanetfilter.c
carloslack: If I ignore if vlan is not present, like you suggested below, then the change is very small, probably won't need a helper

lib/luanetfilter.c
carloslack: the struggle :D 
again, won't be needed (see below)

lib/luanetfilter.c
```diff
@@ -90,7 +90,13 @@ static int luanetfilter_hook_cb(lua_State *L, luanetfilter_t *luanf, struct sk_b
 	else
 		lua_pushnil(L); /* dev may be NULL if hook is LOCAL_OUT */
 
-	if (lua_pcall(L, 2, 2, 0) != LUA_OK) {
+	int narg = 2;
+	if (skb_vlan_tag_present(skb)) {
+		lua_pushinteger(L, skb_vlan_tag_get_id(skb));
+		narg++;
+	}
```
lneto, line 97: We should use the same approach for dev, for consistency ;-)

carloslack: You mean, dropping `narg` and pushing `nil` if vlan not available?

lneto: I meant on `if (de)` do `narg++` as well and suppress pushnil otherwise

lneto: It's the same pattern ;-)

lneto: We could even have an inline for that..

carloslack: But if we also make `dev` an optional arg, I think we'll have to push something so don't mess with expected arguments, on lua side?

lneto: I think you're right, we can have VLAN tag present, but not dev, right? If so, I agree, it might be confusing. I'd leave at this indeed. However, I think we should plan to implement a luaskb binding, so we can move most of this logic to there.. I put some thoughts on this, we can just have a method there to create a luadata, something like `skb:todata()`. I think it should be straight forward to implement and a good way to get more familiarized with lunatik modules.. =)

lneto:
@sheharyaar it would be good to have your review before merging it..

PR #429, opened 2026-02-05
Add basic luachecksum

carloslack:
Added a very basic `checksum` abstraction.

Currently only with a generic data checksum.
When we agree on the basics we can also discuss what other wrappers we may need.


carloslack:
Nevermind this PR, will open a new one on luadata :)

PR #430, opened 2026-02-06
Added lunatik raw checksum op

carloslack:
Example:

```
local str = "ABCD"
local buf = data.new(#str)
buf:setstring(0, "ABCD")
print (buf:checksum())
```

### DIFF
--- a/examples/echod/worker.lua
+++ b/examples/echod/worker.lua
@@ -1,11 +1,9 @@
 --
--- SPDX-FileCopyrightText: (c) 2023-2024 Ring Zero Desenvolvimento de Software LTDA
+-- SPDX-FileCopyrightText: (c) 2023-2026 Ring Zero Desenvolvimento de Software LTDA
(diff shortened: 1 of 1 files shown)


lib/luadata.c
```diff
@@ -161,6 +161,26 @@ static void luadata_skb_resize(lua_State *L, luadata_t *data, size_t new_size)
...
+* @raise Error if the write operation (offset + length of data).
+*/
+static int luadata_checksum(lua_State *L)
+{
+	luadata_t *data = luadata_check(L, 1);
+	lua_Integer offset = luaL_optinteger(L, 2, 0);
+	lua_Integer length = luaL_optinteger(L, 3, data->size - offset);
+	void *value = (void*)luadata_checkbounds(L, 2, data, offset, length);
+
+	/** checksum a single buffer */
```
lneto: LGTM; quite nitpicking, but I think we don't need this comment here..

Issue #420, opened 2026-01-30
sk_buff: Move `skb` ops to dedicated module


carloslack:
Also, there is the possibility of doing this in lua on lunatik.
Measuring performance in both cases can also be a good idea, for hotpaths.

PR #432, opened 2026-02-11
Add skb module

carloslack:
This is about #420

Added a very small skb module, so we can improve upon.
I was also thinking maybe about having __index metatable? So we could from lua just `skb.len` for example.

Let's also discuss what methods we want .

### DIFF
--- a/lib/luaskb.c
+++ b/lib/luaskb.c
@@ -22,6 +22,29 @@ static int luaskb_len(lua_State *L)
 	return 1;
 }
 
+static int luaskb_index(lua_State *L)
+{
+	struct sk_buff *skb = luaskb_check(L, 1);
+	const char *key = luaL_checkstring(L, 2);
+
+	if (!strcmp(key, "len")) {
+		lua_pushinteger(L, skb->len);
+		return 1;
+	}
+
+	if (!strcmp(key, "hdr_len")) {
+		lua_pushinteger(L, skb->hdr_len);
+		return 1;
+	}
+
+	if (!strcmp(key, "truesize")) {
+		lua_pushinteger(L, skb->truesize);
+		return 1;
+	}
+
+	return 0;
+}
+
 static void luaskb_release(void *private)
 {
 	struct sk_buff *skb = (struct sk_buff*)private;
@@ -35,6 +58,7 @@ static const luaL_Reg luaskb_lib[] = {
 
 static const luaL_Reg luaskb_mt[] = {
 	{"__gc", lunatik_deleteobject},
+	{"__index", luaskb_index},
 	{NULL, NULL}
 };
 


lib/luaskb.c
```diff
@@ -0,0 +1,45 @@
...
+#include <linux/kernel.h>
+#include <linux/module.h>
+#include <linux/skbuff.h>
+
+#include <lua.h>
+#include <lualib.h>
+#include <lunatik.h>
+
+static int luaskb_len(lua_State *L) {
+	struct sk_buff *skb = (struct sk_buff *)lua_touserdata(L, 1);
```
lneto: Whe should use lunatik_checkobject here

lib/luaskb.c
```diff
@@ -0,0 +1,45 @@
...
+* SPDX-FileCopyrightText: (c) 2023-2026 Ring Zero Desenvolvimento de Software LTDA
+* SPDX-License-Identifier: MIT OR GPL-2.0-only
+*/
+
+#include <linux/kernel.h>
+#include <linux/module.h>
+#include <linux/skbuff.h>
+
+#include <lua.h>
+#include <lualib.h>
```
lneto: now we don't need them anymore because it's included on lunatik.h.. I think we should update the code base accordingly, especially luaskell

lib/luaskb.c
```diff
@@ -0,0 +1,45 @@
...
+#include <lunatik.h>
+
+static int luaskb_len(lua_State *L) {
```
lneto:
```suggestion
static int luaskb_len(lua_State *L)
{
```

Makefile
```diff
@@ -30,7 +30,8 @@ CONFIG_LUNATIK_RUNTIME ?= y
 CONFIG_LUNATIK_RUN ?= m
 
 LUNATIK_MODULES := DEVICE LINUX NOTIFIER SOCKET RCU THREAD FIB DATA PROBE SYSCALL XDP FIFO XTABLE NETFILTER \
-	COMPLETION CRYPTO_SHASH CRYPTO_SKCIPHER CRYPTO_AEAD CRYPTO_RNG CRYPTO_COMP CPU HID SIGNAL BYTEORDER
+	COMPLETION CRYPTO_SHASH CRYPTO_SKCIPHER CRYPTO_AEAD CRYPTO_RNG CRYPTO_COMP CPU HID SIGNAL BYTEORDER \
+	SKB
```
lneto:
```suggestion
	COMPLETION CRYPTO_SHASH CRYPTO_SKCIPHER CRYPTO_AEAD CRYPTO_RNG CRYPTO_COMP CPU HID SIGNAL BYTEORDER SKB \
```

lib/luaskb.c
carloslack: alright, lemme check that

lib/luaskb.c
```diff
@@ -0,0 +1,80 @@
...
+	.methods = luaskb_mt,
+	.release = luaskb_release,
+	.pointer = true,
+};
+
+LUNATIK_NEWLIB(skb, luaskb_lib, &luaskb_class, NULL);
+
+lunatik_object_t *luaskb_create(struct sk_buff *skb)
+{
+	lunatik_object_t *object = lunatik_createobject(&luaskb_class, sizeof(luaskb_t), false);
```
lneto:
```suggestion
	lunatik_object_t *object = lunatik_createobject(&luaskb_class, 0, false);
```

lib/luaskb.c
```diff
@@ -0,0 +1,80 @@
...
+
+	if (object != NULL) {
+		luaskb_t *data = (luaskb_t *)object->private;
```
lneto:
```suggestion
		struct sk_buff *skb = (struct sk_buff *skb)object->private;
```

lib/luaskb.c
```diff
@@ -0,0 +1,85 @@
...
+static const lunatik_class_t luaskb_class = {
+	.name = "skb",
+	.methods = luaskb_mt,
+	.release = luaskb_release,
+	.pointer = true,
+};
+
+int luaskb_reset(lunatik_object_t *object, void *ptr)
+{
+	lunatik_lock(object);
```
lneto, line 75: why are you locking it by hand instead of using shared = true?

carloslack: Ah thanks, good point, with shared we end up in `lunatik_monitor`, nice!

lib/luaskb.c
```diff
@@ -0,0 +1,110 @@
...
+static int luaskb_len(lua_State *L)
+{
+	struct sk_buff *skb = luaskb_check(L, 1);
+	pr_info_ratelimited("%s:%d: skb=%p %d\n", __func__, __LINE__, skb, skb->len);
+
+	lua_pushinteger(L, skb->len);
+	return 1;
+}
+
+static int luaskb_index(lua_State *L)
```
carloslack, line 25: @lneto , I added this simple index access (fields only), I am wondering if this is a more simple way of exposing these fields to lua. I kept `luaskb_len` for now while we decide.
wdyt?

lneto: I think the simplest way to do it is breaking down the implementation between C and Lua, take a look at the socket library.. we have a lower level lib in C (that is, luasocket) and higher level ones in Lua (that is, inet, raw, etc). We could have the same approach here, where the C binding will provide getters and the Lua library will provide and abstract table interface. In that we way, we can leverage Lua [meta]tables to handle this indexing for us instead of abusing of strcmp'ing ;-). For instance,

```Lua
local skb_in = require("skb.inner") -- couldn't figure a better name, sorry =/

function myhook(skb, ...)
	local skb = skb_in(skb)
	print(skb.len, skb.hdr_len, skb.truesize)
end
```

IMHO, it would be cleaner/easier to create this `skb_in` object, defining it's __index to automatically call the `skb` getter.. just concat'ing a the `key` string and calling it using the underlying `skb` method.. also, we can use `__call` to implement this sort of "typecast" `skb = skb_in(skb)` ;-).

However, I think all those fields you are concerned right now are not the prio for this module. Do you already have a use case for them on a script? Keep in mind that Lua bindings shouldn't be 1:1 in relation to C.

Regarding `len`, I still think it would be better to just define `__len`, so we can simply use `#skb` in the scripts.

lib/luaskb.c
```diff
@@ -0,0 +1,110 @@
...
+	{"__index", luaskb_index},
+	{NULL, NULL}
+};
+
+static const lunatik_class_t luaskb_class = {
+	.name = "skb",
+	.methods = luaskb_mt,
+	.release = luaskb_release,
+	.pointer = true,
+	.shared = true,
```
lneto, line 70: you shouldn't define it as true if you are defining your own __index, it should be overridden if so.. also, I think the idea is not to share skbs, right?

lib/luaskb.c
lneto: update: we have discussed it offline and it shouldn't use monitor..  and this locking is intended for the C side.

lib/luaskb.c
carloslack: Agreed, I think your suggestion makes a lot of sense, with `inet.lua` being a nice example. we can indeed approach it similar way, lemme work on that.
Regarding the actual methods, I am more concentrated in defining the approach than minding the actual methods right now, but will come right after. I think we may have some examples using `skb` buffers and data, but I'll check in more detail. I can also write a more complete useful example along with the changes.

what do you think?

lneto: I think it's the way to go ;-).. however, I would focus on creating methods for the things we are already using first.. mainly the support you have added for VLAN and interface index. Thus, instead of pushing it as callback arguments, we could retrieve it using skb methods.. e.g., `skb:gettag(id)`.

PR #433, opened 2026-02-13
Removed unneeded includes

carloslack:
lua.h is already included from lunatik.h

### DIFF
--- a/lib/luasocket.c
+++ b/lib/luasocket.c
@@ -22,7 +22,6 @@
 * @see socket.inet
 */
 #define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
-#include <linux/module.h>
 #include <linux/version.h>
 #include <linux/string.h>
 #include <linux/net.h>
@@ -31,10 +30,6 @@
 #if (LINUX_VERSION_CODE <= KERNEL_VERSION(6, 1, 0))
 #include <linux/l2tp.h>
 #endif
-
-#include <lua.h>
-#include <lualib.h>
-#include <lauxlib.h>
 
 #include <lunatik.h>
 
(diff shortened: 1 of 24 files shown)


carloslack:
> I think it's a good opportunity to remove lualib and lauxlib altogether ;-), also linux/{kernel, module, version}.. what do you think?

Agreed, I've made the changes, have a look

lib/luabyteorder.c
```diff
@@ -3,10 +3,8 @@
 * SPDX-License-Identifier: MIT OR GPL-2.0-only
 */
 
-#include <linux/module.h>
-#include <linux/byteorder/generic.h>
-
 #include <lunatik.h>
+#include <linux/byteorder/generic.h>
```
lneto, line 7: we shouldn't invert this.. our "non-written" rule is that kernel headers come first ;-)

lib/luasocket.c
```diff
@@ -22,21 +22,15 @@
 * @see socket.inet
 */
 #define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
-#include <linux/module.h>
-#include <linux/version.h>
 #include <linux/string.h>
 #include <linux/net.h>
 #include <linux/un.h>
 #include <net/sock.h>
+#include <lunatik.h>
```
lneto: also here.. we should move lunatik.h to below.. after the \n

lneto (approved):
btw, can you also update the copyright year of the files you changed that aren't updated yet? thanks!

lneto (changes requested):
sorry.. selected "approved" wrongly.. actually, I'm requesting few changes ;-)

PR #434, opened 2026-02-13
add support for non-shared objects in shared class

### LINKED ISSUE #431: add support for "no shared" objects
from matrix discussion:

it's basically to have that shared flag per object and allow the creation of objects with shared = false, when class is shared = true
I think we would need to duplicate metatables for shared classes, one monitored and other not..
when obj->shared == false, we must fail to clone and avoid init lock structures

sneaky-potato:
Introduces a shared flag to `lunatik_object_t` to allow creating non shared objects inside shared classes. This bypasses monitoring for objects guaranteed to be local to a single Lua state.

Results on 1M iterations of `getdata` (script used in #410)

Baseline: **master** branch

| configuration | Average time per call |
| --- | --- |
| No changes (baseline) | 125 ns |
| lunatik_monitorobject call commented out | 32 ns | 

This PR

| lunatik_newobject call mode | Average time per call |
| --- | --- |
| shared = true | 98 ns |
| shared = false | 28 ns |

Fixes #431

### DIFF
--- a/doc/capi.md
+++ b/doc/capi.md
@@ -113,6 +113,32 @@ static ssize_t mydevice_read(struct file *f, char *buf, size_t len, loff_t *off)
 }
 ```
 
+## lunatik\_newobject
+```C
+lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class, size_t size, bool shared);
+```
+_lunatik\_newobject()_ allocates a new Lunatik object and pushes a userdata
+containing a pointer to the object onto the Lua stack.
+If `shared` is _true_, the object will use the monitored metatable for safe
+access Lunatik runtimes. This requires `class->shared = true;` otherwise, it raises a
+Lua error.
+- If `class->sleep` is _true_, it uses a mutex and `GFP_KERNEL`
+- If `class->sleep` is _false_, it uses a spinlock and `GFP_ATOMIC`
+
+It allocates size bytes for the object's private data, unless `class->pointer` is true.
+
+## lunatik\_createobject
+```C
+lunatik_object_t *lunatik_createobject(const lunatik_class_t *class, size_t size, bool sleep, bool shared);
+```
+_lunatik\_createobject()_ creates a Lunatik object independently of any Lua
+state. This is intended for objects created in C context that may be shared
+with Lua runtimes later.
+
+It allocates memory with `GFP_KERNEL` if `sleep` is _true_, or `GFP_ATOMIC` otherwise.
+It returns a pointer to the `lunatik_object_t` on success, or _NULL_ if memory
+allocation fails and _ERR_PTR(-EINVAL)_ if the class is not sharable but API is called with _true_ as `shared`
+
 ## lunatik\_getobject
 ```C
 void lunatik_getobject(lunatik_object_t *object);
--- a/lib/luadata.c
+++ b/lib/luadata.c
@@ -127,7 +127,7 @@ static int luadata_setstring(lua_State *L)
 	lua_Integer offset = luaL_checkinteger(L, 2);
 	const char *str = luaL_checklstring(L, 3, &length);
 	void *ptr = luadata_checkbounds(L, 2, data, offset, length);
-	
+
 	luadata_checkwritable(L, data);
 	memcpy(ptr, str, length);
 	return 0;
@@ -176,7 +176,7 @@ static int luadata_checksum(lua_State *L)
 
 /***
 * Resizes the memory block represented by the data object.
-* If the object is a network packet (SKB), it uses skb_put() to expand 
+* If the object is a network packet (SKB), it uses skb_put() to expand
 * or skb_trim() to shrink the buffer. For raw buffers, it updates the size.
 * @function resize
 * @tparam integer new_size The desired size of the memory block in bytes.
@@ -190,7 +190,7 @@ static int luadata_resize(lua_State *L)
 	luadata_checkwritable(L, data);
 
 	if (data->opt & LUADATA_OPT_SKB)
-		luadata_skb_resize(L, data, new_size); 
+		luadata_skb_resize(L, data, new_size);
 	else if (data->opt & LUADATA_OPT_FREE)
 		data->ptr = lunatik_checknull(L, lunatik_realloc(L, data->ptr, new_size));
 	else
@@ -436,7 +436,7 @@ static inline void luadata_set(luadata_t *data, void *ptr, ptrdiff_t offset, siz
 static int luadata_lnew(lua_State *L)
 {
 	size_t size = (size_t)luaL_checkinteger(L, 1);
-	lunatik_object_t *object = lunatik_newobject(L, &luadata_class, sizeof(luadata_t));
+	lunatik_object_t *object = lunatik_newobject(L, &luadata_class, sizeof(luadata_t), true);
 	luadata_t *data = (luadata_t *)object->private;
 
 	luadata_set(data, lunatik_checkalloc(L, size), 0, size, LUADATA_OPT_FREE);
@@ -445,20 +445,22 @@ static int luadata_lnew(lua_State *L)
 
 LUNATIK_NEWLIB(data, luadata_lib, &luadata_class, NULL);
 
-static inline lunatik_object_t *luadata_create(void *ptr, size_t size, bool sleep, uint8_t opt)
+static inline lunatik_object_t *luadata_create(void *ptr, size_t size, bool sleep, uint8_t opt, bool shared)
 {
-	lunatik_object_t *object = lunatik_createobject(&luadata_class, sizeof(luadata_t), sleep);
+	lunatik_object_t *object = lunatik_createobject(&luadata_class, sizeof(luadata_t), sleep, shared);
 
-	if (object != NULL) {
+	if (!IS_ERR_OR_NULL(object)) {
 		luadata_t *data = (luadata_t *)object->private;
 		luadata_set(data, ptr, 0, size, opt);
 	}
 	return object;
 }
 
-lunatik_object_t *luadata_new(lua_State *L)
+lunatik_object_t *luadata_new(lua_State *L, bool shared)
 {
-	lunatik_object_t *data = lunatik_checknull(L, luadata_create(NULL, 0, false, LUADATA_OPT_NONE));
+	lunatik_object_t *data = lunatik_checknull(L, luadata_create(NULL, 0, false, LUADATA_OPT_NONE, shared));
+	if (IS_ERR(data))
+		luaL_error(L, LUNATIK_ERR_SHARED, luadata_class.name);
 	lunatik_cloneobject(L, data);
 	return data;
 }
--- a/lib/luadevice.c
+++ b/lib/luadevice.c
@@ -355,7 +355,7 @@ static int luadevice_new(lua_State *L)
 	lunatik_checkfield(L, 1, "name", LUA_TSTRING);
 	name = lua_tostring(L, -1);
 
-	object = lunatik_newobject(L, &luadevice_class, sizeof(luadevice_t));
+	object = lunatik_newobject(L, &luadevice_class, sizeof(luadevice_t), false);
 	luadev = (luadevice_t *)object->private;
 
 	memset(luadev, 0, sizeof(luadevice_t));
--- a/lib/luafifo.c
+++ b/lib/luafifo.c
@@ -129,7 +129,7 @@ static const lunatik_class_t luafifo_class = {
 static int luafifo_new(lua_State *L)
 {
 	size_t size = luaL_checkinteger(L, 1);
-	lunatik_object_t *object = lunatik_newobject(L, &luafifo_class, sizeof(struct kfifo));
+	lunatik_object_t *object = lunatik_newobject(L, &luafifo_class, sizeof(struct kfifo), true);
 	gfp_t gfp = lunatik_gfp(lunatik_toruntime(L));
 	int ret;
 
--- a/lib/luanetfilter.c
+++ b/lib/luanetfilter.c
@@ -163,7 +163,7 @@ static const lunatik_class_t luanetfilter_class = {
 static int luanetfilter_register(lua_State *L)
 {
 	luaL_checktype(L, 1, LUA_TTABLE);
-	lunatik_object_t *object = lunatik_newobject(L, &luanetfilter_class , sizeof(luanetfilter_t));
+	lunatik_object_t *object = lunatik_newobject(L, &luanetfilter_class , sizeof(luanetfilter_t), false);
 	luanetfilter_t *nf = (luanetfilter_t *)object->private;
 	nf->runtime = NULL;
 
--- a/lib/luanotifier.c
+++ b/lib/luanotifier.c
@@ -408,7 +408,7 @@ static int luanotifier_new(lua_State *L, luanotifier_register_t register_fn, lua
 
 	luaL_checktype(L, 1, LUA_TFUNCTION); /* callback */
 
-	object = lunatik_newobject(L, &luanotifier_class, sizeof(luanotifier_t));
+	object = lunatik_newobject(L, &luanotifier_class, sizeof(luanotifier_t), false);
 	notifier = (luanotifier_t *)object->private;
 
 	lunatik_setruntime(L, notifier, notifier);
--- a/lib/luaprobe.c
+++ b/lib/luaprobe.c
@@ -236,7 +236,7 @@ static const lunatik_class_t luaprobe_class = {
 
 static int luaprobe_new(lua_State *L)
 {
-	lunatik_object_t *object = lunatik_newobject(L, &luaprobe_class, sizeof(luaprobe_t));
+	lunatik_object_t *object = lunatik_newobject(L, &luaprobe_class, sizeof(luaprobe_t), false);
 	luaprobe_t *probe = (luaprobe_t *)object->private;
 	struct kprobe *kp = &probe->kp;
 	int ret;
--- a/lib/luarcu.c
+++ b/lib/luarcu.c
@@ -372,6 +372,7 @@ static const lunatik_class_t luarcu_class = {
 	.name = "rcu",
 	.methods = luarcu_mt,
 	.release = luarcu_release,
+	.shared = true,
 };
 
 lunatik_object_t *luarcu_newtable(size_t size, bool sleep)
@@ -379,7 +380,7 @@ lunatik_object_t *luarcu_newtable(size_t size, bool sleep)
 	lunatik_object_t *object;
 
 	size = roundup_pow_of_two(size);
-	if ((object = lunatik_createobject(&luarcu_class, luarcu_sizeoftable(size), sleep))!= NULL)
+	if ((object = lunatik_createobject(&luarcu_class, luarcu_sizeoftable(size), sleep, false)) != NULL)
 		luarcu_inittable((luarcu_table_t *)object->private, size);
 	return object;
 }
@@ -419,7 +420,7 @@ EXPORT_SYMBOL(luarcu_newtable);
 static int luarcu_table(lua_State *L)
 {
 	size_t size = roundup_pow_of_two(luaL_optinteger(L, 1, LUARCU_DEFAULT_SIZE));
-	lunatik_object_t *object = lunatik_newobject(L, &luarcu_class, luarcu_sizeoftable(size));
+	lunatik_object_t *object = lunatik_newobject(L, &luarcu_class, luarcu_sizeoftable(size), false);
 
 	luarcu_inittable((luarcu_table_t *)object->private, size);
 	return 1; /* object */
--- a/lib/luaxtable.c
+++ b/lib/luaxtable.c
@@ -315,7 +315,7 @@ static const lunatik_class_t luaxtable_class = {
 static inline lunatik_object_t *luaxtable_new(lua_State *L, int idx, int hook)
 {
 	luaL_checktype(L, idx, LUA_TTABLE);
-	lunatik_object_t *object = lunatik_newobject(L, &luaxtable_class , sizeof(luaxtable_t));
+	lunatik_object_t *object = lunatik_newobject(L, &luaxtable_class , sizeof(luaxtable_t), false);
 	luaxtable_t *xtable = (luaxtable_t *)object->private;
 
 	xtable->type = hook;
--- a/lunatik.h
+++ b/lunatik.h
@@ -91,13 +91,14 @@ typedef struct lunatik_object_s {
 	};
 	bool sleep;
 	gfp_t gfp;
+	bool shared;
 } lunatik_object_t;
 
 extern lunatik_object_t *lunatik_env;
 
 static inline int lunatik_trylock(lunatik_object_t *object)
 {
-	return object->sleep ? mutex_trylock(&object->mutex) : spin_trylock(&object->spin);
+	return unlikely(object->shared) ? (object->sleep ? mutex_trylock(&object->mutex) : spin_trylock(&object->spin)) : 1;
 }
 
 int lunatik_runtime(lunatik_object_t **pruntime, const char *script, bool sleep);
@@ -181,27 +182,29 @@ static inline void lunatik_checkclass(lua_State *L, const lunatik_class_t *class
 		luaL_error(L, "cannot use '%s' class on non-sleepable runtime", class->name);
 }
 
-static inline void lunatik_setclass(lua_State *L, const lunatik_class_t *class)
+static inline void lunatik_setclass(lua_State *L, const lunatik_class_t *class, bool shared)
 {
-	if (luaL_getmetatable(L, class->name) == LUA_TNIL)
-		luaL_error(L, "metatable not found (%s)", class->name);
+	lua_pushfstring(L, shared ? "_%s" : "%s", class->name);
+	if (lua_rawget(L, LUA_REGISTRYINDEX) == LUA_TNIL)
+		luaL_error(L, "metatable not found (%s)", lua_tostring(L, -1));
 	lua_setmetatable(L, -2);
 	lua_pushlightuserdata(L, (void *)class);
 	lua_setiuservalue(L, -2, 1); /* pop class */
 }
 
-static inline void lunatik_setobject(lunatik_object_t *object, const lunatik_class_t *class, bool sleep)
+static inline void lunatik_setobject(lunatik_object_t *object, const lunatik_class_t *class, bool sleep, bool shared)
 {
 	kref_init(&object->kref);
 	object->private = NULL;
 	object->class = class;
 	object->sleep = sleep;
+	object->shared = shared;
 	object->gfp = sleep ? GFP_KERNEL : GFP_ATOMIC;
 	lunatik_newlock(object);
 }
 
-lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class, size_t size);
-lunatik_object_t *lunatik_createobject(const lunatik_class_t *class, size_t size, bool sleep);
+lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class, size_t size, bool shared);
+lunatik_object_t *lunatik_createobject(const lunatik_class_t *class, size_t size, bool sleep, bool shared);
 lunatik_object_t **lunatik_checkpobject(lua_State *L, int ix);
 void lunatik_cloneobject(lua_State *L, lunatik_object_t *object);
 void lunatik_releaseobject(struct kref *kref);
@@ -210,6 +213,7 @@ int lunatik_deleteobject(lua_State *L);
 void lunatik_monitorobject(lua_State *L, const lunatik_class_t *class);
 
 #define LUNATIK_ERR_NULLPTR	"null-pointer dereference"
+#define LUNATIK_ERR_SHARED 	"cannot create shared object from non-shared class '%s'"
 
 #define lunatik_newpobject(L, n)	(lunatik_object_t **)lua_newuserdatauv((L), sizeof(lunatik_object_t *), (n))
 #define lunatik_argchecknull(L, o, i)	luaL_argcheck((L), (o) != NULL, (i), LUNATIK_ERR_NULLPTR)
@@ -238,17 +242,18 @@ static inline bool lunatik_hasindex(lua_State *L, int index)
 	return hasindex;
 }
 
-static inline void lunatik_newclass(lua_State *L, const lunatik_class_t *class)
+static inline void lunatik_newclass(lua_State *L, const lunatik_class_t *class, bool monitored)
 {
-	luaL_newmetatable(L, class->name); /* mt = {} */
+	lua_pushfstring(L, monitored ? "_%s" : "%s", class->name);
+	luaL_newmetatable(L, lua_tostring(L, -1)); /* mt = {} */
 	luaL_setfuncs(L, class->methods, 0);
+	if (monitored)
+		lunatik_monitorobject(L, class);
 	if (!lunatik_hasindex(L, -1)) {
-		if (class->shared)
-			lunatik_monitorobject(L, class);
 		lua_pushvalue(L, -1);  /* push mt */
 		lua_setfield(L, -2, "__index");  /* mt.__index = mt */
 	}
-	lua_pop(L, 1);  /* pop mt */
+	lua_pop(L, 2);  /* pop mt, class name */
 }
 
 static inline lunatik_class_t *lunatik_getclass(lua_State *L, int ix)
@@ -260,8 +265,6 @@ static inline lunatik_class_t *lunatik_getclass(lua_State *L, int ix)
 	}
 	return NULL;
 }
-
-#define lunatik_isobject(L, ix)	(lunatik_getclass((L), (ix)) != NULL)
 
 static inline lunatik_object_t *lunatik_testobject(lua_State *L, int ix)
 {
@@ -292,7 +295,9 @@ int luaopen_##libname(lua_State *L)						\
 	luaL_newlib(L, funcs);							\
 	if (cls) {								\
 		lunatik_checkclass(L, cls);					\
-		lunatik_newclass(L, cls);					\
+		if (cls->shared)							\
+			lunatik_newclass(L, cls, true);			\
+		lunatik_newclass(L, cls, false);			\
 	}									\
 	if (nss)								\
 		lunatik_newnamespaces(L, nss);					\
--- a/lunatik_obj.c
+++ b/lunatik_obj.c
@@ -16,14 +16,18 @@
 	(reg)->func == lunatik_deleteobject || \
 	(reg)->func == lunatik_closeobject)
 
-lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class, size_t size)
+#define lunatik_issharable(class, shared) 	(!(shared) || ((class)->shared))
+
+lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class, size_t size, bool shared)
 {
 	lunatik_object_t **pobject = lunatik_newpobject(L, 1);
 	lunatik_object_t *object = lunatik_checkalloc(L, sizeof(lunatik_object_t));
 
 	lunatik_checkclass(L, class);
-	lunatik_setobject(object, class, class->sleep);
-	lunatik_setclass(L, class);
+	if (!lunatik_issharable(class, shared))
+		luaL_error(L, LUNATIK_ERR_SHARED, class->name);
+	lunatik_setobject(object, class, class->sleep, shared);
+	lunatik_setclass(L, class, shared);
 
 	object->private = class->pointer ? NULL : lunatik_checkalloc(L, size);
 
@@ -32,15 +36,19 @@ lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class, 
 }
 EXPORT_SYMBOL(lunatik_newobject);
 
-lunatik_object_t *lunatik_createobject(const lunatik_class_t *class, size_t size, bool sleep)
+lunatik_object_t *lunatik_createobject(const lunatik_class_t *class, size_t size, bool sleep, bool shared)
 {
 	gfp_t gfp = sleep ? GFP_KERNEL : GFP_ATOMIC;
 	lunatik_object_t *object = (lunatik_object_t *)kmalloc(sizeof(lunatik_object_t), gfp);
 
+	if (!lunatik_issharable(class, shared)) {
+		pr_err(LUNATIK_ERR_SHARED, class->name);
+		return ERR_PTR(-EINVAL);
+	}
 	if (object == NULL)
 		return NULL;
 
-	lunatik_setobject(object, class, sleep);
+	lunatik_setobject(object, class, sleep, shared);
 	if ((object->private = kmalloc(size, gfp)) == NULL) {
 		lunatik_putobject(object);
 		return NULL;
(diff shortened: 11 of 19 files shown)


lunatik.h
```diff
@@ -248,6 +269,18 @@ static inline void lunatik_newclass(lua_State *L, const lunatik_class_t *class)
 		lua_setfield(L, -2, "__index");  /* mt.__index = mt */
 	}
 	lua_pop(L, 1);  /* pop mt */
+
+	if (class->shared) {
+		char buf[64];
```
carloslack: Can it be larger than 64? Can it truncate? Probably not, but anyway, should it be dynamic? 
Also, maybe it is better a constant here

lunatik.h
```diff
@@ -217,6 +234,10 @@ void lunatik_monitorobject(lua_State *L, const lunatik_class_t *class);
 #define lunatik_getobject(o)		kref_get(&(o)->kref)
 #define lunatik_putobject(o)		kref_put(&(o)->kref, lunatik_releaseobject)
 
+/* Convenience wrappers: create object with the sharing policy defined by the class */
+#define lunatik_newsharedobject(L, class, size)		lunatik_newobject((L), (class), (size), true)
+#define lunatik_newnosharedobject(L, class, size)	lunatik_newobject((L), (class), (size), false)
```
carloslack: These are not being used?

lunatik.h
lneto: just to add.. let's not add buffers like this.. in the worst case, let's at least have a define for its size..

lunatik.h
```diff
@@ -180,27 +183,41 @@ static inline void lunatik_checkclass(lua_State *L, const lunatik_class_t *class
...
 {
-	if (luaL_getmetatable(L, class->name) == LUA_TNIL)
-		luaL_error(L, "metatable not found (%s)", class->name);
+	const char *mtname;
+	char buf[64];
+
+	if (class->shared && !shared) {
+		lunatik_noshared_name(buf, class);
+		mtname = buf;
+	} else {
```
lneto: please, keep our code style in mind.. you are well come to propose this change.. but, if so, let's do it across the code base, not in a single place ;-)

lunatik.h
```diff
@@ -90,12 +90,15 @@ typedef struct lunatik_object_s {
...
+	bool shared;
 } lunatik_object_t;
 
 extern lunatik_object_t *lunatik_env;
 
 static inline int lunatik_trylock(lunatik_object_t *object)
 {
+	if (!object->shared)
+		return 1;
 	return object->sleep ? mutex_trylock(&object->mutex) : spin_trylock(&object->spin);
```
```suggestion
return unlikely(object->shared) ?
	(object->sleep ? mutex_trylock(&object->mutex) : spin_trylock(&object->spin)) : 1;
```

lunatik.h
```diff
@@ -248,6 +269,18 @@ static inline void lunatik_newclass(lua_State *L, const lunatik_class_t *class)
 		lua_setfield(L, -2, "__index");  /* mt.__index = mt */
 	}
 	lua_pop(L, 1);  /* pop mt */
+
+	if (class->shared) {
+		char buf[64];
+		lunatik_noshared_name(buf, class);
```
lneto: why using noshared_name here if it's a shared class?

carloslack:
`lunatik_newobject` and others need update as well, for the extra argument

lunatik.h
```diff
@@ -180,27 +183,41 @@ static inline void lunatik_checkclass(lua_State *L, const lunatik_class_t *class
 		luaL_error(L, "cannot use '%s' class on non-sleepable runtime", class->name);
 }
 
-static inline void lunatik_setclass(lua_State *L, const lunatik_class_t *class)
+#define lunatik_noshared_name(buf, class)    \
+    snprintf((buf), sizeof(buf), "_%s", (class)->name)
```
lneto: we can use `lua_pushfstring` on creating the second metatable to avoid this and the static buffer

lunatik.h
lneto: I think the plan is to apply them to the libs.. I'm particular not fond of such APIs, I would rather have a flag field that would handle shared and other properties in the future.. what do you both think?

carloslack: I think these defines ATM are unnecessary. Comparing to the ones above, these new ones are more simple , and in effect I'd just call `lunatik_newobject` with flags with descriptive names.

lunatik.h
sneaky-potato: Nice idea. Resolved in latest push

lunatik.h
sneaky-potato: Resolved in latest push

sneaky-potato:
Hello @lneto, @carloslack 
Right now I am facing the following error while loading lunatik after building:
```shell
[14068.771245] lunatik: attempt to index a string value
[14068.771257] Lua warning:
[14068.771259] error in
[14068.771261] __gc
[14068.771262]  (
[14068.771263] bad argument #1 to '__gc' (rcu expected, got _rcu)
[14068.771264] )

[14068.771288] couldn't create driver runtime
```

To fix this, we need to make `lunatik_checkpobject` metatable-agnostic I think. Let me know your thoughts about refactoring this part.

lneto:
> Hello @lneto, @carloslack Right now I am facing the following error while loading lunatik after building:
> 
> ```shell
> [14068.771245] lunatik: attempt to index a string value
> [14068.771257] Lua warning:
> [14068.771259] error in
> [14068.771261] __gc
> [14068.771262]  (
> [14068.771263] bad argument #1 to '__gc' (rcu expected, got _rcu)
> [14068.771264] )
> 
> [14068.771288] couldn't create driver runtime
> ```
> 
> To fix this, we need to make `lunatik_checkpobject` metatable-agnostic I think. Let me know your thoughts about refactoring this part.



> Hello @lneto, @carloslack Right now I am facing the following error while loading lunatik after building:
> 
> ```shell
> [14068.771245] lunatik: attempt to index a string value
> [14068.771257] Lua warning:
> [14068.771259] error in
> [14068.771261] __gc
> [14068.771262]  (
> [14068.771263] bad argument #1 to '__gc' (rcu expected, got _rcu)
> [14068.771264] )
> 
> [14068.771288] couldn't create driver runtime
> ```
> 
> To fix this, we need to make `lunatik_checkpobject` metatable-agnostic I think. Let me know your thoughts about refactoring this part.

we can just use the right class name here, right? based on obj->shared? we can call tosuerdata first.. what do you think?

https://github.com/luainkernel/lunatik/blob/master/lunatik_obj.c#L58

lneto:
> > Hello @lneto, @carloslack Right now I am facing the following error while loading lunatik after building:
> > ```shell
> > [14068.771245] lunatik: attempt to index a string value
> > [14068.771257] Lua warning:
> > [14068.771259] error in
> > [14068.771261] __gc
> > [14068.771262]  (
> > [14068.771263] bad argument #1 to '__gc' (rcu expected, got _rcu)
> > [14068.771264] )
> > 
> > [14068.771288] couldn't create driver runtime
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > To fix this, we need to make `lunatik_checkpobject` metatable-agnostic I think. Let me know your thoughts about refactoring this part.
> 
> > Hello @lneto, @carloslack Right now I am facing the following error while loading lunatik after building:
> > ```shell
> > [14068.771245] lunatik: attempt to index a string value
> > [14068.771257] Lua warning:
> > [14068.771259] error in
> > [14068.771261] __gc
> > [14068.771262]  (
> > [14068.771263] bad argument #1 to '__gc' (rcu expected, got _rcu)
> > [14068.771264] )
> > 
> > [14068.771288] couldn't create driver runtime
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > To fix this, we need to make `lunatik_checkpobject` metatable-agnostic I think. Let me know your thoughts about refactoring this part.
> 
> we can just use the right class name here, right? based on obj->shared? we can call tosuerdata first.. what do you think?
> 
> https://github.com/luainkernel/lunatik/blob/master/lunatik_obj.c#L58

actually, I think a better idea is to use the second uservalue as the class name.. https://github.com/luainkernel/lunatik/blob/master/lunatik.h#L257

I think it will cleanup things to us, so we first set it on setclass, then we can use it to check whatever is needed..

lunatik.h
```diff
@@ -180,27 +181,36 @@ static inline void lunatik_checkclass(lua_State *L, const lunatik_class_t *class
...
 {
-	if (luaL_getmetatable(L, class->name) == LUA_TNIL)
-		luaL_error(L, "metatable not found (%s)", class->name);
-	lua_setmetatable(L, -2);
+	if (shared) {
+		lua_pushfstring(L, "_%s", class->name);
+	}
+	else {
+		lua_pushstring(L, class->name);
+	}
```
```suggestion
	const char *fmt = shared ? "_%s" : "%s";
	lua_pushfstring(L, fmt, class->name);
```

lunatik.h
```diff
@@ -180,27 +181,36 @@ static inline void lunatik_checkclass(lua_State *L, const lunatik_class_t *class
...
+	if (luaL_getmetatable(L, lua_tostring(L, -1)) == LUA_TNIL)
```
```suggestion
	if (lua_rawget(L, LUA_REGISTRYINDEX, -1) == LUA_TNIL)
```

we can leverage the pushed string.. also, this will pop it afterwards..

lunatik_obj.c
```diff
@@ -55,20 +55,24 @@ lunatik_object_t **lunatik_checkpobject(lua_State *L, int ix)
 	lunatik_class_t *class= lunatik_getclass(L, ix);
 
 	luaL_argcheck(L, class != NULL, ix, "object expected");
-	pobject = (lunatik_object_t **)luaL_checkudata(L, ix, class->name);
+	pobject = (lunatik_object_t **)lua_touserdata(L, ix);
```
lneto: I think it's not enough and we will need to check it's the right class as well.. why do you think it's safe?

lunatik.h
```diff
@@ -237,17 +247,24 @@ static inline bool lunatik_hasindex(lua_State *L, int index)
...
 	luaL_setfuncs(L, class->methods, 0);
+	if (monitored)
+		lunatik_monitorobject(L, class);
 	if (!lunatik_hasindex(L, -1)) {
-		if (class->shared)
-			lunatik_monitorobject(L, class);
-		lua_pushvalue(L, -1);  /* push mt */
-		lua_setfield(L, -2, "__index");  /* mt.__index = mt */
+		lua_pushvalue(L, -1);
+		lua_setfield(L, -2, "__index");
```
lneto: why removing the comments?

lunatik.h
```diff
@@ -237,17 +247,24 @@ static inline bool lunatik_hasindex(lua_State *L, int index)
...
 	}
-	lua_pop(L, 1);  /* pop mt */
+	lua_pop(L, 2); /* pop mt */
+}
+
+static inline void lunatik_newclass(lua_State *L, const lunatik_class_t *class)
```
lneto: I would just add another arg to this function.. I don't think we need this level of encapsulation bellow..

lunatik.h
```diff
@@ -237,17 +247,24 @@ static inline bool lunatik_hasindex(lua_State *L, int index)
...
 	if (!lunatik_hasindex(L, -1)) {
-		if (class->shared)
-			lunatik_monitorobject(L, class);
-		lua_pushvalue(L, -1);  /* push mt */
-		lua_setfield(L, -2, "__index");  /* mt.__index = mt */
+		lua_pushvalue(L, -1);
+		lua_setfield(L, -2, "__index");
 	}
-	lua_pop(L, 1);  /* pop mt */
+	lua_pop(L, 2); /* pop mt */
```
lneto: you are pop'ing two values here and commenting only one..

lunatik.h
sneaky-potato: resolved in latest fixup

lunatik.h
sneaky-potato: nice idea, resolved in latest fixup

lunatik_obj.c
sneaky-potato: Added a pointer validation. I think this should suffice for checking the right class, please let me know your thoughts.

lneto: I need to put more thoughts on this, but it might be indeed enough.. I would move this to an aux, though.. lunatik_checkudata?

lunatik.h
```diff
@@ -180,27 +181,30 @@ static inline void lunatik_checkclass(lua_State *L, const lunatik_class_t *class
...
 }
 
-static inline void lunatik_setclass(lua_State *L, const lunatik_class_t *class)
+static inline void lunatik_setclass(lua_State *L, const lunatik_class_t *class, bool shared)
 {
-	if (luaL_getmetatable(L, class->name) == LUA_TNIL)
-		luaL_error(L, "metatable not found (%s)", class->name);
+	const char *fmt = shared ? "_%s" : "%s";
+	lua_pushfstring(L, fmt, class->name);
+	if (lua_rawget(L, LUA_REGISTRYINDEX) == LUA_TNIL)
```
lneto: perhaps it's better to create an inline aux.. `lunatik_getmetable`, what do you think?

lib/luarcu.c
```diff
@@ -384,7 +384,7 @@ lunatik_object_t *luarcu_newtable(size_t size, bool sleep)
 	lunatik_object_t *object;
 
 	size = roundup_pow_of_two(size);
-	if ((object = lunatik_createobject(&luarcu_class, luarcu_sizeoftable(size), sleep))!= NULL)
+	if ((object = lunatik_createobject(&luarcu_class, luarcu_sizeoftable(size), sleep, true))!= NULL)
```
lneto: it doesn't seem right.. we shouldn't allow RCU to be used with monitors.. we will have a deadlock.. did you run this code? perhaps, we found the reason we need class to have `.shared`, we shouldn't allow classes that handle their own synchronization to create objects with shared = true, otherwise we will double-lock.

lib/luafifo.c
```diff
@@ -135,7 +135,7 @@ static const lunatik_class_t luafifo_class = {
 static int luafifo_new(lua_State *L)
 {
 	size_t size = luaL_checkinteger(L, 1);
-	lunatik_object_t *object = lunatik_newobject(L, &luafifo_class, sizeof(struct kfifo));
+	lunatik_object_t *object = lunatik_newobject(L, &luafifo_class, sizeof(struct kfifo), false);
```
lneto: why false here?

lib/luadevice.c
```diff
@@ -362,7 +362,7 @@ static int luadevice_new(lua_State *L)
 	lunatik_checkfield(L, 1, "name", LUA_TSTRING);
 	name = lua_tostring(L, -1);
 
-	object = lunatik_newobject(L, &luadevice_class, sizeof(luadevice_t));
+	object = lunatik_newobject(L, &luadevice_class, sizeof(luadevice_t), false);
```
lneto, line 358: why false here?

lib/luanetfilter.c
```diff
@@ -167,7 +167,7 @@ static const lunatik_class_t luanetfilter_class = {
 static int luanetfilter_register(lua_State *L)
 {
 	luaL_checktype(L, 1, LUA_TTABLE);
-	lunatik_object_t *object = lunatik_newobject(L, &luanetfilter_class , sizeof(luanetfilter_t));
+	lunatik_object_t *object = lunatik_newobject(L, &luanetfilter_class , sizeof(luanetfilter_t), false);
```
lneto, line 166: why false here?

lib/luanotifier.c
```diff
@@ -414,7 +414,7 @@ static int luanotifier_new(lua_State *L, luanotifier_register_t register_fn, lua
 
 	luaL_checktype(L, 1, LUA_TFUNCTION); /* callback */
 
-	object = lunatik_newobject(L, &luanotifier_class, sizeof(luanotifier_t));
+	object = lunatik_newobject(L, &luanotifier_class, sizeof(luanotifier_t), false);
```
lneto, line 411: why false here?

lib/luaprobe.c
```diff
@@ -241,7 +241,7 @@ static const lunatik_class_t luaprobe_class = {
 
 static int luaprobe_new(lua_State *L)
 {
-	lunatik_object_t *object = lunatik_newobject(L, &luaprobe_class, sizeof(luaprobe_t));
+	lunatik_object_t *object = lunatik_newobject(L, &luaprobe_class, sizeof(luaprobe_t), false);
```
lneto, line 239: why false here?

lib/luaxtable.c
```diff
@@ -319,7 +319,7 @@ static const lunatik_class_t luaxtable_class = {
 static inline lunatik_object_t *luaxtable_new(lua_State *L, int idx, int hook)
 {
 	luaL_checktype(L, idx, LUA_TTABLE);
-	lunatik_object_t *object = lunatik_newobject(L, &luaxtable_class , sizeof(luaxtable_t));
+	lunatik_object_t *object = lunatik_newobject(L, &luaxtable_class , sizeof(luaxtable_t), false);
```
lneto, line 318: why false here?

lunatik_obj.c
```diff
@@ -49,26 +49,38 @@ lunatik_object_t *lunatik_createobject(const lunatik_class_t *class, size_t size
...
+	bool match = (lua_touserdata(L, -1) == (void *)class);
+	lua_pop(L, 1); /* pop userdata */
+	return match ? udata : NULL;
+}
 
-	luaL_argcheck(L, class != NULL, ix, "object expected");
-	pobject = (lunatik_object_t **)luaL_checkudata(L, ix, class->name);
+lunatik_object_t **lunatik_checkpobject(lua_State *L, int ix)
+{
+	lunatik_object_t **pobject = (lunatik_object_t **)lunatik_checkudata(L, ix, lunatik_getclass(L, ix));
```
lneto: getclass also calls lua_getiuservalue https://github.com/luainkernel/lunatik/blob/master/lunatik.h#L256

you ending up comparing the same values here https://github.com/luainkernel/lunatik/pull/434/changes#diff-6dddb85ed3c5b1111e4d90a3fe6d8fad2c346c0a388035fcf59f94011b9688b8R58

right? wouldn't the idea to compare with (*pobject)->class?

lneto: I think we can do something like this..

```C
#define lunatik_isobject(pobject, class)        \
        (class != NULL && pobject != NULL && *pobject != NULL && (*pobject)->class == class)

lunatik_object_t **lunatik_checkpobject(lua_State *L, int ix)
{
        lunatik_object_t **pobject = (lunatik_object_t **)lua_touserdata(L, ix);
        lunatik_class_t *class= lunatik_getclass(L, ix);
        luaL_argcheck(L, lunatik_isobject(pobject, class), ix, "invalid object");
        return pobject;
}
EXPORT_SYMBOL(lunatik_checkpobject);
```

lunatik.h
```diff
@@ -291,7 +297,8 @@ int luaopen_##libname(lua_State *L)						\
 	luaL_newlib(L, funcs);							\
 	if (cls) {								\
 		lunatik_checkclass(L, cls);					\
-		lunatik_newclass(L, cls);					\
+		lunatik_newclass(L, cls, true);				\
```
lneto: we should avoid this when class->shared == false, as per luarcu

lunatik_obj.c
lneto: update: we already have an (unused) isobject macro: https://github.com/luainkernel/lunatik/blob/master/lunatik.h#L264

I think we could do this instead..

```C
static inline bool lunatik_isobject(lua_State *L, int ix, lunatik_object_t *object)
{
        lunatik_class_t *class= lunatik_getclass(L, ix);
        return object && object->class == class
}


lunatik_object_t **lunatik_checkpobject(lua_State *L, int ix)
{
        lunatik_object_t **pobject = (lunatik_object_t **)lua_touserdata(L, ix);
        luaL_argcheck(L, pobject && lunatik_isobject(L, ix, *pobject), ix, "invalid object");
        return pobject;
}
EXPORT_SYMBOL(lunatik_checkpobject);
```

and remove the unused macro.. what do you think?

P.S.: I'm still not that comfortable on non checking its metatable, but I think there's no possible flow where we don't set the class uservalue..

lib/luafifo.c
sneaky-potato: Yeah, this needs to be shared, will update in next push

lib/luarcu.c
sneaky-potato: Right, I agree rcu should not be locked. But we have a `lunatik_cloneobject` call in the rcu module. Setting shared = false will fail the cloneobject call (since I added a restriction that clone must fail when shared is false)

lneto: so, we need to fix that call, right? we cannot simply set a RCU table as shared, otherwise it will use the monitor.. also, we indeed need to allow it to be cloned, RCU tables can be cloned, see lunatik._ENV, for instance.. we cannot break this feature..

sneaky-potato: > also, we indeed need to allow it to be cloned, RCU tables can be cloned, see lunatik._ENV, for instance.. we cannot break this feature..

How about this, we keep class->shared = true for rcu but the object->shared = false.

class | object | remarks
-- | -- | --
shared = false | shared = false | No monitoring exists
shared = true | shared = true | Monitored instance
shared = true | shared = false | Unmonitored instance

RCU fits in the third row:
So we fail the cloneobject only when class->shared == false.

lib/luanetfilter.c
sneaky-potato: we dont need to monitor this object right? Same for luanotifier, luaprobe and luaxtable (these modules do not use the monitor). Please correct me if I am wrong.

lib/luarcu.c
lneto: what happens if class->shared = false and obj->shared = true? No clonable, but monitored?

I think your logics apply, but also it's getting more complex.. perhaps, it's the time to break it into different capabilities and use a bitfield.. the semantics for sleep is also different between class and object.. on class is more like "sleepable".. I'm wondering if `class->capabilities.sleep = true` wouldn't make it clear.. so, it reads as "this class can have objects that sleeps".. then, we can have `.clone` and `.monitor` separated, instead of overloading `.shared`.. what do you think?

P.S: we can figure better names ;-)

lib/luanetfilter.c
lneto: you are absolutely right, my mistake =).. these modules don't expose their objects to Lua, they keep them on kernel subsystems to callback Lua.. sorry for the noise..

just as reference on extending vs embedding ;-)
https://netbsd.org/~lneto/dls14.pdf
https://blog.csdn.net/weixin_33709219/article/details/85543472 (weirdly the original returns 404, only could find a copy on this chinese website)

lib/luarcu.c
sneaky-potato: > what happens if class->shared = false and obj->shared = true? No clonable, but monitored?

This configuration should not be allowed. Because if class->shared is false, I am not registering the other metatable at all (inside `luaopen_##libname()`), so it will cause problems.

> I'm wondering if class->capabilities.sleep = true wouldn't make it clear.. so, it reads as "this class can have objects that sleeps".. then, we can have .clone and .monitor separated, instead of overloading .shared.. what do you think?

Right now classs->shared is used for these things:
- checking if object can be cloned in `lunatik_cloneobject`
- checking if monitored metatable is required in newclass in `luaopen_##libname()` (this is also where I am calling `lunatik_monitorobject`.)
- However the metatable gets used inside `lunatik_setclass` which is invoked by newobject.

So the split could be:

class->shared: signals that `lunatik_monitorobject` should be called during `lunatik_newclass`, registering the other metatable
class->clone: gates `lunatik_cloneobject`, independent of monitoring
object->shared: per-object flag that drives `lunatik_setclass` to pick the right metatable, (must only be true if class->shared is also true )

Does the above make sense?

lneto: yup, it does.. but feel free to do this as a separate PR and stick with your [previous suggestion](https://github.com/luainkernel/lunatik/pull/434#discussion_r2819089159) on this one.. and then we can evolve from there..

lib/luarcu.c
```diff
@@ -372,14 +372,15 @@ static const lunatik_class_t luarcu_class = {
...
+	.shared = true,
 };
 
 lunatik_object_t *luarcu_newtable(size_t size, bool sleep)
 {
 	lunatik_object_t *object;
 
 	size = roundup_pow_of_two(size);
-	if ((object = lunatik_createobject(&luarcu_class, luarcu_sizeoftable(size), sleep))!= NULL)
+	if ((object = lunatik_createobject(&luarcu_class, luarcu_sizeoftable(size), sleep, false)) != NULL)
```
lneto, line 383: I think we should document this behavior..

lunatik_obj.c
```diff
@@ -16,14 +16,14 @@
...
+lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class, size_t size, bool shared)
 {
 	lunatik_object_t **pobject = lunatik_newpobject(L, 1);
 	lunatik_object_t *object = lunatik_checkalloc(L, sizeof(lunatik_object_t));
 
 	lunatik_checkclass(L, class);
-	lunatik_setobject(object, class, class->sleep);
-	lunatik_setclass(L, class);
+	lunatik_setobject(object, class, class->sleep, shared);
+	lunatik_setclass(L, class, shared);
```
lneto, line 30: shouldn't we add a guard for the invalid case here?

lneto (changes requested):
I think we are almost there, could you 1) add some documentation on the shared flag and 2) create an issue to improve the flags in the way we have discussed on this PR? Thanks!

lunatik_obj.c
```diff
@@ -16,14 +16,23 @@
...
 	(reg)->func == lunatik_closeobject)
 
-lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class, size_t size)
+/*
+ * Creates a new object of the given class.
+ * If shared = true, the object will use the monitored metatable and
+ * participate in cross-state sharing. This requires class->shared = true;
+ * passing shared = true for a non-shared class is invalid and will raise
+ * a Lua error.
+ */
```
lneto: I would move this to our C API doc, it's where we explain sleep as well..

lunatik.h
```diff
@@ -181,27 +182,30 @@ static inline void lunatik_checkclass(lua_State *L, const lunatik_class_t *class
 		luaL_error(L, "cannot use '%s' class on non-sleepable runtime", class->name);
 }
 
-static inline void lunatik_setclass(lua_State *L, const lunatik_class_t *class)
+static inline void lunatik_setclass(lua_State *L, const lunatik_class_t *class, bool shared)
 {
-	if (luaL_getmetatable(L, class->name) == LUA_TNIL)
-		luaL_error(L, "metatable not found (%s)", class->name);
+	const char *fmt = shared ? "_%s" : "%s";
+	lua_pushfstring(L, fmt, class->name);
```
```suggestion
	lua_pushfstring(L, shared ? "_%s" : "%s", class->name);
```

lunatik.h
```diff
@@ -238,17 +242,19 @@ static inline bool lunatik_hasindex(lua_State *L, int index)
...
-static inline void lunatik_newclass(lua_State *L, const lunatik_class_t *class)
+static inline void lunatik_newclass(lua_State *L, const lunatik_class_t *class, bool monitored)
 {
-	luaL_newmetatable(L, class->name); /* mt = {} */
+	const char *fmt = monitored ? "_%s" : "%s";
+	lua_pushfstring(L, fmt, class->name);
+	luaL_newmetatable(L, lua_tostring(L, -1)); /* mt = {} */
 	luaL_setfuncs(L, class->methods, 0);
+	if (monitored)
+		lunatik_monitorobject(L, class);
```
lneto, line 251: we are moving this to here because of the RCU case?

lunatik_obj.c
```diff
@@ -32,15 +41,19 @@ lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class,
 }
 EXPORT_SYMBOL(lunatik_newobject);
 
-lunatik_object_t *lunatik_createobject(const lunatik_class_t *class, size_t size, bool sleep)
+lunatik_object_t *lunatik_createobject(const lunatik_class_t *class, size_t size, bool sleep, bool shared)
 {
 	gfp_t gfp = sleep ? GFP_KERNEL : GFP_ATOMIC;
+	if (shared && !class->shared) {
+		pr_err("lunatik: cannot create shared object from non-shared class '%s'\n", class->name);
```
lneto: pr_err will prefix with module name, right? I don't think we need "lunatik: " here.

lunatik_obj.c
```diff
@@ -32,15 +41,19 @@ lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class,
 }
 EXPORT_SYMBOL(lunatik_newobject);
 
-lunatik_object_t *lunatik_createobject(const lunatik_class_t *class, size_t size, bool sleep)
+lunatik_object_t *lunatik_createobject(const lunatik_class_t *class, size_t size, bool sleep, bool shared)
 {
 	gfp_t gfp = sleep ? GFP_KERNEL : GFP_ATOMIC;
+	if (shared && !class->shared) {
```
lneto: we should move it to a macro.. `lunatik_isshared`, `lunatik_issharable`?

lunatik_obj.c
lneto: also, let's not copy and paste error messages around.. let's use the same strategy of LUNATIK_ERR_NULLPTR..

lunatik_obj.c
```diff
@@ -32,15 +41,19 @@ lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class,
...
+		pr_err("lunatik: cannot create shared object from non-shared class '%s'\n", class->name);
+		return NULL;
```
lneto: can you check what will happen on the caller side on this case? I have the impression it will propagate a ENOMEM error.. it might be misleading.. perhaps it's that case of using BUG_ON here..

sneaky-potato: Since this is a development-time logic error (invalid API usage), how about WARN_ON() instead of BUG_ON? Or maybe should we return ERR_PTR(-EINVAL instead of NULL?

> can you check what will happen on the caller side on this case?

I should approach to test this via a standalone kernel module, right? Does our current Makefile support adding out-of-tree test modules?

lneto: > Since this is a development-time logic error (invalid API usage), how about WARN_ON() instead of BUG_ON? Or maybe should we return ERR_PTR(-EINVAL instead of NULL?

I think returning ERR_PTR would be more appropriate.. however, we would need to handle it on every caller..

> I should approach to test this via a standalone kernel module, right? Does our current Makefile support adding out-of-tree test modules?

well, every embedding binding (e.g., luanetfilter, luaxdp, luanotifier) is a out-of-tree kmod..

lunatik.h
sneaky-potato: Yes, if class->shared is true then lunatik_newclass is called with monitored = true and we get another metatable. We dont do this operation if class->shared = false

doc/capi.md
```diff
@@ -113,6 +113,32 @@ static ssize_t mydevice_read(struct file *f, char *buf, size_t len, loff_t *off)
...
 ```
 
+## lunatik\_newobject
+```C
+lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class, size_t size, bool shared);
+```
+_lunatik\_newobject()_ allocates a new Lunatik object and pushes a userdata
+containing a pointer to the object onto the Lua stack.
+If `shared` is _true_, the object will use the monitored metatable for safe
+cross-state access. This requires `class->shared = true;` otherwise, it raises a
```
lneto:
```suggestion
access Lunatik runtimes. This requires `class->shared = true;` otherwise, it raises a
```

doc/capi.md
```diff
@@ -113,6 +113,32 @@ static ssize_t mydevice_read(struct file *f, char *buf, size_t len, loff_t *off)
...
+
+## lunatik\_createobject
+```C
+lunatik_object_t *lunatik_createobject(const lunatik_class_t *class, size_t size, bool sleep, bool shared);
+```
+_lunatik\_createobject()_ creates a Lunatik object independently of any Lua
+state. This is intended for objects created in C context that may be shared
+with Lua runtimes later.
+
+It Uses kmalloc with `GFP_KERNEL` if `sleep` is _true_, or `GFP_ATOMIC` otherwise.
```
lneto:
```suggestion
It allocates memory with `GFP_KERNEL` if `sleep` is _true_, or `GFP_ATOMIC` otherwise.
```

doc/capi.md
```diff
@@ -113,6 +113,32 @@ static ssize_t mydevice_read(struct file *f, char *buf, size_t len, loff_t *off)
...
+It returns a pointer to the `lunatik_object_t` on success, or NULL if memory
+allocation fails or if the class is not sharable
```
lneto:
```suggestion
allocation fails or if the class is not sharable.
```

lunatik_obj.c
```diff
@@ -16,14 +16,19 @@
 	(reg)->func == lunatik_deleteobject || \
 	(reg)->func == lunatik_closeobject)
 
-lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class, size_t size)
+#define lunatik_issharable(class, shared) (!(shared) || ((class)->shared))
+#define LUNATIK_ERR_SHARED "cannot create shared object from non-shared class '%s'"
```
lneto: please, check if you are using tabs here..

doc/capi.md
lneto: actually it returns ERR_PTR(-EINVAL), if trying to create a sharable object out from a not sharable class, right/

lib/luarcu.c
```diff
@@ -372,14 +372,16 @@ static const lunatik_class_t luarcu_class = {
...
+	.shared = true,
 };
 
 lunatik_object_t *luarcu_newtable(size_t size, bool sleep)
 {
 	lunatik_object_t *object;
 
 	size = roundup_pow_of_two(size);
-	if ((object = lunatik_createobject(&luarcu_class, luarcu_sizeoftable(size), sleep))!= NULL)
+	object = lunatik_createobject(&luarcu_class, luarcu_sizeoftable(size), sleep, false);
```
lneto:
```suggestion
	lunatik_object_t *object = lunatik_createobject(&luarcu_class, luarcu_sizeoftable(size), sleep, false);
```

lib/luarcu.c
```diff
@@ -372,14 +372,16 @@ static const lunatik_class_t luarcu_class = {
...
+	if (!IS_ERR_OR_NULL(object))
 		luarcu_inittable((luarcu_table_t *)object->private, size);
 	return object;
```
lneto, line 385: here we propagate the -EINVAL error, right? what happens on the callers?

lib/luadata.c
```diff
@@ -445,20 +445,20 @@ static int luadata_lnew(lua_State *L)
...
 {
-	lunatik_object_t *object = lunatik_createobject(&luadata_class, sizeof(luadata_t), sleep);
+	lunatik_object_t *object = lunatik_createobject(&luadata_class, sizeof(luadata_t), sleep, shared);
 
-	if (object != NULL) {
+	if (!IS_ERR_OR_NULL(object)) {
 		luadata_t *data = (luadata_t *)object->private;
 		luadata_set(data, ptr, 0, size, opt);
 	}
 	return object;
```
lneto, line 456: here we propagate -EINVAL as well..

lib/luadata.c
```diff
@@ -445,20 +445,20 @@ static int luadata_lnew(lua_State *L)
...
 }
 
-lunatik_object_t *luadata_new(lua_State *L)
+lunatik_object_t *luadata_new(lua_State *L, bool shared)
 {
-	lunatik_object_t *data = lunatik_checknull(L, luadata_create(NULL, 0, false, LUADATA_OPT_NONE));
+	lunatik_object_t *data = lunatik_checknull(L, luadata_create(NULL, 0, false, LUADATA_OPT_NONE, shared));
```
lneto: this shouldn't be enough to cover -EINVAL

lib/luarcu.c
sneaky-potato: Handled on caller side

lib/luadata.c
sneaky-potato: Resolved on caller side (luadata_new)

lunatik.h
```diff
@@ -128,14 +129,25 @@ static inline void *lunatik_realloc(lua_State *L, void *ptr, size_t size)
...
 		lunatik_enomem(L);
 	return ptr;
 }
 
-#define lunatik_checkalloc(L, s)	(lunatik_checknull((L), lunatik_malloc((L), (s))))
+#define lunatik_handleerr(ptr)	\
+do {						\
+	typeof(ptr) __ptr = (ptr);		\
+	if (IS_ERR_OR_NULL(__ptr))		\
+		return __ptr ? PTR_ERR(__ptr) : -ENOMEM;	\
```
lneto: we are not doing this sort of implicit returns on Lunatik.. instead, we can have macros to test the conditions then explicitly returning.. the only case we do something similar is on lua error handling through longjmps..

sneaky-potato: this is used only in lunatik_run.c. Let me get rid of this macro and use the following in lunatik_run.c
```shell
diff --git a/lunatik_run.c b/lunatik_run.c
index b0a8d2fe..ba75c955 100644
--- a/lunatik_run.c
+++ b/lunatik_run.c
@@ -18,8 +18,9 @@ static int __init lunatik_run_init(void)
 {
 	int ret = 0;
 
-	if ((lunatik_env = luarcu_newtable(LUARCU_DEFAULT_SIZE, false)) == NULL)
-		return -ENOMEM;
+	lunatik_env = luarcu_newtable(LUARCU_DEFAULT_SIZE, false);
+	if (IS_ERR_OR_NULL(lunatik_env))
+		return lunatik_env ? PTR_ERR(lunatik_env) : -ENOMEM;
 
 	if ((ret = lunatik_runtime(&runtime, "driver", true)) < 0) {
 		pr_err("couldn't create driver runtime\n");

```
Does this make sense?

lneto: shared must be true for env, right? Also, in this case, we shouldn't care about -EINVAL.

lib/luarcu.c
sneaky-potato: Ah, I realized this call will never produce -EINVAL as rcu is a shared class

lunatik.h
sneaky-potato: Right, this call will never produce -EINVAL

lunatik_run.c
```diff
@@ -18,8 +18,9 @@ static int __init lunatik_run_init(void)
 {
 	int ret = 0;
 
-	if ((lunatik_env = luarcu_newtable(LUARCU_DEFAULT_SIZE, false)) == NULL)
-		return -ENOMEM;
+	lunatik_env = luarcu_newtable(LUARCU_DEFAULT_SIZE, false);
+	if (IS_ERR_OR_NULL(lunatik_env))
+		return lunatik_env ? PTR_ERR(lunatik_env) : -ENOMEM;
```
lneto: this pattern makes me think if we shouldn't just return ERR_PTR(-ENOMEM) on createobject.. so we can just pass the error all the through the stack.. what do you think?

sneaky-potato: That is a nice idea, this way we can extend support for more errors easily

sneaky-potato: Currently I have reverted the change on this file as we don't need to handle EINVAL right now.

lunatik.h
```diff
@@ -128,14 +129,18 @@ static inline void *lunatik_realloc(lua_State *L, void *ptr, size_t size)
...
 {
+	if (IS_ERR(ptr))
+		luaL_error(L, LUNATIK_ERR_SHARED);
 	if (ptr == NULL)
 		lunatik_enomem(L);
 	return ptr;
 }
 
-#define lunatik_checkalloc(L, s)	(lunatik_checknull((L), lunatik_malloc((L), (s))))
+#define lunatik_checkalloc(L, s)	(lunatik_checkerr((L), lunatik_malloc((L), (s))))
```
lneto: I don't think we should make this change.. it will interferer with other things.. I would keep it simpler in this PR and just handle shared separated.. also, this checkerr is misleading.. it treats all ptr errors as shared errors.. checkalloc can be used on many other places..

sneaky-potato: I agree, since the scope is not that much. I have reverted this change, and now checking for error explicitly.

lib/luadata.c
```diff
@@ -445,20 +445,23 @@ static int luadata_lnew(lua_State *L)
...
 		luadata_set(data, ptr, 0, size, opt);
 	}
 	return object;
 }
 
-lunatik_object_t *luadata_new(lua_State *L)
+lunatik_object_t *luadata_new(lua_State *L, bool shared)
 {
-	lunatik_object_t *data = lunatik_checknull(L, luadata_create(NULL, 0, false, LUADATA_OPT_NONE));
+	lunatik_object_t *data = luadata_create(NULL, 0, false, LUADATA_OPT_NONE, shared);
```
lneto: why changing this? it would be enough to just add the block below, right?

lunatik_obj.c
```diff
@@ -32,15 +36,18 @@ lunatik_object_t *lunatik_newobject(lua_State *L, const lunatik_class_t *class,
 }
 EXPORT_SYMBOL(lunatik_newobject);
 
-lunatik_object_t *lunatik_createobject(const lunatik_class_t *class, size_t size, bool sleep)
+lunatik_object_t *lunatik_createobject(const lunatik_class_t *class, size_t size, bool sleep, bool shared)
 {
 	gfp_t gfp = sleep ? GFP_KERNEL : GFP_ATOMIC;
 	lunatik_object_t *object = (lunatik_object_t *)kmalloc(sizeof(lunatik_object_t), gfp);
-
```
lneto, line 39: why removing this line? please, don't.

sneaky-potato: oops, my bad

lneto (approved):

PR #438, opened 2026-02-16
add luartree

lneto:


lib/luartree.c
```diff
@@ -0,0 +1,204 @@
...
+		
+		node = luartree_getchild(cur, hash);
+		if (!node)
+			node = luartree_newchild(cur, hash, L);
+		
+		cur = &node->children;
+	}
+	
+	node->is_leaf = true;
+
```
lneto:
```suggestion
```

lib/luartree.c
```diff
@@ -0,0 +1,204 @@
...
+/***
+* Reverse tree using XArray.
+* A tree structure where paths are stored in reverse order for efficient
+* prefix sharing (e.g., domain names, file paths).
+*
+* @module rtree
+*/
+
+#define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+#include <linux/kernel.h>
```
lneto: Unnecessary

lib/luartree.c
```diff
@@ -0,0 +1,204 @@
...
+*/
+static const luaL_Reg luartree_lib[] = {
+	{"new", luartree_new},
+	{NULL, NULL}
+};
+
+static const luaL_Reg luartree_mt[] = {
+	{"__gc", lunatik_deleteobject},
+	{"__close", lunatik_closeobject},
+	{"close", lunatik_closeobject},
```
lneto, line 217: We don't need this.. it's not a resource like a socket or file..

lib/luartree.c
```diff
@@ -0,0 +1,256 @@
...
+* @usage t:insert(42, "ai", "claude", "bar")
+* @usage t:insert(true, "ai", "claude", "baz")
+*/
+static int luartree_insert(lua_State *L)
+{
+	struct xarray *root = luartree_check(L, 1);
+	int n = lua_gettop(L) - 2;
+
+	if (n < 1)
+		luaL_error(L, "insert requires value and path");
```
lneto, line 127: improve this error.. mathlib and lualinux have good examples on throwing error due to #arg, I think

lib/luartree.c
```diff
@@ -0,0 +1,256 @@
...
+
+	struct xarray *xa = root;
+	luartree_node_t *node = NULL;
+
+	for (int i = 3; i <= n + 2; i++) {
+		size_t len;
+		const char *label = luaL_checklstring(L, i, &len);
+		unsigned long hash = lunatik_hash(label, len, 0);
```
lneto, line 135: we probably can create a foreach macro for this

lib/luartree.c
```diff
@@ -0,0 +1,256 @@
...
+
+		node = luartree_getchild(xa, hash);
+		if (!node)
+			goto notfound;
+
+		xa = &node->children;
+	}
+
+	if (node) {
```
lneto, line 178: I think if we reach out this line is because we have found the node..

lib/luartree.c
```diff
@@ -0,0 +1,256 @@
...
+		luartree_pushnode(L, node);
+		return 1;
+	}
+
+notfound:
+	lua_pushnil(L);
+	return 1;
```
lneto, line 185: we can just return 0 here..

lib/luartree.c
```diff
@@ -0,0 +1,256 @@
...
+			lunatik_free(node->value.s);
+		xa_erase(xa, index);
+		lunatik_free(node);
+	}
+}
+
+static void luartree_tonode(lua_State *L, luartree_node_t *node, int idx)
+{
+	if (node->type == LUA_TSTRING)
+		lunatik_free((void *)node->value.s);
```
lneto, line 66: I think it will be better to move this to the caller function

lunatik.h
```diff
@@ -137,6 +137,13 @@ static inline void *lunatik_checknull(lua_State *L, void *ptr)
 
 #define lunatik_checkalloc(L, s)	(lunatik_checknull((L), lunatik_malloc((L), (s))))
 
+static inline void *lunatik_checkzalloc(lua_State *L, size_t size)
+{
+	void *ptr = lunatik_checkalloc(L, size);
```
carloslack, line 142: Can you use  `lunatik_checkzalloc` here? then you could drop the `memset` below.

lib/luartree.c
carloslack: I agree, I may be missing some context here but if we end up in the default case, you'd be freeing `s` without ever assigning a new value to it

lunatik.h
lneto: Wouldn't it be a recursive definition? 🤔

carloslack: ah! now I see it, my bad :)

PR #440, opened 2026-02-20
Moving to task

Suryansh-Dey:
## Move `lualinux_task` and `lualinux_stat` to [configure.lua](cci:7://file:///home/suryansh/dev/GSOC/lunatik/configure.lua:0:0-0:0)

Closes #435

### Summary

Auto-generates `linux.task` and `linux.stat` constants at build time via [configure.lua](cci:7://file:///home/suryansh/dev/GSOC/lunatik/configure.lua:0:0-0:0), removing the hardcoded C arrays from [lualinux.c](cci:7://file:///home/suryansh/dev/GSOC/lunatik/lib/lualinux.c:0:0-0:0).

#### [configure.lua](cci:7://file:///home/suryansh/dev/GSOC/lunatik/configure.lua:0:0-0:0)
- Added **grep-based extraction** for kernel-internal headers (`internal = true` flag on specs). The C preprocessor can't be used on [linux/sched.h](cci:7://file:///lib/modules/6.12.73-1-lts/build/include/linux/sched.h/lib/modules/6.12.73-1-lts/build/include/linux/sched.h:0:0-0:0) or [linux/stat.h](cci:7://file:///lib/modules/6.12.73-1-lts/build/include/linux/stat.h/lib/modules/6.12.73-1-lts/build/include/linux/stat.h:0:0-0:0) because they pull in massive kernel dependency chains that cause hangs. Instead, `#define` lines are extracted directly via file reading and parsed in Lua.
- Added `parse_c_number()` to correctly handle C octal literals (e.g. `00700` for `S_IRWXU`).
- Added `resolve_numeric()` with OR-expression support to resolve composites like `TASK_KILLABLE = (TASK_WAKEKILL | TASK_UNINTERRUPTIBLE)`.
- Added `supplement_constants()` to merge composites from [linux/stat.h](cci:7://file:///lib/modules/6.12.73-1-lts/build/include/linux/stat.h/lib/modules/6.12.73-1-lts/build/include/linux/stat.h:0:0-0:0) into the base constants already extracted from `uapi/linux/stat.h`.
- New spec: [linux/sched.h](cci:7://file:///lib/modules/6.12.73-1-lts/build/include/linux/sched.h/lib/modules/6.12.73-1-lts/build/include/linux/sched.h:0:0-0:0) → `task` module (grep mode, all `TASK_` constants)
[...]

PR #441, opened 2026-02-21
auto generating task

### LINKED ISSUE #435: move lualinux_task and lualinux_stat to configure.lua
these flags could be autogenerated on the build time

Suryansh-Dey:
Closes #435  
Here is phase 1 of the [PR](https://github.com/luainkernel/lunatik/pull/440)  
- Simplify configure.lua (remove stat, signal, eth, supplement logic)
- Restore lualinux_stat in lib/lualinux.c
- Update documentation for task access

Now `make configure` auto generates hexadecimal constants for both task and stat modules.  

Phase 2: Move stat constants  
Phase 3: Move OR'ed flags (composites)
Ready for the feedback.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,8 @@ MODULES_BUILD_PATH ?= ${BTF_INSTALL_PATH}
 MODULES_INSTALL_PATH := ${MODULES_RELEASE_PATH}/kernel
 SCRIPTS_INSTALL_PATH := ${MODULES_PATH}/lua
 INCLUDE_PATH := ${MODULES_BUILD_PATH}/include
+KERNEL_ARCH ?= ${shell uname -m}
+ARCH ?= $(subst x86_64,x86,$(subst aarch64,arm64,$(KERNEL_ARCH)))
 
 LUA ?= lua5.4
 LUA_PATH ?= $(shell $(LUA) -e 'print(package.path:match("([^;]*)/%?%.lua;"))')
@@ -154,7 +156,7 @@ lunatik_sym.h: $(LUA_API) gensymbols.sh
 	${shell CC='$(CC)' ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
 
 configure:
-	CC='$(CC)' "$(LUA)" configure.lua "$(KERNEL_RELEASE)" "$(INCLUDE_PATH)" "$(LUNATIK_MODULES)"
+	CC='$(CC)' "$(LUA)" configure.lua "$(KERNEL_RELEASE)" "$(INCLUDE_PATH)" "$(LUNATIK_MODULES)" "$(ARCH)"
 
 moontastik_install_%:
 	[ $* ] || (echo "usage: make moontastik_install_TARGET" ; exit 1)
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ BTF_INSTALL_PATH = ${MODULES_RELEASE_PATH}/build
 MODULES_BUILD_PATH ?= ${BTF_INSTALL_PATH}
 MODULES_INSTALL_PATH := ${MODULES_RELEASE_PATH}/kernel
 SCRIPTS_INSTALL_PATH := ${MODULES_PATH}/lua
-INCLUDE_PATH := ${MODULES_BUILD_PATH}/include
+
 KERNEL_ARCH ?= ${shell uname -m}
 ARCH ?= $(subst x86_64,x86,$(subst aarch64,arm64,$(KERNEL_ARCH)))
 
@@ -156,7 +156,7 @@ lunatik_sym.h: $(LUA_API) gensymbols.sh
 	${shell CC='$(CC)' ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
 
 configure:
-	CC='$(CC)' "$(LUA)" configure.lua "$(KERNEL_RELEASE)" "$(INCLUDE_PATH)" "$(LUNATIK_MODULES)" "$(ARCH)"
+	CC='$(CC)' "$(LUA)" configure.lua "$(KERNEL_RELEASE)" "$(MODULES_BUILD_PATH)" "$(LUNATIK_MODULES)" "$(ARCH)"
 
 moontastik_install_%:
 	[ $* ] || (echo "usage: make moontastik_install_TARGET" ; exit 1)
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ MODULES_INSTALL_PATH := ${MODULES_RELEASE_PATH}/kernel
 SCRIPTS_INSTALL_PATH := ${MODULES_PATH}/lua
 
 KERNEL_ARCH ?= ${shell uname -m}
-ARCH ?= $(subst x86_64,x86,$(subst aarch64,arm64,$(KERNEL_ARCH)))
+export ARCH ?= $(subst x86_64,x86,$(subst aarch64,arm64,$(KERNEL_ARCH)))
 
 LUA ?= lua5.4
 LUA_PATH ?= $(shell $(LUA) -e 'print(package.path:match("([^;]*)/%?%.lua;"))')
@@ -156,7 +156,7 @@ lunatik_sym.h: $(LUA_API) gensymbols.sh
 	${shell CC='$(CC)' ./gensymbols.sh $(LUA_API) > lunatik_sym.h}
 
 configure:
-	CC='$(CC)' "$(LUA)" configure.lua "$(KERNEL_RELEASE)" "$(MODULES_BUILD_PATH)" "$(LUNATIK_MODULES)" "$(ARCH)"
+	CC='$(CC)' "$(LUA)" configure.lua "$(KERNEL_RELEASE)" "$(MODULES_BUILD_PATH)" "$(LUNATIK_MODULES)"
 
 moontastik_install_%:
 	[ $* ] || (echo "usage: make moontastik_install_TARGET" ; exit 1)
--- a/configure.lua
+++ b/configure.lua
@@ -17,7 +17,7 @@ local specs = {
 		module_name = "eth",
 	},
 	{
-		header = "uapi/linux/stat.h",
+		header = "linux/stat.h",
 		prefix = "S_",
 		module_name = "stat",
 	},
--- a/configure.lua
+++ b/configure.lua
@@ -6,7 +6,7 @@
 local OUTPUT_DIR = "autogen"
 local SCRIPT  = arg[0]
 local KERNEL  = arg[1]
-local INCLUDE = arg[2]
+local BUILD   = arg[2]
 local MODULES = arg[3]
 local ARCH    = arg[4]
 
@@ -38,12 +38,12 @@ local function exit(msg)
 	os.exit(1)
 end
 
-if not KERNEL or not INCLUDE or not MODULES or not ARCH then
-	exit("usage: lua5.4 " .. SCRIPT .. " <KERNEL> <INCLUDE> <MODULES> <ARCH>")
+if not KERNEL or not BUILD or not MODULES or not ARCH then
+	exit("usage: lua5.4 " .. SCRIPT .. " <KERNEL> <BUILD> <MODULES> <ARCH>")
 end
 
 local CC = os.getenv("CC") or "cc"
-local BUILD = INCLUDE:gsub("/include$", "")
+local INCLUDE = BUILD .. "/include"
 local CPP = string.format("%s -E -dM -I%s -I%s/include/generated -I%s/arch/%s/include -I%s/arch/%s/include/generated -include linux/kconfig.h",
 	CC, INCLUDE, BUILD, BUILD, ARCH, BUILD, ARCH)
 
@@ -77,24 +77,18 @@ local function resolve_value(value, prefix, constants, seen)
 	if seen[value] then return nil end
 
 	if tonumber(value) then return tonumber(value) end
+	local clean = value:match("^%((.+)%)$") or value
 
 	-- OR expression: (A | B | ...)
-	local inner = value:match("^%((.+)%)$")
-	if inner then
-		local is_or_expr = true
-		local result = 0
-		for token in inner:gmatch("[^%s|]+") do
-			-- only resolve if token is a number or it starts with the expected prefix
-			if not (tonumber(token) or string.sub(token, 1, #prefix) == prefix) then
-				is_or_expr = false
-				break
-			end
-			local resolved = resolve_value(token, prefix, constants, seen)
-			if not resolved then return nil end
-			result = result | resolved
-		end
-		if is_or_expr then return result end
-	end
+    if clean:find("|") then
+        local result = 0
+        for token in clean:gmatch("[^%s|]+") do
+            local resolved = resolve_value(token, prefix, constants, seen)
+            if not resolved then return nil end
+            result = result | resolved
+        end
+        return result
+    end
 
 	-- macro reference
 	if not constants[value] then return nil end
--- a/configure.lua
+++ b/configure.lua
@@ -8,6 +8,7 @@ local SCRIPT  = arg[0]
 local KERNEL  = arg[1]
 local INCLUDE = arg[2]
 local MODULES = arg[3]
+local ARCH    = arg[4]
 
 local specs = {
 	{
@@ -24,6 +25,11 @@ local specs = {
 		header = "uapi/linux/signal.h",
 		prefix = "SIG",
 		module_name = "signal",
+	},
+	{
+		header = "linux/sched.h",
+		prefix = "TASK_",
+		module_name = "task",
 	}
 }
 
@@ -32,12 +38,14 @@ local function exit(msg)
 	os.exit(1)
 end
 
-if not KERNEL or not INCLUDE or not MODULES then
-	exit("usage: lua5.4 " .. SCRIPT .. " <KERNEL> <INCLUDE> <MODULES>")
+if not KERNEL or not INCLUDE or not MODULES or not ARCH then
+	exit("usage: lua5.4 " .. SCRIPT .. " <KERNEL> <INCLUDE> <MODULES> <ARCH>")
 end
 
 local CC = os.getenv("CC") or "cc"
-local CPP = string.format("%s -E -dM -I%s", CC, INCLUDE)
+local BUILD = INCLUDE:gsub("/include$", "")
+local CPP = string.format("%s -E -dM -I%s -I%s/include/generated -I%s/arch/%s/include -I%s/arch/%s/include/generated -include linux/kconfig.h",
+	CC, INCLUDE, BUILD, BUILD, ARCH, BUILD, ARCH)
 
 local function preprocess(header_path)
 	local cmd = string.format("%s %s/%s", CPP, INCLUDE, header_path)
@@ -52,37 +60,55 @@ local function collect_constants(cpp_output, prefix)
 	local constants = {}
 	local macro_names = {}
 
-	for macro, literal in cpp_output:gmatch("#define%s+(" .. prefix .. "%w+)%s+(%S+)") do
-		constants[macro] = literal
-		table.insert(macro_names, macro)
+	for line in cpp_output:gmatch("[^\n]+") do
+		local macro, value = line:match("#define%s+(" .. prefix .. "[%u%d_]+)%s+(.+)%s*$")
+		if macro and not line:match("#define%s+" .. prefix .. "%w+%(") then
+			constants[macro] = value
+			table.insert(macro_names, macro)
+		end
 	end
 
 	table.sort(macro_names)
 	return constants, macro_names
 end
 
-local function resolve_value(value, constants, prefix, module_name, seen)
-	if tonumber(value) then return value end
-
-	if not constants[value] then return nil end
-
+local function resolve_value(value, prefix, constants, seen)
 	seen = seen or {}
-	-- recursion check: avoid cyclic defines
 	if seen[value] then return nil end
-	seen[value] = true
 
-	if value:find("^" .. prefix) then
-		local stripped = value:gsub("^" .. prefix, "")
-		return string.format('%s["%s"]', module_name, stripped)
+	if tonumber(value) then return tonumber(value) end
+
+	-- OR expression: (A | B | ...)
+	local inner = value:match("^%((.+)%)$")
+	if inner then
+		local is_or_expr = true
+		local result = 0
+		for token in inner:gmatch("[^%s|]+") do
+			-- only resolve if token is a number or it starts with the expected prefix
+			if not (tonumber(token) or string.sub(token, 1, #prefix) == prefix) then
+				is_or_expr = false
+				break
+			end
+			local resolved = resolve_value(token, prefix, constants, seen)
+			if not resolved then return nil end
+			result = result | resolved
+		end
+		if is_or_expr then return result end
 	end
 
-	return resolve_value(constants[value], constants, prefix, module_name, seen)
+	-- macro reference
+	if not constants[value] then return nil end
+
+	seen[value] = true
+	local res = resolve_value(constants[value], prefix, constants, seen)
+	seen[value] = nil
+	return res
 end
 
 local function write_constants(file, module, spec, constants, macro_names)
 	for _, macro in ipairs(macro_names) do
 		local field_name = macro:gsub("^" .. spec.prefix, "")
-		local resolved_value = resolve_value(constants[macro], constants, spec.prefix, spec.module_name)
+		local resolved_value = resolve_value(constants[macro], spec.prefix, constants)
 		if resolved_value then
 			file:write(string.format('%s["%s"]\t= %s\n', spec.module_name, field_name, resolved_value))
 		end
--- a/configure.lua
+++ b/configure.lua
@@ -8,7 +8,7 @@ local SCRIPT  = arg[0]
 local KERNEL  = arg[1]
 local BUILD   = arg[2]
 local MODULES = arg[3]
-local ARCH    = arg[4]
+local ARCH    = os.getenv("ARCH")
 
 local specs = {
 	{
@@ -39,7 +39,7 @@ local function exit(msg)
 end
 
 if not KERNEL or not BUILD or not MODULES or not ARCH then
-	exit("usage: lua5.4 " .. SCRIPT .. " <KERNEL> <BUILD> <MODULES> <ARCH>")
+	exit("usage: ARCH=<ARCH> lua5.4 " .. SCRIPT .. " <KERNEL> <BUILD> <MODULES>")
 end
 
 local CC = os.getenv("CC") or "cc"
--- a/lib/lualinux.c
+++ b/lib/lualinux.c
@@ -88,7 +88,8 @@ static int lualinux_random(lua_State *L)
 * @see linux.task
 * @usage
 *   linux.schedule(1000) -- Sleep for 1 second (interruptible)
-*   linux.schedule(500, linux.task.UNINTERRUPTIBLE) -- Sleep for 0.5 seconds (uninterruptible)
+*   local task = require("linux.task")
+*   linux.schedule(500, task.UNINTERRUPTIBLE) -- Sleep for 0.5 seconds (uninterruptible)
 */
 static int lualinux_schedule(lua_State *L)
 {
@@ -231,27 +232,6 @@ static int lualinux_ifaddr(lua_State *L)
 }
 
 /***
-* Table of task state constants.
-* Exports task state flags from `<linux/sched.h>`. These are used with
-* `linux.schedule()`.
-*
-* @table task
-*   @tfield integer INTERRUPTIBLE Task is waiting for a signal or a resource (sleeping), can be interrupted.
-*   @tfield integer UNINTERRUPTIBLE Task is waiting (sleeping),
-*   cannot be interrupted by signals (except fatal ones if KILLABLE is also implied by context).
-*   @tfield integer KILLABLE Task is waiting (sleeping) like UNINTERRUPTIBLE, but can be interrupted by fatal signals.
-*   @tfield integer IDLE Task is idle, similar to UNINTERRUPTIBLE but avoids loadavg accounting.
-* @see linux.schedule
-*/
-static const lunatik_reg_t lualinux_task[] = {
-	{"INTERRUPTIBLE", TASK_INTERRUPTIBLE},
-	{"UNINTERRUPTIBLE", TASK_UNINTERRUPTIBLE},
-	{"KILLABLE", TASK_KILLABLE},
-	{"IDLE", TASK_IDLE},
-	{NULL, 0}
-};
-
-/***
 * Table of file mode constants.
 * Exports file permission flags from `<linux/stat.h>`. These can be used, for
 * example, with `device.new()` to set the mode of a character device.
@@ -302,7 +282,6 @@ static int lualinux_errname(lua_State *L)
 
 static const lunatik_namespace_t lualinux_flags[] = {
 	{"stat", lualinux_stat},
-	{"task", lualinux_task},
 	{NULL, NULL}
 };
 
--- a/lib/lualinux.c
+++ b/lib/lualinux.c
@@ -13,7 +13,6 @@
 */
 
 #include <linux/random.h>
-#include <linux/stat.h>
 #include <linux/sched.h>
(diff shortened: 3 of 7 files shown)


lib/lualinux.c
```diff
@@ -88,7 +88,9 @@ static int lualinux_random(lua_State *L)
 * @see linux.task
 * @usage
 *   linux.schedule(1000) -- Sleep for 1 second (interruptible)
-*   linux.schedule(500, linux.task.UNINTERRUPTIBLE) -- Sleep for 0.5 seconds (uninterruptible)
+ *
+ *   local task = require("linux.task")
```
```suggestion
*   local task = require("linux.task")
```

please, mind the code style of the files.. read and respect their form.. always double check your patches

configure.lua
```diff
@@ -24,6 +24,13 @@ local specs = {
 		header = "uapi/linux/signal.h",
 		prefix = "SIG",
 		module_name = "signal",
+	},
+	{
+		-- kernel-internal header, extracted via grep (not preprocessor)
```
lneto: why? it's not clear to me why you can't use the preprocessor and need to introduce this new "mode".

Suryansh-Dey: `linux/sched.h` is a kernel-internal header (not under `uapi/`), so it can't be preprocessed with cc -E -dM in the host environment. It pulls in other internal headers like `asm/current.h` that aren't available outside the full kernel build tree. 
```bash
~/dev/GSOC/lunatik task-phased-v2
❮ cc -E -dM -I/lib/modules/$(uname -r)/build/include /lib/modules/$(uname -r)/build/include/linux/sched.h
/lib/modules/6.12.73-1-lts/build/include/linux/sched.h:12:10: fatal error: asm/current.h: No such file or directory
   12 | #include <asm/current.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```

lib/lualinux.c
```diff
@@ -302,7 +284,6 @@ static int lualinux_errname(lua_State *L)
 
 static const lunatik_namespace_t lualinux_flags[] = {
 	{"stat", lualinux_stat},
-	{"task", lualinux_task},
```
lneto, line 305: when removing this you should adjust the code base. Did you run the tests and examples? didn't it break anything?

Suryansh-Dey: `examples/shared.lua` and `examples/echod/daemon.lua`
 both have local task = linux.task. I'll update these to use require("linux.task").

That said, neither file actually uses the task variable — they both call linux.schedule(100) without a state argument, so there's no functional breakage. But I'll clean them up to reflect the new access pattern regardless.

I don't have a kernel test environment set up to run the full examples yet — would you like me to fix these two files and re-push?

Suryansh-Dey:
I did run the relevant tests and nothing seem to break:
```bash
~/dev/GSOC/lunatik task-phased-v2 [!]
❯ sudo lunatik reload
  sudo lunatik status
[sudo] password for suryansh:
lunatik is loaded
luadevice is loaded
lualinux is loaded
luanotifier is loaded
luasocket is loaded
luarcu is loaded
luathread is loaded
luafib is loaded
luadata is loaded
luaprobe is loaded
luasyscall is loaded
luaxdp is loaded
luafifo is loaded
luaxtable is loaded
luanetfilter is loaded
luacompletion is loaded
luacrypto_shash is loaded
luacrypto_skcipher is loaded
luacrypto_aead is loaded
luacrypto_rng is loaded
luacrypto_comp is loaded
luacpu is loaded
luahid is loaded
luasignal is loaded
luabyteorder is loaded
lunatik_run is loaded
```
```bash
~/dev/GSOC/lunatik task-phased-v2 [!]
❯ echo 'local task = require("linux.task"); print(task.INTERRUPTIBLE, task.UNINTERRUPTIBLE, task.KILLABLE, task.IDLE); return "task OK"' | sudo lunatik
  sudo dmesg | tail -5
Lunatik 4.1  Copyright (C) 2023-2026 Ring Zero Desenvolvimento de Software LTDA.
> task OK
> [  610.149922] 2
[  610.149924]
[  610.149925] 258
[  610.149927]
[  610.149928] 1026
```
```bash
~/dev/GSOC/lunatik task-phased-v2 [!]
❯ echo 'local linux = require("linux"); local task = require("linux.task"); linux.schedule(100, task.INTERRUPTIBLE); return "schedule OK"' | sudo lunatik

Lunatik 4.1  Copyright (C) 2023-2026 Ring Zero Desenvolvimento de Software LTDA.
> schedule OK
> ⏎
```
```bash
~/dev/GSOC/lunatik task-phased-v2 [!]
❯ echo 'local linux = require("linux"); print(linux.stat.IRUGO); return "stat OK"' | sudo lunatik
  sudo dmesg | tail -1
Lunatik 4.1  Copyright (C) 2023-2026 Ring Zero Desenvolvimento de Software LTDA.
> stat OK
> [  710.751080] 292
```
Got 292 as expected.

What should I do with `examples/shared.lua` and `example/echod/daemon.lua`? Remove the line to get rid of linter error?

configure.lua
```diff
@@ -61,6 +68,70 @@ local function collect_constants(cpp_output, prefix)
...
+
+	table.sort(macro_names)
+	return constants, macro_names
+end
+
+-- Resolve a value to a numeric result. Handles:
+-- 1. Numeric literals (decimal, hex)
+-- 2. References to other macros (resolved recursively)
+-- 3. OR expressions like (MACRO_A | MACRO_B)
+local function resolve_numeric(value, constants, seen)
```
sneaky-potato: resolve_numeric largely duplicates the logic already present in resolve_value. Both recursively resolve macro references and handle chained defines. Rather than introducing a second function, consider extending resolve_value to handle OR expressions and numeric literals directly?

lneto:
> I did run the relevant tests and nothing seem to break:

you did break the examples, right? we shouldn't break anything.

> What should I do with `examples/shared.lua` and `example/echod/daemon.lua`? Remove the line to get rid of linter error?

you should adjust the examples accordingly.

Suryansh-Dey (commented):
I have moved 
```lua
if tonumber(value) then return value end
```
to 
```lua
	local n = tonumber(value)
	if n then return string.format("0x%08x", n) end
```
To avoid mixing octal and hex. Or should I keep it octal?  
(Git diff seem to glitch a bit there so better you look at the new code instead of looking at the diff)

lib/lualinux.c
lneto: > `examples/shared.lua` and `examples/echod/daemon.lua` both have local task = linux.task. I'll update these to use require("linux.task").
> 
> That said, neither file actually uses the task variable — they both call linux.schedule(100) without a state argument, so there's no functional breakage. But I'll clean them up to reflect the new access pattern regardless.

Removing the require line should be enough then.

> I don't have a kernel test environment set up to run the full examples yet — would you like me to fix these two files and re-push?

I think you should firstly set your test environment before submitting any PR. How are you testing your patch?

configure.lua
lneto: that's not a reason to change this script to use grep instead. We just need to figure the right way to use cpp with this file (probably fixing the include path) or stick with uapi.

lib/lualinux.c
Suryansh-Dey: Earlier I was running `make configure` to test autogen manually. The issue is in master branch, which I didn't touch, I am getting same FAILs as my PR branch:
<details>
<summary>
sudo dmesg | tail -n 100
</summary>

```bash
[31638.400943] FAIL     AEAD AES-128-GCM decrypt with input data too short for tag      /lib/modules/lua/tests/crypto/aead.lua:83: Error code should be 'EBADMSG', got: EBADMSG
               stack traceback:
                /lib/modules/lua/crypto/aead.lua:91: in function 'crypto.aead.decrypt'
                [C]: in global 'pcall'
                /lib/modules/lua/tests/crypto/aead.lua:81: in function </lib/modules/lua/tests/crypto/aead.lua:76>
                [C]: in global 'pcall'
                /lib/modules/lua/util.lua:43: in local 'test'
                /lib/modules/lua/tests/crypto/aead.lua:76: in main chunk
                [C]: in ?
               stack traceback:
                /lib/modules/lua/util.lua:47: in local 'test'
                /lib/modules/lua/tests/crypto/aead.lua:76: in main chunk
                [C]: in ?
[31638.401006] FAIL     AEAD AES-128-GCM decrypt with authentication failure    /lib/modules/lua/tests/crypto/aead.lua:93: Error code should be 'EBADMSG', got: EBADMSG
               stack traceback:                                                                                                                       /lib/modules/lua/crypto/aead.lua:91: in function 'crypto.aead.decrypt'
                [C]: in global 'pcall'
                /lib/modules/lua/tests/crypto/aead.lua:91: in function </lib/modules/lua/tests/crypto/aead.lua:86>
                [C]: in global 'pcall'
                /lib/modules/lua/util.lua:43: in local 'test'
                /lib/modules/lua/tests/crypto/aead.lua:86: in main chunk
                [C]: in ?
               stack traceback:
                /lib/modules/lua/util.lua:47: in local 'test'
                /lib/modules/lua/tests/crypto/aead.lua:86: in main chunk
                [C]: in ?
[31638.416347] PASS     COMP compress empty string (error)
[31638.416391] PASS     COMP decompress empty string (error)
[31638.416433] PASS     COMP compress                                                                                                 [31638.416479] PASS     COMP decompress
[31638.416509] PASS     COMP decompress with larger buffer
[31638.416576] FAIL     COMP decompress with too small buffer (expect error)    /lib/modules/lua/tests/crypto/comp.lua:54: Error code should be 'EINVAL', got: EINVAL
               stack traceback:
                [C]: in global 'pcall'
                /lib/modules/lua/tests/crypto/comp.lua:52: in function </lib/modules/lua/tests/crypto/comp.lua:48>
                [C]: in global 'pcall'
                /lib/modules/lua/util.lua:43: in local 'test'
                /lib/modules/lua/tests/crypto/comp.lua:48: in main chunk
                [C]: in ?
               stack traceback:
                /lib/modules/lua/util.lua:47: in local 'test'
                /lib/modules/lua/tests/crypto/comp.lua:48: in main chunk
                [C]: in ?
[31638.431813] PASS     HKDF test vectors from RFC 5869, Appendix A
[31638.431918] PASS     HKDF-Expand-Label examples from https://quic.xargs.org/#client-initial-keys-calc
[31638.451526] PASS     RNG generate 32 bytes
[31638.456432] PASS     RNG generate 0 bytes (error)                                                                                  [31638.463149] PASS     RNG reset without seed
[31638.469705] PASS     RNG reset with seed
[31638.474583] PASS     RNG info
[31638.479770] PASS     RNG getbytes 0 bytes (error)
[31638.484681] PASS     RNG getbytes (16 bytes)
[31638.489586] PASS     RNG getbytes (32 bytes)
[31638.494882] PASS     RNG getbytes (different from previous)
[31638.511355] PASS     crypto_shash.new and digestsize
[31638.511448] PASS     crypto_shash:digest (single-shot)
[31638.511481] PASS     crypto_shash:init, update, final (multi-part)
[31638.511513] PASS     crypto_shash:finup
[31638.511544] PASS     crypto_shash:export and import functionality
[31638.526737] PASS     SKCIPHER AES-128-CBC encrypt
[31638.526762] PASS     SKCIPHER AES-128-CBC decrypt
[31638.526772] PASS     SKCIPHER AES-128-CBC ivsize and blocksize
[31638.526845] FAIL     SKCIPHER AES-128-CBC setkey with invalid key length     /lib/modules/lua/tests/crypto/skcipher.lua:43: Error code should be 'EINVAL', got: EINVAL
               stack traceback:
                [C]: in global 'pcall'
                /lib/modules/lua/tests/crypto/skcipher.lua:41: in function </lib/modules/lua/tests/crypto/skcipher.lua:39>
                [C]: in global 'pcall'
                /lib/modules/lua/util.lua:43: in local 'test'
                /lib/modules/lua/tests/crypto/skcipher.lua:39: in main chunk
                [C]: in ?
               stack traceback:
                /lib/modules/lua/util.lua:47: in local 'test'
                /lib/modules/lua/tests/crypto/skcipher.lua:39: in main chunk
                [C]: in ?
[31638.526905] PASS     SKCIPHER AES-128-CBC encrypt with incorrect IV length
[31638.526988] PASS     SKCIPHER AES-128-CBC decrypt with incorrect IV length
[31638.527049] FAIL     SKCIPHER AES-128-CBC encrypt with data not multiple of blocksize        /lib/modules/lua/tests/crypto/skcipher.lua:70: Error code should be 'EINVAL', got: EINVAL                                                                                                  stack traceback:
                [C]: in global 'pcall'
                /lib/modules/lua/tests/crypto/skcipher.lua:68: in function </lib/modules/lua/tests/crypto/skcipher.lua:64>
                [C]: in global 'pcall'
                /lib/modules/lua/util.lua:43: in local 'test'
                /lib/modules/lua/tests/crypto/skcipher.lua:64: in main chunk
                [C]: in ?
               stack traceback:
                /lib/modules/lua/util.lua:47: in local 'test'
                /lib/modules/lua/tests/crypto/skcipher.lua:64: in main chunk
                [C]: in ?
[31638.527123] FAIL     SKCIPHER AES-128-CBC decrypt with data not multiple of blocksize        /lib/modules/lua/tests/crypto/skcipher.lua:79: Error code should be 'EINVAL', got: EINVAL
               stack traceback:
                [C]: in global 'pcall'                                                                                                                /lib/modules/lua/tests/crypto/skcipher.lua:77: in function </lib/modules/lua/tests/crypto/skcipher.lua:73>
                [C]: in global 'pcall'
                /lib/modules/lua/util.lua:43: in local 'test'
                /lib/modules/lua/tests/crypto/skcipher.lua:73: in main chunk
                [C]: in ?
               stack traceback:
                /lib/modules/lua/util.lua:47: in local 'test'
                /lib/modules/lua/tests/crypto/skcipher.lua:73: in main chunk
                [C]: in ?
```
</details>
And the fails are like `Error code should be 'EBADMSG', got: EBADMSG`.
Please correct me if I am doing something wrong

configure.lua
Suryansh-Dey: But `TASK_INTERRUPTIBLE` and other `TASK_*` flags are strictly kernel-internal definitions and are not exposed in any `uapi/ headers`.  
Please guide me if I am missing something.

lneto:
@Suryansh-Dey are you still working on this? please, let me know.

Suryansh-Dey:
I have exams tomorrow and day after tomorrow, I will get back to it on 28th feb.

lneto:
> I have exams tomorrow and day after tomorrow, I will get back to it on 28th feb.

good luck! no worries, just wanted to check if we should keep this PR opened.. thanks!

configure.lua
```diff
@@ -52,39 +61,47 @@ local function collect_constants(cpp_output, prefix)
...
 	end
 
 	table.sort(macro_names)
 	return constants, macro_names
 end
 
-local function resolve_value(value, constants, prefix, module_name, seen)
-	if tonumber(value) then return value end
+local function resolve_value(value, constants, seen)
+	if tonumber(value) then return tonumber(value) end
```
lneto, line 79: Why?

configure.lua
```diff
@@ -52,39 +61,47 @@ local function collect_constants(cpp_output, prefix)
...
+	local inner = value:match("^%((.+)%)$")
+	if inner and inner:match("^[%w_%s|]+$") then
+		local result = 0
+		for component in inner:gmatch("[%w_]+") do
+			local resolved = resolve_value(constants[component] or component, constants, seen)
+			if not resolved then return nil end
+			result = result | resolved
+		end
+		return result
+	end
```
lneto: Did you test it, is it working?

configure.lua
```diff
@@ -52,39 +61,47 @@ local function collect_constants(cpp_output, prefix)
...
 end
 
 local function write_constants(file, module, spec, constants, macro_names)
 	for _, macro in ipairs(macro_names) do
 		local field_name = macro:gsub("^" .. spec.prefix, "")
-		local resolved_value = resolve_value(constants[macro], constants, spec.prefix, spec.module_name)
+		local resolved_value = resolve_value(constants[macro], constants)
 		if resolved_value then
-			file:write(string.format('%s["%s"]\t= %s\n', spec.module_name, field_name, resolved_value))
+			file:write(string.format('%s["%s"]\t= 0x%08x\n', spec.module_name, field_name, resolved_value))
```
lneto: Why this?

lneto:
@sneaky-potato can you please review it? It's your baby ;-)

configure.lua
Suryansh-Dey: To keep all the flags in hex representation.

configure.lua
Suryansh-Dey: Because write_constants now uses string.format('0x%08x', resolved_value) which requires an actual number, not a string. The original returned the raw string "0x00000001" and wrote it with %s.  
We need it for:
- Format it consistently as 0x%08x
- Use the bitwise OR operator result | resolved for composite constants like TASK_KILLABLE

configure.lua
Suryansh-Dey: Yes, make configure generates correct flag values in task.lua. Rest I asked @sneaky-potato at matrix for guide in relevant testing

configure.lua
```diff
@@ -52,39 +61,47 @@ local function collect_constants(cpp_output, prefix)
 	local constants = {}
 	local macro_names = {}
 
-	for macro, literal in cpp_output:gmatch("#define%s+(" .. prefix .. "%w+)%s+(%S+)") do
-		constants[macro] = literal
-		table.insert(macro_names, macro)
+	for line in cpp_output:gmatch("[^\n]+") do
+		local macro, value = line:match("#define%s+(" .. prefix .. "[%u%d_]+)%s+(.+)%s*$")
```
sneaky-potato, line 64: why changing the regex pattern? I think the original regex is sufficient? Let's not complicate the pattern if it is not required.

configure.lua
```diff
@@ -52,39 +61,47 @@ local function collect_constants(cpp_output, prefix)
...
 end
 
-local function resolve_value(value, constants, prefix, module_name, seen)
-	if tonumber(value) then return value end
+local function resolve_value(value, constants, seen)
+	if tonumber(value) then return tonumber(value) end
+
+	-- OR expression: (A | B | ...) — only if it contains just identifiers and |
+	local inner = value:match("^%((.+)%)$")
+	if inner and inner:match("^[%w_%s|]+$") then
```
sneaky-potato: Why matching with underscore here in this regex? I don't understand why did you remove the prefix usage from here to match the macro name? For instance what if future macros do not have underscores.

EDIT: Is it because your OR logic supports macros that are composed of flags from different "families" or headers?

configure.lua
```diff
@@ -52,39 +61,47 @@ local function collect_constants(cpp_output, prefix)
...
+			constants[macro] = value
+			table.insert(macro_names, macro)
+		end
 	end
 
 	table.sort(macro_names)
 	return constants, macro_names
 end
 
-local function resolve_value(value, constants, prefix, module_name, seen)
```
sneaky-potato, line 64: lets use the prefix and not hard-code the regex to match with underscores here?

configure.lua
Suryansh-Dey: The change to [%u%d_]+ was to explicitly filter out two lowercase offsets (TASK_stack_canary and TASK_threadsp) that were leaking in from asm-offsets.h due to the broad TASK_ prefix. This is safer.

configure.lua
```diff
@@ -37,7 +42,11 @@ if not KERNEL or not INCLUDE or not MODULES then
 end
 
 local CC = os.getenv("CC") or "cc"
-local CPP = string.format("%s -E -dM -I%s", CC, INCLUDE)
+local BUILD = INCLUDE:gsub("/include$", "")
+local ARCH_PIPE <close> = io.popen("uname -m | sed 's/x86_64/x86/;s/aarch64/arm64/'")
```
sneaky-potato: consider receiving this constant from the Makefile itself? This pattern has been maintained for other constants as well because Makefile already computes paths and all.

Suryansh-Dey:
Done, I will squash the commits at the end. Should I move on for moving stas constants and all?

configure.lua
sneaky-potato: Please tell me what issues are being observed in the old regex pattern here? Is it dropping some defines?

Suryansh-Dey: no, it's over-matching.  
With the TASK_ prefix, the preprocessor output from `asm-offsets.h` includes defines like `TASK_stack_canary` and `TASK_threadsp` that have lowercase characters. These aren't task/state flags and they leak in because %w+ matches lowercase letters too.

Changing to [%u%d_]+ filters these out since all the actual task state constants (TASK_RUNNING, TASK_INTERRUPTIBLE, etc.) use only uppercase letters, digits, and underscores.

configure.lua
```diff
@@ -52,39 +60,53 @@ local function collect_constants(cpp_output, prefix)
...
 	return constants, macro_names
 end
 
-local function resolve_value(value, constants, prefix, module_name, seen)
-	if tonumber(value) then return value end
+local function resolve_value(value, prefix, constants, seen)
+	if tonumber(value) then return tonumber(value) end
+
+	-- OR expression: (A | B | ...)
+	local inner = value:match("^%((.+)%)$")
```
sneaky-potato: `seen` table must be defined before this block, right?

configure.lua
```diff
@@ -52,39 +60,53 @@ local function collect_constants(cpp_output, prefix)
...
+	if inner then
+		local is_or_expr = true
+		local result = 0
+		for token in inner:gmatch("[^%s|]+") do
+			-- only resolve if token is a number or it starts with the expected prefix
+			if not (tonumber(token) or string.sub(token, 1, #prefix) == prefix) then
+				is_or_expr = false
+				break
+			end
+			local resolved = resolve_value(constants[token] or token, prefix, constants, seen)
```
sneaky-potato: Currently, if seen is nil at the top level (first call), each component initializes its own independent seen in the calls of resolve_value. For it to work correctly, seen must be defined at the top of this function I think.

configure.lua
```diff
@@ -52,39 +60,53 @@ local function collect_constants(cpp_output, prefix)
...
 end
 
 local function write_constants(file, module, spec, constants, macro_names)
 	for _, macro in ipairs(macro_names) do
 		local field_name = macro:gsub("^" .. spec.prefix, "")
-		local resolved_value = resolve_value(constants[macro], constants, spec.prefix, spec.module_name)
+		local resolved_value = resolve_value(constants[macro], spec.prefix, constants)
 		if resolved_value then
-			file:write(string.format('%s["%s"]\t= %s\n', spec.module_name, field_name, resolved_value))
+			file:write(string.format('%s["%s"]\t= 0x%08x\n', spec.module_name, field_name, resolved_value))
```
sneaky-potato: what's the justification for hex over decimal? If there isn't one, I think %d or %s is simpler and avoids truncation issue. (Any value above 0xFFFFFFFF will get truncated with 0x%08x)

configure.lua
Suryansh-Dey: Hmm, It does look a bit dangerous. sorry

configure.lua
Suryansh-Dey: Ok, reverted back to decimal. I thought hex is more standard representation for flags

configure.lua
Suryansh-Dey: Yes, I have added some safety checks without compromising performance.

configure.lua
```diff
@@ -52,37 +60,54 @@ local function collect_constants(cpp_output, prefix)
...
-local function resolve_value(value, constants, prefix, module_name, seen)
-	if tonumber(value) then return value end
-
-	if not constants[value] then return nil end
-
+local function resolve_value(value, prefix, constants, seen)
+	if tonumber(value) then return tonumber(value) end
 	seen = seen or {}
-	-- recursion check: avoid cyclic defines
-	if seen[value] then return nil end
```
sneaky-potato, line 71: Please don't move this. Keep this block above the main loop only, it will allow the method to return early if the macro has been seen.

Suryansh-Dey: ok

Makefile
```diff
@@ -12,6 +12,7 @@ MODULES_BUILD_PATH ?= ${BTF_INSTALL_PATH}
 MODULES_INSTALL_PATH := ${MODULES_RELEASE_PATH}/kernel
 SCRIPTS_INSTALL_PATH := ${MODULES_PATH}/lua
 INCLUDE_PATH := ${MODULES_BUILD_PATH}/include
+ARCH ?= $(shell uname -m | sed 's/x86_64/x86/;s/aarch64/arm64/')
```
sneaky-potato: One improvement: instead of piping to another instance of sed, I think we can combine the patterns
```suggestion
ARCH ?= $(shell uname -m | sed -e 's/x86_64/x86/' -e 's/aarch64/arm64/')
```

But then we also could use Makefile builtin functions instead of sed here, like so
```suggestion
KERNEL_ARCH ?= ${shell uname -m}
ARCH ?= $(subst x86_64,x86,$(subst aarch64,arm64,$(KERNEL_ARCH)))
```

(Naming can be better for the variables)
What do you think?

Suryansh-Dey: Okay second one would be more efficient since we don't spawn an extra process for `sed`.

configure.lua
```diff
@@ -17,7 +17,7 @@ local specs = {
 		module_name = "eth",
 	},
 	{
-		header = "uapi/linux/stat.h",
```
Suryansh-Dey, line 19: Switching to `linux/stat.h` ensures the CPP sees both the UAPI base constants and the internal compound ones like `S_IRUGO = (S_IRUSR|S_IRGRP|S_IROTH)`, which our resolve_value function then resolves into their final numeric values.

lib/lualinux.c
lneto: see this discussion here.. it sounds like a regression on master.. https://github.com/luainkernel/lunatik/pull/465/changes#r2901813814

it's a good thing for you to look at and proper fix it.. I mentioned the original commit that might have introduced this problem on the channel..

Suryansh-Dey: I have described the issue to sneaky-potato. He said he will fix it.

Suryansh-Dey:
All the tests are passing now. If the PR is ready to merge and no changes needed I can squash the commits. Please let me know

configure.lua
```diff
@@ -52,37 +60,55 @@ local function collect_constants(cpp_output, prefix)
...
+			if not (tonumber(token) or string.sub(token, 1, #prefix) == prefix) then
+				is_or_expr = false
+				break
+			end
+			local resolved = resolve_value(token, prefix, constants, seen)
+			if not resolved then return nil end
+			result = result | resolved
+		end
+		if is_or_expr then return result end
 	end
```
sneaky-potato: Lets make this block a bit simpler and easier to read
Could you double check if this works? It does seem to work.
```suggestion
	local clean = value:match("^%((.+)%)$") or value

	-- OR expression: (A | B | ...)
    if clean:find("|") then
        local result = 0
        for token in clean:gmatch("[^%s|]+") do
            local resolved = resolve_value(token, prefix, constants, seen)
            if not resolved then return nil end
            result = result | resolved
        end
        return result
    end
```

Suryansh-Dey: Yes it does output the same. Because the CPP output only contains #defines matching the prefix (since collect_constants already filters by prefix), we don't need to filter it here.

configure.lua
```diff
@@ -32,12 +38,14 @@ local function exit(msg)
...
-if not KERNEL or not INCLUDE or not MODULES then
-	exit("usage: lua5.4 " .. SCRIPT .. " <KERNEL> <INCLUDE> <MODULES>")
+if not KERNEL or not INCLUDE or not MODULES or not ARCH then
+	exit("usage: lua5.4 " .. SCRIPT .. " <KERNEL> <INCLUDE> <MODULES> <ARCH>")
 end
 
 local CC = os.getenv("CC") or "cc"
-local CPP = string.format("%s -E -dM -I%s", CC, INCLUDE)
+local BUILD = INCLUDE:gsub("/include$", "")
+local CPP = string.format("%s -E -dM -I%s -I%s/include/generated -I%s/arch/%s/include -I%s/arch/%s/include/generated -include linux/kconfig.h",
```
sneaky-potato: I feel we should accept `BUILD` as a command line argument instead of `INCLUDE` to the script. 
`INCLUDE` would then become `BUILD .. "/include"`
This way we can construct the other paths as well instead of relying on string manipulation.
 
If you notice in Makefile we already have a variable `${MODULES_BUILD_PATH}`
What do you think?

Suryansh-Dey: as you wish

configure.lua
```diff
@@ -52,37 +60,49 @@ local function collect_constants(cpp_output, prefix)
...
-		return string.format('%s["%s"]', module_name, stripped)
-	end
+	if tonumber(value) then return tonumber(value) end
+	local clean = value:match("^%((.+)%)$") or value
+
+	-- OR expression: (A | B | ...)
+    if clean:find("|") then
+        local result = 0
+        for token in clean:gmatch("[^%s|]+") do
+            local resolved = resolve_value(token, prefix, constants, seen)
```
sneaky-potato, line 86: Here we are not setting `seen[value] = true` right before this `resolve_value` call. This might cause problems right?

Makefile
```diff
@@ -12,6 +12,8 @@ MODULES_BUILD_PATH ?= ${BTF_INSTALL_PATH}
 MODULES_INSTALL_PATH := ${MODULES_RELEASE_PATH}/kernel
 SCRIPTS_INSTALL_PATH := ${MODULES_PATH}/lua
 INCLUDE_PATH := ${MODULES_BUILD_PATH}/include
```
sneaky-potato: This is not used anywhere now, we can safely remove this

Makefile
```diff
@@ -12,6 +12,8 @@ MODULES_BUILD_PATH ?= ${BTF_INSTALL_PATH}
...
+KERNEL_ARCH ?= ${shell uname -m}
+ARCH ?= $(subst x86_64,x86,$(subst aarch64,arm64,$(KERNEL_ARCH)))
```
sneaky-potato: I am having mixed feelings here. We already do similar logic to find architecture here in [KBuild](https://github.com/luainkernel/lunatik/blob/d765e9a796a80428524e3dbc4c7bd8acef2457c8/Kbuild#L16-L17). Maybe export this as environment variable so that both the Kbuild and configure.lua simply "consume" instead of recomputing it.
@lneto what do you think?

lneto: We should never reinvent the wheel.. BTW, we should care about cross compiling; particularly, this should work with OpenWRT feed as well..

configure.lua
lneto: it seems odd to me to convert it twice.. perhaps you should use tonumber on the caller and if type is number here, return it.. not sure.. but you should clean up your logics.. it seems very convoluted to me..

configure.lua
lneto: > as you wish

don't you agree with the suggestions? if so, please make your point..

Suryansh-Dey: No, I do agree. I also think it is cleaner.

configure.lua
Suryansh-Dey: Actually, this won't cause any problems.  
The `value` at this specific point is the literal OR expression string itself. We only need to track the actual macro names in the `seen` table to detect cycles. Since each individual token inside the OR expression (like TASK_A and TASK_B) gets passed back into resolve_value recursively, they will hit the `seen[value] = true` block at the bottom when they are resolved as macro references. If an OR expression contains a macro that forms a cycle, it will still be caught perfectly.  
Please correct me If I missed something

Makefile
Suryansh-Dey:
```
MODULES_INSTALL_PATH := ${MODULES_RELEASE_PATH}/kernel
SCRIPTS_INSTALL_PATH := ${MODULES_PATH}/lua
```
They get used later in the makefile below, so we can't remove them.  
But yes I should have removed line 14. My bad.

Makefile
Suryansh-Dey: What should I do then?

lneto: I think @sneaky-potato's idea is way to do it, just make sure it works on OpenWRT feed. It must be tested there as well.

Suryansh-Dey: Could you please test if works on OpenWRT feed. I am not sure of that thing, how it works.

lneto: I can help if you would like to do it, but not do it for you. For merging this PR, we need to make sure we don't break it. For building it you can just use our OpenWRT feed pointing to your commit. You probably will need to adjust the makefile there.

PR #445, opened 2026-02-25
Issue #403 : Add safe unregister methods to luanetfilter, luahid, and luaxdp

### LINKED ISSUE #403: add unregister methods to luanetfilter, luaxdp, luaxtables, luahid

ayushactiveat:
Fixes #403

this adds the explicit unregister() methods to luanetfilter, luahid, and luaxdp.

### DIFF
--- a/lib/luahid.c
+++ b/lib/luahid.c
@@ -70,8 +70,27 @@ static const luaL_Reg luahid_lib[] = {
 	{NULL, NULL},
 };
 
+static int luahid_unregister(lua_State *L)
+{
+	lunatik_object_t *object = lunatik_checkobject(L, 1);
+	luahid_t *hid = (luahid_t *)object->private;
+
+	lunatik_lock(object);
+	if (hid->registered) {
+		hid_unregister_driver(&hid->driver);
+		hid->registered = false;
+	}
+	lunatik_unlock(object);
+
+	if (lunatik_isruntime(L, hid))
+		lunatik_unregisterobject(L, object);
+
+	return 0;
+}
+
 static const luaL_Reg luahid_mt[] = {
 	{"__gc", lunatik_deleteobject},
+	{"unregister", luahid_unregister},
 	{NULL, NULL},
 };
 
--- a/lib/luanetfilter.c
+++ b/lib/luanetfilter.c
@@ -132,8 +132,27 @@ static unsigned int luanetfilter_hook(void *priv, struct sk_buff *skb, const str
 	return luanetfilter_docall(luanf, skb);
 }
 
+static int luanetfilter_unregister(lua_State *L)
+{
+	lunatik_object_t *object = lunatik_checkobject(L, 1);
+	luanetfilter_t *nf = (luanetfilter_t *)object->private;
+
+	lunatik_lock(object);
+	if (nf->nfops.hook) {
+		nf_unregister_net_hook(&init_net, &nf->nfops);
+		nf->nfops.hook = NULL;
+	}
+	lunatik_unlock(object);
+
+	if (lunatik_isruntime(L, nf))
+		lunatik_unregisterobject(L, object);
+
+	return 0;
+}
+
 static const luaL_Reg luanetfilter_mt[] = {
 	{"__gc", lunatik_deleteobject},
+	{"unregister", luanetfilter_unregister},
 	{NULL, NULL}
 };
 
@@ -198,10 +217,14 @@ static void luanetfilter_release(void *private)
 {
 	luanetfilter_t *nf = (luanetfilter_t *)private;
 	lunatik_object_t *runtime = nf->runtime;
+
 	if (runtime == NULL)
 		return;
 
-	nf_unregister_net_hook(&init_net, &nf->nfops);
+	if (nf->nfops.hook) {
+		nf_unregister_net_hook(&init_net, &nf->nfops);
+		nf->nfops.hook = NULL;
+	}
 	luadata_detach(runtime, nf, skb);
 	lunatik_putobject(runtime);
 	nf->runtime = NULL;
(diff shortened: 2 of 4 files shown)


lneto (changes requested):
please, mind our code style.. tabs, etc

lib/luahid.c
```diff
@@ -70,8 +70,29 @@ static const luaL_Reg luahid_lib[] = {
...
+	lunatik_lock(object);
+	runtime = hid->runtime;
+	if (hid->registered) {
+		hid_unregister_driver(&hid->driver);
+		hid->registered = false;
+	}
+	lunatik_unlock(object);
+
+	if (runtime != NULL && lunatik_toruntime(L) == runtime)
+		lunatik_unregisterobject(L, object);
```
lneto: why doing this outside the lock?

lib/luanetfilter.c
```diff
@@ -132,8 +132,31 @@ static unsigned int luanetfilter_hook(void *priv, struct sk_buff *skb, const str
...
+        luanetfilter_t *nf = (luanetfilter_t *)object->private;
+        lunatik_object_t *runtime;
+
+        lunatik_lock(object);
+        runtime = nf->runtime;
+        if (runtime != NULL) {
+                nf_unregister_net_hook(&init_net, &nf->nfops);
+                luadata_detach(runtime, nf, skb);
+                lunatik_putobject(runtime);
+                nf->runtime = NULL;
```
lneto: shouldn't we do this on hid as well?

lneto (commented):
did you test your patch?

lib/luanetfilter.c
ayushactiveat: missed that, I'll do that right away

ayushactiveat:
> did you test your patch?

i tried a full local full compilati, but due to limitation on my machine relied purely on logic, i'll just spin a proper Lm now, i'll definitely make sure to do it locally first from now on

lib/luahid.c
ayushactiveat: i thought lunatik_unreg. may trigger the gc  and cause deadlock if lock was acquired. but yes ig that introduces a race condition. I'll just move it back

ayushactiveat:
hi @lneto , I've made the changes and tested the patch, let me know if you need some other change..

lib/luahid.c
lneto: I believe you were right on the first place; why did you change the original code? I don't see the race condition you mentioned..

btw, we should move this macro to lunatik.h and just use it.. https://github.com/luainkernel/lunatik/blob/master/lib/luanotifier.c#L187

also, take a look at notifier:stop(), it should be a similar pattern here..

lneto:
> > did you test your patch?
> 
> i tried a full local full compilati, but due to limitation on my machine relied purely on logic, i'll just spin a proper Lm now, i'll definitely make sure to do it locally first from now on

please, don't do this again. It's not ok to submit a patch without testing it.

lib/luaxdp.c
```diff
@@ -227,6 +227,7 @@ static const luaL_Reg luaxdp_lib[] = {
 #if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 4, 0))
 	{"attach", luaxdp_attach},
 	{"detach", luaxdp_detach},
+	{"unregister", luaxdp_detach},
```
lneto: I don't think we need this alias

lib/luanetfilter.c
```diff
@@ -132,8 +132,30 @@ static unsigned int luanetfilter_hook(void *priv, struct sk_buff *skb, const str
...
+static int luanetfilter_unregister(lua_State *L)
+{
+	lunatik_object_t *object = lunatik_checkobject(L, 1);
+	luanetfilter_t *nf = (luanetfilter_t *)object->private;
+	lunatik_object_t *runtime;
+
+	lunatik_lock(object);
+	runtime = nf->runtime;
+	if (runtime != NULL) {
+		nf_unregister_net_hook(&init_net, &nf->nfops);
```
lneto: what will happen when the object is collected? this function will be called twice? what happened on your tests and how did you test it?

ayushactiveat:
> > > did you test your patch?
> > 
> > 
> > i tried a full local full compilati, but due to limitation on my machine relied purely on logic, i'll just spin a proper Lm now, i'll definitely make sure to do it locally first from now on
> 
> please, don't do this again. It's not ok to submit a patch without testing it.

sure thing

lib/luanetfilter.c
ayushactiveat: yes and earlier i did a compile only, well now, I've refactored both luanetfilter and luahid to mirror the notifier-stop .  when object is collected, the release function checks flags and bypasses the second unregister call.

ayushactiveat:
hi @lneto , let me knowb if this is fine and regarding the testing - i have compile tested this patch against the headers , am currently building a qemu environ... to preoperly test the gc cycle. i wanted to push the corrected logic for you to see, but I will follow up with the live runtime test results before asking for final merge

lneto:
> hi @lneto , let me knowb if this is fine and regarding the testing - i have compile tested this patch against the headers , am currently building a qemu environ... to preoperly test the gc cycle. i wanted to push the corrected logic for you to see, but I will follow up with the live runtime test results before asking for final merge

please, only submit patches for review when you have properly test them. It's not enough to compile. You need to actual validate your changes. Please, provide the script you have used as test cases.

lib/luanetfilter.c
```diff
@@ -198,10 +217,14 @@ static void luanetfilter_release(void *private)
 {
 	luanetfilter_t *nf = (luanetfilter_t *)private;
 	lunatik_object_t *runtime = nf->runtime;
+
```
lneto, line 220: why changing this line? please, restrict your patch to relevant changes; avoid cosmetic ones. Always double check it.. when you just throw non-relevant changes or even worse when you don't run and test your changes, you are not being respectful of reviewers time. Please, keep this in mind.

PR #455, opened 2026-03-03
add luaskb: pointer class wrapping struct sk_buff 

lneto:


lib/luaskb.c
```diff
@@ -0,0 +1,187 @@
...
+{
+	lunatik_object_t *object = lunatik_createobject(&luaskb_class, 0, false, false);
+
+	if (!IS_ERR_OR_NULL(object))
+		object->private = skb;
+	return object;
+}
+EXPORT_SYMBOL(luaskb_create);
+
+void luaskb_reset(lunatik_object_t *object, struct sk_buff *skb)
```
lneto: we can move it to a inline on .h

lib/luaskb.c
```diff
@@ -0,0 +1,187 @@
...
+	if (mac) {
+		offset = skb_mac_offset(skb);
+		size   = skb_headlen(skb) + skb_mac_header_len(skb);
+	}
+	else {
+		offset = 0;
+		size   = skb_headlen(skb);
+	}
+
+	data = luadata_new(L, false);
```
lneto: we can "cache" the associated data object to avoid realloc'ing on every hook call..

lib/luanetfilter.c
```diff
@@ -202,7 +179,10 @@ static void luanetfilter_release(void *private)
 		return;
 
 	nf_unregister_net_hook(&init_net, &nf->nfops);
-	luadata_detach(runtime, nf, skb);
```
lneto, line 205: we should add attach/detach APIs for skb

lib/luaskb.c
```diff
@@ -0,0 +1,212 @@
+/*
```
lneto, line 1: we need to add this to LDoc.. and it should be a claude skill

lib/luanetfilter.c
```diff
@@ -222,4 +202,3 @@ module_init(luanetfilter_init);
 module_exit(luanetfilter_exit);
 MODULE_LICENSE("Dual MIT/GPL");
 MODULE_AUTHOR("Mohammad Shehar Yaar Tausif <sheharyaar48@gmail.com>");
-
```
lneto, line 225: =p

lib/luaskb.c
```diff
@@ -0,0 +1,212 @@
...
+* Provides access to Linux socket buffer fields and operations from Lua.
+* @module skb
+*/
+
+#define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+#include <linux/if_vlan.h>
+#include <linux/netdevice.h>
+#include <linux/skbuff.h>
+
+#include <lunatik.h>
```
lneto: don't need it

lib/luaskb.c
```diff
@@ -0,0 +1,212 @@
...
+}
+
+static void __exit luaskb_exit(void)
+{
+}
+
+module_init(luaskb_init);
+module_exit(luaskb_exit);
+MODULE_LICENSE("Dual MIT/GPL");
+MODULE_AUTHOR("Lourival Vieira Neto <lourival.neto@ringzero.com.br>");
```
lneto, line 249: add other authors here..

lib/luaskb.c
```diff
@@ -0,0 +1,221 @@
...
+
+static inline struct sk_buff *luaskb_check(lua_State *L, int ix)
+{
+	lunatik_object_t *object = lunatik_checkobject(L, ix);
+	luaskb_t *s = (luaskb_t *)object->private;
+	luaL_argcheck(L, s != NULL && s->skb != NULL, ix, "'skb' expected");
+	return s->skb;
+}
+
+
```
lneto, line 24: We don't do \n\n in the middle of the files.. please, fix and add it to CLAUDE.md

lib/luaskb.c
```diff
@@ -0,0 +1,221 @@
...
+#define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+#include <linux/if_vlan.h>
+#include <linux/netdevice.h>
+#include <linux/skbuff.h>
+
+#include "luadata.h"
+#include "luaskb.h"
+
+static inline struct sk_buff *luaskb_check(lua_State *L, int ix)
```
lneto: You should use PRIVATE CHECKER; add it to CLAUDE.md

lib/luaskb.c
```diff
@@ -0,0 +1,221 @@
...
+* With "mac", includes the MAC header (L2): offset = skb_mac_offset,
+* size includes skb_mac_header_len.
+* @function data
+* @tparam[opt] string layer Pass "mac" to include the L2 MAC header.
+* @treturn data A data object pointing into the skb buffer.
+* @raise Error if skb linearization fails or layer is invalid.
+*/
+static int luaskb_data(lua_State *L)
+{
+	lunatik_object_t *object = lunatik_toobject(L, 1);
```
lneto: Why not check()?

lib/luaskb.c
```diff
@@ -0,0 +1,221 @@
...
+	luaskb_t *s = (luaskb_t *)object->private;
```
lneto: Why luaskb_t here?

lneto: Please review the entire patch for this.

lib/luaskb.c
```diff
@@ -0,0 +1,231 @@
...
+	{"checksum", luaskb_checksum},
+	{"forward",  luaskb_forward},
+	{NULL, NULL}
+};
+
+static const lunatik_class_t luaskb_class = {
+	.name    = "skb",
+	.methods = luaskb_mt,
+	.release = luaskb_release,
+	.sleep   = false,
```
carloslack, line 203: do we need to set these? I thought it'd be false unless explicitly set to `true`

examples/dnsblock/nf_dnsblock.lua
```diff
@@ -13,11 +13,12 @@ local hooks = nf.inet_hooks
 local priority = nf.ip_priority
 
 local function dnsblock_hook(skb)
-	local proto = skb:getuint8(9)
-	local ihl = (skb:getuint8(0) & 0x0F)
+	local pkt = skb:data()
```
carloslack: Maybe having a short comment here, describing the empty version against the `mac` one, in addition to the comment in the C part (same for dnsdoctor)

lib/luaskb.c
```diff
@@ -0,0 +1,231 @@
...
+* @function ifindex
+* @treturn integer Interface index, or nil if not available.
+*/
+static int luaskb_ifindex(lua_State *L)
+{
+	luaskb_t *lskb = luaskb_check(L, 1);
+	luaskb_checkskb(L, lskb);
+	struct net_device *dev = lskb->skb->dev;
+
+	if (dev)
```
carloslack: Could be a ternary

lib/luaskb.c
```diff
@@ -0,0 +1,231 @@
...
+* With "mac", includes the MAC header (L2): offset = skb_mac_offset,
+* size includes skb_mac_header_len.
+* @function data
+* @tparam[opt] string layer Pass "mac" to include the L2 MAC header.
+* @treturn data A data object pointing into the skb buffer.
+* @raise Error if skb linearization fails or layer is invalid.
+*/
+static int luaskb_data(lua_State *L)
+{
+	luaskb_t *lskb = luaskb_check(L, 1);
```
carloslack, line 75: `l2skb` ?

lib/luadata.h
```diff
@@ -11,7 +11,6 @@
 #define	LUADATA_OPT_NONE	0x00
 #define	LUADATA_OPT_READONLY	0x01
 #define	LUADATA_OPT_FREE	0x02
```
carloslack, line 13: I think we could use shifting here, like
   
``` 
LUADATA_OPT_NONE = 1 << 0, 
LUADATA_OPT_READONLY = 1 << 1, 
...
```

and could be an enum IMO

examples/dnsblock/nf_dnsblock.lua
lneto: hm.. what's in your mind? I'd rather adding a new parm to let the code be more descriptive than adding a comment explaining the API..

lib/luadata.h
lneto: agreed.. but I wouldn't mix it up with the changes of this PR.. would you like to do this as a separate one? it's a good idea to handle other flags in the code base in the same way..

lib/luadata.h
```diff
@@ -27,18 +26,11 @@ static inline void luadata_close(lunatik_object_t *object)
...
-
-#define luadata_detach(runtime, obj, field)			\
-do {								\
-	lua_State *L = lunatik_getstate(runtime);		\
-	if (L != NULL) /* might be called on lunatik_stop */	\
-		lunatik_unregister(L, obj->field);		\
-	obj->field = NULL; 					\
+	obj->field = luadata_new((L), false);	\
+	lunatik_register((L), -1, obj->field);	\
+	lua_pop((L), 1);			\
```
lneto: this change doesn't make sense to me..

lib/luaskb.c
lneto: as in "lua to skb"? I don't think it is the case.. we have this pattern on writing lua version of the vars as l<foo> or l_<foo>.. I don't mind much to change it to l_skb, but I don't think we should use `2` here.. what do you think? I'm open to other naming to.. =)

lib/luaskb.c
lneto: indeed, it will be; but I think as we grow these settings (e.g., the previous @sneaky-potato work), we should be as explicit as we can.. btw, perhaps it's a good task for you to clean up our flags.. I thought on have a single flag field in the class and object structs, using bitfield or bitmask.. what do you think?

lib/luaskb.c
lneto: I don't think so, right? they are voids.. for ternary we would need return values, right? please, correct me if I'm wrong.. btw, it's your code copied in here.. hehe ;-)

lib/luaskb.c
```diff
@@ -0,0 +1,231 @@
...
+
+/***
+* Returns the network interface index.
+* @function ifindex
+* @treturn integer Interface index, or nil if not available.
+*/
+static int luaskb_ifindex(lua_State *L)
+{
+	luaskb_t *lskb = luaskb_check(L, 1);
+	luaskb_checkskb(L, lskb);
```
lneto: repetitive pattern.. we should create an inline for this

lib/luaskb.c
carloslack: no, you are right, checked, both are void

lib/luadata.h
carloslack: fair enough, [this issue ](https://github.com/luainkernel/lunatik/issues/457)

lib/luaskb.c
carloslack: but wait, for the sake of curiosity, this should work

```dev ? lua_pushinteger(L, dev->ifindex) : lua_pushnil(L);```

but it is unusual syntax indeed, I would not advocate to use this ahah

File: lib/luaskb.c (hunk 12 above)
Comment (carloslack) [reply lib/luaskb.c:203]: let me take on that!

File: lib/luaskb.c [hunk 19]
```diff
@@ -0,0 +1,244 @@
...
+}
+
+/***
+* Recomputes IP, TCP, and UDP checksums for the skb.
+* Recalculates the IP header checksum unconditionally. If the protocol
+* is TCP or UDP, also recalculates the transport-layer checksum using
+* the pseudo-header (addresses, length, protocol).
+* @function rechecksum
+*/
+static int luaskb_rechecksum(lua_State *L)
```
lneto: mixed feelings about this name.. perhaps `skb:checksum` it's better.. as it's applied to an existent skb, it should be clear it's a "recalculation".. @carloslack what do you think?

lib/luaskb.c
```diff
@@ -0,0 +1,244 @@
...
+	ip_send_check(iph);
+
+	if (iph->protocol == IPPROTO_UDP) {
+		struct udphdr *udph = udp_hdr(skb);
+		unsigned int udplen = ntohs(udph->len);
+		udph->check = 0;
+		udph->check = csum_tcpudp_magic(iph->saddr, iph->daddr,
+			udplen, IPPROTO_UDP,
+			skb_checksum(skb, iphlen, udplen, 0));
+	}
```
lneto, line 141: we can have a macro to generalize that

lib/luaskb.c
```diff
@@ -0,0 +1,244 @@
...
+	size_t size = skb_headlen(skb);
+
+	if (mac) {
+		luaL_argcheck(L, skb_mac_header_was_set(skb), 2, "MAC header not set");
+		ptr  += skb_mac_offset(skb);
+		size += skb_mac_header_len(skb);
+	}
+
+	luadata_reset(data, ptr, 0, size, LUADATA_OPT_NONE);
+	lunatik_getregistry(L, data);
```
lneto:
```suggestion
	lunatik_getregistry(L, data); /* push data */
```

lib/luaskb.c
```diff
@@ -0,0 +1,244 @@
...
+	{"forward",  luaskb_forward},
+	{NULL, NULL}
+};
+
+static const lunatik_class_t luaskb_class = {
+	.name    = "skb",
+	.methods = luaskb_mt,
+	.release = luaskb_release,
+	.sleep   = false,
+	.shared  = true,
```
lneto, line 204: btw, @carloslack @sneaky-potato, this confuses me all the time.. we need `.shared = true` because of this [clone call](https://github.com/luainkernel/lunatik/pull/455/changes#diff-b81700b4acc9ecbfbbaf7098836465268f32f7e699edb44bbad806de90b8661cR220), but the object should be set with `.shared = false`, as in [here](https://github.com/luainkernel/lunatik/pull/455/changes#diff-b81700b4acc9ecbfbbaf7098836465268f32f7e699edb44bbad806de90b8661cR217), to not allow it to be shared afterwords or calling the monitor (imposing overhead on the hotpath).. IMHO, we should find better names fot this flags.. perhaps class should be `.clone = true` and object `.monitor = false`.. what do you think?

lib/luaskb.c
```diff
@@ -0,0 +1,244 @@
...
+};
+
+lunatik_object_t *luaskb_new(lua_State *L)
+{
+	lunatik_object_t *object = lunatik_allocobject(L, &luaskb_class, luaskb_t, false, false);
+	luaskb_t *lskb = (luaskb_t *)object->private;
+	lskb->skb = NULL;
```
lneto: we don't need this, right? if we are not memsetting the struct to zero on allocobject, we should IMHO

lib/luaskb.c
```diff
@@ -0,0 +1,244 @@
...
+	lunatik_cloneobject(L, object);
```
lneto: perhaps it should be moved to allocobject.. also, it would be good to have a comment justifying why we need this clone.. BTW, it's because it will be called from C through attach.. pretty much the "embedding case"

lib/luaskb.c
```diff
@@ -0,0 +1,244 @@
...
+	lskb->data = luadata_new(L, false);
+	lunatik_getobject(lskb->data);
+	lunatik_register(L, -1, lskb->data);
```
lneto, line 229: can't we use registerobject here?

lib/luaskb.c
carloslack: Like we were discussing earlier, maybe this could be a `flags` with `SKB_MONITOR_FLAG` , `SKB_CLONE_FLAG` and etc (please ignore the flag names, it is just to illustrate) 

But that would cause the weirdness of, if user wants to mess, of setting two or more concurrent flags, wdyt?

lib/luaskb.c
carloslack: I agree with simple `checksum` , if it recalculates, does the caller must know?

lib/luaskb.c
carloslack: Given the reasoning, `lskb` should be fine!
thanks :)

examples/dnsblock/nf_dnsblock.lua
carloslack: The user can always check how it works, by looking at the documentation elsewhere and the code itself.

It is just that it is supposed to be as helpful as possible to the user, if a comment helps, ok, if not let it be, IMO.
Bottom line is that I would not add an extra argument, it is more simple as it is

lib/luaskb.c
lneto: perhaps we should have a enum for this as you suggest at some point.. e.g., `LUNATIK_SHARE_MONITOR`,  `LUNATIK_SHARE_CLONE`, `LUNATIK_SHARE_NONE` and use a single field in the struct for this flag (`.shared`)? a tri-state in practice.. what do you think? perhaps we should discuss this on the issue itself ;-)..

examples/dnsblock/nf_dnsblock.lua
lneto: I think having two options will be a good way to document.. especially an example.. I changed it to..

```suggestion
	local pkt = skb:data("net")
```

so we can have "net" and "mac", what do you think?

and we still have the default case as "net".. so, `skb:data("net")` is equivalent to `skb:data()`

lib/luaskb.c
lneto: no, actually..

lib/luaskb.c
```diff
@@ -0,0 +1,230 @@
...
+
+#define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+#include <linux/skbuff.h>
+#include <linux/if_vlan.h>
+#include <linux/netdevice.h>
+#include <linux/tcp.h>
+#include <linux/udp.h>
+#include <net/ip.h>
+
+#include "luadata.h"
```
lneto: we don't need this

lib/luaskb.c
```diff
@@ -0,0 +1,253 @@
...
+static void __exit luaskb_exit(void)
+{
+}
+
+module_init(luaskb_init);
+module_exit(luaskb_exit);
+MODULE_LICENSE("Dual MIT/GPL");
+MODULE_AUTHOR("Lourival Vieira Neto <lourival.neto@ringzero.com.br>");
+MODULE_AUTHOR("Carlos Carvalho <carloslack@gmail.com>");
+MODULE_AUTHOR("Mohammad Shehar Yaar Tausif <sheharyaar48@gmail.com>");
```
lneto, line 251: @sheharyaar I added you as author because I borrowed some ideas from luanetfilter and previous discussions.. let me know if it's alright ;-)

lib/luaskb.c
```diff
@@ -0,0 +1,255 @@
...
+* Returns an independent copy of the skb with its own data buffer.
+* @function copy
+* @treturn skb
+* @raise if copy allocation fails
+*/
+static int luaskb_copy(lua_State *L)
+{
+	luaskb_t *lskb = luaskb_check(L, 1);
+	struct sk_buff *nskb = lunatik_checknull(L, skb_copy(lskb->skb, GFP_ATOMIC));
+	lunatik_object_t *object = lunatik_newobject(L, &luaskb_class, sizeof(luaskb_t), false);
```
lneto, line 216: if we fail here, we're gonna leak nskb

lib/luaskb.c
```diff
@@ -0,0 +1,252 @@
...
+{
+	luaskb_t *lskb = luaskb_check(L, 1);
+	struct sk_buff *skb = lskb->skb;
+	luaskb_pushoptinteger(L, skb_vlan_tag_present(skb), skb_vlan_tag_get_id(skb));
+	return 1;
+}
+
+/***
+* @function data
+* @tparam[opt] string layer "net" (default, L3) or "mac" (L2, includes MAC header)
```
carloslack, line 69: What exactly is `default` here?

lneto: default is "net", that is, L3 headers.. `:data()` is equivalent to `:data("net")`

PR #459, opened 2026-03-05
prevent GC finalizers from running under spinlock in lunatik_monitor

lneto:
Shared object methods are wrapped by lunatik_monitor(), which holds a spinlock during lua_pcall(). Any Lua allocation inside that call can trigger an incremental GC step, which may finalize objects whose release callbacks sleep (e.g. packet_release() calling synchronize_net()), causing "scheduling while atomic" crashes.

Stop the GC before acquiring the lock and restart it after releasing, so finalizers only run outside atomic context.

### DIFF
--- a/lunatik_obj.c
+++ b/lunatik_obj.c
@@ -158,9 +158,11 @@ static int lunatik_monitor(lua_State *L)
 	lua_pushvalue(L, lua_upvalueindex(1)); /* method */
 	lua_insert(L, 1); /* stack: method, object, args */
 
+	lua_gc(L, LUA_GCSTOP);
 	lunatik_lock(object);
 	ret = lua_pcall(L, n, LUA_MULTRET, 0);
 	lunatik_unlock(object);
+	lua_gc(L, LUA_GCRESTART);
 
 	if (ret != LUA_OK) {
 		const char *method = lua_tostring(L, lua_upvalueindex(2));


sneaky-potato:
This makes me wonder, is lunatik_monitor the only place in lunatik where lua_pcall is invoked while holding a spinlock, or are there other call sites that need the same treatment?

lneto:
> This makes me wonder, is lunatik_monitor the only place in lunatik where lua_pcall is invoked while holding a spinlock, or are there other call sites that need the same treatment?

perhaps it would be possible to reach it out through lunatik_run(), using a script that do something like..
```
sleepable_object = nil
collectgarbage()
```

could you try that? I need to put more thoughts on this..

lneto:
> This makes me wonder, is lunatik_monitor the only place in lunatik where lua_pcall is invoked while holding a spinlock, or are there other call sites that need the same treatment?



> > This makes me wonder, is lunatik_monitor the only place in lunatik where lua_pcall is invoked while holding a spinlock, or are there other call sites that need the same treatment?
> 
> perhaps it would be possible to reach it out through lunatik_run(), using a script that do something like..
> 
> ```
> sleepable_object = nil
> collectgarbage()
> ```
> 
> could you try that? I need to put more thoughts on this..

Actually this is not possible..

lneto:
> > This makes me wonder, is lunatik_monitor the only place in lunatik where lua_pcall is invoked while holding a spinlock, or are there other call sites that need the same treatment?
> 
> > > This makes me wonder, is lunatik_monitor the only place in lunatik where lua_pcall is invoked while holding a spinlock, or are there other call sites that need the same treatment?
> > 
> > 
> > perhaps it would be possible to reach it out through lunatik_run(), using a script that do something like..
> > ```
> > sleepable_object = nil
> > collectgarbage()
> > ```  
> >   
> > could you try that? I need to put more thoughts on this..
> 
> Actually this is not possible..

the problem is the opposite, actually, using sleep = false objects on sleep = true runtimes.. if I got it correctly, it can only happen on calling Lua while holding a spinlock and having sleep = true objects to be collected and triggering the GC, and again if I correctly understood it, it can only happen on monitors of sleep = false objects on sleep = true runtimes.. see this example:

```Lua
 -- Create a fifo (shared=true, sleep=false → lunatik_monitor uses spinlock)
  -- and drop an AF_PACKET socket reference before calling a fifo method.

  local fifo   = require("fifo")
  local linux  = require("linux")
  local thread = require("thread")

  local f = fifo.new(64)

  return function()
      while not thread.shouldstop() do
          -- AF_PACKET socket dropped without close() → sits in GC heap
          socket.new(socket.af.PACKET, socket.sock.RAW, 0)

          -- lunatik_monitor acquires spinlock, calls lua_pcall
          -- allocation inside triggers GC → packet_release() → synchronize_net() → CRASH
          f:write(string.rep("x", 32))

          linux.schedule(10)
      end
  end
```

if runtime->sleep = false, then all objects must be sleep = false, we won't have sleepable finalizers in this case.

If runtime->sleep = true, then the only spinlocks we have are on sleep = false objects; it can be on monitors, or on their methods explicitly if they don't use monitors. In this case, I think the only object that is shared, non-sleepable and not monitored is RCU, which doesn't call Lua holding a spinlock.. so it's safe..

sneaky-potato:
> If runtime->sleep = true, then the only spinlocks we have are on sleep = false objects; it can be on monitors, or on their methods explicitly if they don't use monitors. In this case, I think the only object that is shared, non-sleepable and not monitored is RCU, which doesn't call Lua holding a spinlock.. so it's safe..

Right that makes sense, I think we can go ahead and merge this. I ran my benchmark script on this branch, the metrics also look fine.

PR #462, opened 2026-03-07
lunatik: use modprobe -r and document module unload order

lneto:
LUNATIK_MODULES load order matters: modules are unloaded in reverse, so each module must follow its dependencies in the list.


Makefile
```diff
@@ -29,7 +29,9 @@ CONFIG_LUNATIK ?= m
 CONFIG_LUNATIK_RUNTIME ?= y
 CONFIG_LUNATIK_RUN ?= m
 
-LUNATIK_MODULES := DEVICE LINUX NOTIFIER SOCKET RCU THREAD FIB DATA PROBE SYSCALL XDP FIFO NETFILTER \
+# Order matters: modules are loaded left-to-right and unloaded right-to-left (rmmod).
```
carloslack, line 32: Thanks! good point!

bin/lunatik
```diff
@@ -75,7 +75,7 @@ lunatik.commands = {
 		local first = #lunatik.modules
 		for i = first, 1, -1 do
 			local module = lunatik.modules[i]
-			lunatik.sh("rmmod $MODULE", module, true)
+			lunatik.sh("{ modprobe -r $MODULE 2>/dev/null || rmmod $MODULE 2>/dev/null; }", module, true)
```
carloslack, line 78: Do we need to discard errors? It could useful to keep them

lneto: we need to fallback, right? but I agree with you, perhaps we should simplify and leave this line as it was.. it's not a good thing to simple discard it and it would be a bit complex to configure it properly beforehand to detect we have `modprobe -r` in the box..

lneto:
it should be integrated along with luaskb

PR #485, opened 2026-03-12
lunatik: move object API to lunatik_obj.h

lneto:
Extract object-related macros, inlines, and prototypes from lunatik.h into a new lunatik_obj.h, included at the bottom of lunatik.h before lunatik_val.h. Mirrors the lunatik_val.h pattern.


lneto:
Closing this: the branch is 413 commits behind master and the API it moves no longer exists in that shape, so it rebases into a rewrite rather than a move. The split is still wanted and will be redone against the current lunatik.h, cut by what the bindings actually include rather than by topic.

PR #503, opened 2026-03-15
issue#444 move object definitions to lunatik_obj.h

rajsaksham-32:
Moved lunatik_class_t, lunatik_object_t, lunatik_opt_t and the LUNATIK_OPT_* defines out of lunatik.h into a new header lunatik_obj.h. 

I tested the changes and checked that all .c files go through lunatik.h which pulls in lunatik_obj.h, so all existing usage works. Also made sure there are no circular includes.

I kept the same license header, include guard style and indentation as the rest of the codebase.

I used AI for formatting the text.

### DIFF
--- a/lunatik.h
+++ b/lunatik.h
@@ -6,50 +6,11 @@
 #ifndef lunatik_h
 #define lunatik_h
 
-#include <linux/mutex.h>
-#include <linux/spinlock.h>
-#include <linux/slab.h>
-#include <linux/kref.h>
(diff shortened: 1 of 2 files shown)


lneto (changes requested):
- it's not enough, we need all definitions, APIs, macros, etc related to object to be moved;
- the code style is wrong, see how we use \n on our files and what's wrong with yours;
- AI assisted commits should include the tool as co-author.

PR #509, opened 2026-03-16
lunatik: add lunatik_assert for internal state checks

lneto:
Replace luaL_argcheck(L, cond, 1, msg) used for internal object state checks with a new lunatik_assert(L, cond, msg) macro that calls luaL_error directly, avoiding the misleading "bad argument #1" prefix.

Applied to: skb not-set, FRAGLIST GSO, linearization failure, forward preconditions in luaskb; read-only check in luadata; module-load and softirq-context guards in luathread.

### DIFF
--- a/lib/luaskb.c
+++ b/lib/luaskb.c
@@ -28,9 +28,9 @@ LUNATIK_PRIVATECHECKER(luaskb_check, luaskb_t *,
 
 /* FRAGLIST GSO skbs hold segments in frag_list; skb_copy refuses to copy
  * them (ambiguous semantics: copy the container or the segments?). */
-#define luaskb_checkfraglist(L, lskb, ix)				\
-	luaL_argcheck(L, !(skb_shinfo((lskb)->skb)->gso_type &		\
-		SKB_GSO_FRAGLIST), (ix), "FRAGLIST GSO skbs cannot be copied")
+#define luaskb_checkfraglist(L, lskb)					\
+	lunatik_assert(L, !(skb_shinfo((lskb)->skb)->gso_type &		\
+		SKB_GSO_FRAGLIST), "FRAGLIST GSO skbs cannot be copied")
 
 #define luaskb_csum4(skb, iph, iphlen)					\
 	csum_tcpudp_magic((iph)->saddr, (iph)->daddr,			\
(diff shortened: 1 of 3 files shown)


lib/luaskb.c
```diff
@@ -28,9 +28,9 @@ LUNATIK_PRIVATECHECKER(luaskb_check, luaskb_t *,
 
 /* FRAGLIST GSO skbs hold segments in frag_list; skb_copy refuses to copy
  * them (ambiguous semantics: copy the container or the segments?). */
-#define luaskb_checkfraglist(L, lskb, ix)				\
-	luaL_argcheck(L, !(skb_shinfo((lskb)->skb)->gso_type &		\
-		SKB_GSO_FRAGLIST), (ix), "FRAGLIST GSO skbs cannot be copied")
+#define luaskb_checkfraglist(L, lskb)					\
```
carloslack, line 31: style?

lneto: beacause of `\` alingment?

lunatik.h
```diff
@@ -139,6 +139,12 @@ static inline void *lunatik_realloc(lua_State *L, void *ptr, size_t size)
 
 #define lunatik_enomem(L)	luaL_error((L), "not enough memory")
 
+static inline void lunatik_assert(lua_State *L, bool cond, const char *msg)
```
lneto, line 142: I'm not sure it's gonna work because of __no_return.. need to check.. however, I'm still in doubt of this new API.. problem I see with argcheck everywhere is the "bad self" error message.. but I'm not sure it's worth.. what do you think?

carloslack: Weirdly it could work if you make `(void)0` so there would be no type mismatch, but it is a gimmick , you are right, nothing beats a simple `if` 

But I haven't seen this error yet, is it causing anything to misbehave or the objective is to improve UX?

lib/luaskb.c
carloslack: No, it looks it is a space after `)` , is it github thing?

lunatik.h
lneto: > Weirdly it could work if you make (void)0 so there would be no type mismatch, but it is a gimmick , you are right, nothing beats a simple if

not against it actually, if you think it's clearer..

> But I haven't seen this error yet, is it causing anything to misbehave or the objective is to improve UX?

it's just to improve the error message.. I'm getting this on skb:copy():
```
calling 'copy' on bad self (FRAGLIST GSO skbs cannot be copied)
```

not sure if this change works just to have something like:
```
FRAGLIST GSO skbs cannot be copied
```

carloslack: hmm in this case we are also avoiding a call to `luaL_argerror` (that eventually call luaL_error) right? And going straight to `luaL_error` ?

lneto: Yeah.. argerror is the one that adds "bad self"..

carloslack: Apparently it is worth changing, no call to `luaL_argerror` , you don't need the `arg` argument for these cases and the bad self error message is artificial for this case, IMO , what do you think?

PR #512, opened 2026-03-19
luarcu: add string support using external strings and kref

carloslack:
- Add lunatik_string_t (kref + len + flexible char data[]) to lunatik_val.h
- Extend lunatik_value_t union with lunatik_string_t *string member
- lunatik_checkvalue: handle LUA_TSTRING by kmalloc'ing a lunatik_string_t, copying the string data out of Lua's GC reach, and kref_init'ing it
- lunatik_pushvalue: handle LUA_TSTRING via lua_pushexternalstring with a custom alloc callback (lunatik_string_alloc) that kref_put's on GC; kref_get is called before push so Lua's GC holds one reference
- Add lunatik_putvalue to release the caller's string reference after lunatik_pushvalue (no-op for objects, which transfer their ref)
- luarcu_newentry/luarcu_free/luarcu_getvalue: mirror existing object kref_get/kref_put pattern for string values
- luarcu_index: call lunatik_putvalue to release luarcu_getvalue's kref
- luarcu_newindex: save ret before releasing string ref via lunatik_putvalue
- luarcu_map: generalize to lunatik_value_t, handle string kref alongside existing object handling; refactor map_handle/map_call accordingly

### DIFF
--- a/lib/luarcu.c
+++ b/lib/luarcu.c
@@ -89,6 +89,8 @@ static luarcu_entry_t *luarcu_newentry(const char *key, size_t keylen, lunatik_v
 	entry->value = *value;
 	if (lunatik_isuserdata(value))
 		lunatik_getobject(value->object);
+	else if (lunatik_isstring(value))
+		lunatik_getstring(value->string);
 	return entry;
 }
 
@@ -96,6 +98,8 @@ static inline void luarcu_free(luarcu_entry_t *entry)
 {
 	if (lunatik_isuserdata(&entry->value))
 		lunatik_putobject(entry->value.object);
+	else if (lunatik_isstring(&entry->value))
+		lunatik_putstring(entry->value.string);
 	kfree_rcu(entry, rcu);
 }
 
@@ -114,6 +118,8 @@ void luarcu_getvalue(lunatik_object_t *table, const char *key, size_t keylen, lu
 		*value = entry->value;
 		if (lunatik_isuserdata(value))
 			lunatik_getobject(value->object);
+		else if (lunatik_isstring(value))
+			lunatik_getstring(value->string);
 	}
 	rcu_read_unlock();
 }
@@ -169,6 +175,7 @@ static int luarcu_index(lua_State *L)
 
 	luarcu_getvalue(table, key, keylen, &value);
 	lunatik_pushvalue(L, &value);
+	lunatik_putvalue(&value);
 	return 1; /* value */
 }
 
@@ -188,7 +195,9 @@ static int luarcu_newindex(lua_State *L)
 
 	lunatik_value_t value;
 	lunatik_checkvalue(L, 3, &value);
-	if (luarcu_setvalue(table, key, keylen, &value) < 0)
+	int ret = luarcu_setvalue(table, key, keylen, &value);
+	lunatik_putvalue(&value);
+	if (ret < 0)
 		luaL_error(L, "not enough memory");
 	return 0;
 }
@@ -215,20 +224,23 @@ static inline void luarcu_inittable(luarcu_table_t *table, size_t size)
 static int luarcu_map_handle(lua_State *L)
 {
 	const char *key = (const char *)lua_touserdata(L, 2);
-	lunatik_object_t *value = (lunatik_object_t *)lua_touserdata(L, 3);
+	lunatik_value_t *value = (lunatik_value_t *)lua_touserdata(L, 3);
 
 	BUG_ON(!key || !value);
 
 	lua_pop(L, 2); /* key, value */
 
 	lua_pushstring(L, key);
-	lunatik_pushobject(L, value);
+	if (lunatik_isuserdata(value))
+		lunatik_pushobject(L, value->object);
+	else
+		lunatik_pushvalue(L, value);
 	lua_call(L, 2, 0);
 
 	return 0;
 }
 
-static inline int luarcu_map_call(lua_State *L, int cb, const char *key, lunatik_object_t *value)
+static inline int luarcu_map_call(lua_State *L, int cb, const char *key, lunatik_value_t *value)
 {
 	lua_pushcfunction(L, luarcu_map_handle);
 	lua_pushvalue(L, cb);
@@ -257,15 +269,22 @@ static int luarcu_map(lua_State *L)
 	rcu_read_lock();
 	luarcu_foreach(table, bucket, n, entry) {
 		char key[LUARCU_MAXKEY];
+		lunatik_value_t value = entry->value;
 
 		strncpy(key, entry->key, LUARCU_MAXKEY);
 		key[LUARCU_MAXKEY - 1] = '\0';
-		lunatik_object_t *value = entry->value.object;
-		lunatik_getobject(value);
+
+		if (lunatik_isuserdata(&value))
+			lunatik_getobject(value.object);
+		else if (lunatik_isstring(&value))
+			lunatik_getstring(value.string);
 
 		rcu_read_unlock();
-		int ret = luarcu_map_call(L, 1, key, value);
-		lunatik_putobject(value);
+		int ret = luarcu_map_call(L, 1, key, &value);
+		if (lunatik_isuserdata(&value))
+			lunatik_putobject(value.object);
+		else if (lunatik_isstring(&value))
+			lunatik_putstring(value.string);
 		if (ret != LUA_OK)
 			lua_error(L);
 		rcu_read_lock();
--- a/lib/luarcu.c
+++ b/lib/luarcu.c
@@ -87,19 +87,13 @@ static luarcu_entry_t *luarcu_newentry(const char *key, size_t keylen, lunatik_v
 	strncpy(entry->key, key, keylen);
 	entry->key[keylen] = '\0';
 	entry->value = *value;
-	if (lunatik_isuserdata(value))
-		lunatik_getobject(value->object);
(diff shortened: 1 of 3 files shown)


lunatik_val.c
```diff
@@ -17,6 +30,18 @@ void lunatik_checkvalue(lua_State *L, int ix, lunatik_value_t *value)
 	case LUA_TNUMBER:
 		value->integer = lua_tointeger(L, ix);
 		break;
+	case LUA_TSTRING: {
+		size_t len;
+		const char *s = lua_tolstring(L, ix, &len);
+		lunatik_string_t *str = kmalloc(sizeof(*str) + len + 1, GFP_ATOMIC);
+		if (unlikely(!str))
+			luaL_argerror(L, ix, "not enough memory");
```
carloslack: This one could be the upcoming `lunatik_assert`

lunatik_val.h
```diff
@@ -6,16 +6,35 @@
...
+	char data[];
+} lunatik_string_t;
+
 typedef struct lunatik_value_s {
 	int type;
 	union {
 		int boolean;
 		lua_Integer integer;
 		lunatik_object_t *object;
+		lunatik_string_t *string;
```
carloslack, line 20: Here we have it added in the union. It can be checked with new `lunatik_isstring`
There the addition of `lunatik_getstring`,  and `lunatik_putstring`  that when called end up `kfree`ing the memory in the callback & update refcount.

lib/luarcu.c
```diff
@@ -257,15 +269,22 @@ static int luarcu_map(lua_State *L)
...
+			lunatik_getstring(value.string);
 
 		rcu_read_unlock();
-		int ret = luarcu_map_call(L, 1, key, value);
-		lunatik_putobject(value);
+		int ret = luarcu_map_call(L, 1, key, &value);
+		if (lunatik_isuserdata(&value))
+			lunatik_putobject(value.object);
+		else if (lunatik_isstring(&value))
+			lunatik_putstring(value.string);
```
carloslack: Repeating code here - could be a macro (passing `get` or `put`) or an inline

lunatik_val.c
lneto: we have enomem() already

lib/luarcu.c
```diff
@@ -89,13 +89,17 @@ static luarcu_entry_t *luarcu_newentry(const char *key, size_t keylen, lunatik_v
...
+	else if (lunatik_isstring(value))
+		lunatik_getstring(value->string);
 	return entry;
 }
 
 static inline void luarcu_free(luarcu_entry_t *entry)
 {
 	if (lunatik_isuserdata(&entry->value))
 		lunatik_putobject(entry->value.object);
+	else if (lunatik_isstring(&entry->value))
```
lneto: we can put value into a local var

lib/luarcu.c
```diff
@@ -114,6 +118,8 @@ void luarcu_getvalue(lunatik_object_t *table, const char *key, size_t keylen, lu
 		*value = entry->value;
 		if (lunatik_isuserdata(value))
 			lunatik_getobject(value->object);
+		else if (lunatik_isstring(value))
+			lunatik_getstring(value->string);
```
lneto: this pattern is repeated quite a lot.. we can move it to a function.. perhaps on lunatik_val module?

lib/luarcu.c
```diff
@@ -169,6 +175,7 @@ static int luarcu_index(lua_State *L)
 
 	luarcu_getvalue(table, key, keylen, &value);
 	lunatik_pushvalue(L, &value);
+	lunatik_putvalue(&value);
```
lneto, line 169: it doesn't seem right

lib/luarcu.c
```diff
@@ -188,7 +195,9 @@ static int luarcu_newindex(lua_State *L)
 
 	lunatik_value_t value;
 	lunatik_checkvalue(L, 3, &value);
-	if (luarcu_setvalue(table, key, keylen, &value) < 0)
+	int ret = luarcu_setvalue(table, key, keylen, &value);
+	lunatik_putvalue(&value);
```
lneto, line 190: also this.. why we are putting value here?

lib/luarcu.c
```diff
@@ -215,20 +224,23 @@ static inline void luarcu_inittable(luarcu_table_t *table, size_t size)
 static int luarcu_map_handle(lua_State *L)
 {
 	const char *key = (const char *)lua_touserdata(L, 2);
-	lunatik_object_t *value = (lunatik_object_t *)lua_touserdata(L, 3);
+	lunatik_value_t *value = (lunatik_value_t *)lua_touserdata(L, 3);
```
lneto: was map() broken for values? we should have a separate fix for this, I think

lneto: also, it would be good to add a regression test..

lunatik_val.c
```diff
@@ -17,6 +30,18 @@ void lunatik_checkvalue(lua_State *L, int ix, lunatik_value_t *value)
 	case LUA_TNUMBER:
 		value->integer = lua_tointeger(L, ix);
 		break;
+	case LUA_TSTRING: {
+		size_t len;
+		const char *s = lua_tolstring(L, ix, &len);
+		lunatik_string_t *str = kmalloc(sizeof(*str) + len + 1, GFP_ATOMIC);
```
lneto: we should use lunatik_malloc here, right? we would need to check if we have already a dup API, I got the impression we do have..

lunatik_val.c
```diff
@@ -5,6 +5,19 @@
...
 
+void lunatik_freestring(struct kref *kref)
+{
+	kfree(container_of(kref, lunatik_string_t, kref));
+}
+EXPORT_SYMBOL(lunatik_freestring);
+
+static void *lunatik_string_alloc(void *ud, void *ptr, size_t osize, size_t nsize)
+{
+	if (nsize == 0)
```
lneto: I'm not sure it's enough.. we need to check the API contract..

lunatik_val.c
```diff
@@ -5,6 +5,19 @@
 
 #include "lunatik.h"
 
+void lunatik_freestring(struct kref *kref)
+{
+	kfree(container_of(kref, lunatik_string_t, kref));
```
lneto: we also need to free the buffer, right?

lunatik_val.h
```diff
@@ -6,16 +6,35 @@
 #ifndef lunatik_val_h
 #define lunatik_val_h
 
+typedef struct lunatik_string_s {
+	struct kref kref;
+	size_t len;
```
lneto: do we really need size? we are dealing with null-terminated strings only

lunatik_val.h
```diff
@@ -6,16 +6,35 @@
...
 
 #define lunatik_isuserdata(v)	((v)->type == LUA_TUSERDATA)
+#define lunatik_isstring(v)	((v)->type == LUA_TSTRING)
+
+void lunatik_freestring(struct kref *kref);
+#define lunatik_getstring(s)	kref_get(&(s)->kref)
+#define lunatik_putstring(s)	kref_put(&(s)->kref, lunatik_freestring)
+
+/* release value's reference after lunatik_pushvalue (strings only; objects transfer their ref) */
+static inline void lunatik_putvalue(lunatik_value_t *value)
```
lneto, line 48: this one could abstract both string and object

lib/luarcu.c
lneto: @carloslack I added a fix for this here.. https://github.com/luainkernel/lunatik/pull/515/

lunatik_val.c
```diff
@@ -5,6 +5,28 @@
 
 #include "lunatik.h"
 
+void lunatik_freestring(struct kref *kref)
+{
+	lunatik_string_t *str = container_of(kref, lunatik_string_t, kref);
+	lunatik_free(str->data);
```
carloslack, line 11: need to make sure data is valid..

lunatik_val.c
```diff
@@ -5,6 +5,28 @@
...
+	lunatik_free(str);
+}
+EXPORT_SYMBOL(lunatik_freestring);
+
+static inline lunatik_string_t *lunatik_newstring(lua_State *L, int ix)
+{
+	const char *s = luaL_checkstring(L, ix);
+	lunatik_string_t *str = lunatik_checkalloc(L, sizeof(*str));
+	str->data = lunatik_malloc(L, strlen(s) + 1);
+	if (unlikely(!str->data)) {
```
carloslack, line 21: alright, it is handled here

lunatik_val.h
```diff
@@ -6,16 +6,50 @@
...
 } lunatik_value_t;
 
 #define lunatik_isuserdata(v)	((v)->type == LUA_TUSERDATA)
+#define lunatik_isstring(v)	((v)->type == LUA_TSTRING)
+
+void lunatik_freestring(struct kref *kref);
+#define lunatik_getstring(s)	kref_get(&(s)->kref)
+#define lunatik_putstring(s)	kref_put(&(s)->kref, lunatik_freestring)
+
+static inline void lunatik_holdvalue(lunatik_value_t *value)
```
carloslack, line 31: could still be a local macro...

lib/luarcu.c
carloslack: the string can leak because caller's ref is never released
after `luarcu_setvalue` I think the count is `2` (caller holds one)

PR #518, opened 2026-03-24
docs: add Code_style.md

Physic69:
Adds Code_style.md to formalize project-wide C coding standards

### DIFF
--- a/Code_style.md
+++ b/Code_style.md
@@ -1,6 +1,6 @@
 # Lunatik C Code Style Guide
 
-This guide outlines the code style conventions observed in the `lunatik.h` header file, which should be followed for consistency across the Lunatik project. As a general rule, we tend to stick to [Linux kernel rules](https://www.kernel.org/doc/html/latest/process/coding-style.html).
+This guide outlines the code style conventions observed in the Lunatik project, which should be followed for consistency. As a general rule, we tend to stick to [Linux kernel rules](https://www.kernel.org/doc/html/latest/process/coding-style.html).
 
 ## File Headers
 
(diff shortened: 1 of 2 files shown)


Code_style.md
```diff
@@ -0,0 +1,235 @@
...
+Includes should be grouped, with system/kernel headers first, followed by Lua headers, and then project-specific headers. Use an empty line to separate groups (e.g., [`lunatik.h`](lunatik.h:9-16)).
+
+```c
+#include <linux/mutex.h>
+#include <linux/spinlock.h>
+#include <linux/slab.h>
+#include <linux/kref.h>
+
+#include <lua.h>
+#include <lauxlib.h>
```
lneto: now we demand only lunatik.h

Code_style.md
```diff
@@ -0,0 +1,235 @@
...
+Break long conditional expressions after logical operators, using two tabs for indentation of the continuation (e.g., [`lunatik.h`](lunatik.h:110-111)).
+
+```c
+return unlikely(lunatik_ismonitor(object->opt)) ?
+		(lunatik_issoftirq(object->opt) ? spin_trylock(&object->spin) : mutex_trylock(&object->mutex)) : 1;
+```
+
+## Line breaks, new lines and file Endings
+
+All source files must end with one empty line. Files should not contain multiple consecutive empty lines.
```
lneto: what do you think about this? I'm thinking on dropping this requirement.. I notice Claude Code, at least, has some trouble on sticking to this rule.. I'm not sure the benefit of having this rule pays off...

@carloslack would be good to have your input on this, as well.. ;-)

Code_style.md
```diff
@@ -0,0 +1,235 @@
+# Lunatik C Code Style Guide
+
+This guide outlines the code style conventions observed in the `lunatik.h` header file, which should be followed for consistency across the Lunatik project. As a general rule, we tend to stick to [Linux kernel rules](https://www.kernel.org/doc/html/latest/process/coding-style.html).
```
lneto: wondering if it wouldn't be easier to just base our style on something existent, like this one, and just ask Claude to document the differences..

Code_style.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Control Structures (if, for, while, do-while)
+
+Control structures like `if`, `for`, `while`, and `do-while` loops use K&R style
+(e.g., [`lunatik.h`](lunatik.h:20-25), [`lunatik.h`](lunatik.h:47-52)).
+
+```c
+if (condition) {
+	/* ... */
+}
```
lneto, line 56: another thing that I notice Claude has some difficult is on..

```
if (...) {
...
}
else
```

it always get back to this..

```
if (...) {
...
} else
```

thinking o dropping this requirement as well..

Code_style.md
Physic69: Have you tried putting these rules through Claude Code first? It might help us see which ones it consistently struggles with, and then we can decide whether to drop or relax them...

lneto: Many times.. it seems to forget about this all the time

Code_style.md
Physic69: i assume, dropping this requirement means, sticking to linux kernel style...

Code_style.md
Physic69: I think we can relax this a bit. git expects a for new line att EOF  for proper maintenance, so my suggestion: 
`All source files must end with a single newline. Avoid using more than one consecutive empty line within source files. `
That way it’s simple, and not too strict. WDYT?

Issue #561, opened 2026-05-01
[GSoC 2026] Create a generic lunatik_bpf_run API


sneaky-potato:
Issue to track ongoing GSoC work for Lunatik eBPF abstraction.

The goals are:
- [ ] Introduce a generic API that can be consumed by multiple subsystems using bpf kfuncs
- [ ] Refactor luaxdp to use the generic interface
- [ ] Documentation

Design notes:
I revisited the design I proposed in the plan but I realized there is a registry lookup that needs to be handled as well, so these are the revised responsibilities:
1. luaxdp needs to supply 3 facts
    - how the input needs to be passed: via `luaxdp_push_ctx`
    - how the output needs to be read: via `luaxdp_read_result`
    - by what name the luaxdp callback needs to be registered: via `registry_key`
2. lunatik_bpf_run
    - retrieves the runtime
    - retrieves the registered callback using registry_key
    - calls push_ctx to setup input
    - calls the actual handler (via `lua_pcall`)
    - calls read_result and sets up the output

```c
struct lunatik_bpf_call {
    lua_CFunction         registry_key;
    lunatik_bpf_input_cb  push_ctx;
    void                 *ctx;
    lunatik_bpf_output_cb read_result;
    void                 *result;
};

// calls the callback using registry_key
static int lunatik_bpf_handler(lua_State *L, struct lunatik_bpf_call *call); 

// exposed to luaxdp, luatc and others
int lunatik_bpf_run(lunatik_object_t *runtimes,
                    const char *key, size_t keylen,
                    lua_CFunction         registry_key,
                    lunatik_bpf_input_cb  push_ctx,  void *ctx,
                    lunatik_bpf_output_cb read_result, void *result);
```

PR #562, opened 2026-05-02
Introduce generic bpf API and refactor luaxdp 

sneaky-potato:
This patch:
- introduces a generic lunatik_bpf_run API
- refactor luaxdp to use the generic interface

related to #561

### DIFF
--- a/lib/luaxdp.c
+++ b/lib/luaxdp.c
@@ -23,6 +23,7 @@
 
 #include "luarcu.h"
 #include "luadata.h"
+#include "lunatik_bpf.h"
 
 #if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 4, 0))
 #include <linux/btf.h>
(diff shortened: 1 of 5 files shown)


lib/luaxdp.c
```diff
@@ -231,7 +277,14 @@ static const luaL_Reg luaxdp_lib[] = {
 	{NULL, NULL}
 };
 
-LUNATIK_NEWLIB(xdp, luaxdp_lib, NULL);
+int luaopen_xdp(lua_State *L);
```
lneto: not exactly what I meant, actually my suggestion was to create an actual lunatik class..

lib/luaxdp.c
```diff
@@ -39,59 +40,125 @@ __diag_ignore_all("-Wmissing-prototypes",
...
+	.opt     = LUNATIK_OPT_SOFTIRQ | LUNATIK_OPT_SINGLE,
+};
+
+static int luaxdp_push_ctx(lua_State *L, void *raw_ctx)
+{
+	luaxdp_ctx_t *ctx = (luaxdp_ctx_t *)raw_ctx;
+
+	lunatik_require(L, &luaxdp_ctx_class);
+	lunatik_object_t *object = lunatik_newobject(L, &luaxdp_ctx_class, sizeof(luaxdp_ctx_t), LUNATIK_OPT_NONE);
+	luaxdp_ctx_t *lctx = (luaxdp_ctx_t *)object->private;
```
lneto: why not using a pointer directly?

sneaky-potato: Ah yes, I think I did some over engineering here. We can use the pointer directly.

autogen.lua
```diff
@@ -248,7 +248,7 @@ function extract.parse()
 				current = { name = name, prefix = prefix, entries = {} }
 				table.insert(modules[top], current)
 			else
-				local sym, val = ascii:match('^%-%>(%S+)%s+(%-?%d+)%s+%S+$')
+				local sym, val = ascii:match('^%-%>(%S+)%s+%$?(%-?%d+)%s+%S+$')
```
lneto, line 251: ?

lneto: don't do this sort of thing through PRs, just rebase the your base branch on master

lib/luaxdp.c
```diff
@@ -39,59 +40,121 @@ __diag_ignore_all("-Wmissing-prototypes",
...
+LUNATIK_OPENER(xdp);
+static const lunatik_class_t luaxdp_ctx_class = {
+	.name    = "xdp.ctx",
+	.methods = luaxdp_ctx_mt,
+	.release = luaxdp_ctx_release,
+	.opener  = luaopen_xdp,
+	.opt     = LUNATIK_OPT_SOFTIRQ | LUNATIK_OPT_SINGLE,
+};
+
+static int luaxdp_push_ctx(lua_State *L, void *raw_ctx)
```
lneto: I'm not sure we need such callback.. it seems to me that we could have a template as on the rest of the framework and the specifics go through methods.. I think we could push a single object to lua tied to the module and then we call methods for every other operation..

sneaky-potato: Right, I get how this pattern is going to force future modules to construct the ctx object each time. How about this:  we store the ctx userdata in the registry at attach time. The luaxdp_callback retrieves the ctx object from the registry, refreshes the pointers, then invokes the user's Lua function ?

lneto: I think it's worth to try.. let's see how it looks.. now it's time to run experiments and proper design our API..

sneaky-potato: Okay I tried implementing this design, right now running in some stack issues. But this made me think about the push_ctx handler. Right now luaxdp has one argument (ctx) to be used by the handler, some other binding might have something else. So why do we want to push a *single* object in the generic layer? How about we let the caller decide how what arguments to be pushed.

lneto: For me, to have a generic interface, that is, a Lunatik object being pushed; and to leverage the framework.

> How about we let the caller decide how what arguments to be pushed.

what's the case for multiple arguments? why not using Lunatik object as the abstraction layer?

the way I see, every eBPF/Lunatik module should expose their own class as we do for kmod/Lunatik modules; but again, I might be wrong and not getting the whole picture here.. =)

sneaky-potato: I had to really convince myself how pushing the object will work with callback, but now I understand. This would indeed be the better design that we let the _object_ itself be the interface rather than pushing arguments.

sneaky-potato:
Right now facing an error:
```shell
[  204.825853] lunatik: /lib/modules/lua/examples/filter/sni.lua:25: bad argument #1 to 'getbyte' (out of bounds)
```

I tried two things:
1. previous approach- no luaxdp_handler, directly pass the lua callback to lunatik_bpf_run and run it with the object. The signature would be `int lunatik_bpf_run(lunatik_object_t *runtime, lua_CFunction registry_key, lunatik_object_t *object)`
But I ran into the above error when working with `ctx:packet()` and `ctx:argument()` because I was doing reset and clear on luadata outside runtime lock I think (in `bpf_luaxdp_run`).

2. Current approach- I thought the error would be because we should interact with packet and argument only after runtime lock is acquired, so I introduced another layer: `luaxdp_handler` which would make the actual lua callback using pcall(L, 1, 0, 0) and keep the generic layer lean
signature would be:  `int lunatik_bpf_run(lunatik_object_t *runtime, lua_CFunction registry_key)`
in the second case we are registering lua callback and luaxdp_handler both in the registry and `luaxdp_handler` calls lua callback after setting up `packet` and `argument`

But, I am still getting the same out of bounds error. Need to further investigate it, I believe is bug is specifically in luadata_reset not correctly reflecting the xdp_buff data bounds.
The implementation works with simplest drop handler:
```lua
xdp.attach(
    function(ctx) {
        ctx:set_action(action.DROP)
    }
)
```

lneto:
> (...)
> But, I am still getting the same out of bounds error. Need to further investigate it, I believe is bug is specifically in luadata_reset not correctly reflecting the xdp_buff data bounds. The implementation works with simplest drop handler:

I'd put some logs to see the offsets and lengths you are using..

sneaky-potato:
> > (...)
> > But, I am still getting the same out of bounds error. Need to further investigate it, I believe is bug is specifically in luadata_reset not correctly reflecting the xdp_buff data bounds. The implementation works with simplest drop handler:
> 
> I'd put some logs to see the offsets and lengths you are using..

Found the issue, I was returning the wrong value in case of ctx:packet()

sneaky-potato:
There appears to be another issue...
the Lua parser correctly extracts the SNI and calls `ctx:set_action(action.DROP)`. On the C side, the action is also updated correctly, before the Lua callback it is `XDP_PASS (2)` and after the callback it becomes `XDP_DROP (1)`.

```shell
May 11 17:57:09 elmon kernel: luaxdp: before run action=2
May 11 17:57:09 elmon kernel: sni:
May 11 17:57:09 elmon kernel:
May 11 17:57:09 elmon kernel: ebpf.io
May 11 17:57:09 elmon kernel: verdict:
May 11 17:57:09 elmon kernel:
May 11 17:57:09 elmon kernel: DROP
May 11 17:57:09 elmon kernel: filter_sni: ebpf.io DROP
May 11 17:57:09 elmon kernel: luaxdp: after run action=1
May 11 17:57:09 elmon kernel: luaxdp: before run action=2
May 11 17:57:09 elmon kernel: luaxdp: after run action=2
May 11 17:57:09 elmon kernel: luaxdp: before run action=2
May 11 17:57:09 elmon kernel: luaxdp: after run action=2
May 11 17:57:09 elmon kernel: luaxdp: before run action=2
May 11 17:57:09 elmon kernel: luaxdp: after run action=2
May 11 17:57:09 elmon kernel: luaxdp: before run action=2
May 11 17:57:09 elmon kernel: luaxdp: after run action=2
May 11 17:57:09 elmon kernel: luaxdp: before run action=2
May 11 17:57:09 elmon kernel: luaxdp: after run action=2
May 11 17:57:10 elmon kernel: luaxdp: before run action=2
May 11 17:57:10 elmon kernel: luaxdp: after run action=2
May 11 17:57:10 elmon kernel: luaxdp: before run action=2
May 11 17:57:10 elmon kernel: luaxdp: after run action=2
```

The HTTPS request by curl still succeeds after what appears to be a retry/retransmission by curl. So the DROP decision seems to be reached correctly, but the traffic is not actually being blocked consistently.

I added the logs before and after calling the generic layer.
```c
...
    pr_info("before run action=%d\n", action);
    lunatik_bpf_run(runtime, luaxdp_handler);
    pr_info("after run action=%d\n", action);
...
```

Tested the setup using instructions given in README
```shell
sudo xdp-loader load -m skb docker0 examples/filter/https.o
sudo docker run --rm -it alpine/curl https://ebpf.io    #returns the ful HTML and succeeds
```

sneaky-potato:
Hi @lneto i have pushed a change using macro pattern for the API, it is very thin right now, I think we'll evolve as we add more modules. I have tested it, sni.lua is refactored to use object oriented style.

lib/luaxdp.c
```diff
@@ -37,61 +38,133 @@ __diag_ignore_all("-Wmissing-prototypes",
...
+	lctx->arg__sz = ctx->arg__sz;
+	lctx->action  = ctx->action;
 
-	if (lunatik_getregistry(L, luaxdp_callback) != LUA_TFUNCTION) {
-		pr_err("couldn't find callback");
-		goto out;
+	luadata_reset(lctx->packet, lctx->xdp->data, lctx->xdp->data_end - lctx->xdp->data, LUADATA_OPT_KEEP);
+	luadata_reset(lctx->argument, lctx->arg, lctx->arg__sz, LUADATA_OPT_KEEP);
+
+	if (lunatik_getregistry(L, &luaxdp_callback_key) != LUA_TFUNCTION) {
```
lneto: Can't this be a lunatik_ebpf helper?

lib/luaxdp.c
```diff
@@ -37,61 +38,133 @@ __diag_ignore_all("-Wmissing-prototypes",
...
+	.opt     = LUNATIK_OPT_SOFTIRQ | LUNATIK_OPT_SINGLE,
+};
+
+static int luaxdp_handler(lua_State *L, luaxdp_ctx_t *ctx)
 {
-	int action = -1;
-	int status;
+	lunatik_require(L, &luaxdp_ctx_class);
+
+	if (lunatik_getregistry(L, &luaxdp_env_key) != LUA_TUSERDATA) {
```
lneto: Also this (can't be a helper)? I think we will need these patterns on all eBPF binding, right?

lib/luaxdp.c
```diff
@@ -106,24 +179,30 @@ static inline int luaxdp_checkruntimes(void)
...
 	lunatik_object_t *runtime;
-	struct xdp_buff *ctx = (struct xdp_buff *)xdp_ctx;
-	int action = -1;
+	int action = XDP_PASS;
 	size_t keylen = key__sz - 1;
 
 	if (unlikely(luaxdp_checkruntimes() != 0)) {
 		pr_err("couldn't find _ENV.runtimes\n");
-		goto out;
+		return XDP_PASS;
```
lneto: Why changing this pattern along this patch? It seems cosmetics to me.

lib/luaxdp.c
```diff
@@ -37,61 +38,133 @@ __diag_ignore_all("-Wmissing-prototypes",
...
+	*ctx->action = luaL_checkinteger(L, 2);
+	return 0;
+}
 
-	luadata_clear(buffer);
-	luadata_clear(argument);
-	return 1;
+static const luaL_Reg luaxdp_ctx_mt[] = {
+	{"__gc",        lunatik_deleteobject},
+	{"packet",      luaxdp_ctx_packet},
```
lneto: I wouldn't use "ctx" namespace, I'd keep our pattern here and use the module name only.. e.g., luaxdp_argument and luaxdp_action.. I wouldn't also call it set_action, I would call it simply action, instead

lib/lunatik_bpf.h
```diff
@@ -0,0 +1,15 @@
...
+ * SPDX-License-Identifier: MIT OR GPL-2.0-only
+ */
+#ifndef LUNATIK_BPF_H
+#define LUNATIK_BPF_H
+
+#include "lunatik.h"
+
+#define lunatik_bpf_run(runtime, handler, ret, ...)				\
+do {					\
+	lunatik_run((runtime), (handler), (ret), ##__VA_ARGS__);	\
```
lneto: it's not clear to me why we need this indirection here.. why not calling lunatik_run directly?

sneaky-potato: Yeah, I realized, fixing this in latest push

lib/luaxdp.c
sneaky-potato: that's a good idea!

sneaky-potato: I have added a macro lunatik_bpf_run which would read the callback and context from registry and do a pcall.

sneaky-potato:
Hello @lneto 
I did some tests, for this branch and the callback based approach. I tested the push_ctx callback approach, code is in [sneaky-potato/lunatik_bpf_run_userdata](https://github.com/luainkernel/lunatik/tree/sneaky-potato/lunatik_bpf_run_userdata) branch.
```c
int lunatik_bpf_run(lunatik_object_t *runtime, lua_CFunction registry_key, lunatik_bpf_push_t push_ctx, void *ctx)
```

```shell
kprobe:bpf_luaxdp_run
{
    // Record start time per thread ID (handles softirq concurrency safely)
    @start[tid] = nsecs;
}

kretprobe:bpf_luaxdp_run
/@start[tid]/
{
    $latency = nsecs - @start[tid];

    // 1. Log the absolute average, min, and max
    @avg_latency_ns = avg($latency);
    @max_latency_ns = max($latency);
    @min_latency_ns = min($latency);

    // 2. Generate a power-of-2 histogram of the latency distribution
    @latency_distribution_ns = hist($latency);

    // Clean up the start map entry
    delete(@start[tid]);
}
```

I tested the drop everything handler: 
```lua
local function filter_sni(ctx)
	ctx:action(action.PASS)
end

xdp.attach(filter_sni)
```

Metrics:
1. callback approach
@avg_latency_ns: 41889
@max_latency_ns: 76891
@min_latency_ns: 25116

2. Current branch
@avg_latency_ns: 17853
@max_latency_ns: 28620
@min_latency_ns: 15814

Issue #564, opened 2026-05-07
Kernel 7.0 : segfault when launching lunatik


jperon:
When launching lunatik on kernel 7.0, I get two segfault messages, and the following trace : 

<details>
<summary> click to see kernel trace</summary>
<pre><code>
mai 07 09:13:46 peron kernel: lunatik: loading out-of-tree module taints kernel.
mai 07 09:13:46 peron kernel: lunatik: module verification failed: signature and/or required key missing - tainting kernel
mai 07 09:13:47 peron kernel: Missing ENDBR: kallsyms_lookup_name+0x4/0xc0
mai 07 09:13:47 peron kernel: ------------[ cut here ]------------
mai 07 09:13:47 peron kernel: kernel BUG at arch/x86/kernel/cet.c:133!
mai 07 09:13:47 peron kernel: Oops: invalid opcode: 0000 [#1] SMP NOPTI
mai 07 09:13:47 peron kernel: CPU: 0 UID: 0 PID: 4838 Comm: modprobe Tainted: G S         OE       7.0.3+deb14-amd64 #1 PREEMPT(lazy)  Debian 7.0.3-1 
mai 07 09:13:47 peron kernel: Tainted: [S]=CPU_OUT_OF_SPEC, [O]=OOT_MODULE, [E]=UNSIGNED_MODULE
mai 07 09:13:47 peron kernel: Hardware name: CHUWI Innovation And Technology(ShenZhen)co.,Ltd MiniBook X/Default string, BIOS DNN20 V2.22 06/12/2024
mai 07 09:13:47 peron kernel: RIP: 0010:exc_control_protection+0x30f/0x320
mai 07 09:13:47 peron kernel: Code: e8 b9 09 00 00 00 48 8b 95 80 00 00 00 be 81 00 00 00 48 c7 c7 01 79 c4 92 e8 3d 3e 1f ff 80 a5 8a 00 00 00 fb e9 e1 fd ff ff <0f> 0b 66 2e 0f 1f 84 00 00 00 00 00 0f 1f 44 00 00 90 90 90 90 90
mai 07 09:13:47 peron kernel: RSP: 0018:ffffd19cc968f9e0 EFLAGS: 00010002
mai 07 09:13:47 peron kernel: RAX: 000000000000002c RBX: 0000000000000003 RCX: 0000000000000000
mai 07 09:13:47 peron kernel: RDX: 0000000000000000 RSI: 0000000000000001 RDI: ffff888eb3a1d2c0
[...]

PR #568, opened 2026-05-09
crypto: avoid source staging buffer in aead and skcipher

lneto:
Map AEAD and SKCIPHER inputs through scatterlists pointing directly at
  the Lua input strings, instead of memcpy-ing them into a contiguous
  buffer used for in-place encrypt/decrypt. Reduces transient memory
  pressure, allowing the operations to succeed under GFP_ATOMIC where
  the previous duplicate of the source data could fail (the scenario
  reported in #567 for aead).

  Common boilerplate (IV checking, request alloc, request free) is
  pulled out to `luacrypto.h` so both submodules stay aligned by
  construction. Two regression tests are added covering the empty-AAD
  and empty-output paths exercised by the refactor.

  Counter-proposal to #567: same motivation, applied to skcipher as
  well and structured around shared helpers.

### DIFF
--- a/lib/luacrypto.h
+++ b/lib/luacrypto.h
@@ -30,6 +30,34 @@ static void luacrypto_##name##_release(void *private)		\
 		obj_free(obj);					\
 }
 
+static inline u8 *luacrypto_checkiv(lua_State *L, int idx, size_t expected)
+{
+	size_t iv_len;
+	const char *iv = luaL_checklstring(L, idx, &iv_len);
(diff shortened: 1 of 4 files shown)


lneto:
hey @jperon, I have a counter-proposal.. please let me know what you think; thanks!

jperon (approved):
Once rebased on master (because of #564), it compiles and runs successfully. Thanks !

PR #571, opened 2026-05-22
Add luatc binding

sneaky-potato:
This patch adds the following
- tc constants in autogen
- tc binding for supporting bpf_luatc_run kfunc
- tc related metadata in luaskb
- tc classifier example


autogen/specs.lua
```diff
@@ -27,6 +27,8 @@ return {
 		desc = "Network device notifier event types." },
 	{ header = "uapi/linux/bpf.h",              prefix = "XDP_",       module = "xdp",
 		desc = "XDP verdicts and flags." },
+	{ header = "uapi/linux/pkt_cls.h",          prefix = "TC_",        module = "tc",
+		desc = "TC verdicst and flags." },
```
lneto: Typo

lib/luaskb.c
```diff
@@ -206,23 +255,69 @@ static void luaskb_release(void *private)
...
+	lua_pushvalue(L, 2);
+	lua_rawget(L, -2);
+	if (!lua_isnil(L, -1))
+		return 1;
+	lua_pop(L, 2);
+
+	const char *key = luaL_checkstring(L, 2);
+	for (const luaL_Reg *f = luaskb_getters; f->name; f++)
+		if (strcmp(f->name, key) == 0)
+			return f->func(L);
```
lneto: I think we have discussed this on matrix.. we should have a table instead of comparing strings.. btw, I think a better approach would be to have a table as an field of skb (e.g., `skb.attr`) for using `__index` and `__newindex` so we can avoid the verification on the top and a second table stored in the lib table itself to  access getters and setters CFunctions directly, instead of comparing strings like this. What do you think?

lib/luatc.c
```diff
@@ -0,0 +1,350 @@
...
+__bpf_kfunc_start_defs();
+#else
+__diag_push();
+__diag_ignore_all("-Wmissing-prototypes",
+                  "Global kfuncs as their definitions will be in BTF");
+#endif
+
+static char luatc_env_key;
+
+static lunatik_object_t *luatc_runtimes = NULL;
```
lneto, line 39: I think we should move this to our common layer, having something like `lunatik_ebpf_runtimes`.. btw, I was taking a look at the integration branch.. I don't think the header belongs to `lib/`, `lunatik_*` should be placed at `/`, also.. I would rename it to `lunatik_ebpf.h`, `bpf` is misnomer IMHO

lib/luatc.c
```diff
@@ -0,0 +1,350 @@
...
+*   end
+*   tc.attach(my_packet_processor)
+*
+*   -- In eBPF C code, to call the above Lua function:
+*   -- char rt_key[] = "my_tc_handler.lua"; // Key matches the script name
+*   -- int verdict = bpf_luatc_run(rt_key, sizeof(rt_key), ctx);
+* @see data
+* @within tc
+*/
+static int luatc_attach(lua_State *L)
```
lneto, line 249: isn't there anything from attach on XDP and TC that we can generaliza to the lunatik_ebpf layer?

lib/luaskb.c
lneto: perhaps, a better idea is to just expose getters and setters here and implement the table abstraction in Lua, e.g., `skb/attr.lua`

sneaky-potato: > perhaps, a better idea is to just expose getters and setters here and implement the table abstraction in Lua, e.g., skb/attr.lua

I agree, this way C side will be straight forward

lib/luatc.c
sneaky-potato: Working on a fixup for this

lib/luaskb.c
sneaky-potato: Added `lib/skb/attr.lua` to handle the table abstraction

lib/luatc.c
sneaky-potato: I have moved the following to the generic layer till now:
- runtime lookup
- BPF kfunc init and set macros

lib/skb/attr.lua
```diff
@@ -0,0 +1,47 @@
...
+-- This module provides a higher-level abstraction over the `skb` module.
+--
+-- @module skb.attr
+-- @see skb
+--
+
+local getters = {
+	mark     = function(s) return s:getmark()     end,
+	priority = function(s) return s:getpriority() end,
+}
```
lneto: we don't need this, right.. you can just `_skb["get" .. attr]`, if needed we can memoize this as well

lib/luaskb.c
```diff
@@ -195,6 +195,55 @@ static int luaskb_forward(lua_State *L)
...
+}
+
+static int luaskb_getpriority(lua_State *L)
+{
+	luaskb_t *lskb = luaskb_check(L, 1);
+	lua_pushinteger(L, lskb->skb->priority);
+	return 1;
+}
+
+static int luaskb_setpriority(lua_State *L)
```
lneto: I believe we have a macro for setting integers, if not, we can create one.. the same accounts for getters.. let's avoid repeat code and have "templates" instead

lib/luaskb.c
```diff
@@ -195,6 +195,55 @@ static int luaskb_forward(lua_State *L)
...
+}
+
+static int luaskb_protocol(lua_State *L)
+{
+	luaskb_t *lskb = luaskb_check(L, 1);
+	lua_pushinteger(L, ntohs(lskb->skb->protocol));
+	return 1;
+}
+
+static int luaskb_proto(lua_State *L)
```
lneto: mixed feelings regarding this.. what's special regarding protocol? why exposing only it [from header] through a call? currently, we are typically parsing the header with luadata.

lunatik_ebpf.h
```diff
@@ -0,0 +1,87 @@
...
+		pr_err("couldn't find the context object\n"); \
+		return -1; \
+	} \
+	lunatik_object_t *obj = (lunatik_object_t *)lunatik_toobject((L), -1); \
+	(out_ptr) = obj->private; \
+} while (0)
+
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 7, 0))
+#define LUNATIK_EBPF_START_DEFS() __bpf_kfunc_start_defs()
+#define LUNATIK_EBPF_END_DEFS()   __bpf_kfunc_end_defs()
```
lneto: I'd recommend LUNATIK_EBPF_{START, END}

lunatik_ebpf.h
```diff
@@ -0,0 +1,87 @@
...
+	static const struct btf_kfunc_id_set bpf_lua##subsys##_kfunc_set = { \
+		.owner = THIS_MODULE, \
+		.set   = &bpf_lua##subsys##_set, \
+	};
+
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 4, 0))
+#define LUNATIK_EBPF_KFUNC_INIT(subsys, prog_type) \
+	return register_btf_kfunc_id_set(prog_type, &bpf_lua##subsys##_kfunc_set);
+#else
+#define LUNATIK_EBPF_KFUNC_INIT(subsys, prog_type) return 0;
```
lneto, line 80: question: can't we bundle many of these macros into a single one for lunatik? e.g., LUNATIK_EBPF_KFUNC?

lunatik_ebpf.h
```diff
@@ -0,0 +1,87 @@
...
+	BTF_ID_FLAGS(func, kfunc) \
+	BTF_KFUNCS_END(bpf_lua##subsys##_set)
+#else
+#define LUNATIK_EBPF_KFUNC_SET(subsys, kfunc) \
+	BTF_SET8_START(bpf_lua##subsys##_set) \
+	BTF_ID_FLAGS(func, kfunc) \
+	BTF_SET8_END(bpf_lua##subsys##_set)
+#endif
+
+#define LUNATIK_EBPF_DEFINE_KFUNC_SET(subsys, kfunc) \
```
lneto: I'd at least start all of them with LUNATIK_EBPF_KFUNC_*

lunatik_ebpf.h
```diff
@@ -0,0 +1,87 @@
...
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 9, 0))
+#define LUNATIK_EBPF_KFUNC_SET(subsys, kfunc) \
+	BTF_KFUNCS_START(bpf_lua##subsys##_set) \
+	BTF_ID_FLAGS(func, kfunc) \
+	BTF_KFUNCS_END(bpf_lua##subsys##_set)
+#else
+#define LUNATIK_EBPF_KFUNC_SET(subsys, kfunc) \
+	BTF_SET8_START(bpf_lua##subsys##_set) \
+	BTF_ID_FLAGS(func, kfunc) \
+	BTF_SET8_END(bpf_lua##subsys##_set)
```
lneto: it seems it can be simplified with a intermediary definition for using BTF_KFUNCS_START or BTF_SET8_START.. let's simplify this header the mos we can, avoiding too much code repetition

lunatik_ebpf.h
```diff
@@ -0,0 +1,87 @@
+/*
+* SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+* SPDX-License-Identifier: MIT OR GPL-2.0-only
+*/
+
+#ifndef LUNATIK_BPF_H
+#define LUNATIK_BPF_H
```
lneto: EBPF, right?

lib/luaskb.c
sneaky-potato: Ah, that must be from an older push. We do not need proto for an initial demonstration, I can get rid of this (proto and protocol), let's add this only if we need this in future.

lunatik_ebpf.h
sneaky-potato: These macros are required at different steps inside the binding. The overall picture:
```c
LUNATIK_EBPF_START();
__bpf_kfunc int bpf_luaxdp_run(..) { ... } /* define the kfunc */
LUNATIK_EBPF_END();
LUNATIK_EBPF_KFUNC_DEFINE_SET();   /* define the kfunc set */
...
static int __init luaxdp_init(void)
{
	LUNATIK_EBPF_KFUNC_INIT(xdp, BPF_PROG_TYPE_XDP);
}
```

I can get rid of one macro (LUNATIK_EBPF_KFUNC_SET) since we don't need that exposed and merge it.

lib/luaskb.c
sneaky-potato: I tried using lunatik_setinteger but it caused an "attempt to index a number" error. It appears to be pulling fields from a Lua table on the stack, whereas our setters receive the value directly as a stack argument...

lneto: yes, I see.. it expects a table as argument.. not sure if it's a good idea to leverage it by having something as `skb:setattr{<name> = value}`, e.g., `skb:setattr{priority = 1}`, so we can use the macro directly and perhaps clean our code a bit.. or if it's better to have your own `luaskb_setinteger` macro here to avoid this repetition.. what do you think?

lunatik_ebpf.h
lneto: LGTM

examples/sniclassify/sni.lua
```diff
@@ -0,0 +1,85 @@
...
+					log(sni, classid)
+					skb.priority = classid
+					break
+				end
+			end
+			ctx:action(action.ACT_OK)
+			return
+		end
+		extension = data_off + short(extension + 2)
+	end
```
lneto, line 81: btw, did you copy it from another example? if yes, we should give the credit/copyright

lib/luaskb.c
sneaky-potato: I think we can do our own luaskb_setinteger macro here to avoid repetition. Having a table will make things a bit awkward I think.

examples/sniclassify/sni.lua
sneaky-potato: added a fixup, if you think it's okay, shall we merge this?

examples/sniclassify/sni.lua
```diff
@@ -0,0 +1,86 @@
+--
+-- SPDX-FileCopyrightText: (c) 2024-2026 Ring Zero Desenvolvimento de Software LTDA
```
lneto, line 3: you should stack your copyright on top..

PR #572, opened 2026-06-04
Introduce luatask and luasched

sneaky-potato:
This patch 
- introduces luatask module (to be used in future by luasched)
- refactor existing luathread to start using luatask
- introduce luasched
- add a scheduling example

### DIFF
--- a/lib/luatask.c
+++ b/lib/luatask.c
@@ -1,0 +1,196 @@
+/*
+* SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+* SPDX-License-Identifier: MIT OR GPL-2.0-only
+*/
+
+/***
+* Linux task interface.
+* @module task
+*/
+
+#define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+#include <linux/sched.h>
+#include <linux/sched/task.h>
+#include <linux/kthread.h>
+
+#include "luatask.h"
+
+LUNATIK_PRIVATECHECKER(luatask_check, struct task_struct *,
+	luaL_argcheck(L, private != NULL, ix, "task is not set");
+);
+
+/***
+* Returns the command name (comm) of the task, i.e., the executable.
+* This is truncated to TASK_COMM_LEN (16) characters by the kernel.
+* @function comm
+* @treturn string command name of the task
+*/
+static int luatask_comm(lua_State *L)
+{
+	struct task_struct *task = luatask_check(L, 1);
+	lua_pushstring(L, task->comm);
+	return 1;
+}
+
+/***
+* Returns the process ID (PID) of the task.
+* For threads, this is the thread ID (TID) unique to each thread.
+* @function pid
+* @treturn integer pid of the task
+*/
+static int luatask_pid(lua_State *L)
+{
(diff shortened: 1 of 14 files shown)


lib/luatask.h
```diff
@@ -0,0 +1,20 @@
...
+*/
+
+#ifndef luatask_h
+#define luatask_h
+
+#include <lunatik.h>
+
+typedef struct {
+	struct task_struct *task;
+} luatask_t;
```
lneto: why not using the internal struct directly?

lib/luatask.c
```diff
@@ -0,0 +1,167 @@
...
+{
+	luatask_t *ltask = (luatask_t *)private;
+	if (ltask->task) {
+		put_task_struct(ltask->task);
+		ltask->task = NULL;
+	}
+}
+
+static const luaL_Reg luatask_lib[] = {
+	{NULL, NULL}
```
lneto, line 118: it seems weird to have a lib registered without any function.. why do we need a lib? what's expected on require?

lib/luatask.h
sneaky-potato: Good catch! Made the relevant changes.

lib/luatask.c
sneaky-potato: Hmm, I agree, we can remove this for now. But I was thinking something along the lines of capturing current running task.
Something like
```lua
local task = require("task")
local t = task.current
print(t:pid(), t:comm())
```

But I do not have a usecase for now, maybe working on sched will show more clarity.
EDIT: I realized luathread already captures the current running process.

lneto: it seems a legit usage for me.. perhaps we can move this task handling from luathread to here.. but you should need to evaluate it properly beforehand ;-)

sneaky-potato: For example:
luathread has a current method defined [here](https://github.com/luainkernel/lunatik/blob/8249338d818ab88c200c6936b55f2b88bcef24d2/lib/luathread.c#L203-L220), but the runtime is always NULL, so why not move this to luatask.
Similarly for shouldstop defined [here](https://github.com/luainkernel/lunatik/blob/8249338d818ab88c200c6936b55f2b88bcef24d2/lib/luathread.c#L74-L78) which only works on `current` macro (which is a `task_struct *`)

lneto: I agree on current, but not sure about shouldstop, it seems to be specific to kthreads

sneaky-potato:
Couple of things here:
- I have made changes in autogen.lua as some scx constants are huge, so storing them as numbers, lua is not able to interpret them. So I am storing them in hex format
- Right now, running workload.lua is crashing lunatik. I suspect ownership issues in attach / handler. I will keep working on it.

sneaky-potato:
Hi @lneto 
It looks like sched_ext callbacks can execute in contexts that are not compatible with the BH locking used by the softirq runtime so the hardirq runtime appears to be the correct choice. I was getting this warning before machine freeze:
```sh
WARNING: CPU: X PID: 0 at kernel/softirq.c:429 __local_bh_enable_ip+0x4e/0x70
```

Changing the sched runtime to `LUNATIK_OPT_HARDIRQ` has solved the freeze and warning. Pushed latest changes.

PR #576, opened 2026-06-14
crypto: pre-allocate skcipher/aead request and IV per cipher object (v2)

lneto:
skcipher/aead allocated the request and the IV buffer on every
  encrypt/decrypt. Under memory pressure in softirq context this fails
  with ENOMEM (notably `ecb` during QUIC decryption in ipparse).

  Hold the tfm, the request and the IV buffer (all fixed-size for a given
  tfm) in a `luacrypto_ctx_t` allocated once at creation time. The hot
  path no longer performs any fallible allocation besides the output
  buffer. The runtime lock already serializes operations on a cipher
  object, so the request and IV are safe to reuse across calls.

  The skcipher output buffer gains one byte for the NUL terminator written
  by `lunatik_pushstring()`, fixing a one-byte overflow on the previous
  exact-size allocation.

  The shared context lets skcipher and aead generate their `checkctx`/`check`
  accessors from a single `LUACRYPTO_CHECKER` macro instead of repeating the
  boilerplate in each file.

  Adds skcipher/aead stress round-trip tests.

### DIFF
--- a/lib/luacrypto.h
+++ b/lib/luacrypto.h
@@ -38,6 +38,14 @@ typedef struct luacrypto_ctx_s {
 	u8 *iv;
 } luacrypto_ctx_t;
 
+/* checkctx returns the context; check returns ctx->tfm for tfm-only methods. */
+#define LUACRYPTO_CHECKER(name, T)							\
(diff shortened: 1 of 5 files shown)


jperon (approved):
Just tested : working for me, LGTM.

PR #587, opened 2026-06-21
Add ebpf maps binding

sneaky-potato:
Draft of the ebpf maps support

1. Create the map using bpftool and pin it to bpffs
```sh
$ bpftool map create /sys/fs/bpf/flow_cache type hash key 3 value 3 entries 128 \
name flow_cache
```

2. Verify the map
```sh
$ bpftool map show pinned /sys/fs/bpf/flow_cache
```

3. Update the map and add a key
```sh
$ bpftool map update \
    pinned /sys/fs/bpf/flow_cache \
    key hex 66 6f 6f \
    value hex 62 61 72
```

4. Lookup for this key to verify update
```sh
$ bpftool map lookup \
    pinned /sys/fs/bpf/flow_cache \
    key hex 66 6f 6f
```

5. Now we can use the map via bpffs path
```lua
local map = require("ebpf.map")

local flow = map.open("/sys/fs/bpf/flow_cache")

local val = flow:lookup("foo")

if val then
    print(val)
end

-- flow:update(key, val)
-- flow:delete(key)
```

### DIFF
--- a/lib/luaebpf_map.c
+++ b/lib/luaebpf_map.c
@@ -1,0 +1,316 @@
+/*
+* SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+* SPDX-License-Identifier: MIT OR GPL-2.0-only
+*/
+
+/***
+* Lua interface for eBPF maps.
+*
+* This module provides generic key-value access to pinned eBPF maps,
+* supporting types that implement standard lookup, update, deletion,
+* and iteration operations, including:
+*   - BPF_MAP_TYPE_HASH
+*   - BPF_MAP_TYPE_ARRAY
+*   - BPF_MAP_TYPE_LRU_HASH
+*
+* Keys and values are exchanged as packed Lua strings whose sizes must
+* match the map's configured key and value sizes.
+*
+* @module ebpf.map
+*/
+
+#define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+
+#include <linux/bpf.h>
+#include <linux/namei.h>
+#include <linux/fs.h>
+
+#include <lunatik.h>
+
+LUNATIK_PRIVATECHECKER(luaebpf_map_check, struct bpf_map *,
+	luaL_argcheck(L, private != NULL, ix, "bpf map is not set");
+	/* map->ops should be set on creation; crash if broken */
+	BUG_ON(!private->ops);
+);
+
+#define luaebpf_map_checksize(L, size, expected, ix, arg) \
+	luaL_argcheck((L), (size) == (expected), (ix), "invalid " arg " size")
+
+#define luaebpf_map_checkoptsize(L, size, expected, ix, arg) \
+    luaL_argcheck((L), ((size) == (expected)) || ((size) == 0), (ix), "invalid " arg " size")
+
+#define luaebpf_map_is_valid_update_flag(flags) \
+	((flags) == BPF_ANY || (flags) == BPF_NOEXIST || (flags) == BPF_EXIST)
+
+#define luaebpf_map_push_clean(L, ret, buffer, buffer_size) \
+do { 												\
+	if ((ret) == 0) 								\
+		lua_pushlstring(L, buffer, buffer_size); 	\
+	kfree(buffer); 									\
+	if (ret == -ENOENT) 							\
+		lua_pushnil(L); 							\
+	else if (ret) 									\
+		lunatik_throw(L, ret); 						\
+} while (0)
+
+static int luaebpf_map_pushresult(lua_State *L, long ret)
+{
+	if (ret == -ENOENT) {
+		lua_pushboolean(L, 0);
+		return 1;
+	}
+	if (ret)
+		lunatik_throw(L, ret);
+	lua_pushboolean(L, 1);
+	return 1;
+}
+
+static void luaebpf_map_release(void *private);
+
+static inline struct bpf_map *luaebpf_map_frompath(const struct path *path)
+{
+	struct inode *inode = d_inode(path->dentry);
+	return inode ? inode->i_private : NULL;
+}
+
+static struct bpf_map *luaebpf_map_get(const char *pathname)
+{
+	struct path path;
+
+	if (kern_path(pathname, LOOKUP_FOLLOW, &path))
+		return NULL;
+
+	struct bpf_map *map = luaebpf_map_frompath(&path);
+	if (map)
+		bpf_map_inc(map);
+
+	path_put(&path);
+	return map;
+}
+
+/**
+* Looks up a key from the map.
+* @function lookup
+* @tparam map map
+* @tparam string key Packed key
+* @treturn string value Packed value, or `nil` if the key is not present
+*/
+static int luaebpf_map_lookup(lua_State *L)
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+	luaebpf_map_checksize(L, key_size, map->key_size, 2, "key");
+
+	void *value = map->ops->map_lookup_elem(map, (void *)key);
+	if (!value) {
+		lua_pushnil(L);
+	} else {
+		lua_pushlstring(L, value, map->value_size);
+	}
+	return 1;
+}
+
+/**
+* Updates a key in the map.
+* @function update
+* @tparam map map
+* @tparam string key Packed key.
+* @tparam string value Packed value.
+* @tparam[opt=BPF_ANY] integer flags Update flags. One of:
+*   - `BPF_ANY` (default): Create a new element or update an existing one.
+*   - `BPF_NOEXIST`: Create a new element only if the key does not exist.
+*   - `BPF_EXIST`: Update an existing element only if the key exists.
+* @treturn boolean success
+* @raise Error if the operation is not permitted by the map.
+*/
+static int luaebpf_map_update(lua_State *L)
+{
+	struct bpf_map *map= luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+	size_t value_size;
+	const char *value = luaL_checklstring(L, 3, &value_size);
+
+	luaebpf_map_checksize(L, key_size, map->key_size, 2, "key");
+	luaebpf_map_checksize(L, value_size, map->value_size, 3, "value");
+	u64 flags = luaL_optinteger(L, 4, BPF_ANY);
+
+	if (!luaebpf_map_is_valid_update_flag(flags))
+		luaL_argerror(L, 4, "invalid update flag");
+
+	long ret = map->ops->map_update_elem(map, (void *)key, (void *)value, flags);
+	return luaebpf_map_pushresult(L, ret);
+}
+
+/**
+* Deletes a key from the map.
+* @function delete
+* @tparam map map
+* @tparam string key Packed key.
+* @treturn boolean success
+* @raise Error if the operation is not permitted by the map.
+*/
+static int luaebpf_map_delete(lua_State *L)
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+
+	luaebpf_map_checksize(L, key_size, map->key_size, 2, "key");
+
+	long ret = map->ops->map_delete_elem(map, (void *)key);
+	return luaebpf_map_pushresult(L, ret);
+}
+
+/**
+* Looks up and deletes a key from the map.
+* @function remove
+* @tparam map map
+* @tparam string key Packed key.
+* @treturn string value Packed value, or `nil` if the key is not present.
+* @raise Error if the operation is not permitted by the map or memory allocation fails.
+*/
+static int luaebpf_map_remove(lua_State *L)
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+
+	luaebpf_map_checksize(L, key_size, map->key_size, 2, "key");
+
+	char *value = lunatik_checkalloc(L, map->value_size);
+	int ret = map->ops->map_lookup_and_delete_elem(map, (void *)key, (void *)value, 0);
+
+	luaebpf_map_push_clean(L, ret, value, map->value_size);
+
+	return 1;
+}
+
+/**
+* Returns the next key in the map.
+* @function next
+* @tparam map map
+* @tparam[opt] string key Packed key, returns the first key if nothing is passed.
+* @treturn string next_key Packed next key, or `nil` if there are no more keys.
+* @raise Error if the operation is not permitted by the map or memory allocation fails.
+* @usage
+*   local map = require("ebpf.map")
+*   local flow = map.open("/sys/fs/bpf/flow_cache")
+*   local key = flow:next()
+*   while key do
+*   	print(key)
+*   	key = flow:next(key)
+*   end
+*   flow:close()
+*/
+static int luaebpf_map_get_next_key(lua_State *L)
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_optlstring(L, 2, NULL, &key_size);
+
+	luaebpf_map_checkoptsize(L, key_size, map->key_size, 2, "key");
+	void *next_key = lunatik_checkalloc(L, map->key_size);
+	long ret = map->ops->map_get_next_key(map, (void *)key, next_key);
+
+	luaebpf_map_push_clean(L, ret, next_key, map->key_size);
+
+	return 1;
+}
+
+/**
+* Releases the map reference.
+* @function close
+*/
+static int luaebpf_map_close(lua_State *L)
+{
+	lunatik_object_t *obj = lunatik_checkobject(L, 1);
+	struct bpf_map *map = obj->private;
+	if (map) {
+		obj->private = NULL;
+		bpf_map_put(map);
+	}
+	return 0;
+}
+
+static const luaL_Reg luaebpf_map_mt[] = {
+	{"lookup", luaebpf_map_lookup},
+	{"update", luaebpf_map_update},
+	{"delete", luaebpf_map_delete},
+	{"remove", luaebpf_map_remove},
+	{"next",   luaebpf_map_get_next_key},
+	{"close",  luaebpf_map_close},
+	{"__gc",   lunatik_deleteobject},
+	{NULL, NULL}
+};
+
+LUNATIK_OPENER(ebpf_map);
+static const lunatik_class_t luaebpf_map_class = {
+	.name = "ebpf_map",
+	.methods = luaebpf_map_mt,
+	.release = luaebpf_map_release,
+	.opener = luaopen_ebpf_map,
+	.opt = LUNATIK_OPT_EXTERNAL,
+};
+
+/**
+* Opens a map from pinned bpffs path.
+*
+* @function open
+* @tparam string path Path to a pinned eBPF map.
+* @treturn map map Opened map handle.
+* @usage
+*   local map = require("ebpf.map")
+*   local bpf = require("linux.bpf")
+*   local counter = map.open("/sys/fs/bpf/counters")
+*   -- 32-bit key/value encoded as packed strings.
+*   local key = string.pack("I4", 1)
+*   local value = string.pack("I4", 42)
+*   assert(counter:update(key, value, bpf.ANY))
+*   local result = counter:lookup(key)
+*   if result then
+*       print(string.unpack("I4", result))
+*   end
+*   counter:delete(key)
+*   counter:close()
+*/
+static int luaebpf_map_open(lua_State *L)
+{
+	const char *pathname = luaL_checkstring(L, 1);
+	struct bpf_map *map = luaebpf_map_get(pathname);
+	luaL_argcheck(L, map != NULL, 1, "failed to open bpf map");
+	lunatik_object_t *object = lunatik_newobject(L, &luaebpf_map_class, 0, LUNATIK_OPT_NONE);
+	object->private = map;
+	return 1;
+}
+
+static const luaL_Reg luaebpf_map_lib[] = {
+	{"open", luaebpf_map_open},
+	{NULL, NULL},
+};
+
+static void luaebpf_map_release(void *private)
+{
+	struct bpf_map *map = (struct bpf_map *)private;
+	if (map)
+		bpf_map_put(map);
+}
+
+LUNATIK_CLASSES(ebpf_map, &luaebpf_map_class);
+LUNATIK_NEWLIB(ebpf_map, luaebpf_map_lib, luaebpf_map_classes);
+
+static int __init luaebpf_map_init(void)
+{
+	return 0;
+}
+
+static void __exit luaebpf_map_exit(void)
+{
+}
+
+module_init(luaebpf_map_init);
+module_exit(luaebpf_map_exit);
+MODULE_LICENSE("Dual MIT/GPL");
+MODULE_AUTHOR("Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>");
+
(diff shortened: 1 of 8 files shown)


lib/luaebpf_map.c
```diff
@@ -0,0 +1,222 @@
+/*
+* SPDX-FileCopyrightText: (c) 2024-2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
```
lneto: why this date?

lib/luaebpf_map.c
```diff
@@ -0,0 +1,222 @@
...
+static struct bpf_map *luaebpf_map_get(const char *pathname)
+{
+	struct path path;
+	struct inode *inode;
+	struct bpf_map *map = NULL;
+
+	if (kern_path(pathname, LOOKUP_FOLLOW, &path))
+		return NULL;
+
+	inode = d_inode(path.dentry);
```
lneto: it's up to you, but we are trying to use C99 decl on lunatik.. that is, declaring variables on use

lib/luaebpf_map.c
```diff
@@ -0,0 +1,222 @@
...
+
+	if (inode)
+		map = inode->i_private;
+
+	if (map)
+		bpf_map_inc(map);
```
lneto: perhaps..

```suggestion
	if (((inode = d_inode(path.dentry) != NULL) && ((map = inode->i_private) != NULL))
		bpf_map_inc(map);		
```

in that, I'd use C99 decl.. it's up to you.. both ways are good to me.. =)

lib/luaebpf_map.c
```diff
@@ -0,0 +1,222 @@
...
+
+	void *value = map->ops->map_lookup_elem(map, (void *)key);
+
+	if (!value) {
+		lua_pushnil(L);
+		return 1;
+	}
+
+	lua_pushlstring(L, value, map->value_size);
+
```
lneto:
```suggestion
```

lib/luaebpf_map.c
```diff
@@ -0,0 +1,222 @@
...
+	luaL_argcheck(L, key_size == map->key_size, 2, "invalid key size");
+
+	void *value = map->ops->map_lookup_elem(map, (void *)key);
+
+	if (!value) {
+		lua_pushnil(L);
+		return 1;
+	}
+
+	lua_pushlstring(L, value, map->value_size);
```
lneto:
```suggestion
	if (!value)
		lua_pushnil(L);
	else
		lua_pushlstring(L, value, map->value_size);
```

lib/luaebpf_map.c
```diff
@@ -0,0 +1,222 @@
...
+ */
+static int luaebpf_map_lookup(lua_State *L)
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+
+	luaL_argcheck(L, key_size == map->key_size, 2, "invalid key size");
+
+	void *value = map->ops->map_lookup_elem(map, (void *)key);
```
lneto, line 105: is this safe?

lib/luaebpf_map.c
```diff
@@ -0,0 +1,222 @@
...
+ * @tparam string key Packed key
+ * @treturn string value Packed value, or `nil` if the key is not present
+ */
+static int luaebpf_map_lookup(lua_State *L)
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+
+	luaL_argcheck(L, key_size == map->key_size, 2, "invalid key size");
```
lneto: also this.. is this safe?

lib/luaebpf_map.c
```diff
@@ -0,0 +1,222 @@
...
+	size_t value_size;
+	const char *value = luaL_checklstring(L, 3, &value_size);
+
+	luaL_argcheck(L, key_size == map->key_size, 2, "invalid key size");
+	luaL_argcheck(L, value_size == map->value_size, 3, "invalid value size");
+
+	int ret = map->ops->map_update_elem(map, (void *)key, (void *)value, BPF_ANY);
+
+	lua_pushboolean(L, ret == 0);
+
```
lneto, line 141:
```suggestion
```

lib/luaebpf_map.c
```diff
@@ -0,0 +1,222 @@
...
+	const char *key = luaL_checklstring(L, 2, &key_size);
+
+	long ret;
+
+	luaL_argcheck(L, key_size == map->key_size, 2, "invalid key size");
+
+	ret = map->ops->map_delete_elem(map, (void *)key);
+
+	lua_pushboolean(L, ret == 0);
+
```
lneto, line 189:
```suggestion
```

lib/luaebpf_map.c
```diff
@@ -0,0 +1,222 @@
...
+ */
+static int luaebpf_map_delete(lua_State *L)
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+
+	long ret;
+
+	luaL_argcheck(L, key_size == map->key_size, 2, "invalid key size");
```
lneto: we can create a macro for this argcheck.. I'd look for similar cases as well..

lib/luaebpf_map.c
```diff
@@ -0,0 +1,222 @@
...
+ * @tparam string key Packed key.
+ * @treturn boolean success
+ */
+static int luaebpf_map_delete(lua_State *L)
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+
+	long ret;
```
lneto: C99 decl would be better here, IMO

lib/luaebpf_map.c
```diff
@@ -0,0 +1,222 @@
...
+ * @tparam map map
+ * @tparam string key Packed key.
+ * @treturn boolean success
+ */
+static int luaebpf_map_delete(lua_State *L)
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+
```
lneto, line 159:
```suggestion
```

lib/luaebpf_map.c
```diff
@@ -0,0 +1,222 @@
...
+	.methods = luaebpf_map_mt,
+	.release = luaebpf_map_release,
+	.opener = luaopen_ebpf_map,
+	.opt = LUNATIK_OPT_SOFTIRQ | LUNATIK_OPT_SINGLE,
+};
+
+/**
+ * Opens a map from path.
+ *
+ * Currently only hash maps (`BPF_MAP_TYPE_HASH`) are supported.
```
lneto: what would take to support other types?

lneto: and what happens if we try to access a different map type?

lib/luaebpf_map.c
```diff
@@ -0,0 +1,222 @@
...
+	{NULL, NULL}
+};
+
+LUNATIK_OPENER(ebpf_map);
+static const lunatik_class_t luaebpf_map_class = {
+	.name = "ebpf_map",
+	.methods = luaebpf_map_mt,
+	.release = luaebpf_map_release,
+	.opener = luaopen_ebpf_map,
+	.opt = LUNATIK_OPT_SOFTIRQ | LUNATIK_OPT_SINGLE,
```
lneto: wouldn't be useful to share a map across runtimes? or would it be better to open it separately? what's the rational here?

sneaky-potato: the map itself has its own lifetime management. I understand that LUNATIK_OPT_SINGLE only prevents sharing of the lunatik wrapper object. If the users can simply open the map by path in each runtime, then each runtime gets its own lunatik wrapper. How will sharing the wrapper object provide benefit?

I can already access the map via open in two different Lua scripts. (Just trying to understand if we'll have cross runtime synchronization issues if marked sharable)

lib/luaebpf_map.c
sneaky-potato: code does not actually enforce this fact. Right now opwn will accept any pinned BPF map. And since methods are generic they would succeed too. There are some conditions: like ARRAY map does not support delete. And then certain maps like `BPF_MAP_TYPE_QUEUE` uses push / pop instead of lookup, update.

> what would take to support other types?

I don't think it would be that hard to support other types. Jus need to add map specific restrictions and we'd be good to go.

lneto: but would will fail safely over it? what would really happen if the user try to open and use an ARRAY, for instance? Also, would it change the API to support it?

sneaky-potato: > but would will fail safely over it? what would really happen if the user try to open and use an ARRAY, for instance? Also, would it change the API to support it?

the API fails safely if the map like ARRAY does not support delete method. 
specifically the following script prints `false`
```lua
print("delete:", flow:delete(key)) -- flow is array type map, prints fals
```

lib/luaebpf_map.c
sneaky-potato: The binding verifies that the Lua string length exactly matches the map's key size before passing it to the map operations. Lua guarantees the buffer is valid for that length, so the kernel should never read beyond the buffer. But do you have any scenarios in mind where this operation might cause trouble?

lneto: I mean assuming the pointer is not NULL

sneaky-potato: > I mean assuming the pointer is not NULL

I think that always will be ensured because of `luaebpf_map_check` call to retrieve the map pointer (if map is NULL it will raise an error: bpf map is not set)

lneto: please, just double check

lib/luaebpf_map.c
lneto: what would be the error message? and what about the other opertations? will them succeed? we need to think the API through and cover all the usages

sneaky-potato: Okay so there are errors which the delete operation may return. -ENOENT if key is missing and -EINVAL if the operation is not supported (example using delete on array bpf map). Our Lua binding should respect these errors.

lib/luaebpf_map.c
```diff
@@ -0,0 +1,303 @@
...
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+
+	luaL_argcheck(L, key_size == map->key_size, 2, "invalid key size");
+
+	char *value = kmalloc(map->value_size, GFP_KERNEL);
+	if (!value) {
+		return luaL_error(L, "out of memory allocating value buffer");
+	}
```
lneto:
```suggestion
	if (!value)
		luaL_error(L, "out of memory allocating value buffer");
```

we don't use the idiom `return luaL_error(...)` on lunatik because kbuild warnings

lib/luaebpf_map.c
```diff
@@ -0,0 +1,303 @@
...
+ */
+static int luaebpf_map_lookup_and_delete(lua_State *L)
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+
+	luaL_argcheck(L, key_size == map->key_size, 2, "invalid key size");
+
+	char *value = kmalloc(map->value_size, GFP_KERNEL);
```
lneto: why not using lunatik_checkalloc here?

lib/luaebpf_map.c
```diff
@@ -0,0 +1,303 @@
...
+}
+
+/**
+ * Looks up and deletes a key from the map.
+ * @function lookup_and_delete
+ * @tparam map map
+ * @tparam string key Packed key.
+ * @treturn string value Packed value, or `nil` if the key is not present.
+ */
+static int luaebpf_map_lookup_and_delete(lua_State *L)
```
lneto: can't we call this "remove"? I don't think we need to have 1:1 wrappers

lib/luaebpf_map.c
```diff
@@ -0,0 +1,303 @@
...
+
+	if (ret == -ENOENT) {
+		kfree(value);
+		lua_pushnil(L);
+		return 1;
+	}
+
+	if (ret) {
+		kfree(value);
+		return luaL_error(L, "lookup_and_delete failed: %ld", ret);
```
lneto: why not using lunatik_error?

lib/luaebpf_map.c
```diff
@@ -0,0 +1,303 @@
...
+	luaL_argcheck(L, key_size == map->key_size, 2, "invalid key size");
+
+	char *value = kmalloc(map->value_size, GFP_KERNEL);
+	if (!value) {
+		return luaL_error(L, "out of memory allocating value buffer");
+	}
+	int ret = map->ops->map_lookup_and_delete_elem(map, (void *)key, (void *)value, 0);
+
+	if (ret == -ENOENT) {
+		kfree(value);
```
lneto: I would use goto for error handling..

lib/luaebpf_map.c
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	if (ret) {
+		kfree(next_key);
+		return luaL_error(L, "get_next_key failed: %ld", ret);
+	}
+
+	lua_pushlstring(L, next_key, map->key_size);
+	kfree(next_key);
+	return 1;
```
lneto: let's avoid repeating code like this.. let's use aux functions or macro templates

lib/luaebpf_map.c
```diff
@@ -0,0 +1,303 @@
...
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+
+	luaL_argcheck(L, key_size == map->key_size, 2, "invalid key size");
+
+	char *value = kmalloc(map->value_size, GFP_KERNEL);
+	if (!value) {
+		return luaL_error(L, "out of memory allocating value buffer");
```
lneto: why not returning the standard allocation error from lunatik? btw, I would review this entire module looking for the missed APIs from the framework, avoiding re-inventing the wheel.

lib/luaebpf_map.c
```diff
@@ -0,0 +1,303 @@
...
+static struct bpf_map *luaebpf_map_get(const char *pathname)
+{
+	struct path path;
+
+	if (kern_path(pathname, LOOKUP_FOLLOW, &path))
+		return NULL;
+
+	struct inode *inode = NULL;
+	struct bpf_map *map = NULL;
+	if (((inode = d_inode(path.dentry)) != NULL) && ((map = inode->i_private) != NULL))
```
lneto: I would move this predicate to a macro or an inline, even if we use it only here it would make the code more readable, IMO

lib/luaebpf_map.c
```diff
@@ -0,0 +1,303 @@
...
+ * @treturn string next_key Packed next key, or `nil` if there are no more keys.
+ */
+static int luaebpf_map_get_next_key(lua_State *L)
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_optlstring(L, 2, NULL, &key_size);
+	if (key != NULL) {
+		luaL_argcheck(L, key_size == map->key_size, 2, "invalid key size");
+	}
```
lneto: please, mind our code style

lib/luaebpf_map.c
```diff
@@ -0,0 +1,303 @@
...
+ * @function get_next_key
+ * @tparam map map
+ * @tparam[opt] string key Packed key, or `nil` for the first key.
+ * @treturn string next_key Packed next key, or `nil` if there are no more keys.
+ */
+static int luaebpf_map_get_next_key(lua_State *L)
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_optlstring(L, 2, NULL, &key_size);
```
lneto, line 211: can't we use "" instead of NULL here? we would avoid the if below

lneto: we can move the test to the argcheck call anyway.. != NULL && ..., right?

lib/luaebpf_map.c
```diff
@@ -0,0 +1,303 @@
...
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+
+	luaL_argcheck(L, key_size == map->key_size, 2, "invalid key size");
+
+	long ret = map->ops->map_delete_elem(map, (void *)key);
+	if (ret == -EINVAL) {
+		luaL_error(L, "operation not permitted");
+	}
```
lneto: lunatik_error?

lib/luaebpf_map.c
```diff
@@ -0,0 +1,303 @@
...
+		return luaL_error(L, "lookup_and_delete failed: %ld", ret);
+	}
+	lua_pushlstring(L, value, map->value_size);
+	kfree(value);
+	return 1;
+}
+
+/**
+ * Returns the next key in the map.
+ * Passing `nil` returns the first key.
```
lneto: it's not new, it's none, right?

lib/luaebpf_map.c
sneaky-potato: Yes you are right, the kernel panics right now with this script
```lua
local flow = map.open("/sys/fs/bpf/flow_cache")
-- flow:update("fool", "bazk")
flow.update(nil, "fool", "bazk")
```

I think we need a patch in privatecheck macro itself to handle this. It assumes obj is non null and dereferences the private field, [ref](https://github.com/luainkernel/lunatik/blob/4f2007f8d93366d118b6b8491caf4e70bca47bfd/lunatik.h#L338)

lib/luaebpf_map.c
sneaky-potato: That's a good idea! will push a fixup to handle this

lib/luaebpf_map.c
sneaky-potato: > can't we use "" instead of NULL here? we would avoid the if below

The kernel API expects a NULL in the method, but we can accept key_size = 0 in argcheck
```c
	size_t key_size;
	const char *key = luaL_optlstring(L, 2, NULL, &key_size);
	luaL_argcheck(L, (key_size == map->key_size) || (key_size == 0), 2, "invalid key size");

```

lib/luaebpf_map.c
sneaky-potato: Handled in latest fixup

lib/luaebpf_map.c
sneaky-potato: resolved in latest push

sneaky-potato:
Resolved merge conflicts.

lib/luaebpf_map.c
```diff
@@ -0,0 +1,311 @@
...
+*   	key = flow:get_next_key(key)
+*   end
+*   flow:close()
+*/
+static int luaebpf_map_get_next_key(lua_State *L)
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_optlstring(L, 2, NULL, &key_size);
+	luaL_argcheck(L, (key_size == map->key_size) || (key_size == 0), 2, "invalid key size");
```
lneto: I would move this condition to a macro.. it would be more readable

lib/luaebpf_map.c
```diff
@@ -0,0 +1,311 @@
...
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+	size_t value_size;
+	const char *value = luaL_checklstring(L, 3, &value_size);
+
+	luaebpf_map_checksize(L, key_size, map->key_size, 2, "key");
+	luaebpf_map_checksize(L, value_size, map->value_size, 3, "value");
+	u64 flags = luaL_optinteger(L, 4, BPF_ANY);
+
+	if (flags != BPF_ANY && flags != BPF_NOEXIST && flags != BPF_EXIST)
```
lneto: I would move this predicate to a macro

lib/luaebpf_map.c
```diff
@@ -0,0 +1,311 @@
...
+	lua_pushboolean(L, 1);
+	return 1;
+}
+
+#define luaebpf_map_checkret(L, ret, cleanup) \
+do { \
+	if ((ret) == -ENOENT) { \
+		cleanup; \
+		lua_pushnil(L); \
+		return 1; \
```
lneto: we're not using macros as this.. shadowing the function return..

lneto: also, please mind the `\` alignment on macros like this..

lib/luaebpf_map.c
```diff
@@ -0,0 +1,311 @@
...
+	}
+	return 0;
+}
+
+static const luaL_Reg luaebpf_map_mt[] = {
+	{"lookup",       luaebpf_map_lookup},
+	{"update",       luaebpf_map_update},
+	{"delete",       luaebpf_map_delete},
+	{"remove",       luaebpf_map_remove},
+	{"get_next_key", luaebpf_map_get_next_key},
```
lneto: next?

lib/luaebpf_map.c
```diff
@@ -0,0 +1,311 @@
...
+#define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+
+#include <linux/bpf.h>
+#include <linux/namei.h>
+#include <linux/fs.h>
+
+#include <lunatik.h>
+
+LUNATIK_PRIVATECHECKER(luaebpf_map_check, struct bpf_map *,
+	luaL_argcheck(L, private != NULL, ix, "bpf map is not set");
```
lneto, line 31: shouldn't we check if ops is not NULL as well? is it guaranteed to be set?

lib/luaebpf_map.c
sneaky-potato: Simplified the macro and adjusted `\` alignment

lib/luaebpf_map.c
sneaky-potato: I think `map->ops` is going to be non-NULL for any successfully allocated map in the kernel, [source](https://elixir.bootlin.com/linux/v7.1.2/source/kernel/bpf/syscall.c#L1505). I think kernel prevents an unconfigured map object from ever existing.

lneto: perhaps we can just add a BUG_ON to sleep well ;-)

sneaky-potato: fair enough, I've added a bug on just in case.

lib/luaebpf_map.c
marcelstanley: @sneaky-potato 👆🏼

sneaky-potato:
Rebased and cleaned the commits

lib/luaebpf_map.c
```diff
@@ -0,0 +1,317 @@
...
+	/* map->ops should be set on creation; crash if broken */
```
lneto, line 32: I don't think we need this comment.. the macro below is clear enough

lib/luaebpf_map.c
```diff
@@ -0,0 +1,317 @@
...
+#define luaebpf_map_push_clean(L, ret, buffer, buffer_size) \
+do { 												\
+	if ((ret) == 0) 								\
+		lua_pushlstring(L, buffer, buffer_size); 	\
+	kfree(buffer); 									\
+	if (ret == -ENOENT) 							\
+		lua_pushnil(L); 							\
+	else if (ret) 									\
+		lunatik_throw(L, ret); 						\
+} while (0)
```
lneto, line 54: I think this would be cleaner as an inline.. why does this need to be a macro?

lib/luaebpf_map.c
```diff
@@ -0,0 +1,317 @@
...
+{
+	struct bpf_map *map = luaebpf_map_check(L, 1);
+	size_t key_size;
+	const char *key = luaL_checklstring(L, 2, &key_size);
+	luaebpf_map_checksize(L, key_size, map->key_size, 2, "key");
+
+	void *value = map->ops->map_lookup_elem(map, (void *)key);
+	if (!value) {
+		lua_pushnil(L);
+	} else {
```
lneto, line 108: mind our style

lneto:
#630

PR #588, opened 2026-06-28
netlink: AF_NETLINK support (socket, rtnetlink, generic netlink)

lneto:
Adds AF_NETLINK support to Lunatik, split across a thin C module and Lua helpers.

## C module (`lib/luanetlink.c`)

Two object classes:

- `netlink` — sleepable AF_NETLINK socket: `new(proto)`, `send`, `recv`, `close`.
- `netlink.channel` — kernel-side multicast channel whose `broadcast` is
  softirq-safe (netfilter, XDP), for kernel-to-userspace telemetry (same
  mechanism as NFLOG). Created from a sleepable context, usable from a shared
  softirq runtime.

Message building and parsing is left to Lua; the C layer only handles socket
lifecycle and raw send/recv.

## Constants (`autogen/specs.lua`)

Protocol numbers, flags and attribute types are generated from kernel headers
into `linux.netlink`, `linux.rtnetlink` and `linux.genl` (the `IFLA_` set is
curated to top-level link attributes to avoid per-driver nested enums).

## Lua modules

- `netlink.message` — shared nlmsghdr/nlattr (de)serialization: attribute
  builders, message encoder, stateless iterators over messages and attributes,
  a sequence counter, multipart receive and error checking.
- `netlink.rt` — rtnetlink: `route_dump`, `link_dump`, `addr_dump`,
  `route_add`, `route_del`.
- `netlink.genl` — generic netlink: cached family resolution, single and dump
  command dispatch.

## Tests

`tests/netlink/` exercises the read-only paths against loopback and the generic
netlink controller: `link_dump`, `addr_dump`, `route_dump`, `genl_family`.
Registered in the suite runner.

## Notes

- `netlink.channel`'s softirq broadcast path is not yet covered by a test (needs
  a netfilter hook plus a userspace subscriber).
- nl80211 is left as a follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #590, opened 2026-06-28
netlink: generic netlink (netlink.genl)

lneto:
Second of three stacked PRs splitting netlink (#588). **Based on the socket-core PR**, not master.

Adds `netlink.genl`: opens a `NETLINK_GENERIC` socket, resolves family names to ids, and dispatches single and dump commands, parsing each response into `{cmd, attrs}` messages. Constants come from `linux.genl`. Includes a `genl_family` test resolving the always-present `nlctrl` controller.


lneto:

PR #589, opened 2026-06-28
netlink: AF_NETLINK socket core + rtnetlink

lneto:
First of three stacked PRs splitting the original netlink support (#588) for easier review.

Adds the AF_NETLINK socket core and the rtnetlink layer:

- `lib/luanetlink.c` — a thin `netlink` socket class (`new`/`send`/`recv`/`close`); message building/parsing is left to Lua.
- `autogen/specs.lua` — `linux.netlink` (proto/flag/type) and `linux.rtnetlink` (rtm/rta/rtn/rtprot/scope/table/ifla/ifa) constants from kernel headers.
- `lib/netlink/message.lua` — shared `nlmsghdr`/`nlattr` (de)serialization.
- `lib/netlink/rt.lua` — rtnetlink: `route_dump`/`link_dump`/`addr_dump`/`route_add`/`route_del`.
- `tests/netlink/` — read-only dumps against loopback and an add/delete round-trip.

Stacked follow-ups (based on this branch): generic netlink, and `netlink.channel`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #592, opened 2026-06-29
netlink: AF_NETLINK socket core

lneto:
Splits the former core PR (#589) into smaller, self-contained pieces. **This is the base of the stack.**

Adds the raw AF_NETLINK socket:

- `lib/luanetlink.c` — a thin `netlink` socket class (`new`/`send`/`recv`/`close`); message building/parsing is left to Lua.
- `autogen/specs.lua` — `linux.netlink` (proto/flag/type) constants from kernel headers.
- `tests/netlink/socket` — opens a `NETLINK_ROUTE` socket and round-trips a hand-built `RTM_GETLINK` dump.

Stacked on top of this: the shared message module, then rtnetlink.


lneto:

PR #594, opened 2026-06-29
netlink: rtnetlink module (netlink.rt)

lneto:
Part of the #589 split. **Based on the message PR.**

Adds `netlink.rt`: rtnetlink over the netlink socket and message helpers — `route_dump`/`link_dump`/`addr_dump` (returning plain lists) and `route_add`/`route_del`. Plus `linux.rtnetlink` constants. Tests: dumps against loopback and an add/delete round-trip on an isolated table.


lneto:

PR #606, opened 2026-07-02
socket: add setsockopt

### LINKED ISSUE #603: luasocket: bounded/interruptible receive (setsockopt + SO_RCVTIMEO)
`socket:receive()` blocks with no way to bound the wait: `luasocket` exposes no `setsockopt`, so there is no `SO_RCVTIMEO`. A blocking receive on a socket that never gets data blocks the caller indefinitely — harmless for request/response with the kernel (the reply always comes), but it means a blocking receive cannot be made interruptible, e.g. inside a `spawn`ed kernel thread (which must poll `thread.shouldstop()`), and offers no defense against a pathological no-reply.

This is a general socket concern (TCP/UDP/UNIX/netlink alike), surfaced while auditing the netlink stack (`netlink.rt`/`netlink.genl` do blocking request/response). Proposal: add `socket:setsockopt` (or a receive-timeout option) to `luasocket` so a receive can be time-bounded, enabling interruptible/non-hanging receives for all blocking socket users.

Note: `MSG_DONTWAIT` is already available via the `receive` flags for a non-blocking poll, but a timeout (`SO_RCVTIMEO`) is the ergonomic fit for bounded blocking reads.

lneto:
Resolves #603.

`sock:setsockopt(level, optname, value)` sets a socket option: an integer value is passed as the common `int` payload, and a string carries the option's packed binary payload — e.g. a `struct __kernel_sock_timeval` for `SO_RCVTIMEO_NEW`, bounding a blocking receive so it can be made interruptible (a `spawn`ed kthread polling `thread.shouldstop()`) and defending against a no-reply peer:

```lua
local timeval = struct(sk.layout.timeval)
sock:setsockopt(sk.sol.SOCKET, sk.so.RCVTIMEO_NEW, timeval:pack(0, 500 * 1000))
```

### Design notes
- **Routing mirrors `setsockopt(2)`**: `SOL_SOCKET` goes through the network core (`sock_setsockopt`) and any other level through the protocol handler (`sock->ops->setsockopt`). The distinction matters: netlink's handler rejects non-`SOL_NETLINK` levels with `-ENOPROTOOPT`, so a protocol-handler-only implementation cannot set `SO_RCVTIMEO` on a netlink socket.
- **Constants and layout from autogen**: `linux.socket.sol` (`SOL_*`), `linux.socket.so` (`SO_*`) and the `timeval` payload layout (`struct __kernel_sock_timeval`, the y2038-safe variant) are extracted from the kernel headers; a small autogen `as` field names the layout's Lua key.

### DIFF
--- a/lib/luasocket.c
+++ b/lib/luasocket.c
@@ -34,6 +34,10 @@
 
 #include <lunatik.h>
 
+#if LINUX_VERSION_CODE < KERNEL_VERSION(6, 7, 0)
+typedef int (*luasocket_setter_t)(struct socket *, int, int, sockptr_t, unsigned int);
+#endif
+
 #define luasocket_msgaddr(msg, addr, size)	\
 do {						\
 	msg.msg_namelen = size;			\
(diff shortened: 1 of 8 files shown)


lib/luasocket.c
```diff
@@ -378,6 +382,49 @@ LUASOCKET_NEWGETTER(sockname);
...
+{
+	struct socket *socket = luasocket_check(L, 1);
+	int level = (int)luaL_checkinteger(L, 2);
+	int optname = (int)luaL_checkinteger(L, 3);
+	int value;
+	size_t len;
+	const char *optval;
+
+	if (lua_type(L, 4) == LUA_TSTRING)
+		optval = lua_tolstring(L, 4, &len);
```
lneto, line 411: is it safe to pass this around without copying?

PR #628, opened 2026-07-23
Add qos example

sneaky-potato:


examples/qos/tc.lua
```diff
@@ -0,0 +1,65 @@
...
+local map     = require("ebpf.map")
+
+local TC_H_MAKE = function(maj, min) return (maj << 16) | min end
+
+local stats <close> = map.open("/sys/fs/bpf/flow_stats")
+
+-- struct flow_stats {
+--     u64 packets;
+--     u32 avg_pkt_size;
+-- };
```
lneto, line 18: now we can have named struct fields on autogen; won't it help here?

sneaky-potato: flow_stats is currently only defined in the example BPF program, not in a kernel header. Are you suggesting to move it to a shared header so autogen can generate a layout for it, or some other mechanism?

lneto: I wasn't suggesting that, but I think it's a good idea to leverage autogen for this.

examples/qos/tc.lua
```diff
@@ -0,0 +1,65 @@
...
+
+	stats:update(key, value)
+
+	if avg < 256 then
+		skb.priority = TC_H_MAKE(1, 0x10)
+	elseif avg < 800 then
+		skb.priority = TC_H_MAKE(1, 0x20)
+	else
+		skb.priority = TC_H_MAKE(1, 0x30)
+	end
```
lneto:
```suggestion
	local mask = (avg < 256 and 0x10) or (avg < 800 and 0x20) or 0x30
	skb.priority = TC_H_MAKE(1, mask)
```

I think this is more Lua-ish; couldn't find a better name for mask though. (Just a suggestion ;-)

examples/qos/tc.lua
```diff
@@ -0,0 +1,65 @@
...
+-- SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+local tc      = require("tc")
+local action  = require("linux.tc")
+local skbattr = require("skb.attr")
+local map     = require("ebpf.map")
+
+local TC_H_MAKE = function(maj, min) return (maj << 16) | min end
```
lneto, line 11: why the caps?

sneaky-potato: this is a defined kernel macro ([/include/uapi/linux/pkt_sched.h#L72](https://elixir.bootlin.com/linux/v7.1.4/source/include/uapi/linux/pkt_sched.h#L72))

lneto: I think it's fine to leave it as is by now, but I think we should have a better way to export such macros.. perhaps leveraging autogen as well.. it is worth creating an issue at least..

PR #630, opened 2026-07-26
Add bpf maps binding (consolidated review of #587)

lneto:
Consolidated review of #587 (@sneaky-potato's ebpf maps binding): the original commits rebased onto current master, with the review folded into them. The module is now shaped like crypto: a single `bpf` module (matching the `linux.bpf` constants, as xdp/`linux.xdp`) with `bpf.map(path)` as the constructor.

```lua
local bpf = require("bpf")

local flow = bpf.map("/sys/fs/bpf/flow_cache")
flow:update(key, value)          -- BPF_ANY by default
local v = flow:lookup(key)
flow:delete(key)
for k in flow.next, flow do      -- next mirrors Lua's next(t, k)
	print(k)
end
```

Changes on top of #587, addressing the open review threads:

**Safety/correctness**
- `bpf.map()` validates that the pinned object is really a bpf map: `bpf_map_iops` is resolved via `lunatik_lookup()` at module init and compared against the inode ops, the same discrimination `bpf_inode_type()` does. Previously any path's `i_private` (pinned progs/links, files on other filesystems) would be treated as a map and get `bpf_map_inc()`'d. Map types are restricted to the documented set (HASH, ARRAY, LRU_HASH); other types have different op shapes (percpu copies, push/peek, object references) and can be added deliberately later.
- All map ops run under `rcu_read_lock()` (through a single `luabpf_map_call` template), mirroring the syscall path. `lookup` copies the value out with `copy_map_value()`/`check_and_init_map_value()` instead of pushing the live element, which could be freed by a concurrent delete.
- `remove()` guards against map types that lack `map_lookup_and_delete_elem` (e.g. arrays), which was a NULL dereference; `next()` rejects empty keys, which previously caused an out-of-bounds read.
- The object is created before the map reference is taken, so an allocation error cannot leak it; `kern_path()` errors are propagated as errno names. On interrupt-context runtimes `bpf.map()` is only allowed during script load (`lunatik_isready` scheme, as in luathread) since path lookup may sleep; handlers use the handle opened at load.

**API**
- `update` returns `false` when the flag condition is not met (`BPF_EXIST` on an absent key, `BPF_NOEXIST` on a present one) and raises on real errors, symmetric with `delete` on a missing key.
- `map:info()` exposes `type`, `key_size`, `value_size` and `max_entries` (`struct bpf_map_info` naming) and `#map` returns `max_entries`, so scripts can pack keys/values without out-of-band knowledge.
- `next` has the same signature as Lua's `next(t, k)`, so `for k in m.next, m do` iterates the map directly.

**Framework conformance**
- Value/key buffers use `luaL_Buffer` (as the crypto modules do), removing the manual alloc/free paths — including a `kfree()` of runtime-allocated (kvmalloc-backed) memory. `close` uses `lunatik_closeobject`; redundant checks dropped; result pushing and key/value size checks factored into helpers. LDoc fixed to the `/***` markers the base uses with a proper `@type` block.
- Missing Kconfig entry added (`depends on BPF_SYSCALL`); the `linux.bpf` autogen spec is narrowed to the update flags and supported `MAP_TYPE_` constants via the specs' `include` mechanism, instead of ~690 `BPF_`-prefixed constants.

**Tests** (suite `bpf`, wired into `tests/run.sh` and the READMEs; skips without bpftool)
- Per-type coverage: hash (full API and flag semantics, lifecycle after `close`, `info`/`#`, invalid key/value sizes and flags, open validation), array (packed index round-trip, out-of-range edges, `delete`/`remove` rejection, full iteration) and LRU hash.
- A softirq-runtime script exercises the atomic-context allocator path, and the harness cross-checks a Lua-written value with `bpftool map lookup`.


lib/luabpf.c
```diff
@@ -0,0 +1,361 @@
...
+* @treturn boolean success; `false` when the flag condition is not met
+* (`BPF_EXIST` on an absent key, `BPF_NOEXIST` on a present one).
+* @raise Error if the operation is not permitted by the map.
+*/
+static int luabpf_map_update(lua_State *L)
+{
+	struct bpf_map *map = luabpf_map_check(L, 1);
+	const char *key = luabpf_map_checkkey(L, map, 2);
+	const char *value = luabpf_map_checkvalue(L, map, 3);
+	u64 flags = luaL_optinteger(L, 4, BPF_ANY);
```
sneaky-potato, line 161: Should we add some validation on the flag being passed here? I believe we have an exhaustive list of flags which can be passed on update. Thoughts?

lneto: Would it fail safely over this? If yes, I think we should let the user take care of it as in other parts of the framework (e.g., socket). Otherwise, we must validate it.

sneaky-potato: Yes, it does fail safely with the error code EINVAL.

sneaky-potato (approved):
LGTM. I tested it locally and it worked well, including tc/softirq context.

sneaky-potato (approved):
I was able to test the implementation in softirq with tc, LGTM.

PR #633, opened 2026-07-28
bpf: add bpf.map, a high-level layer over maps

lneto:
A Lua-level view over pinned maps, on top of the typed constructors from #636. It mirrors them one for one, taking the packing spec of what the map holds:

```lua
local map = require("bpf.map")

local counters <close> = map.hash("/sys/fs/bpf/counters", "I4", "I4")
counters[1] = 42
counters[1] = nil                 -- delete
for key, value in pairs(counters) do print(key, value) end

local events <close> = map.queue("/sys/fs/bpf/events", "I8")
events:push(1234)
print(events:pop())
```

Two shapes, because the map kind decides who owns the name space. In a key-value map the key is the user's, so a proxy driven by `__index`/`__newindex` cannot carry methods: `t:close()` would be a lookup of the key `"close"`. Operations come from outside, as in Lua's own `table` library and as `rcu.map` already does for `rcu.table`. A queue or a stack has no key, nothing of the name space belongs to the user, and the natural form is the one `fifo` uses in the base: `q:push(v)`, `q:pop()`.

Specs are positional, in the order every eBPF API uses (`bpf_map_update_elem(map, key, value)`, `bpf_map_lookup_elem(map, key)`, `bpftool map update ... key ... value ...`), so a keyless map takes one spec and a keyed map takes two. A spec is a `string.pack` format (`"I4"`, `"c6"`, `"I2I2"`) or a `struct` codec: it says how many bytes, validated against `key_size` and `value_size` on open, and how to read them. Single-value formats take and return scalars; multi-value formats and `struct` codecs take and return arrays, in field order.

No `__len`: the raw `#` is `max_entries`, which in a table proxy reads as the entry count and in a queue as the number queued. eBPF exposes no count, so the capacity stays where it has its uapi name, `info(t).max_entries`.

Conditional updates (`BPF_NOEXIST`/`BPF_EXIST`) and the boolean results stay in the raw API: `t[k] = v` has no return value to carry them.

Tests cover the three table kinds and both keyless kinds: scalar, multi-value and `struct` specs, assignment, `nil` delete, `pairs`, `<close>`, keys named after module functions, FIFO and LIFO order, empty `nil`, full `false`, `info`, spec size validation and the cross-type rejection. Verified on 6.8.0-124: bpf suite `pass:8 fail:0`, full suite `pass:72 fail:0`.

### DIFF
--- a/lib/bpf/map.lua
+++ b/lib/bpf/map.lua
@@ -1,0 +1,309 @@
+--
+-- SPDX-FileCopyrightText: (c) 2026 Ring Zero Desenvolvimento de Software LTDA
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+---
+-- High-level view over pinned eBPF maps.
+-- Each constructor mirrors the one in `bpf`, taking the packing spec of what
+-- the map holds. Key-value maps become table proxies: indexing looks values
+-- up, assignment updates, assigning `nil` deletes, and `pairs` iterates.
+-- Keyless maps become objects, with `push`, `pop` and `peek`.
+--
+-- Keys and values are opaque fixed-size buffers to eBPF. A spec is either a
+-- `string.pack` format (e.g. `"I4"`, `"c6"`) or a `struct` codec, and says both
+-- how many bytes it takes, validated against the map on open, and how to read
+-- them. Specs that pack a single value take and return it as a scalar;
+-- multi-value formats and `struct` codecs take and return an array of values,
+-- in field order.
+--
+-- Operations on a table proxy come from outside it (`close` and `info` are
+-- module functions, as in Lua's `table` library or `rcu.map`), so any key the
+-- spec can encode is a valid map key.
+--
+-- Conditional updates (`BPF_NOEXIST`/`BPF_EXIST`) and the boolean results
+-- belong to the raw API.
+-- @module bpf.map
+-- @see bpf
+-- @usage
+-- local map = require("bpf.map")
+--
+-- local counters <close> = map.hash("/sys/fs/bpf/counters", "I4", "I4")
+-- counters[1] = 42
+-- print(counters[1])
+-- counters[1] = nil                 -- delete
+-- for key, value in pairs(counters) do print(key, value) end
+--
+-- local events <close> = map.queue("/sys/fs/bpf/events", "I8")
+-- events:push(1234)
+-- print(events:pop())
+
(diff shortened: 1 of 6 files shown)


lib/bpf/map.lua
```diff
@@ -0,0 +1,195 @@
...
+-- that behaves like a Lua table backed by the map: indexing looks values up,
+-- assignment updates, assigning `nil` deletes, and `#` is the map's
+-- `max_entries` -- mirroring `rcu.table` ergonomics over an eBPF backend.
+--
+-- Each spec entry is either a `string.pack` format (e.g. `"I4"`) or a
+-- `struct` codec; its packed size is validated against the map on open.
+-- Specs that pack a single value take and return it as a scalar; multi-value
+-- formats and `struct` codecs take and return an array of values, in field
+-- order. Method names (`pairs`, `close`, `info`) shadow equal scalar string
+-- keys; use the raw `bpf` module for such maps.
```
sneaky-potato: Does this mean I cannot create a key like "pairs" or "close", because t["close"] will return the method instead?

lneto: yup.. I've added a fixup to remove the methods.. so we can use the sabe approach of table library.. let me know what you think..

lib/bpf/map.lua
```diff
@@ -0,0 +1,292 @@
...
+	__close = map.close,
+}
+
+local function opentable(open, pathname, keyspec, valuespec)
+	local handle = open(pathname)
+	local info = handle:info()
+	local tbl = setmetatable({}, view)
+	states[tbl] = {
+		map = handle,
+		key = newcodec(keyspec, "key", info.key_size),
```
sneaky-potato, line 181: Is there a chance of memory leak here? If `newcodec()` raises an error (in case of mismatched spec) will `handle` require explicit closing? Just wondering... 
Same observation with `openqueue`

lneto: it's not a leak.. the handle is a userdata with `__gc`, so if `newcodec()` raises it becomes garbage and the release puts the map ref on the next collection.. deferred, not lost.. and the bpffs pin holds the map anyway.. it's the same reliance on GC as the error path of `luabpf_open` itself.. if you prefer a deterministic release, a `pcall` closing the handle before re-raising costs a few lines, but I'd rather keep the GC path for simplicity

sneaky-potato (approved):

PR #634, opened 2026-07-28
luabpf: add support for queue, stack maps

sneaky-potato:
- Queues and stacks do not support `map_update_elem`, `map_get_next_key`. Calling these methods returns in `-EINVAL` (ref: [kernel/bpf/queue_stack_maps.c](https://elixir.bootlin.com/linux/v7.1.4/source/kernel/bpf/queue_stack_maps.c#L264))
- They return NULL for `map_lookup_elem`
- They support `map_pop_elem`, `map_push_elem` and `map_peek_elem`
- Add Lua test suites covering push, pop, peek FIFO/LIFO behavior, unsupported operations, and map metadata for BPF queue and stack map types.

### DIFF
--- a/lib/luabpf.c
+++ b/lib/luabpf.c
@@ -6,17 +6,24 @@
 /***
 * Lua interface to eBPF.
 *
-* This module provides generic key-value access to pinned eBPF maps,
-* supporting types that implement standard lookup, update, deletion,
-* and iteration operations:
+* This module provides access to pinned eBPF maps of two kinds:
+*
+* Key-value maps, via `lookup`, `update`, `delete`, `remove`, and `next`:
 *   - BPF_MAP_TYPE_HASH
 *   - BPF_MAP_TYPE_ARRAY
 *   - BPF_MAP_TYPE_LRU_HASH
 *
-* Other map types are rejected when opened.
+* Queue/stack maps, via `push`, `pop`, and `peek`:
+*   - BPF_MAP_TYPE_QUEUE (FIFO)
+*   - BPF_MAP_TYPE_STACK (LIFO)
+*
+* Other map types are rejected when opened. Calling an operation that a
+* map type does not implement (e.g. `lookup` on a queue) raises an
+* `EOPNOTSUPP` error.
 *
 * Keys and values are exchanged as packed Lua strings whose sizes must
-* match the map's configured key and value sizes.
+* match the map's configured key and value sizes. Queue/stack maps have
+* no keys, so only `value_size` applies to them.
 *
 * @module bpf
 */
@@ -32,7 +39,8 @@
 LUNATIK_PRIVATECHECKER(luabpf_map_check, struct bpf_map *);
 
 #define luabpf_map_issupported(type)	\
-	((type) == BPF_MAP_TYPE_HASH || (type) == BPF_MAP_TYPE_ARRAY || (type) == BPF_MAP_TYPE_LRU_HASH)
+	((type) == BPF_MAP_TYPE_HASH || (type) == BPF_MAP_TYPE_ARRAY || (type) == BPF_MAP_TYPE_LRU_HASH || \
+	 (type) == BPF_MAP_TYPE_QUEUE || (type) == BPF_MAP_TYPE_STACK)
 
 static const struct inode_operations *luabpf_map_iops;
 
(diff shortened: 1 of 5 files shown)


lib/luabpf.c
```diff
@@ -233,6 +241,61 @@ static int luabpf_map_next(lua_State *L)
...
+* @raise Error if the map does not support this operation (e.g. hash/array maps).
+*/
+static int luabpf_map_push(lua_State *L)
+{
+	struct bpf_map *map = luabpf_map_check(L, 1);
+	const char *value = luabpf_map_checkvalue(L, map, 2);
+	u64 flags = luaL_optinteger(L, 3, 0);
+
+	long ret;
+	luabpf_map_call(ret, map->ops->map_push_elem(map, (void *)value, flags));
```
lneto, line 260: hash/array/lru do not implement `map_push_elem`/`map_pop_elem`/`map_peek_elem` — the ops are NULL pointers (check `hashtab.c`/`arraymap.c`), so `hashmap:push(v)` is a NULL dereference and oopses the kernel. Same class we fixed in `remove()` with the `EOPNOTSUPP` guard; with 4 users now, I would factor a checker, e.g.:
```c
static inline void luabpf_map_checkop(lua_State *L, const void *op)
{
	if (op == NULL)
		lunatik_throw(L, -EOPNOTSUPP);
}
```
and call it before each use (remove() becomes a user too). The tests do not catch this because they only call push/pop/peek on queue/stack maps.

lib/luabpf.c
```diff
@@ -6,17 +6,24 @@
...
 *   - BPF_MAP_TYPE_ARRAY
 *   - BPF_MAP_TYPE_LRU_HASH
 *
-* Other map types are rejected when opened.
+* Queue/stack maps, via `push`, `pop`, and `peek`:
+*   - BPF_MAP_TYPE_QUEUE (FIFO)
+*   - BPF_MAP_TYPE_STACK (LIFO)
+*
+* Other map types are rejected when opened. Calling an operation that a
+* map type does not implement (e.g. `lookup` on a queue) raises an
```
lneto, line 21: this is not what the kernel does: on queue/stack the key-value ops are present as stubs — `lookup` returns NULL (so we return `nil`, and the test below asserts exactly that) and `update`/`delete`/`get_next_key` return `-EINVAL` (so we raise `EINVAL`). `EOPNOTSUPP` only happens where the op pointer is NULL and we guard it (today only `remove`; push/pop/peek on key-value maps once the guard above is added). Let's document the real taxonomy: missing op -> EOPNOTSUPP, kernel stub -> EINVAL, queue lookup -> nil.

tests/bpf/run.sh
```diff
@@ -31,7 +33,7 @@ command -v bpftool > /dev/null 2>&1 || skip "bpf: bpftool unavailable"
 cleanup()
 {
 	lunatik stop "$SOFTIRQ_SCRIPT" 2>/dev/null
-	rm -f "$MAP" "$ARRAY_MAP" "$LRU_MAP" "$PERCPU_MAP"
+	rm -f "$MAP" "$ARRAY_MAP" "$LRU_MAP" "$PERCPU_MAP" "$QUEUE_MAP"
```
lneto, line 36: `$STACK_MAP` is missing here — the stack pin leaks after the suite.

lib/luabpf.c
```diff
@@ -233,6 +241,61 @@ static int luabpf_map_next(lua_State *L)
 	return luabpf_map_pushbuffer(L, &B, map->key_size, ret);
 }
 
+/***
+* Pushes a value onto a queue or stack map.
+* @function push
+* @tparam string value Packed value.
+* @tparam[opt=0] integer flags Update flags. `BPF_EXIST` may be set to
```
lneto, line 248: the default 0 is `BPF_ANY` — I would name the constant, as in `update`. Also, `BPF_EXIST` (overwrite the oldest when full) is documented but not covered by the tests.

tests/bpf/queue.lua
```diff
@@ -0,0 +1,111 @@
...
+	m:close()
+end)
+
+test("bpf.map queue push full queue raises", function()
+	local m = map(path)
+	drain(m)
+	for i = 1, #m do
+		assert(m:push("xyz"))
+	end
+	assert(not pcall(m.push, m, "overflow"), "expected error when queue is full")
```
lneto, line 67: this pins full-queue as a raise (`E2BIG`). Fine by me, but note the alternative: treating `E2BIG` as `false` (a capacity condition, like the flag conditions in `update`) would allow `while m:push(v)` loops. If we keep the raise, let's keep it deliberately.

lneto (commented):
Reviewed against the 6.8 sources. One blocker (NULL op dereference), one doc mismatch and one test leak below; two smaller points inline as well.

Two general notes: (1) subject prefix — we have been using `bpf:` for this module (see the log); (2) whichever lands second between this and #633 should decide what the high-level `bpf.map` layer does with queue/stack maps (they have no key-value semantics for the proxy).

lneto:
#635

PR #635, opened 2026-07-28
luabpf: queue and stack maps (consolidated review of #634)

lneto:
Consolidated review of #634 (@sneaky-potato's queue/stack maps support): the original commits plus review fixups on top, all findings validated against the 6.8 sources (`kernel/bpf/queue_stack_maps.c`, `hashtab.c`, `arraymap.c`).

**Fixes**

- **NULL op dereference (kernel oops)**: hash/array/lru do not implement `map_push_elem`/`map_pop_elem`/`map_peek_elem` — the ops-table fields are absent, i.e. NULL pointers — so `hashmap:push(v)` dereferenced NULL. Guarded with a `luabpf_map_checkop()` helper (`EOPNOTSUPP`), which `remove()` now uses as well (same guard it already had, factored).
- **Error taxonomy documented as the kernel behaves**: on queue/stack the key-value ops are *stubs*, not missing — `lookup` returns NULL (our API returns `nil`) and `update`/`delete`/`get_next_key` return `-EINVAL` (queue_stack_maps.c:235-257) — while `EOPNOTSUPP` is raised only for genuinely missing ops (`remove` on queue/stack, `push`/`pop`/`peek` on key-value maps). The module doc claimed `EOPNOTSUPP` across the board.
- **Full-queue test was passing for the wrong reason**: it pushed an 8-byte value into a 3-byte map, so `pcall` was catching the value-size check, not `E2BIG`; now it pushes a right-sized value and pins the actual full-map error.
- **Stack map pin leaked**: `cleanup()` removed the queue pin but not `$STACK_MAP`.
- **`BPF_EXIST` covered**: new test pins the overwrite-oldest semantics (push advances `qs->tail`, dropping the oldest; the next pop returns the following element).
- `push` flags default documented as `BPF_ANY` (it was `0`), matching `update`.

**Also in this review round**: pushing to a full map without `BPF_EXIST` now returns `false` instead of raising — a capacity condition, like the flag conditions in `update`, so producers can branch without `pcall` (specific to `push`: array's out-of-range `E2BIG` still raises). `pop`/`peek` are factored into a `LUABPF_MAP_TAKER` template (they differed only in the op called). Tests gained the push/pop/peek rejection on key-value maps (regression for the NULL-op guard), invalid push flags, and the queue/stack entries in the test docs (they were missing).

**Out of scope**: the interaction with the high-level `bpf.map` layer (#633) — whichever lands second should decide what the table proxy does with keyless maps.


tests/bpf/queue.lua
```diff
@@ -0,0 +1,132 @@
...
+	local m = map(path)
+	drain(m)
+	for i = 1, #m do
+		assert(m:push("xyz"))
+	end
+	assert(m:push("ovr") == false, "expected false when queue is full")
+	m:close()
+end)
+
+test("bpf.map queue push BPF_EXIST overwrites the oldest when full", function()
```
sneaky-potato, line 71: I noticed we don't have an equivalent test in stack. For stack as well we can pass BPF_EXIST flag in push, ref: [/kernel/bpf/queue_stack_maps.c#L193](https://elixir.bootlin.com/linux/v7.1.4/source/kernel/bpf/queue_stack_maps.c#L193) and test for the oldest element in LIFO.

lneto: good catch! added in 0620b809, along with the NOEXIST test that was missing too.. note that on a full stack BPF_EXIST drops the bottom (the element that would pop last) and the new one still lands on top.. the test pins exactly that

sneaky-potato (approved):
The tests look good, approved.

PR #636, opened 2026-07-29
bpf: split map kinds into typed constructors

lneto:
Stacked on #635. Splits the map support into one typed constructor per map type, mirroring the module's growth plan (typed constructors per object, as prog/link will be) and the crypto layout (per-type classes under one lib).

```lua
local bpf = require("bpf")

local flows = bpf.hash("/sys/fs/bpf/flow_cache")
local jobs  = bpf.queue("/sys/fs/bpf/job_queue")    -- FIFO
local undo  = bpf.stack("/sys/fs/bpf/undo_stack")   -- LIFO
```

- `bpf.hash()`, `bpf.array()` and `bpf.lru_hash()` return key-value handles; `bpf.queue()` and `bpf.stack()` return handles exposing only `push`/`pop`/`peek` (plus `info`, `close`, `#`). Constructor names match the bpftool map type names; each kind shares one class internally (the methods are identical; pop order comes from the map type), and all three constructors share the opener — each one only validates the map type.
- A mismatched type now fails **at open** ("not a queue map") and a wrong-kind operation is simply **absent from the handle**, instead of surfacing later as `EOPNOTSUPP`, `EINVAL` or `nil` depending on how the kernel declines it; the error-taxonomy paragraph disappears from the docs.
- The NULL-op guards stay as pure crash-safety (method transplant across handles is still expressible in Lua).
- Tests: the five stubbed-operation tests become two stronger ones per kind — the operations are absent from the handle, and the constructors reject the other kinds in every direction — using only the maps the suite already pins.


lib/luabpf.c
```diff
@@ -365,25 +366,55 @@ static const luaL_Reg luabpf_map_mt[] = {
...
+	object->private = map;
+	luaL_argcheck(L, map != NULL, 1, "not a bpf map");
+	return map;
+}
+
+#define LUABPF_OPENER(name, class, TYPE)					\
+static int luabpf_##name##_open(lua_State *L)					\
+{										\
+	struct bpf_map *map = luabpf_open(L, &luabpf_##class##_class);		\
+	luaL_argcheck(L, luabpf_map_istype(map->map_type, TYPE), 1,		\
```
sneaky-potato: Just a thought but, if somebody passes a valid pinned stack path to bpf.queue(), it will resolve the map pointer, increase it's ref count and then check for map type and throw an error. Although gc will handle it eventually but I think it would be safe to not create the lunatik object if map type is wrong. Thoughts?

lib/luabpf.c
```diff
@@ -394,35 +425,72 @@ static const lunatik_class_t luabpf_map_class = {
...
-	luaL_argcheck(L, map != NULL, 1, "not a bpf map");
-	luaL_argcheck(L, luabpf_map_issupported(map->map_type), 1, "unsupported map type");
-	return 1;
-}
+/***
+* Opens a stack (LIFO) map from a pinned bpffs path.
+* Same interrupt-context restriction as `hash`.
+* @function stack
+* @tparam string path Path to a pinned eBPF stack map.
+* @treturn queue Opened stack handle.
```
sneaky-potato: I tried building the docs, to check how the `@ treturn queue` will look like for stack. It is redirecting to queue(path) method. Same case for the hash return on array and lru hash (they redirect to hash(path) method which is used for hash maps, not array and lru_hash)

sneaky-potato (approved):

PR #642, opened 2026-08-02
Add support for multiple runtimes (perpcu)

sneaky-potato:
Adds an opt-in percpu mode for Lunatik scripts. Instead of a single shared runtime instance, a percpu script gets one dedicated runtime per online CPU, and eBPF-triggered lookups (luasched, luatc, luaxdp) resolve to the runtime matching the CPU the hook is executing on.

### DIFF
--- a/bin/lunatik
+++ b/bin/lunatik
@@ -159,11 +159,16 @@ if #arg >= 1 then
 	local tokens = {run = true, spawn = true, stop = true, list = true}
 	local needs_script = {run = true, spawn = true, stop = true}
 	if not tokens[token] or (needs_script[token] and not script) then
-		print("usage: lunatik [load|unload|reload|status|test|list] [run|spawn|stop <script>]")
+		print("usage: lunatik [load|unload|reload|status|test|list] [run|spawn|stop <script>] [percpu]")
 		os.exit(false)
 	end
 	ensure_loaded()
-	local parm  = arg[3] and (', "' .. arg[3] .. '"') or ""
+	local percpu  = arg[3] == "percpu" or arg[4] == "percpu"
+	local context = (arg[3] ~= "percpu" and arg[3]) or (arg[4] ~= "percpu" and arg[4]) or nil
+	local parm    = context and string.format(', "%s"', context) or ""
+	if percpu then
+		parm = (context and parm or ", nil") .. ", true"
+	end
 	local ret   = token == "list" and "return" or ""
 	local chunk = string.format("%s lunatik.runner.%s('%s'%s)", ret, token, script, parm)
 	print(dostring(chunk))
--- a/bin/lunatik
+++ b/bin/lunatik
@@ -153,13 +153,21 @@ if cmd then
 	os.exit()
 end
 
+local USAGE = [[
+usage:
+	lunatik run <script> [softirq|hardirq] [percpu]
+	lunatik spawn <script>
+	lunatik stop <script>
+	lunatik [load|unload|reload|status|test|list]
+]]
+
 if #arg >= 1 then
 	local token  = arg[1]
 	local script = arg[2]
 	local tokens = {run = true, spawn = true, stop = true, list = true}
 	local needs_script = {run = true, spawn = true, stop = true}
 	if not tokens[token] or (needs_script[token] and not script) then
-		print("usage: lunatik [load|unload|reload|status|test|list] [run|spawn|stop <script>] [percpu]")
+		print(USAGE)
 		os.exit(false)
 	end
 	ensure_loaded()
--- a/bin/lunatik
+++ b/bin/lunatik
@@ -155,10 +155,10 @@ end
 
 local USAGE = [[
 usage:
+	lunatik [load|unload|reload|status|test|list]
 	lunatik run <script> [softirq|hardirq] [percpu]
 	lunatik spawn <script>
 	lunatik stop <script>
-	lunatik [load|unload|reload|status|test|list]
 ]]
 
 if #arg >= 1 then
--- a/lib/lualinux.c
+++ b/lib/lualinux.c
@@ -249,6 +249,21 @@ static int lualinux_errname(lua_State *L)
     return 1;
 }
 
+/***
+* Returns the number of possible CPU ids.
+* CPU ids range from `0` to `numcpus() - 1`; the online ones are a subset,
+* so an id in that range may be offline.
+* @function numcpus
+* @treturn integer number of possible CPU ids
+* @usage
+* for cpu = 0, linux.numcpus() - 1 do print(cpu) end
+*/
+static int lualinux_numcpus(lua_State *L)
+{
+	lua_pushinteger(L, nr_cpu_ids);
+	return 1;
+}
+
 static const luaL_Reg lualinux_lib[] = {
 	{"random", lualinux_random},
 	{"schedule", lualinux_schedule},
@@ -259,6 +274,7 @@ static const luaL_Reg lualinux_lib[] = {
 	{"ifindex", lualinux_ifindex},
 	{"ifaddr", lualinux_ifaddr},
 	{"errname", lualinux_errname},
+	{"numcpus", lualinux_numcpus},
 	{NULL, NULL}
 };
 
--- a/lib/luaxdp.c
+++ b/lib/luaxdp.c
@@ -38,6 +38,7 @@ __diag_ignore_all("-Wmissing-prototypes",
 #endif
 
 static lunatik_object_t *luaxdp_runtimes = NULL;
+static lunatik_object_t *luaxdp_percpu = NULL;
 
 static inline lunatik_object_t *luaxdp_pushdata(lua_State *L, int upvalue, void *ptr, size_t size)
 {
@@ -96,11 +97,14 @@ out:
 
 static inline int luaxdp_checkruntimes(void)
 {
-	static const char key[] = "runtimes";
-	if (luaxdp_runtimes == NULL &&
-	   (luaxdp_runtimes = luarcu_getobject(lunatik_env, key, sizeof(key) - 1)) == NULL)
-		return -1;
-	return 0;
+	static const char runtimes_key[] = "runtimes";
+	static const char percpu_key[] = "percpu";
+
+	if (luaxdp_runtimes == NULL)
+		luaxdp_runtimes = luarcu_getobject(lunatik_env, runtimes_key, sizeof(runtimes_key) - 1);
+	if (luaxdp_percpu == NULL)
+		luaxdp_percpu = luarcu_getobject(lunatik_env, percpu_key, sizeof(percpu_key) - 1);
+	return luaxdp_runtimes != NULL && luaxdp_percpu != NULL ? 0 : -1;
 }
 
 __bpf_kfunc int bpf_luaxdp_run(char *key, size_t key__sz, struct xdp_md *xdp_ctx, void *arg, size_t arg__sz)
@@ -109,13 +113,21 @@ __bpf_kfunc int bpf_luaxdp_run(char *key, size_t key__sz, struct xdp_md *xdp_ctx
 	struct xdp_buff *ctx = (struct xdp_buff *)xdp_ctx;
 	int action = -1;
 	size_t keylen = key__sz - 1;
+	lunatik_value_t value;
+	char cpu_key[LUARCU_MAXKEY];
 
 	if (unlikely(luaxdp_checkruntimes() != 0)) {
-		pr_err("couldn't find _ENV.runtimes\n");
+		pr_err("couldn't find _ENV.runtimes or _ENV.percpu\n");
 		goto out;
 	}
 
 	key[keylen] = '\0';
+	luarcu_getvalue(luaxdp_percpu, key, keylen, &value);
+	if (value.type == LUA_TBOOLEAN && value.boolean) {
+		keylen = scnprintf(cpu_key, sizeof(cpu_key), "%s:%d", key, raw_smp_processor_id());
+		key = cpu_key;
+	}
+
 	if ((runtime = luarcu_getobject(luaxdp_runtimes, key, keylen)) == NULL) {
 		pr_err("couldn't find runtime '%s'\n", key);
 		goto out;
--- a/lib/luaxdp.c
+++ b/lib/luaxdp.c
@@ -96,9 +96,9 @@ out:
 
 static inline int luaxdp_checkruntimes(void)
 {
-	const char *key = "runtimes";
+	static const char key[] = "runtimes";
 	if (luaxdp_runtimes == NULL &&
-	   (luaxdp_runtimes = luarcu_getobject(lunatik_env, key, sizeof(key))) == NULL)
+	   (luaxdp_runtimes = luarcu_getobject(lunatik_env, key, sizeof(key) - 1)) == NULL)
 		return -1;
 	return 0;
 }
--- a/lib/luaxdp.c
+++ b/lib/luaxdp.c
@@ -259,6 +259,8 @@ static void __exit luaxdp_exit(void)
 #if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 4, 0))
 	if (luaxdp_runtimes != NULL)
 		lunatik_putobject(luaxdp_runtimes);
+	if (luaxdp_percpu != NULL)
+		lunatik_putobject(luaxdp_percpu);
 #endif
 }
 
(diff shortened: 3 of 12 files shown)


lib/lualinux.c
```diff
@@ -249,6 +249,17 @@ static int lualinux_errname(lua_State *L)
...
 }
 
+/***
+ * Returns the number of CPUs.
+ * @function numcpus
+ * @treturn integer number of CPUs
+ */
+static int lualinux_numcpus(lua_State *L)
+{
+	lua_pushinteger(L, num_online_cpus());
```
lneto: `num_online_cpus()` is the *count* of online CPUs, not the id space. `cpu_online_mask` is described in `cpumask.h` as a "dynamic subset of cpu_present_mask", so ids have holes and `raw_smp_processor_id()` can return an id `>= num_online_cpus()`.

Reproduced on a 6-CPU box with `echo 0 > /sys/devices/system/cpu/cpu3/online`: `run <script> percpu` registers `:0` through `:4`, so **cpu5 is online with no runtime** (its hooks take the `-ENOENT` path) while cpu3, which is offline, has one. `nr_cpu_ids` is the id-space bound and does not change at runtime, so it also keeps `stop` symmetric with `run` even if CPUs go up or down in between.

Addressed in `review/642` by [`679c8817`](https://github.com/luainkernel/lunatik/commit/679c8817d0788b9eb635552829a746e1b6b697d3).

lib/lunatik/runner.lua
```diff
@@ -53,6 +65,9 @@ end
 -- @treturn userdata kernel thread object.
 function runner.spawn(script, ...)
 	local runtime = runner.run(script, ...)
+	if not runtime then
+		error("spawn does not support percpu scripts")
```
lneto, line 122: The rejection happens after `runner.run` already created the N runtimes and set `env.percpu[script]`, so a refused `spawn` leaves the script fully registered. Checking `percpu` before calling `run` keeps the failure side-effect free.

Addressed in `review/642` by [`d98a943a`](https://github.com/luainkernel/lunatik/commit/d98a943a3b94ac826174dc6b5ad3a99950b32a08).

lib/lunatik/runner.lua
```diff
@@ -73,13 +88,30 @@ local function stop(registry, script)
 	end
 end
```
lneto: `stop_percpu` handles the explicit `runner.stop`, but `runner.shutdown` never reaches it: it iterates `env.runtimes`, whose keys for a percpu script are `script:0`, `script:1`, ..., so each runtime is stopped through the generic path and `env.percpu[script]` survives. From then on `runner.run(script)` raises "is already running" with nothing running. Iterating `env.percpu` first in `shutdown` stops them as a unit and clears the flag.

Addressed in `review/642` by [`ee613b1c`](https://github.com/luainkernel/lunatik/commit/ee613b1ce7aa03cd964af218a20145d18efba867).

lib/lunatik/runner.lua
```diff
@@ -28,19 +29,30 @@ local function trim(script) -- drop ".lua" file extension
 	return script:gsub("(%w+).lua", "%1")
 end
 
+local function key(script, cpu)
+	return script .. ":" .. cpu
```
lneto, line 41: These keys leak into `runner.list`: a percpu script shows up as N entries (`script:0, script:1, ...`) instead of one, and `lunatik list` is the documented way to see what is running. Listing `env.percpu` and filtering the `:` keys out of `env.runtimes` keeps one name per script.

Addressed in `review/642` by [`254f84cf`](https://github.com/luainkernel/lunatik/commit/254f84cfb3b1fdade4bca696c480f96abc489af9).

lib/lunatik/runner.lua
```diff
@@ -107,6 +139,7 @@ end
 function runner.startup()
```
lneto, line 174: The `if env.runtimes then return end` guard predates `env.percpu`: on an env created before this change (e.g. core kept loaded across an upgrade), `runtimes` exists, so `percpu` is never created and `checkenv` then fails every lookup. Creating each table only if missing keeps startup idempotent without that coupling.

Addressed in `review/642` by [`8ceb15fa`](https://github.com/luainkernel/lunatik/commit/8ceb15faa9a4e08bb630fb40c262abe3f979f600).

lunatik_ebpf.h
```diff
@@ -8,26 +8,45 @@
...
 	else { \
+		lunatik_value_t value; \
+		char cpu_key[PERCPU_KEY_LEN]; \
+		luarcu_getvalue((percpu), (key), (keylen), (&value)); \
+		if (value.type == LUA_TBOOLEAN && value.boolean) { \
+			if (key_sz + PERCPU_SUFFIX_LEN > PERCPU_KEY_LEN) { \
+				pr_err("runtime name is too long\n"); \
+				goto lunatik_ebpf_lookup_skip; \
+			} \
+			keylen = snprintf(cpu_key, PERCPU_KEY_LEN, "%s:%d", (key), (cpu)); \
```
lneto: Kernel `snprintf` returns what *would* have been written, not what was (`lib/vsprintf.c`: "If the return is greater than or equal to @size, the resulting string is truncated"), so on truncation `keylen` describes more bytes than the buffer holds and the lookup silently misses. `scnprintf` returns what was actually written and removes the need for the length guard entirely, and with it the `goto`.

Two more things in this block:
- the label is at function scope, so a second use of the macro in the same function stops compiling (only one use per kfunc today);
- `key = cpu_key` points the caller's argument at a buffer that dies at the end of the statement expression, so any caller that touches `key` afterwards reads dead stack (none do today). A local `lookup` pointer avoids both.

On the sizing: `PERCPU_SUFFIX_LEN 4` silently assumes fewer than 1000 CPUs (Ubuntu ships `NR_CPUS` 8192). These keys live in a `luarcu` table, whose own key bound is `LUARCU_MAXKEY` — a key that does not fit there could not have been registered in the first place — so sizing `cpu_key` by `LUARCU_MAXKEY` removes the ad-hoc constants entirely (`luarcu_map` already stacks a buffer of that size). The macros also have no module prefix. In the master port, where this lives in a plain function, the guard and the `goto` disappear for free.


Addressed in `review/642` by [`48e65d23`](https://github.com/luainkernel/lunatik/commit/48e65d2399636a7effc4e235d43da7c4c7e6591b).

lunatik_ebpf.h
```diff
@@ -8,26 +8,45 @@
...
+	static const char runtimes_key[] = "runtimes"; \
+	static const char percpu_key[] = "percpu"; \
 	if ((runtimes) == NULL) \
-		(runtimes) = luarcu_getobject(lunatik_env, key, sizeof(key)); \
-	(runtimes) != NULL ? 0 : -1; \
+		(runtimes) = luarcu_getobject(lunatik_env, runtimes_key, sizeof(runtimes_key) - 1); \
+	if ((percpu) == NULL) \
+		(percpu) = luarcu_getobject(lunatik_env, percpu_key, sizeof(percpu_key) - 1); \
+	(runtimes) && (percpu); \
 })
```
lneto: Note the old code passed `sizeof(key)` on a `const char *`, the size of the pointer, which matches `strlen("runtimes")` only by coincidence on 64-bit. The static-array form is the correct one. Since this fix is independent of percpu, split it into its own commit.

Addressed in `review/642` by [`7ce12e25`](https://github.com/luainkernel/lunatik/commit/7ce12e257b7e606f83922cacc6ac731c0312937b).

bin/lunatik
```diff
@@ -159,11 +159,16 @@ if #arg >= 1 then
...
 	if not tokens[token] or (needs_script[token] and not script) then
-		print("usage: lunatik [load|unload|reload|status|test|list] [run|spawn|stop <script>]")
+		print("usage: lunatik [load|unload|reload|status|test|list] [run|spawn|stop <script>] [percpu]")
 		os.exit(false)
 	end
 	ensure_loaded()
-	local parm  = arg[3] and (', "' .. arg[3] .. '"') or ""
+	local percpu  = arg[3] == "percpu" or arg[4] == "percpu"
+	local context = (arg[3] ~= "percpu" and arg[3]) or (arg[4] ~= "percpu" and arg[4]) or nil
+	local parm  = (percpu or context) and (', ' .. (context and ('"' .. context .. '"') or "nil")) or ""
```
lneto: Three decisions packed into one expression. Splitting the `percpu` case out reads closer to what it builds:

```lua
local parm    = context and string.format(', "%s"', context) or ""
if percpu then
	parm = (context and parm or ", nil") .. ", true"
end
```

Addressed in `review/642` by [`1399fa7e`](https://github.com/luainkernel/lunatik/commit/1399fa7efd531670cc811233c7a0f0a915691b0f).

lneto (changes requested):
The percpu mode works, and the `checkenv` rework fixes a real latent bug on the way (see the inline note). What needs addressing before merge is the CPU id space, which I could reproduce, plus a few lifecycle gaps in the runner.

I pushed the changes I'm proposing to [`review/642`](https://github.com/luainkernel/lunatik/tree/review/642), rebuilt **on master** rather than on `gsoc26`: the runner half cherry-picks cleanly (your commits, untouched, plus my fixups), and the bindings half lands on `luaxdp.c`, the only eBPF-dispatched binding master has. Full suite passes there (75/75), including a new `runtime/percpu` test that covers the runner side: one runtime per possible CPU id under the exact keys the kfuncs look up, listing, stop-and-rerun, and the spawn refusal.

One thing that is *not* about this PR but that you'll hit: on the `gsoc26` base, unloading `lunatik_run` after the thread suite reproducibly dies with `kernel BUG at mm/slub.c:553` (`lunatik_stop -> lua_close -> GCTM -> lunatik_releaseobject -> kvfree`), second run onward. Master does not reproduce it, nor does master + this patch, so it comes from the base branch, not from percpu.

```
kernel BUG at mm/slub.c:553!
Call trace:
 __slab_free / kfree / kvfree
 lunatik_releaseobject [lunatik]
 lunatik_deleteobject [lunatik]
 GCTM / luaC_freeallobjects / close_state / lua_close [lunatik]
 lunatik_stop [lunatik]
 lunatik_run_exit [lunatik_run]
```

What is still missing is a test: this changes runtime resolution for every eBPF-dispatched binding and ships nothing that exercises it. [`review/642`](https://github.com/luainkernel/lunatik/tree/review/642) has a `runtime/percpu` test covering the runner side: one runtime per possible CPU id under the keys the kfuncs look up, listing, stop-and-rerun, and the spawn refusal.

sneaky-potato:
Hi @lneto I picked up the fixup commits from review branch and added a check for runtime failure in case of percpu mode. Also added some comments for recently added local functions.

lib/luaxdp.c
```diff
@@ -29,124 +30,152 @@
...
-__diag_push();
-__diag_ignore_all("-Wmissing-prototypes",
-                  "Global kfuncs as their definitions will be in BTF");
-#endif
+LUNATIK_EBPF_START();
+
+static char luaxdp_env_key;
 
 static lunatik_object_t *luaxdp_runtimes = NULL;
+static lunatik_object_t *luaxdp_percpu = NULL;
```
lneto, line 38: `luarcu_getobject` takes a reference on the object it returns (`luarcu_getvalue` does `lunatik_getobject` on userdata values), and the module exits only put `*_runtimes`: the cached `_ENV.percpu` table leaks a reference on every unload, in the three bindings. Each exit needs the symmetric put for `*_percpu`.

Addressed in `review/642` by [`ca67331e`](https://github.com/luainkernel/lunatik/commit/ca67331e0fd0e9b3713e6f12637107de3f2f2900).

lib/lunatik/runner.lua
```diff
@@ -28,19 +29,30 @@ local function trim(script) -- drop ".lua" file extension
...
-function runner.run(script, ...)
+function runner.run(script, context, percpu, ...)
 	local script = trim(script)
-	if env.runtimes[script] then
+	if env.runtimes[script] or env.percpu[script] then
 		error(string.format("%s is already running", script))
 	end
-	local runtime = lunatik.runtime(script, ...)
+	if percpu then
+		for cpu = 0, linux.numcpus() - 1 do
```
lneto, line 49: The script runs once per instance, so a mid-loop failure is not hypothetical: an error at instance k propagates out with instances 0..k-1 registered and `env.percpu[script]` not yet set. `stop_percpu` checks that flag, so `runner.stop` never reaches them: they are left running with no way to stop them short of a reload. The loop needs to roll back the instances already created before re-raising.

Addressed in `review/642` by [`82a5e6fe`](https://github.com/luainkernel/lunatik/commit/82a5e6fe39a4fc8033a694b2ae71de5215378513), with a regression test ([`1c833c12`](https://github.com/luainkernel/lunatik/commit/1c833c129885f4581c9624a92613a85f69fb724e)) that fails a script on its second instance and asserts nothing survives.

lib/lunatik/runner.lua
```diff
@@ -73,13 +88,30 @@ local function stop(registry, script)
...
+end
+
 --- Stops a running script and its associated thread, if any.
 -- It attempts to stop the thread first, then the runtime.
 -- @tparam string script name of the script to stop. The ".lua" extension will be trimmed.
 function runner.stop(script)
 	local script = trim(script)
 	stop(env.threads, script)
 	stop(env.runtimes, script)
+	if env.percpu[script] then
```
lneto, line 112: `lunatik stop <script>:0` reaches the generic path: it stops and unregisters that one instance, `env.percpu[script]` stays set, and from then on `run` refuses the name while `stop` by name only cleans what is left. The per-CPU keys are an implementation detail; stopping one should be refused.

Addressed in `review/642` by [`f307fe48`](https://github.com/luainkernel/lunatik/commit/f307fe4870d2ba809ce1569a416ac4f0c40d39ad).

lunatik_ebpf.h
```diff
@@ -0,0 +1,103 @@
...
+			if (key_sz + PERCPU_SUFFIX_LEN > PERCPU_KEY_LEN) { \
+				pr_err("runtime name is too long\n"); \
+				goto lunatik_ebpf_lookup_skip; \
+			} \
+			keylen = snprintf(cpu_key, PERCPU_KEY_LEN, "%s:%d", (key), (cpu)); \
+			key = cpu_key; \
+		} \
+		runtime = luarcu_getobject((runtimes), (key), keylen); \
+		if (runtime == NULL) \
+			pr_err("couldn't find runtime '%s'\n", (key)); \
```
lneto, line 47: This logs per packet at XDP rates, and a percpu stop widens the exposure: the instances are torn down one by one, and every packet on a CPU whose instance is already gone lands here until the flag is cleared. The class predates this series, but percpu multiplies it by N. It deserves `pr_err_ratelimited`, as a separate commit.

Addressed in `review/642` by [`af581c9f`](https://github.com/luainkernel/lunatik/commit/af581c9f288ea69106cc4811e2dfd44dae1c7e60).

lib/lunatik/runner.lua
```diff
@@ -28,19 +29,30 @@ local function trim(script) -- drop ".lua" file extension
...
 
+local function key(script, cpu)
+	return script .. ":" .. cpu
+end
+
 --- Runs a Lunatik script in the current context.
 -- Creates a new Lunatik runtime for the given script and registers it.
 -- Throws an error if a script with the same name is already running.
 -- @tparam string script path or name of the Lua script to run. The ".lua" extension will be trimmed.
 -- @tparam[opt] string context Execution context: `"process"` (default) or `"softirq"` (for netfilter/XDP hooks).
```
lneto, line 40: Nothing restricts `percpu` to eBPF-dispatched scripts, and only those benefit: a netfilter script run percpu registers N hooks that all fire for every packet, and a mangle script would apply its rewrite N times, silently. There is no mechanical guard the runner could apply (it cannot know what the script registers), so the docs must carry it: one runtime per CPU id, the script runs once per instance, meant for scripts the eBPF bindings dispatch. The enforcement at the registration points is tracked in #675.

Addressed in `review/642` by [`868a7960`](https://github.com/luainkernel/lunatik/commit/868a79602c999e90cc3288e3e3fe82877ba6ca94) (docblock and README wording).

lneto (changes requested):
Second pass: lifecycle and failure paths. Each comment links the commit in [`review/642`](https://github.com/luainkernel/lunatik/tree/review/642) that addresses it. The enforcement of per-CPU semantics at the registration points is tracked in #675.

lib/luaxdp.c
sneaky-potato: This is already resolved in the latest push. Related [fixup](https://github.com/luainkernel/lunatik/pull/642/changes/040237e84c719b4daa929ce94fa628fb12c9d5ae).

lneto:
#676

PR #678, opened 2026-08-05
lunatik: enforce per-CPU semantics at the registration points

### LINKED ISSUE #675: percpu: enforce per-CPU semantics at the registration points
## Context

The percpu mode (#642, consolidated in [`review/642`](https://github.com/luainkernel/lunatik/tree/review/642)) creates one runtime per possible CPU id, registered as `<script>:<cpu>`, and the eBPF bindings resolve a call to the instance of the CPU the hook runs on. The script body runs once per instance.

## Problem

Nothing restricts `lunatik run <script> percpu` to scripts designed for it, and the load-time side effects run N times. The registrations differ in kind:

- `xdp.attach` (and the other eBPF bindings) store the callback in the instance's own registry: nothing global, one callback per runtime, dispatch already picks the right one. Correct by construction.
- `netfilter.register` installs a global hook per call: N hooks on the same chain, all firing for every packet. A counting script counts each packet N times; a mangle script applies its rewrite N times, silently. No error is raised anywhere.
- named resources (`device.new`, notifiers) fail on the second instance with EEXIST, and the rollback aborts the run cleanly. This class defends itself.

The dangerous case produces no error, and the script cannot be trusted to know better: any script-level declaration is trust-based and decides at the wrong layer. The binding is the one that knows whether its registration is per-runtime or global. The enforcement belongs there.

## Design

1. **The runtime carries its instance CPU id.** The runner creates percpu instances passing the cpu; a plain runtime has none. `lunatik_toruntime(L)` already gives every binding its runtime at registration time, so this is the only plumbing the rest needs.

2. **Each registration point applies its policy:**
   - eBPF `attach`: per-runtime state, nothing to do.
   - `netfilter.register` in a percpu instance: **affinity at dispatch**. The hook only acts on packets being processed on its instance's CPU and is transparent (`ACCEPT`) otherwise. `luanetfilter_docall` already has the early-out shape for this, next to the `mark` filter. This turns percpu netfilter from a footgun into a feature: per-CPU hook state with no locking, at the cost of N-1 cheap early-outs per packet.
   - registrations that cannot be made percpu-correct: refuse in a percpu instance, through a shared helper (`lunatik_checkruntime`-style), adopted binding by binding starting with the ones that hurt.

3. **Expose the id to Lua.** Once the runtime carries the cpu, a script can read its instance id, to shard state or name resources per instance. Comes for free and makes percpu-aware scripts natural to write.

## Acceptance criteria

- A percpu instance knows its CPU id, and a plain runtime has none.
- A netfilter hook registered by a percpu instance acts only on its own CPU's packets and is transparent to the rest of the chain otherwise.
- A registration that opted for refusal fails at load with an error naming percpu, not at packet time.
- A script can read its instance's CPU id.
- `tests/runtime/percpu.sh` covers the netfilter affinity and the refusal path.
- The `runner.run` docblock and the README describe the contract.

lneto:
Resolves #675: the registration points now apply the percpu policy, instead of trusting the script.

The runtime carries its instance CPU id in the state's extraspace, stored before the script is loaded so load-time registrations can read it, and `lunatik.cpu()` exposes it to the script (on the stub library as well: percpu scripts are usually softirq). `runner.run` no longer forwards its vararg tail, since a stray extra would silently claim an instance id for a plain runtime.

With the id in place, every registration a percpu instance cannot own is refused at load, with an error naming percpu, through a shared `lunatik_checkpercpu` helper:

- constructors whose registration is global: `device`, `notifier`, `probe`, `hid`;
- `netfilter.register`, which has no per-CPU dispatch: each instance would register its own hook and every packet would be handled once per instance. An earlier version of this branch shipped affinity at dispatch instead, each hook accepting everything not on its CPU, but that is exact only where the hook runs in softirq; on the process-context paths a migration mid-chain can handle a packet twice or not at all, so it was dropped rather than merged as a half measure. The exact form, a single hook per script dispatching to the instance of the current CPU, needs group identity in the core and builds on the percpu groups work (#693).

`syscall` was audited and left alone (address lookup only); the eBPF `attach` and the per-runtime objects need nothing.

Tests, each validated by breaking the code and watching it fail: `lunatik.cpu()` coverage in the percpu suite (every id stamped, a plain runtime sees `nil`), and the refusals for `device.new` and `netfilter.register`, failing at load in a percpu instance with a clean rollback, with the same scripts running plain.

Tested on 6.8.0-136: full suite green with clean dmesg; builds clean against 5.15 and 6.8 headers.

### DIFF
--- a/lunatik_core.c
+++ b/lunatik_core.c
@@ -157,12 +157,30 @@ static int lunatik_lresume(lua_State *L)
 	return nresults;
 }
 
+/***
+* Returns the CPU id of a percpu runtime instance.
+* Ranges from `0` to `linux.numcpus() - 1`; see `runner.run`.
+* @function cpu
+* @treturn integer instance CPU id, or `nil` on a plain runtime
+* @within lunatik
+*/
+static int lunatik_cpu(lua_State *L)
+{
+	if (lunatik_ispercpu(L))
+		lua_pushinteger(L, lunatik_getcpu(L));
+	else
+		lua_pushnil(L);
+	return 1;
+}
+
 static const luaL_Reg lunatik_lib[] = {
 	{"runtime", lunatik_lruntime},
+	{"cpu", lunatik_cpu},
 	{NULL, NULL}
 };
 
 static const luaL_Reg lunatik_stub_lib[] = {
+	{"cpu", lunatik_cpu},
 	{NULL, NULL}
 };
 
@@ -225,7 +243,7 @@ static int lunatik_runscript(lua_State *L)
 	return 1; /* callback */
 }
 
-static int lunatik_newruntime(lunatik_object_t **pruntime, lua_State *Lfrom, const char *script, lunatik_opt_t opt)
+static int lunatik_newruntime(lunatik_object_t **pruntime, lua_State *Lfrom, const char *script, lunatik_opt_t opt, int cpu)
 {
 	lunatik_object_t *runtime;
 	lua_State *L;
@@ -244,6 +262,7 @@ static int lunatik_newruntime(lunatik_object_t **pruntime, lua_State *Lfrom, con
 	lunatik_setobject(runtime, &lunatik_class, opt);
 	lunatik_toruntime(L) = runtime;
 	lunatik_extra(L)->ready = false;
+	lunatik_extra(L)->cpu = cpu;
 
 	runtime->gfp = GFP_KERNEL; /* might use kvmalloc while running in process */
 	lua_setallocf(L, lunatik_alloc, runtime);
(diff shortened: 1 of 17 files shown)


lunatik_core.c
```diff
@@ -271,7 +290,7 @@ static int lunatik_newruntime(lunatik_object_t **pruntime, lua_State *Lfrom, con
 
 int lunatik_runtime(lunatik_object_t **pruntime, const char *script, lunatik_opt_t opt)
 {
-	return lunatik_newruntime(pruntime, NULL, script, opt);
+	return lunatik_newruntime(pruntime, NULL, script, opt, -1);
```
sneaky-potato: Could we reuse `LUNATIK_CPU_NONE` here? I suppose `-1` here means no cpu instance.

sneaky-potato:
Hi @lneto the PR description talks about cpu affinity at dispatch but the code changes refuse percpu mode for netfilter. I think we should update the PR description just so that it remains correct if someone references it in future.

lunatik_core.c
lneto: Yes. That call is the one site the sweep missed when the sentinel was named. Fixed: https://github.com/luainkernel/lunatik/commit/6b5d4e3f

lneto: good catch, btw.. ;-) thanks!

sneaky-potato (approved):
LGTM 💪🏻

PR #685, opened 2026-08-06
Add common eBPF helper and migrate XDP

sneaky-potato:
This PR introduces a shared eBPF helper header to consolidate common functionality across Lunatik eBPF subsystems and refactors the XDP callback API to use a context object.

XDP callback API is refactored to pass a context object to Lua callbacks instead of individual arguments. This groups the callback state into a single object, simplifies the interface, and makes it easier to extend in the future.

## Breaking Change
`xdp.attach()` now expects a callback that accepts a single `ctx` parameter.
The `ctx` object exposes the packet and callback argument via `ctx:packet()` and `ctx:argument()`. User can set xdp action via `ctx:action()`

Existing callbacks of the form:
```lua
local function filter_sni(packet, arg)
    ...
end
xdp.attach(filter_sni)
```

must be updated to:
```lua
local function filter_sni(ctx)
    local packet = ctx:packet()
    local arg = ctx:argument()
    ...
    ctx:action(action.PASS)
end
xdp.attach(filter_sni)
```


lib/luaxdp.c
```diff
@@ -20,142 +20,160 @@
...
 {
-	static const char runtimes_key[] = "runtimes";
-	static const char percpu_key[] = "percpu";
-
-	if (luaxdp_runtimes == NULL)
-		luaxdp_runtimes = luarcu_getobject(lunatik_env, runtimes_key, sizeof(runtimes_key) - 1);
-	if (luaxdp_percpu == NULL)
-		luaxdp_percpu = luarcu_getobject(lunatik_env, percpu_key, sizeof(percpu_key) - 1);
-	return luaxdp_runtimes != NULL && luaxdp_percpu != NULL ? 0 : -1;
+	luaxdp_ctx_t *lctx = lunatik_ebpf_getctx(L, &luaxdp_env_key);
```
lneto, line 132: `lunatik_ebpf_getctx` returns NULL when the key is not in the registry, which is the case for a script that ran but never called `xdp.attach`, or that has detached. The handler dereferences it straight away, so an eBPF program calling `bpf_luaxdp_run` with the name of any running Lunatik script panics the machine. The old code checked `!= LUA_TFUNCTION` and bailed out.

Addressed in `review/685` by [`79877b4a`](https://github.com/luainkernel/lunatik/commit/79877b4a315d1aba8677ae732595bbce4a8b6c4e).

lib/luaxdp.c
```diff
@@ -169,7 +187,11 @@ static const struct btf_kfunc_id_set bpf_luaxdp_kfunc_set = {
 */
 static int luaxdp_detach(lua_State *L)
 {
-	lunatik_unregister(L, luaxdp_callback);
+	luaxdp_ctx_t *lctx = lunatik_ebpf_getctx(L, &luaxdp_env_key);
+	luaL_unref(L, LUA_REGISTRYINDEX, lctx->callback_ref);
```
lneto, line 208: Same NULL here: `xdp.detach()` without a previous `attach` dereferences it on `lctx->callback_ref`. Reproduced with a two-line script.

While in this function: it unrefs the callback and unregisters the ctx key, but not the `packet` and `argument` entries `attach` registered, so they stay in the registry and an attach/detach cycle accumulates a pair each time. The base keeps this kind of cleanup in `detach` — `release` should be a noop — which is what `lunatik_detach`/`luadata_attach` are for.

Addressed in `review/685` by [`1361f9ec`](https://github.com/luainkernel/lunatik/commit/1361f9ec032669187de4f92e85d18e94f2b02d78).

lib/luaxdp.c
```diff
@@ -20,142 +20,160 @@
...
+	return ret;
 }
 
 __bpf_kfunc int bpf_luaxdp_run(char *key, size_t key__sz, struct xdp_md *xdp_ctx, void *arg, size_t arg__sz)
 {
-	lunatik_object_t *runtime;
-	struct xdp_buff *ctx = (struct xdp_buff *)xdp_ctx;
-	int action = -1;
-	size_t keylen = key__sz - 1;
+	int action = XDP_PASS;
```
lneto: This changes policy: it used to be `-1`, so a failed lookup, a missing callback or a Lua error left the verdict to the eBPF program. `XDP_PASS` makes all of them pass the packet, which is fail-open for a filter.

Restored to `-1` in `review/685` by [`e5c2f63a`](https://github.com/luainkernel/lunatik/commit/e5c2f63a51c07a5647de93bbdffc9175a9139724), with the docblock stating it.

lunatik_ebpf.h
```diff
@@ -0,0 +1,89 @@
...
+	static const char runtimes_key[] = "runtimes";
+	static const char percpu_key[] = "percpu";
+	if (*runtimes == NULL)
+		*runtimes = luarcu_getobject(lunatik_env, runtimes_key, sizeof(runtimes_key) - 1);
+	if (*percpu == NULL)
+		*percpu = luarcu_getobject(lunatik_env, percpu_key, sizeof(percpu_key) - 1);
+	return *runtimes && *percpu ? 0 : -1;
+}
+
+static inline lunatik_object_t *lunatik_ebpf_lookupruntime(lunatik_object_t **runtimes, lunatik_object_t **percpu, char *key, size_t key_sz, int cpuid)
```
lneto: 151 columns; the limit is 120, and `luaxdp.c` has one at 128 as well.

Also, the shared header includes `"lib/luarcu.h"`, reaching into `lib/` from the root, where the modules include `"luarcu.h"` themselves.

Addressed in `review/685` by [`8f010642`](https://github.com/luainkernel/lunatik/commit/8f010642ba6cfb1c0a9d1f7685f46443f74a4ea3) (line lengths only; the include path is your call).

lib/luaxdp.c
```diff
@@ -20,142 +20,160 @@
...
+}
 
-	if (lunatik_getregistry(L, luaxdp_callback) != LUA_TFUNCTION) {
-		pr_err("couldn't find callback");
-		goto out;
-	}
+static const luaL_Reg luaxdp_mt[] = {
+	{"__gc",		lunatik_deleteobject},
+	{"packet",		luaxdp_packet},
+	{"argument",	luaxdp_arg},
```
lneto: `luaxdp_arg` is registered as `"argument"`; the C name and the Lua name should match. Two style items nearby: `luaxdp_handler_cleanup` opens its brace on the same line, and one line inside `luaxdp_arg` is indented with spaces.

Addressed in `review/685` by [`d4d3ac9f`](https://github.com/luainkernel/lunatik/commit/d4d3ac9fd7728c30a17269300ebd9cb87abce252).

lib/luaxdp.c
```diff
@@ -20,142 +20,160 @@
...
-	return data;
-}
+LUNATIK_PRIVATECHECKER(luaxdp_ctx_check, luaxdp_ctx_t *,
+	luaL_argcheck(L, private->xdp != NULL, ix, "ctx is not set");
+);
 
-static int luaxdp_callback(lua_State *L)
+/***
+* Returns the packet data buffer for the current XDP context.
+* @function packet
```
lneto: The new class has methods but no `@type` block, so LDoc attaches `packet`, `argument` and `action` to the module instead of the type.

Addressed in `review/685` by [`ab78d7fa`](https://github.com/luainkernel/lunatik/commit/ab78d7fa333d51ad51833500beb5c0a7f0751e12), which adds `@type xdp_ctx` and anchors the three as `xdp_ctx:<method>`; the type cannot be named `xdp` or it collides with the module.

lneto (changes requested):
The shared header removes real duplication between the bindings, and the ctx object closes correctly against use after the callback returns: the secondary check in `luaxdp_ctx_check` plus the cleanup that nulls `xdp`/`action` cover the pointer to the kfunc's stack variable, and `SINGLE` keeps the ctx out of `_ENV`.

What needs addressing are two NULL dereferences that panic the machine, a registry leak on detach, and a change of verdict policy that is not in the description. Each comment links the commit in [`review/685`](https://github.com/luainkernel/lunatik/tree/review/685) that addresses it, one commit per finding.

One gap worth naming: none of this is covered by a test. The XDP dispatch path has no harness today, so the fixes here were verified by reading and, for the detach path, by running a two-line script.

tests/xdp/test_xdp.sh
```diff
@@ -0,0 +1,83 @@
...
+# SPDX-License-Identifier: MIT OR GPL-2.0-only
+#
+
+# Tests the XDP dispatch path by loading an XDP program with bpftool,
+# attaching it to an interface, and running a Lunatik XDP script.
+#
+# Usage: sudo bash tests/xdp/run.sh
+
+MODULE="luaxdp"
+IFACE="docker0"
```
lneto: The test hard-depends on `docker0` and a `docker run alpine`: on a host without Docker it fails instead of testing, and the first run pulls an image off the network. A veth pair with the peer in a netns needs nothing from the host and makes the ping path deterministic: https://github.com/luainkernel/lunatik/commit/91644367

tests/xdp/test_xdp.sh
```diff
@@ -0,0 +1,83 @@
...
+	exit 0
+}
+
+command -v bpftool > /dev/null 2>&1 || {
+	echo "# SKIP: bpftool not available"
+	ktap_totals
+	exit 0
+}
+
+cleanup()
```
lneto, line 42: The Lua scripts are never stopped, so the second run of the suite fails with "already running", and `cleanup` leaves the runtimes behind. Stop them in `cleanup` and after each case, and run `cleanup` once up front, as the other suites do: https://github.com/luainkernel/lunatik/commit/0b0dceac

tests/xdp/test_xdp.sh
```diff
@@ -0,0 +1,83 @@
...
+{
+	bpftool net detach xdp dev "$IFACE" 2>/dev/null
+	rm -f "${PIN}_PASS"
+	rm -f "${PIN}_DROP"
+}
+
+trap cleanup EXIT
+
+make -C tests/xdp || { ktap_fail "failed to build XDP program"; ktap_totals; exit 1; }
```
lneto: `make -C tests/xdp` and `bpftool prog load "tests/xdp/$obj"` resolve from the caller's cwd, so the test only works from the repo root, and never from the install dir, where `lunatik test xdp` runs it. Use the suite's own directory: https://github.com/luainkernel/lunatik/commit/1cd33c2b

tests/xdp/test_xdp.sh
```diff
@@ -0,0 +1,83 @@
...
+IFACE="docker0"
+PIN="/sys/fs/bpf/xdp"
+TARGET=$(ip -4 addr show "$IFACE" | awk '/inet / {sub(/\/.*/, "", $2); print $2}')
+
+source "$(dirname "$(readlink -f "$0")")/../lib.sh"
+
+ktap_header
+ktap_plan 1
+
+cat /sys/module/$MODULE/refcnt > /dev/null 2>&1 || {
```
lneto: A module built without BTF (the default `make` when the build dir has no vmlinux) loads fine, but `bpftool prog load` then fails to resolve the kfunc: `extern (func ksym) 'bpf_luaxdp_run': not found in kernel or module BTFs`. That is an environment limitation, not a regression, so it must skip, like the missing module and missing bpftool; clang gets the same treatment. The plan is also printed twice (`1..1` then `1..2`), which is not KTAP, and the skip paths emit no test lines: plan once and use `ktap_skip`: https://github.com/luainkernel/lunatik/commit/50701571

tests/xdp/test_xdp.sh
```diff
@@ -0,0 +1,83 @@
...
+	echo "# SKIP: bpftool not available"
+	ktap_totals
+	exit 0
+}
+
+cleanup()
+{
+	bpftool net detach xdp dev "$IFACE" 2>/dev/null
+	rm -f "${PIN}_PASS"
+	rm -f "${PIN}_DROP"
```
lneto: The drop case pins `${PIN}_drop` but `cleanup` removes `${PIN}_DROP`, leaking the pin on failure paths: https://github.com/luainkernel/lunatik/commit/2d6a139d

tests/xdp/test_xdp.sh
```diff
@@ -0,0 +1,83 @@
...
+
+#
+# SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+# SPDX-License-Identifier: MIT OR GPL-2.0-only
+#
+
+# Tests the XDP dispatch path by loading an XDP program with bpftool,
+# attaching it to an interface, and running a Lunatik XDP script.
+#
+# Usage: sudo bash tests/xdp/run.sh
```
lneto: The usage line names `run.sh` but the file is `test_xdp.sh`, there is a stray blank line inside the license block, and the `.lua` files are missing the one-line pointer back to the `.sh` the test convention asks for: https://github.com/luainkernel/lunatik/commit/9d770815

tests/xdp/Makefile
```diff
@@ -0,0 +1,17 @@
...
+	bpftool btf dump file /sys/kernel/btf/vmlinux format c > $@
+
+xdp_pass.bpf.o: xdp_pass.bpf.c vmlinux.h
+	clang -target bpf -Wall -O2 -c -g $<
+
+xdp_drop.bpf.o: xdp_drop.bpf.c vmlinux.h
+	clang -target bpf -Wall -O2 -c -g $<
+
+clean:
+	rm -f vmlinux.h xdp.bpf.o
```
lneto: `clean` removes `xdp.bpf.o`, which no rule produces, and leaves the two objects the rules do produce: https://github.com/luainkernel/lunatik/commit/030a57f8

lib/luaxdp.c
```diff
@@ -223,11 +262,25 @@ static int luaxdp_attach(lua_State *L)
 	lunatik_checkruntime(L, LUNATIK_OPT_SOFTIRQ);
 	luaL_checktype(L, 1, LUA_TFUNCTION); /* callback */
```
lneto, line 264: Attaching twice leaks the previous callback: the old context object is replaced in the registry and collected, but its `callback_ref` stays in the registry table for the life of the state. Route a re-attach through `detach` first: https://github.com/luainkernel/lunatik/commit/94a30c04

lib/luaxdp.c
```diff
@@ -12,150 +12,180 @@
...
-	}
-
-	lunatik_run(runtime, luaxdp_handler, action, ctx, arg, arg__sz);
+	luaxdp_ctx_t ctx = {
+		.xdp     = (struct xdp_buff *)xdp_ctx,
+		.arg     = arg,
+		.arg__sz = arg__sz,
+		.action  = &action,
+	};
+	int ret;
```
lneto: `int ret;` is declared mid-block between the ctx initializer and the call; declarations go at the top with `action`: https://github.com/luainkernel/lunatik/commit/1024d55f

lib/luaxdp.c
```diff
@@ -12,150 +12,180 @@
...
-	}
+	luaxdp_ctx_t *ctx = luaxdp_ctx_check(L, 1);
+	*ctx->action = luaL_checkinteger(L, 2);
+	return 0;
+}
 
-	luadata_clear(buffer);
-	luadata_clear(argument);
-	return 1;
+static const luaL_Reg luaxdp_mt[] = {
```
lneto, line 96: The method table is tab-aligned; the codebase writes `{"name", fn},` with a single space: https://github.com/luainkernel/lunatik/commit/8ac0c79d

lunatik_ebpf.h
```diff
@@ -0,0 +1,90 @@
...
+	if (!runtime) {
+		char cpu_key[LUARCU_MAXKEY];
+		size_t cpulen;
+		cpulen = scnprintf(cpu_key, sizeof(cpu_key), "%s:%d", key, cpuid);
+		runtime = luarcu_getobject(*percpu, cpu_key, cpulen);
+	}
+	return runtime;
+}
+
+static inline void *lunatik_ebpf_getctx(lua_State *L, char *env_key)
```
lneto, line 43: `lunatik_ebpf_getctx` pops on failure but leaves the userdata on the stack on success. That is deliberate, since the caller passes it to the callback, but invisible at the prototype. One line saying so: https://github.com/luainkernel/lunatik/commit/9fbd1716

lunatik_ebpf.h
```diff
@@ -0,0 +1,90 @@
...
+	static const struct btf_kfunc_id_set bpf_lua##subsys##_kfunc_set = { \
+		.owner = THIS_MODULE, \
+		.set   = &bpf_lua##subsys##_set, \
+	};
+
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 4, 0))
+#define LUNATIK_EBPF_KFUNC_INIT(subsys, prog_type) \
+	return register_btf_kfunc_id_set(prog_type, &bpf_lua##subsys##_kfunc_set);
+#else
+#define LUNATIK_EBPF_KFUNC_INIT(subsys, prog_type) return 0;
```
lneto: `LUNATIK_EBPF_KFUNC_INIT` expands to a `return` statement, hiding control flow inside a macro. Make it a template that generates the whole init function, as `LUNATIK_NEWLIB` does for the opener: https://github.com/luainkernel/lunatik/commit/5016f3f1

tests/xdp/pass.lua
```diff
@@ -0,0 +1,32 @@
+--
+-- SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+
+local xdp    = require("xdp")
+local action = require("linux.xdp")
+
+local function test_pass(ctx)
+	if not ctx then
```
lneto: The `if not ctx/packet/argument` branches cannot execute: those calls raise instead of returning nil, so the guards are dead code that misstates the API's contract. Calling the methods is the assertion: https://github.com/luainkernel/lunatik/commit/b608615d

lunatik_ebpf.h
```diff
@@ -0,0 +1,90 @@
...
+	}
+	return runtime;
+}
+
+static inline void *lunatik_ebpf_getctx(lua_State *L, char *env_key)
+{
+	lunatik_object_t *obj;
+	if (lunatik_getregistry(L, env_key) != LUA_TUSERDATA) {
+		lua_pop(L, 1);
+		pr_err("couldn't find the context object\n");
```
lneto: After `xdp.detach`, every packet that still reaches the kfunc logs `couldn't find the context object` unratelimited, a documented, normal state flooding the log at line rate: https://github.com/luainkernel/lunatik/commit/ac2cb0aa

lneto (changes requested):
Second round. The first-round changes were incorporated; the findings below are on the new material, mostly the test suite. Each comment links a commit in `review/685`: corrections are fixups of the commit they fix, and the coverage additions at the end are standalone commits. With the whole branch applied, the suite passes here end to end, twice in a row, from the source tree and from the install dir, skips cleanly when the module is unloaded, and the full run is 102/102.

The suite is not reachable: `tests/run.sh` does not run it, the README suite list does not name it, and `tests_install` does not install the Makefile and `.bpf.c`, so the installed tree cannot build the programs. A test is not done until the runner and the docs know about it: https://github.com/luainkernel/lunatik/commit/63d8a9fb

Coverage is the matrix of operation by outcome: `detach` was untested, the argument path never carried data (the program passes `NULL, 0`), the plain-name kfunc lookup never ran, and the refusal on a sleepable runtime was uncovered. These are additions rather than corrections, so the review branch carries them as standalone commits on top of the fixups; take them as they are or as a starting point: https://github.com/luainkernel/lunatik/commit/6dc4fb81, https://github.com/luainkernel/lunatik/commit/1315f689, https://github.com/luainkernel/lunatik/commit/44ea155a. The neighbor pinning that makes the drop and detach assertions deterministic is a correction of the substrate and stays a fixup: https://github.com/luainkernel/lunatik/commit/88092278

lneto (commented):
Third round, on the fold. The branch matches the review branch byte for byte, the merge with current master is clean, and the full suite runs green here, with the xdp suite passing twice in a row. Two asks before merge, neither about the code:

1. Rebase onto current master; the base is twelve commits behind.

2. Two commits lost their story in the fold. `xdp: refactor callback api to use ctx object` changes what every XDP script sees, and the diff does not say what a script must do differently; `lunatik: add shared ebpf helper header` does not say who the header is for. Suggested bodies, take or rewrite:

For the refactor:

> The callback now receives a single ctx object instead of the packet and argument buffers: `ctx:packet()` and `ctx:argument()` return them, and the verdict is set through `ctx:action()` instead of being returned, so a callback that sets no action leaves the verdict to the eBPF program. The context is valid only while the callback runs; its methods raise afterwards.

For the header:

> luaxdp carried the code every eBPF-facing module will need: looking up the runtime by the key the eBPF program passes, fetching the context object from the registry, and the version guards around kfunc registration. Move it into `lunatik_ebpf.h`, so the next module exposing a kfunc includes the header instead of copying luaxdp.

On the `examples/filter/Makefile` observation: confirmed. The target writes `vmlinux.h` but is named `vmlinux`, so it is never up to date and the dump reruns on every make; `https.o` also does not depend on the header it includes. That predates this pull request, so it takes its own one. Send it with the target renamed to `vmlinux.h` and the prerequisites following; `tests/xdp/Makefile` already has the right shape to copy.

sneaky-potato:
Rebased on current master, reworded the commits to include description

lneto (approved):

PR #708, opened 2026-08-10
examples/filter: fix vmlinux.h dependency for https.o

sneaky-potato:
Rename the vmlinux target to vmlinux.h and make https.o depend on it, so bpftool regenerates the header only when needed instead of running unconditionally on every build.


lneto (commented):
Verified here: a clean build generates both artifacts, the second and third runs are no-ops, and clean removes them. One thing: the copyright year of a modified file extends to the current one; fixup ready: https://github.com/luainkernel/lunatik/commit/2e2a5b00

lneto (changes requested):
The copyright fixup above is required before merge.

lneto (approved):

PR #713, opened 2026-08-11
tc: add binding with tests

sneaky-potato:
Add the luatc binding with TC test suite covering the Lunatik TC/eBPF dispatch path.

The tests exercise TC programs attached to a veth pair and verify that Lua callbacks can inspect the skb, set TC verdicts, and handle runtime lifecycle case


examples/sniclassify/sni.lua
```diff
@@ -0,0 +1,116 @@
+--
+-- SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+-- SPDX-FileCopyrightText: (c) 2024-2026 Ring Zero Desenvolvimento de Software LTDA
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+-- Based on https://github.com/luainkernel/lunatik/blob/master/examples/filter/sni.lua
```
lneto: I wouldn't duplicate such code, but would move it to a common place..

lib/luaskb.c
```diff
@@ -244,21 +288,26 @@ static const luaL_Reg luaskb_lib[] = {
...
+	{"ifindex",     luaskb_ifindex},
+	{"vlan",        luaskb_vlan},
+	{"data",        luaskb_data},
+	{"resize",      luaskb_resize},
+	{"checksum",    luaskb_checksum},
+	{"forward",     luaskb_forward},
+	{"copy",        luaskb_copy},
 #if defined(CONFIG_NF_CONNTRACK_MARK)
-	{"connmark", luaskb_connmark},
+	{"connmark",    luaskb_connmark},
```
lneto: changes on formatting and style should come as a separate PR

lib/luatc.c
```diff
@@ -0,0 +1,295 @@
...
+	.methods = luatc_mt,
+	.release = luatc_release,
+	.opener  = luaopen_tc,
+	.opt     = LUNATIK_OPT_SOFTIRQ | LUNATIK_OPT_SINGLE,
+};
+
+static void luatc_handler_cleanup(luatc_ctx_t *lctx)
+{
+	luaskb_t *lskb = (luaskb_t *)lctx->skb_obj->private;
+	lskb->skb = NULL;
```
lneto: `luatc_handler_cleanup` nulls `lskb->skb` and `lctx->action`, but not `lctx->skb` — the field `luatc_ctx_check` guards on. So the ctx stays "armed" after the callback returns and the `@type` doc's "methods raise once the callback returns" does not hold; luaxdp nulls its guard field (`lctx->xdp`) in the same spot. The fixup nulls it and clears through `luaskb_clear` instead of reaching into `->private`: https://github.com/luainkernel/lunatik/commit/646fe0e1

lib/luatc.c
```diff
@@ -0,0 +1,295 @@
...
+	{"__gc", lunatik_deleteobject},
+	{"skb", luatc_skb},
+	{"action", luatc_action},
+	{NULL, NULL}
+};
+
+static void luatc_release(void *private)
+{
+	luatc_ctx_t *lctx = (luatc_ctx_t *)private;
+	if (lctx->skb_obj)
```
lneto: `attach` takes `lunatik_getobject(ctx->skb_obj)`, and release only `luaskb_clear`s it, never `lunatik_putobject` — so the skb object's reference is never dropped and it leaks. luaxdp's release balances the get with `luadata_close` (clear + put). Fixup adds the put: https://github.com/luainkernel/lunatik/commit/1837c1e2

lib/luatc.c
```diff
@@ -0,0 +1,295 @@
...
+*     local skb = ctx:skb()
+*     print("Packet received, size:", #skb)
+*     ctx:action(action.ACT_OK)
+*     return
+*   end
+*   tc.attach(my_traffic_shaper)
+*
+*   -- In eBPF C code, to call the above Lua function:
+*   -- char rt_key[] = "my_tc_handler.lua"; // Key matches the script name
+*   -- int verdict = bpf_luatc_run(rt_key, sizeof(rt_key), ctx, NULL, 0);
```
lneto: The eBPF example calls `bpf_luatc_run(rt_key, sizeof(rt_key), ctx, NULL, 0)` — five arguments copied from the XDP doc; the kfunc takes three (`key, key__sz, skb`): https://github.com/luainkernel/lunatik/commit/b74170b5

lib/luatc.c
```diff
@@ -0,0 +1,295 @@
...
+		lunatik_putobject(luatc_runtimes);
+	if (luatc_percpu != NULL)
+		lunatik_putobject(luatc_percpu);
+#endif
+}
+
+module_init(luatc_init);
+module_exit(luatc_exit);
+MODULE_LICENSE("Dual MIT/GPL");
+MODULE_AUTHOR("Ashwani Kumar Kamal <ashwanikamal/im421@gmail.com>");
```
lneto: `ashwanikamal/im421@gmail.com` — a slash where the dot goes: https://github.com/luainkernel/lunatik/commit/a1176d5c

lib/luaskb.c
```diff
@@ -228,6 +228,50 @@ static int luaskb_connmark(lua_State *L)
 }
 #endif /* CONFIG_NF_CONNTRACK_MARK */
 
+#define luaskb_getinteger(name, field) \
+static int luaskb_get##name(lua_State *l) \
```
lneto: The new get/set macros declare `lua_State *l`, lowercase, where the whole file uses `L`; there is also a stray double blank line before `LUNATIK_OPENER(skb)`. Fixup: https://github.com/luainkernel/lunatik/commit/07f298f2

tests/tc/test_tc.sh
```diff
@@ -0,0 +1,164 @@
...
+ip neigh replace "$TARGET" lladdr "$PEER_MAC" dev "$IFACE" nud permanent
+ip netns exec "$NETNS" ip neigh replace "$HOST" lladdr "$MAC" dev "$PEER" nud permanent
+
+run_case()
+{
+	local obj="$1" script="$2" expect_reachable="$3" label="$4" title="$5"
+	shift 5
+
+	tc qdisc add dev "$IFACE" clsact 2>/dev/null
+	tc filter add dev "$IFACE" egress bpf da obj "$DIR/$obj" sec classifier ||
```
lneto: The suite loads with `tc ... bpf da obj`, which uses iproute2's bundled libbpf; on stock iproute2 (5.15 here) that is too old to load a modern object on 6.8, so every case fails to attach. Load with bpftool and attach the pinned program (`tc ... bpf da object-pinned`), as the XDP suite uses bpftool — the suite then passes on stock iproute2. `detach_case` and `zerokey_case` also referenced `$label`, a local only in `run_case`, printing an empty label; header said "efress". Fixups: https://github.com/luainkernel/lunatik/commit/211db02e , https://github.com/luainkernel/lunatik/commit/7e434625

lneto (changes requested):
The binding mirrors `luaxdp`, and its two divergences from it are the two substantive findings. Each comment links a fixup in `review/713`; with all of them the full suite is green here (114/114).

One caveat worth stating plainly: the suite did not attach on this machine at first, because it loads with `tc ... bpf da obj`, which uses iproute2's bundled libbpf (here 5.15 / libbpf 0.5.0, too old to load a BPF object on 6.8). The binding itself is fine: the same object loads under bpftool v7.4.0 and the whole dispatch runs (drop blocks the ping, pass reads the packet and passes it). The loader is one of the fixups below; see the test-suite comment.

`bpf_luatc_run` still takes the two-table `lunatik_ebpf_lookupruntime`, correct against current master (it inherits the zero-key guard), but it will conflict with the percpu-groups branch that collapses that lookup to one table, exactly as luaxdp does — whichever merges second pays that rebase.

lib/luatc.c
```diff
@@ -0,0 +1,295 @@
...
+*   end
+*   tc.attach(my_traffic_shaper)
+*
+*   -- In eBPF C code, to call the above Lua function:
+*   -- char rt_key[] = "my_tc_handler.lua"; // Key matches the script name
+*   -- int verdict = bpf_luatc_run(rt_key, sizeof(rt_key), ctx, NULL, 0);
+* @see skb
+* @within tc
+*/
+static int luatc_attach(lua_State *L)
```
lneto, line 212: `luatc.c` is a near-copy of `luaxdp.c`: the callback-binding skeleton — env key, ctx guard, `action`, `release`, `handler`, the kfunc's lookup-build-run-put, `attach`/`detach` — is identical; only the ctx payload, the accessor methods, and the per-call field reset differ. `lunatik_ebpf.h` already holds the shared lookup and init; this skeleton is the next layer to move there. It belongs in its own pull request landing before this one: lift the skeleton out of luaxdp first, then luatc here is written against it rather than copied from it.

lneto (commented):
A second round, on architecture and coverage rather than correctness.

tests/tc/pass.lua
```diff
@@ -0,0 +1,29 @@
...
+local action  = require("linux.tc")
+local skbattr = require("skb.attr")
+
+local ETH_HI   <const> = 12 -- ethertype, then IPv4 protocol
+local ETH_LO   <const> = 13
+local IPPROTO  <const> = 23
+local ICMP     <const> = 1
+
+local function test_pass(ctx)
+	local skb = skbattr(ctx:skb())
```
lneto, line 17: The attribute accessors (`getmark`/`setmark`/`getpriority`/`setpriority`) and `skb.attr` had no test. The fixup extends this pass case to round-trip mark and priority through them, and it discriminates — breaking the read fails the case: https://github.com/luainkernel/lunatik/commit/8b0c261d

tests/tc/test_tc.sh
```diff
@@ -0,0 +1,177 @@
...
+	tc_unload
+	lunatik stop tests/tc/detach > /dev/null 2>&1
+
+	dmesg_since | grep -qF "tc detach test pass" || { ktap_fail "tc detach: callback did not run"; return 1; }
+	ktap_pass "tc detach: callback stops firing and traffic resumes"
+}
+
+zerokey_case()
+{
+	tc_load zerokey.bpf.o ||
```
lneto: `zerokey_case` loads `zerokey.bpf.o`, but the Makefile builds `tc_zerokey.bpf.o` like every other case — so it never loads its program and fails regardless of the zero-key path it means to exercise. Fixup: https://github.com/luainkernel/lunatik/commit/0f725467

lib/luatc.c
```diff
@@ -0,0 +1,258 @@
...
+	lunatik_object_t *object = lunatik_newobject(L, &luatc_class, sizeof(luatc_ctx_t), LUNATIK_OPT_NONE);
+	luatc_ctx_t *ctx = (luatc_ctx_t *)object->private;
+
+	ctx->skb_obj = luaskb_new(L);
+	lunatik_getobject(ctx->skb_obj);
+	lunatik_register(L, -1, ctx->skb_obj);
+	lua_pop(L, 1);
+
+	lua_pushvalue(L, 1);
+	ctx->callback_ref = luaL_ref(L, LUA_REGISTRYINDEX);
```
lneto: The binding already uses the shared layer everywhere else (`getctx`, `invoke`, `LUNATIK_EBPF_RUN`); `attach`/`detach` are the only spots left open-coding the callback ref-and-register that `lunatik_ebpf_attach`/`lunatik_ebpf_detach` provide. Call them, as `luaxdp` does. Fixup: https://github.com/luainkernel/lunatik/commit/e283cd08

lib/luatc.c
```diff
@@ -0,0 +1,248 @@
...
+*   tc.attach(my_traffic_shaper)
+*
+*   -- In eBPF C code, to call the above Lua function:
+*   -- char rt_key[] = "my_tc_handler.lua"; // Key matches the script name
+*   -- int verdict = bpf_luatc_run(rt_key, sizeof(rt_key), skb);
+* @see skb
+* @within tc
+*/
+static int luatc_attach(lua_State *L)
+{
```
lneto, line 213: `attach` does not detach the previous binding first, so re-attaching leaks the old ctx (pinned in the registry by its `skb_obj`) and its callback ref — the env-key registration is overwritten, but nothing releases the old one. `luaxdp_attach` calls `luaxdp_detach(L)` for exactly this. Fixup: https://github.com/luainkernel/lunatik/commit/5625d49b

lib/luatc.c
```diff
@@ -0,0 +1,248 @@
...
+
+static int luatc_handler(lua_State *L, luatc_ctx_t *ctx)
+{
+	luatc_ctx_t *lctx = lunatik_ebpf_getctx(L);
+	int ret = 0;
+
+	if (lctx == NULL)
+		return -1;
+
+	luaskb_t *lskb = (luaskb_t *)lctx->skb_obj->private;
```
lneto, line 118: The handler open-codes the skb reset by reaching into `skb_obj->private`; `luaskb_reset` (used by `luanetfilter`) does that. The fixup uses it, through a `struct sk_buff *skb` local — the macro's `skb` parameter aliases the `->skb` field it assigns, so it needs an argument literally named `skb`. Fixup: https://github.com/luainkernel/lunatik/commit/0ad913e9

lib/luaskb.c
```diff
@@ -228,6 +228,50 @@ static int luaskb_connmark(lua_State *L)
...
+{ \
+	luaskb_t *lskb = luaskb_check(L, 1); \
+	lskb->skb->field = (u32)luaL_checkinteger(L, 2); \
+	return 0; \
+}
+
+/***
+* Gets the packet mark.
+* @function getmark
+* @return skb->mark
```
lneto, line 250: The mark/priority accessors document with `@return`/`@param`; the rest of the file and the base use typed `@treturn integer`/`@tparam integer`. Fixup: https://github.com/luainkernel/lunatik/commit/2de30869

lneto (changes requested):

lneto (approved):
The three findings are folded: `luatc_detach(L)` on re-attach, `luaskb_reset` through the `skb` local, and the typed accessor docs. Good to go.

lneto (changes requested):
**The mark/priority attribute feature doesn't match how the tree shapes accessors or object modules.**

*C accessors.* `getmark`/`setmark`/`getpriority`/`setpriority` are get/set pairs, but luaskb already exposes a writable scalar attribute — `connmark` — as a single overloaded method: `connmark()` reads, `connmark(v)` writes. luadata's `getX`/`setX` are *indexed* accessors (`getuint8(offset)`), where read and write can't be told apart by arity; a plain attribute can, and `connmark` is the shape to follow. Folded to overloaded `mark`/`priority` on one macro that derives the field width with `typeof` rather than a hardcoded `(u32)` — four methods become two.

*`skb.attr`.* It returns a bare function; object-producing modules return a class or namespace table (`socket.inet`, `bpf.map`, `hkdf`), a bare-function return being for builders like `class`/`struct`. It names its metatable `mt` and its field `_skb`, where the tree names by role and holds the handle in a named field (`socket`, `tfm`); it carries no `@classmod`/`@type`; and its `__index` builds a closure on every method fetch and dispatches through `"get" .. key` — per packet, in a softirq path. Folded to a `@classmod` view modeled on `bpf.map`'s `view` and `hkdf`: a named `attr` table, the skb in a `skb` field reachable as `view.skb`, `attr.new` to construct, and a fixed attribute set looked up once.

https://github.com/luainkernel/lunatik/commit/63f1e6a7d9df0f4aa88385c769089539a937467a

The one caller passes the raw skb for `:data()` and constructs with `.new`:

https://github.com/luainkernel/lunatik/commit/e6520be5071cc40ee8ec9542b49baf24b6d10d90

lneto:
Following up on the fold above — a few refinements to the same two commits:

- The accessor macro caches `lskb->skb` into a local `skb`, as the other methods do (`ifindex`, `vlan`, `connmark`), so the body reads `skb->field` instead of repeating the chain.
- In `skb.attr` the guard is now a plain allow-list (`fields`), checked the same way on both sides: reading or writing an attribute the view does not define raises, where a read previously returned `nil` silently — a typo'd read misclassifies as quietly as a typo'd write. The name→method map was an identity map, so it (and the comment explaining it) is gone; the key is the method name.

https://github.com/luainkernel/lunatik/commit/0edb795d6f3e500123e2298c1d08acce6c02edcd

lneto:
Another pass, comparing `luatc.c` against `luaxdp.c` and checking the branch's base.

- **Rebase needed.** The branch is a commit behind `master` (the single-CPU percpu skip, `91172f4c`), so the diff carries a spurious reversal of `tests/runtime/percpu.sh`. Rebasing onto current `master` before the autosquash drops it; the fixups below do not touch that file, so they apply on the rebased base.
- **`luatc_handler_cleanup` is not `static inline`** — its `luaxdp` counterpart is. https://github.com/luainkernel/lunatik/commit/c4b9dda7ae7cd8b4cc62f2752f36c36f1faaa459
- **The docs read "an TC"**, copied from `luaxdp`'s "an XDP"; TC takes "a". https://github.com/luainkernel/lunatik/commit/992b33a36b15fa844f03d27cf578bbbaa10ca938
- **`tests/tc/attach_sleepable.lua`'s SPDX header is missing the closing `>`** on the email. https://github.com/luainkernel/lunatik/commit/eb214e1042b1c49b3fe743b1145a7a916aa70936

lneto:
Another pass over the eBPF programs and the build integration.

- **The BPF license line was inconsistent.** `tc_zerokey` declared `LICENSE[] = "GPL"` while the other three tc programs use `_license[] = "Dual MIT/GPL"` (the form `examples/filter/https.c` uses) — aligned zerokey up to the dual form so all four match. https://github.com/luainkernel/lunatik/commit/40692d37973007566e95b85ba34a7a1fb657ad2e
- **"Traffic Controller" → "Traffic Control".** Linux `tc` is traffic *control* (`man tc`); the module doc and the Kconfig help both read "Controller". https://github.com/luainkernel/lunatik/commit/31a2b3cb7e806b7e09137c7f943cc5310a676dc1
- The `xdp` test programs still declared `GPL`; bringing them to `Dual MIT/GPL` is #738.
- Note: the Kconfig commit carries a stray trailing-whitespace change on the file's last line, outside the TC addition.

Measured against the sibling and found conforming (no finding): `int ret; ret = ...` (matches the xdp test programs), `tests/tc/run.sh` (identical to `tests/xdp/run.sh`), the `config.ld` placement, and the `autogen/specs.lua` entry.

lneto:
Two more, and the deep pass is otherwise clean.

- **`test_tc.sh` skips on missing BTF, `bpftool`, and `clang`, but not `tc`** — which it then relies on for `tc qdisc`/`tc filter`, so a host without iproute2's `tc` fails the suite instead of skipping it. The README already documents the skip. https://github.com/luainkernel/lunatik/commit/0039e02f6b1b51b83aba22a7da8286f0a2b9384c
- **`luaskb.h` has no `luaskb_close`** mirroring `luadata.h`'s, so `luatc_release` open-codes `luaskb_clear` + `lunatik_putobject`. Added it and used it, so the release matches `luaxdp`'s one-line `luadata_close`. https://github.com/luainkernel/lunatik/commit/33f9aec324609f8238ca43fe787c7751496490b2

The object lifecycle otherwise checks out: `attach`/`detach`/`release` mirror `luaxdp` (the `getobject` is needed because `release` touches `skb_obj` after a possible detach), the refcounts balance, and re-attach is safe. With the fixups folded and the branch rebased onto `master` (it trails the single-CPU percpu skip), I have no outstanding findings.

@sneaky-potato — the branch has the skb.attr and luatc fixups folded in, but the more recent ones aren't yet: the BPF license (`Dual MIT/GPL`, matching the rest of the tree), the "Traffic Control" spelling, the `tc` skip guard above, and `luaskb_close`. Could you fold those and rebase onto `master`, and give a shout when it's ready so we can merge?

sneaky-potato:
Hi @lneto 
> Note: the Kconfig commit carries a stray trailing-whitespace change on the file's last line, outside the TC addition.

Removed this change from the relevant commit. Pushed the requested fixups, Thanks a lot for this thorough review.

lneto:
@sneaky-potato — re-reviewed the updated branch. The rebase onto `master` and most of the fixups are in (license `Dual MIT/GPL`, the `tc` skip guard, `luaskb_close`, the skb.attr reshape). Three small leftovers:

- `lib/luatc.c` still opens with "Linux Traffic **Controller** (TC)" — should be "Traffic Control" (Linux `tc` is traffic control).
- The `attach` doc still reads "**An** `tc_ctx`" — should be "A `tc_ctx`"; the three "a TC" spots got fixed, this one was missed.
- `luatc_release` now wraps a single statement in braces. A single-statement `if` drops them:

```c
if (lctx->skb_obj)
	luaskb_close(lctx->skb_obj);
```

Ping me when these are in and I'll do a final pass.

lneto:
@sneaky-potato — all three are in and the branch is rebased onto `master`. Built it and ran the suites here: tc 5/5, skb 3/3, nothing left on my side. Looks good to merge. 👍

lneto (approved):
All the review fixups are folded in and the branch is rebased onto `master`. Built it and ran the suites here: tc 5/5, skb 3/3, no findings left. LGTM.

PR #716, opened 2026-08-13
ebpf: centralize common helpers and reuse in xdp

sneaky-potato:
This refactors the eBPF bindings to move common runtime, callback, registration, and teardown logic into the shared `lunatik_ebpf.h` layer.

### DIFF
--- a/lunatik_ebpf.h
+++ b/lunatik_ebpf.h
@@ -8,6 +8,10 @@
 
 #include "lib/luarcu.h"
 
+static char lunatik_ebpf_env_key;
+static lunatik_object_t *lunatik_ebpf_runtimes = NULL;
+static lunatik_object_t *lunatik_ebpf_percpu = NULL;
+
 /* a zero size is allowed by the verifier and would underflow the length */
 static inline int lunatik_ebpf_checkkey(char *key, size_t key_sz, size_t *keylen)
 {
@@ -18,44 +22,43 @@ static inline int lunatik_ebpf_checkkey(char *key, size_t key_sz, size_t *keylen
 	return 0;
 }
 
-static inline int lunatik_ebpf_checkruntimes(lunatik_object_t **runtimes, lunatik_object_t **percpu)
+static inline int lunatik_ebpf_checkruntimes(void)
 {
 	static const char runtimes_key[] = "runtimes";
 	static const char percpu_key[] = "percpu";
-	if (*runtimes == NULL)
-		*runtimes = luarcu_getobject(lunatik_env, runtimes_key, sizeof(runtimes_key) - 1);
-	if (*percpu == NULL)
-		*percpu = luarcu_getobject(lunatik_env, percpu_key, sizeof(percpu_key) - 1);
-	return *runtimes && *percpu ? 0 : -1;
+	if (lunatik_ebpf_runtimes == NULL)
+		lunatik_ebpf_runtimes = luarcu_getobject(lunatik_env, runtimes_key, sizeof(runtimes_key) - 1);
+	if (lunatik_ebpf_percpu == NULL)
+		lunatik_ebpf_percpu = luarcu_getobject(lunatik_env, percpu_key, sizeof(percpu_key) - 1);
+	return lunatik_ebpf_runtimes && lunatik_ebpf_percpu ? 0 : -1;
 }
 
-static inline lunatik_object_t *lunatik_ebpf_lookupruntime(lunatik_object_t **runtimes,
-		lunatik_object_t **percpu, char *key, size_t key_sz, int cpuid)
+static inline lunatik_object_t *lunatik_ebpf_lookupruntime(char *key, size_t key_sz, int cpuid)
 {
 	lunatik_object_t *runtime = NULL;
 	size_t keylen;
 
 	if (lunatik_ebpf_checkkey(key, key_sz, &keylen) != 0)
 		return NULL;
-	if (unlikely(lunatik_ebpf_checkruntimes(runtimes, percpu) != 0)) {
+	if (unlikely(lunatik_ebpf_checkruntimes() != 0)) {
 		pr_err_ratelimited("couldn't find _ENV.runtimes or _ENV.percpu\n");
 		return NULL;
 	}
-	runtime = luarcu_getobject(*runtimes, key, keylen);
+	runtime = luarcu_getobject(lunatik_ebpf_runtimes, key, keylen);
 	if (!runtime) {
 		char cpu_key[LUARCU_MAXKEY];
 		size_t cpulen;
 		cpulen = scnprintf(cpu_key, sizeof(cpu_key), "%s:%d", key, cpuid);
-		runtime = luarcu_getobject(*percpu, cpu_key, cpulen);
+		runtime = luarcu_getobject(lunatik_ebpf_percpu, cpu_key, cpulen);
 	}
 	return runtime;
 }
 
 /* on success the context userdata stays on the stack, ready to be passed to the callback */
-static inline void *lunatik_ebpf_getctx(lua_State *L, char *env_key)
+static inline void *lunatik_ebpf_getctx(lua_State *L)
 {
 	lunatik_object_t *obj;
-	if (lunatik_getregistry(L, env_key) != LUA_TUSERDATA) {
+	if (lunatik_getregistry(L, &lunatik_ebpf_env_key) != LUA_TUSERDATA) {
 		lua_pop(L, 1);
 		pr_err_ratelimited("couldn't find the context object\n");
 		return NULL;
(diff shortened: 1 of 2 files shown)


lunatik_ebpf.h
```diff
@@ -64,6 +64,65 @@ static inline void *lunatik_ebpf_getctx(lua_State *L, char *env_key)
...
+	*callback_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+
+	lunatik_register(L, -1, env_key);
+	lua_pop(L, 1);
+}
+
+/* releases the callback reference taken by lunatik_ebpf_attach(); the ctx
+ * userdata left on the stack by lunatik_ebpf_getctx() is popped here, callers
+ * still need to lunatik_unregister() any sub-objects and 'env_key' itself */
+static inline void lunatik_ebpf_detach(lua_State *L, int *callback_ref)
```
lneto, line 102: The PR summary says it moves the common logic into the shared layer, yet `lunatik_ebpf_attach` and `lunatik_ebpf_detach` are added and never called: `luaxdp_attach`/`luaxdp_detach` still hand-roll the `pushvalue`/`luaL_ref`/`register` and `unref`/`pop` those helpers exist to replace. A helper introduced and left unused in the same change, sitting next to the very duplication it was meant to remove, should have been caught before review — grep the new symbols for a caller before sending. Fixup wires them in: https://github.com/luainkernel/lunatik/commit/739ad5a8

lunatik_ebpf.h
```diff
@@ -93,17 +152,37 @@ static inline void *lunatik_ebpf_getctx(lua_State *L, char *env_key)
 	};
 
 #if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 4, 0))
+#define LUNATIK_EBPF_NEWLIB(name, lib, class) \
```
lneto: The two version branches name the classes array differently (`lua##name##_classes` via `LUNATIK_CLASSES` on >=6.4, `name##_classes` on <6.4). Both compile and are self-consistent, so it is only a cosmetic inconsistency inside the macro — worth aligning, not blocking.

lneto (changes requested):

lunatik_ebpf.h
lneto, line 102: `lunatik_ebpf_attach(L, callback_ref, env_key)` registers the ctx under `env_key`, but `lunatik_ebpf_detach(L, callback_ref)` does not unregister it — the caller has to remember a separate `lunatik_unregister(env_key)`. The pair reads as symmetric and is not: an author who calls `attach(..., env_key)` and `detach(...)` leaks the registration. This is a shared layer the next bindings build on, so its API needs more care than a one-off: a helper pair should mirror, and an argument on one side should appear on the other. Fixup gives `detach` the `env_key` and the unregister so the two mirror; sub-objects stay the caller's. https://github.com/luainkernel/lunatik/commit/7758d0c5 , https://github.com/luainkernel/lunatik/commit/2a89086f

lunatik_ebpf.h
```diff
@@ -93,17 +152,37 @@ static inline void *lunatik_ebpf_getctx(lua_State *L, char *env_key)
...
+	LUNATIK_CLASSES(name, class); \
```
lneto: `LUNATIK_EBPF_NEWLIB` names its first argument `name`, where every sibling macro (`STATE`/`RUN`/`KFUNC_INIT`/`EXIT`) calls the same thing `subsys`; its two version-branches even named the classes array differently (`lua<name>_classes` vs `<name>_classes`). Fixup unifies on `subsys`: https://github.com/luainkernel/lunatik/commit/da7ec916

lneto (changes requested):
A second round, on the shared layer's design. Two more, left for a separate change because they touch symbols already on master from the previous ebpf PR, not this one's: `env_key` is typed `char *` through `getctx`/`attach`/`detach` where it is only an opaque registry-key address (`void *` says so), and `LUNATIK_BTF_SET_START/END` drops the `LUNATIK_EBPF_` prefix the rest carry. And one note without a fixup: `LUNATIK_EBPF_RUN` discards the handler's return (`__ret` is write-only; the verdict travels through `ctx->action`), inherited from xdp and now in the macro.

lunatik_ebpf.h
sneaky-potato: Resolved in latest push

lunatik_ebpf.h
sneaky-potato: I was thinking as `lunatik_<subsys>_env_key` is going to be there in all 3 eBPF bindings, how about we don't expose it through the macros from lunatik_ebpf.h
And design API like this
`lunatik_ebpf_detach(L, &lctx->callback_ref);`

sneaky-potato: Resolved in latest push

lunatik_ebpf.h
sneaky-potato: Resolved in latest push

lunatik_ebpf.h
lneto: Yes, that is the better shape, and it supersedes my symmetry fixup. A single `static char lunatik_ebpf_env_key` in the header has internal linkage, so each binding's `.ko` gets its own instance — a unique registry key per module without the per-subsys name. `attach`, `detach` and `getctx` then drop the argument, and the asymmetry goes with it rather than being patched. Two bindings in one runtime stay distinct, each registered under its own module's address. The same trick takes `runtimes`/`percpu` out of `STATE` if you want to go further; only the symbol-name macros (the kfunc set, the init) still need `subsys`.

lneto (approved):
The earlier fixups are folded, and it builds and passes here — XDP suite 5/5, clean dmesg, no behaviour change. Approving as it stands.

The `env_key` redesign from the thread (a single header static, dropping the argument from `attach`/`detach`/`getctx`) is an improvement, not a blocker — do you want to fold it into this PR before merge, or land this and do it in a new one? Either works. The `LUNATIK_BTF_SET_*` prefix, which touches a symbol already on master, belongs with that follow-up whichever you choose.

sneaky-potato:
Hi @lneto, I have added a fixup for the env_key. Let's take care of it now itself. We can create another PR for the `LUNATIK_BTF_SET_*` prefix,

lunatik_ebpf.h
```diff
@@ -64,6 +68,59 @@ static inline void *lunatik_ebpf_getctx(lua_State *L, char *env_key)
...
+{
+	luaL_unref(L, LUA_REGISTRYINDEX, *callback_ref);
+	*callback_ref = LUA_NOREF;
+	lunatik_unregister(L, &lunatik_ebpf_env_key);
+	lua_pop(L, 1);
+}
+
+/* looks up the runtime for 'key'/'key_sz', runs 'handler' with 'ctxp' on it,
+ * and releases the runtime; a no-op if no matching runtime is found */
+#define LUNATIK_EBPF_RUN(subsys, key, key_sz, handler, ctxp) \
```
lneto: The redesign internalized `env_key`: `getctx`/`attach`/`detach` reference the file-static directly and dropped the parameter. `runtimes`/`percpu` became file-statics too, but stayed threaded through `checkruntimes`/`lookupruntime` as out-params, and `LUNATIK_EBPF_RUN` still takes a `subsys` it no longer uses. Finish it the same way — `checkruntimes`/`lookupruntime` reference the statics, and `LUNATIK_EBPF_RUN(key, key_sz, handler, ctxp)` drops both the plumbing and `subsys`. Fixups: https://github.com/luainkernel/lunatik/commit/ee49ff22 (header), https://github.com/luainkernel/lunatik/commit/59354dcc (the luaxdp caller).

sneaky-potato: Yeah, missed it, thanks.

lneto:
> I have added a fixup for the env_key. Let's take care of it now itself. We can create another PR for the `LUNATIK_BTF_SET_*` prefix

Agreed: env_key here, and the `LUNATIK_BTF_SET_*` prefix in a separate PR — opened #722 to track it. On the env_key fixup itself — it covers the char key, but `runtimes`/`percpu` became file-statics in the same change and are still threaded through `checkruntimes`/`lookupruntime`, with `LUNATIK_EBPF_RUN` carrying a now-unused `subsys`. See the inline comment; fixups https://github.com/luainkernel/lunatik/commit/ee49ff22 and https://github.com/luainkernel/lunatik/commit/59354dcc finish that half too.

lneto (approved):
Internalization is complete: `checkruntimes`/`lookupruntime` reference the file-statics and `LUNATIK_EBPF_RUN` dropped `subsys`. Builds clean here, and the fixups are squashed into the two commits. env_key lands here; `LUNATIK_BTF_SET_*` is tracked in #722.

PR #717, opened 2026-08-13
examples: add sniclassify tc/qos

sneaky-potato:
Add a tc demo script.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,10 +133,12 @@ scripts_uninstall:
 
 ebpf:
 	${MAKE} -C examples/filter
+	${MAKE} -C examples/sniclassify
 
 ebpf_install:
 	${MKDIR} ${LUNATIK_EBPF_INSTALL_PATH}
 	${INSTALL} -m 0644 examples/filter/https.o ${LUNATIK_EBPF_INSTALL_PATH}/
+	${INSTALL} -m 0644 examples/sniclassify/classify.o ${LUNATIK_EBPF_INSTALL_PATH}/
 
 ebpf_uninstall:
 	${RM} -r ${LUNATIK_EBPF_INSTALL_PATH}
--- a/README.md
+++ b/README.md
@@ -459,6 +459,51 @@ ip netns exec tcpreject curl --connect-timeout 2 https://[2001:4860:4860::8888]
 sudo examples/tcpreject/cleanup.sh
 ```
 
+### sniclassify
+
+[sniclassify](examples/sniclassify) is a kernel extension composed by
+a TC/eBPF classifier program attached on egress,
+a Lua kernel script to classify [SNI](https://datatracker.ietf.org/doc/html/rfc3546#section-3.1) traffic.
+This kernel extension extracts server name and assigns traffic
+classes according to a Lua [policy table](examples/sniclassify/sni.lua#18).
+
+Install and load the classifier:
+
+```sh
+sudo make btf_install         # needed to export the 'bpf_luatc_run' kfunc
+sudo make examples_install    # installs examples
+make ebpf                     # builds the TC/eBPF program
+sudo make ebpf_install        # installs the TC/eBPF program
+sudo lunatik run examples/sniclassify/sni softirq percpu
+```
+
+Configure HTB classes:
+```
+sudo tc filter del dev eth0 egress 2>/dev/null
+sudo tc qdisc del dev eth0 clsact 2>/dev/null
+sudo tc qdisc del dev eth0 root 2>/dev/null
+sudo tc qdisc add dev eth0 root handle 1: htb default 20
+sudo tc class add dev eth0 parent 1: classid 1:1 htb rate 100mbit ceil 100mbit
+sudo tc class add dev eth0 parent 1:1 classid 1:10 htb rate 50mbit ceil 100mbit prio 1
+sudo tc class add dev eth0 parent 1:1 classid 1:20 htb rate 30mbit ceil 100mbit prio 2
+sudo tc class add dev eth0 parent 1:1 classid 1:30 htb rate 20mbit ceil 100mbit prio 3
+```
+
+Attach the TC/eBPF classifier on egress:
+```
+sudo tc qdisc add dev eth0 clsact
+sudo tc filter add dev eth0 egress bpf da obj examples/sniclassify/classify.o sec classifier
+```
+
+The classifier inspects outbound TLS ClientHello packets, extracts the SNI
+field, and assigns a traffic class according to the Lua policy table.
+
+Verify and test:
+```
+sudo tc filter show dev eth0
+sudo journalctl -ft kernel
+```
+
 ### gesture
 
 [gesture](examples/gesture.lua)
--- a/README.md
+++ b/README.md
@@ -465,34 +465,25 @@ sudo examples/tcpreject/cleanup.sh
 a TC/eBPF classifier program attached on egress,
 a Lua kernel script to classify [SNI](https://datatracker.ietf.org/doc/html/rfc3546#section-3.1) traffic.
 This kernel extension extracts server name and assigns traffic
-classes according to a Lua [policy table](examples/sniclassify/sni.lua#18).
+classes according to a Lua [policy table](examples/sniclassify/sni.lua#L26).
 
-Install and load the classifier:
+Install the classifier:
 
 ```sh
 sudo make btf_install         # needed to export the 'bpf_luatc_run' kfunc
 sudo make examples_install    # installs examples
 make ebpf                     # builds the TC/eBPF program
 sudo make ebpf_install        # installs the TC/eBPF program
-sudo lunatik run examples/sniclassify/sni softirq percpu
 ```
 
-Configure HTB classes:
-```
-sudo tc filter del dev eth0 egress 2>/dev/null
-sudo tc qdisc del dev eth0 clsact 2>/dev/null
-sudo tc qdisc del dev eth0 root 2>/dev/null
-sudo tc qdisc add dev eth0 root handle 1: htb default 20
-sudo tc class add dev eth0 parent 1: classid 1:1 htb rate 100mbit ceil 100mbit
-sudo tc class add dev eth0 parent 1:1 classid 1:10 htb rate 50mbit ceil 100mbit prio 1
-sudo tc class add dev eth0 parent 1:1 classid 1:20 htb rate 30mbit ceil 100mbit prio 2
-sudo tc class add dev eth0 parent 1:1 classid 1:30 htb rate 20mbit ceil 100mbit prio 3
+Run the classifier and set up the HTB classes on an interface:
+```sh
+sudo ./examples/sniclassify/setup.sh eth0
 ```
 
-Attach the TC/eBPF classifier on egress:
-```
-sudo tc qdisc add dev eth0 clsact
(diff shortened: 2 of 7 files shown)


lneto (changes requested):
This was already raised on #713 and left unaddressed: the SNI parser must not be duplicated. This PR still ships a verbatim second copy of the ClientHello parser from `examples/filter/sni.lua`. Please fold that feedback in rather than carry it forward — the duplication should be resolved before this goes in.

The parser belongs in `examples/common/sni.lua`; PR #721 extracts it and moves `filter/sni.lua` onto it. Rebase this example there and `require("examples.common.sni")`. Fixup 11d49163 shows the sniclassify side, which also drops the `unpacker` closures and flattens the walk.

lneto (approved):
The SNI-parser duplication I flagged is resolved: this now uses `require("examples.common.sni")` (the shared parser merged in #721), and the example reads cleanly. Good to go, on top of #713.

lneto (approved):
Still good after the rebase onto #713 — unchanged, uses the shared `examples.common.sni`.

lneto (changes requested):
@sneaky-potato — re-reviewed against the merged #713.

- **Rebase needed.** The branch still carries the three original (pre-fixup) #713 commits, so it is not mergeable and would reintroduce the un-fixed binding. Rebase onto master so this PR is just the sniclassify commit. (Rebased on review/717 for reference.)

- **flow_cache is never populated.** classify.c declares the map and looks it up, but nothing writes to it — every packet misses and re-runs the Lua classifier, so the cache does nothing. Fixed by caching the classified priority after the run: https://github.com/luainkernel/lunatik/commit/be4f743668098e74feb13fc5a48edab7e4f73100

- **README typo** — "the classfier" -> "classifier": https://github.com/luainkernel/lunatik/commit/f875a3abec05f93e8adff780c53b6dc1d65ebe3a

- **README, double filter** — the tc filter is added twice: the HTB block adds it on `egress`, then the "Attach ... on egress" step adds it again with `parent 1:` (the heading says egress but the command uses `parent 1:`). One is redundant; which attach point does the example want?

sneaky-potato:
> README, double filter — the tc filter is added twice: the HTB block adds it on egress, then the "Attach ... on egress" step adds it again with parent 1: (the heading says egress but the command uses parent 1:). One is redundant; which attach point does the example want?

The second block is the redundant one.
The example wants the clsact/egress attach point. I'll make the necessary changes.

lneto (approved):
@sneaky-potato — all addressed: rebased onto master (one commit), `flow_cache` populated, the typo fixed, and the redundant filter dropped (nice consolidation of the egress attach under its own heading). Built it here — the modules and the sniclassify BPF compile, tc suite 5/5. LGTM.

lneto (changes requested):
@sneaky-potato — re-reviewed at runtime, not just the build: the callback crashes with the merged `skb.attr`.

- `skbattr(ctx:skb())` — `skb.attr` is now a class; the module table isn't callable ("attempt to call a table value"). Use `skbattr.new(ctx:skb())`.
- `skb:data()` on the view — the reshaped `skb.attr` is a pure attribute view (mark/priority) and does not forward methods, so `:data()` raises "skb has no attribute 'data'". Keep the raw skb for methods:

```lua
local raw    = ctx:skb()
local skb    = skbattr.new(raw)
local packet = raw:data()
-- skb.priority = classid  (the view is for the attribute)
```

Both confirmed by running it; my earlier approval was premature — the example has no test of its own, so its callback never ran under the suite.

Please run the example end-to-end before the next push — load it and drive an HTTPS flow through it. With nothing covering it, a broken callback like this lands green; the API it leans on changed under it, and only running it surfaces that.

sneaky-potato:
Hi @lneto, pushed the required changes. I was wondering on the following 2 points:
- Right now README uses `eth0` as the interface name. Also the commands are kind of long, so maybe contain the commands in a shell file which takes the interface name as argument.
- There are no cleanup steps in the README, it takes 3 commands to do the cleanup, we could include them in the clean step in Makefile
```shell
sudo $TC filter del dev $IF egress 2>/dev/null
sudo $TC qdisc del dev $IF clsact 2>/dev/null
sudo $TC qdisc del dev $IF root 2>/dev/null
```
Need suggestions on how to improve on these two points.

lneto (changes requested):
The current asks are in the visible review comment below: https://github.com/luainkernel/lunatik/pull/717#issuecomment-5402967591

lneto:
The two runtime bugs are fixed, thanks.

On your two questions:

1. Interface name and long commands: yes, move the setup into a small shell script that takes the interface as an argument, the way `tcpreject/cleanup.sh` already does. The README then reduces to `sudo ./setup.sh eth0`.

2. Cleanup: same answer, a `cleanup.sh <iface>`, not the Makefile `clean` target. `make clean` removes build artifacts and has no idea which interface the qdisc lives on, so teardown that needs a runtime argument does not belong there. `tcpreject/cleanup.sh` is the precedent.

A closer pass turned up three more points:

- `policy` matches with `host:match("netflix%.com")`. An unanchored Lua pattern is a substring search, not a suffix match. `("evilnetflix.com"):match("netflix%.com")` and `("netflix.com.attacker.io"):match("netflix%.com")` both return `netflix.com`, so both hosts are classified as netflix. The `set` module's labeled flavor (`set.labeled` plus `set:match`) does proper `.`-delimited suffix matching and returns the member's label, which is exactly this domain-to-classid mapping, at `O(d log n)` instead of a scan over every pattern.

- The offsets in `tcp_payload` (`+9`, `+2`/`+3`, `+12`, `& 0x0f`) are inline magic numbers. `examples/common/sni.lua` names its parsing offsets as `<const>`; please follow that.

- `TC_H_MAKE` is upper-case. Lua functions in the tree are lower-case.

One more thing, not on you: the Lua re-parsing the packet the eBPF program already parsed is real overhead, but `bpf_luatc_run` had no argument channel, unlike xdp, so the callback had no other way to receive the offset. #744 adds that channel. Once it lands, the eBPF side can pass the payload offset and `tcp_payload` can go away, the way `examples/filter/sni` reads its offset from `ctx:argument()`.

lneto:
@sneaky-potato the two runtime bugs are fixed, thanks.

> Right now README uses `eth0` as the interface name. Also the commands are kind of long, so maybe contain the commands in a shell file which takes the interface name as argument.

Yes — a small `setup.sh <iface>` (and `cleanup.sh <iface>`) taking the interface as an argument, the way `tcpreject/cleanup.sh` already does. The README then reduces to `sudo ./setup.sh eth0`.

> There are no cleanup steps in the README, it takes 3 commands to do the cleanup, we could include them in the clean step in Makefile

A `cleanup.sh <iface>`, not the Makefile `clean` target — `make clean` removes build artifacts and does not know which interface the qdisc lives on, so teardown that needs a runtime argument does not belong there. `tcpreject/cleanup.sh` is the precedent.

Two changes from a closer pass, each as a fixup on `review/717`:

- **`policy` matches by substring, not suffix.** `host:match("netflix%.com")` is an unanchored Lua pattern, so `("evilnetflix.com"):match("netflix%.com")` and `("netflix.com.attacker.io"):match("netflix%.com")` both match and get classified as netflix. https://github.com/luainkernel/lunatik/commit/6dc3c250603b5b8d85f9daadaadb5dd9f14be046 switches `policy` to the `set` module's labeled flavor: `set.labeled{...}` plus `policy:match(host)` does `.`-delimited suffix matching and returns the member's label (the classid), which rejects those two and is `O(d log n)` instead of a scan over every pattern. The set is built at load, in process context, so the allocation is fine.

- **Magic offsets in `tcp_payload`.** https://github.com/luainkernel/lunatik/commit/30f9603d398aabeca9a240fc837404c40e513dfa names the `+9`, `& 0x0f`, `+2`/`+3` and `+12` as `IP_PROTO`, `IHL_MASK`, `TCP_DPORT`, `TCP_DOFF`, the way `examples/common/sni.lua` names its parsing offsets.

One more thing, not on you: the Lua re-derives the payload offset the eBPF program already computed — a duplicated parse (I have not measured its cost; it is a handful of byte reads, so this is about the redundancy, not a hot path). `bpf_luatc_run` had no argument channel, unlike xdp, so the callback had no other way to receive the offset. #744 adds that channel; once it lands, the eBPF side can pass it and `tcp_payload` can go away, the way `examples/filter/sni` reads its offset from `ctx:argument()`.

lneto:
@sneaky-potato both scripts and the README change, as one extra commit (not a fixup, so take or adapt it): https://github.com/luainkernel/lunatik/commit/3f5c61c6e8252a549a0e1b60180d4e13e13409db

- **setup.sh `<iface>`** runs the classifier, builds the HTB tree (the `default 20` root and the `1:1`/`1:10`/`1:20`/`1:30` classes) and attaches the filter on egress, all parameterized by the interface. It loads the object with `bpftool` and attaches it `object-pinned`, the way the tc test suite does, rather than `tc ... bpf da obj`: on my machine the latter fails with "failed to find valid kernel BTF" (older iproute2 with no vmlinux ELF), and the suite already avoids it for that reason.
- **cleanup.sh `<iface>`** is `lunatik stop` plus the three `tc del` lines and the pin, in the shape of `tcpreject/cleanup.sh`.
- The README's two `tc` blocks then reduce to `sudo ./examples/sniclassify/setup.sh eth0`, torn down with `sudo ./examples/sniclassify/cleanup.sh eth0`.

I ran both end to end on a dummy interface: setup attaches the classifier and starts the runtime cleanly, cleanup tears it all down. I did not drive real TLS through it, so I exercised the setup and teardown, not the SNI classification itself.

lneto:
@sneaky-potato a fuller review pass, all minor:

- The flow cache stored `skb->priority` after every `bpf_luatc_run`, including when it returned `< 0` (runtime absent), pinning that flow to the unclassified priority. https://github.com/luainkernel/lunatik/commit/5b9ccf95336c8969bce9153a67c2f671d34f2fe6 caches only when the call succeeded.
- `bpf_luatc_run` is declared with three args here; #744 adds `(arg, arg_sz)`. That change stays out of this PR — the extern is correct against `master` today — but when both land it needs `..., skb, NULL, 0`. Flagging so it is not missed at merge.
- The cache is keyed by `skb->hash`. Locally-generated TCP carries `sk_txhash`, so it is per-flow in practice; a zero hash or a collision would share an entry. Worth a note on the assumption if you want it explicit.
- The HTB rates print `sch_htb: quantum ... is big` on setup (cosmetic; an explicit `quantum` or a higher `r2q` silences it), and the example has no test of its own.

I ran it end to end: a real ClientHello with SNI `netflix.com` lands in class `1:30`.

sneaky-potato:
> @sneaky-potato both scripts and the README change, as one extra commit (not a fixup, so take or adapt it): https://github.com/luainkernel/lunatik/commit/3f5c61c6e8252a549a0e1b60180d4e13e13409db

Verified, the setup and cleanup scripts look good!

lneto (approved):
@sneaky-potato all the review points are folded, and I ran it end to end: a real ClientHello with SNI `netflix.com` classifies into `1:30`, across CPUs, with no errors on a clean run. Approving.

One coordination note for the merge, not a change here: `classify.c` declares `bpf_luatc_run` with three args, correct against `master` today; when #744 lands it takes `(arg, arg_sz)`, so this extern and the call need `..., skb, NULL, 0` at that point.

PR #721, opened 2026-08-15
examples: share the TLS SNI parser between the filtering examples

lneto:
The `filter` (xdp) and `sniclassify` (tc, #717) examples both parse the SNI from a TLS ClientHello with the same code. Extract that parser to `examples/common/sni.lua` (`require("examples.common.sni")`, taking `(packet, base)` and returning the host name) and move `filter/sni.lua` onto it. #717's sniclassify then requires the same module instead of carrying a second copy.

Test plan:
- Loaded both examples in softirq — they attach cleanly, no dmesg errors.
- Verified the extracted parser is equivalent to the original: a crafted ClientHello fed through both the old closure parser and the new module returns the same SNI, and both return nil on a non-ClientHello.


sneaky-potato (approved):
This looks good, approved from my side.

Issue #723, opened 2026-08-16
Epic: kernel TLS binding — kTLS and a TLS tunnel in Lua


lneto:
Lunatik can open kernel sockets but cannot speak TLS. This epic tracks a kernel TLS (kTLS) binding so
Lua scripts can key a connected socket for TLS, read and write plaintext while the kernel does the
record-layer crypto, delegate the handshake to the userspace `tlshd` daemon through the kernel
handshake upcall, and splice plaintext between sockets — enough to build a TLS proxy or an in-kernel
TLS tunnel in Lua.

## The fact that decides the design

The kernel implements the TLS **record** layer, not the **handshake**:

> There is no TLS handshake implementation in the Linux kernel, by design. Every honest "in-kernel
> TLS" — nginx `SSL_sendfile`, HAProxy, Cilium's Envoy-based visibility, `tlshd` — keeps the
> handshake, certificate validation and policy in userspace and only kernelizes record crypto and byte
> movement.

So a "TLS tunnel in Lua" is not a kernel TLS stack; it is a **plaintext-splicing proxy over
already-keyed kTLS sockets**, with the handshake sourced from userspace. Design against that, or you
end up forking `net/tls` like the 2020 GSoC project did.

## Incremental, and half the foundation already exists

This is built incrementally, and the early phases rebase parked work rather than starting from zero.
A `claude_tls` branch (unmerged, ~84 commits behind) already has `socket:setsockopt`, a `tls`
crypto-info packer, and a handshake-upcall binding — but not on `master` (`master`'s `luasocket`
has no `setsockopt` at all), and without the data path. The later phases add plaintext I/O, the
handshake plumbing, and the tunnel. Each phase is a shippable pull request.

kTLS is a TCP ULP; a socket carries exactly one ULP, and we ride the kernel's `tls` ULP. Writing a
*Lua* ULP (Lua as an upper-layer protocol for arbitrary L7, the successor to Pedro Tammela's 2019
`ulp-lua`) is a worthwhile **separate** project, not this one — it cannot share a socket with `tls`.

## Documents

Everything a contributor needs is in the repository, under `doc/design/ktls/`:

- [`plan.md`](https://github.com/luainkernel/lunatik/blob/master/doc/design/ktls/plan.md), current state (with the prior work), gap analysis, the incremental phases, non goals, risks
[...]

PR #730, opened 2026-08-16
docs: add the kernel TLS design notes

lneto:
Design notes for the kernel TLS (kTLS) epic (#723), in the same shape as the other
`doc/design/<theme>/` sets: plan, proposed API, verified kernel reference, test strategy.

`doc/design/ktls/` covers a kTLS binding so Lua scripts can key a connected socket for TLS, read and
write plaintext while the kernel does the record-layer crypto, delegate the handshake to `tlshd`
through the kernel handshake upcall, and splice plaintext between sockets — enough for a TLS proxy or
an in-kernel TLS tunnel in Lua.

The design turns on one fact, stated up front so a contributor does not rediscover it the hard way:

> The Linux kernel implements the TLS record layer but not the handshake. Every honest "in-kernel TLS"
> — nginx `SSL_sendfile`, HAProxy, Cilium's Envoy-based visibility, `tlshd` — keeps the handshake,
> certificate validation and policy in userspace and only kernelizes record crypto and byte movement.
> So a "TLS tunnel in Lua" is a plaintext-splicing proxy over already-keyed kTLS sockets, not a kernel
> TLS stack — design against that, or you fork `net/tls` like the 2020 GSoC project did.

It is built incrementally, and half the foundation already exists on a parked `claude_tls` branch
(`socket:setsockopt`, a `tls` crypto-info packer, a handshake-upcall binding — but not on `master`,
and without the data path). The early phases rebase that; the later phases add plaintext I/O with
control records, the handshake plumbing, and the tunnel. kTLS is a TCP ULP and a socket carries one
ULP, so we ride the kernel's `tls` ULP; a Lua ULP (the successor to Pedro Tammela's `ulp-lua`) is a
separate project, kept as a non goal.

The `kernel-notes.md` facts were verified against Linux 6.8 source and the running kernel's exported
symbols: the two-step setsockopt keying, the module-callable handshake upcall (`tls_client_hello_*`,
all plain `EXPORT_SYMBOL`), the `-EIO`-without-a-control-buffer gotcha on the plaintext read path, and
the version drift (TLS 1.3 KeyUpdate lands in 6.14, absent in 6.8).

### DIFF
--- a/doc/design/ktls/README.md
+++ b/doc/design/ktls/README.md
@@ -1,0 +1,43 @@
+# Kernel TLS binding
+
+Working documents for the `ktls` binding: a Lua interface to Linux kernel TLS, so kernel scripts can
+key a connected socket for TLS, read and write plaintext while the kernel does the record-layer
+crypto, delegate the handshake to the userspace `tlshd` daemon through the kernel handshake upcall,
+and splice plaintext between sockets — enough to build a TLS proxy or an in-kernel TLS tunnel in Lua.
+
+They exist so that a new contributor, with or without an AI coding assistant, can pick the work up
+without first learning where kTLS ends and the TLS handshake begins.
+
+| Document | What it is for |
+|----------|----------------|
+| [plan.md](plan.md) | Current state (including the prior work), gap analysis, the incremental phases, non goals, risks, definition of done |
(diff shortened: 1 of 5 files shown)


lneto:

PR #731, opened 2026-08-16
eBPF: clean up the shared binding header

### LINKED ISSUE #722: eBPF: align the LUNATIK_BTF_SET_* macro prefix with the LUNATIK_EBPF_ family
In `lunatik_ebpf.h` every macro carries the `LUNATIK_EBPF_` prefix (`LUNATIK_EBPF_RUN`, `LUNATIK_EBPF_START`/`_END`, `LUNATIK_EBPF_KFUNC_DEFINE_SET`, `LUNATIK_EBPF_NEWLIB`, `LUNATIK_EBPF_KFUNC_INIT`, `LUNATIK_EBPF_EXIT`) — except `LUNATIK_BTF_SET_START`/`LUNATIK_BTF_SET_END`, which use `LUNATIK_BTF_`.

Rename them to `LUNATIK_EBPF_BTF_SET_START`/`_END` for a consistent prefix and update the one use inside `LUNATIK_EBPF_KFUNC_DEFINE_SET`.

Deferred out of #716 to keep that PR to the env_key change (see https://github.com/luainkernel/lunatik/pull/716#issuecomment-5303101466).

lneto:
Cleanup of `lunatik_ebpf.h` after #716.

- Pass the callback's stack index to `attach` (named `ix`, the base's convention) instead of hardcoding argument 1 in the shared helper.
- Drop the redundant callback type check in `invoke`: the callback is validated as a function at attach, so it is always one here, and `lua_pcall` would catch a non-callable anyway.
- Name the callback reference `cb`, matching the base (luarcu, luahid); `callback_ref` was the only `_ref` name in the tree.
- Simplify the `checkruntimes` result to `!(a && b)` (the caller only tests against zero).
- Drop the block comments the base keeps off internal `static inline` helpers — lunatik.h carries none — keeping only `checkkey`'s.
- Initialize `cpulen` at declaration (C99).
- Prefix the `BTF_SET` macros with `LUNATIK_EBPF_`, aligning them with the family. Closes #722.

Test plan:
- `make` builds `luaxdp.ko` clean against the changed header.

### DIFF
--- a/lunatik_ebpf.h
+++ b/lunatik_ebpf.h
@@ -65,9 +65,9 @@ static inline void *lunatik_ebpf_getctx(lua_State *L)
 	return obj->private;
 }
 
-static inline int lunatik_ebpf_invoke(lua_State *L, int callback_ref)
+static inline int lunatik_ebpf_invoke(lua_State *L, int cb)
 {
-	lua_rawgeti(L, LUA_REGISTRYINDEX, callback_ref);
+	lua_rawgeti(L, LUA_REGISTRYINDEX, cb);
 	lua_insert(L, -2);
 	if (lua_pcall(L, 1, 0, 0) != LUA_OK) {
 		pr_err_ratelimited("%s\n", lua_tostring(L, -1));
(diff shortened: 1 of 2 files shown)


sneaky-potato (approved):
Looks good, especially the `stack index to attach` change. Thanks a lot for this!

lunatik_ebpf.h
```diff
@@ -69,12 +69,6 @@ static inline void *lunatik_ebpf_getctx(lua_State *L)
 static inline int lunatik_ebpf_invoke(lua_State *L, int callback_ref)
 {
 	lua_rawgeti(L, LUA_REGISTRYINDEX, callback_ref);
-	if (!lua_isfunction(L, -1)) {
-		pr_err_ratelimited("callback_ref is not a function\n");
-		lua_pop(L, 2);
-		return -1;
-	}
```
sneaky-potato, line 76: Verified, the attach fails if the callback is not a function, this is a nice improvement. Thanks!

sneaky-potato (approved):

PR #734, opened 2026-08-18
tests/tc: cover re-attaching a callback in the same runtime

lneto:
New coverage stacked on #713: a tc re-attach case. The script attaches one callback, then a second in the same runtime; the ping passes because only the last callback runs. xdp has no equivalent test.

Note: on #713's current code re-attaching leaks the previous binding (see the review comment there). This test exercises the re-attach path and the replace semantics; it passes both before and after that fix, since the leak is not functionally observable (the env-key registration is simply overwritten).

Test plan:
- `sudo lunatik test tc` -> 6/6, the new `tc reattach` case passes.

### DIFF
--- a/tests/tc/Makefile
+++ b/tests/tc/Makefile
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
 # SPDX-License-Identifier: MIT OR GPL-2.0-only
 
-all: tc_pass.bpf.o tc_drop.bpf.o tc_detach.bpf.o tc_zerokey.bpf.o
+all: tc_pass.bpf.o tc_drop.bpf.o tc_detach.bpf.o tc_zerokey.bpf.o tc_reattach.bpf.o
 
 vmlinux.h:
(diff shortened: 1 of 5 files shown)


tests/tc/Makefile
```diff
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
 # SPDX-License-Identifier: MIT OR GPL-2.0-only
 
-all: tc_pass.bpf.o tc_drop.bpf.o tc_detach.bpf.o tc_zerokey.bpf.o
+all: tc_pass.bpf.o tc_drop.bpf.o tc_detach.bpf.o tc_zerokey.bpf.o tc_reattach.bpf.o
```
sneaky-potato, line 4: Do we have a plan of adding an equivalent test for xdp?

sneaky-potato (approved):
This looks good.

tests/tc/Makefile
lneto: hm.. didn't think about that.. can you please create an issue for XDP?

PR #737, opened 2026-08-20
lunatik: unload every loaded module on reload, not just the configured set

lneto:
A module left loaded from another build — a different config, or a feature branch — can pin the core and wedge the next `reload` when the configured set no longer lists it.

`unload` now unloads every module holding `lunatik.ko` (from `/sys/module/lunatik/holders`) and `lunatik` itself, peeling by dependency: each round `rmmod`s the modules the kernel reports ready (`refcnt 0`) and counts what actually came off, repeating until a round removes nothing. `refcnt` counts both symbol users and the runtime's `require`/`module_get` pins — which the holders graph does not show — so it is the complete readiness signal: no order is hardcoded, and a module that cannot be removed ends the loop instead of spinning it.

So a leftover the current config no longer lists comes off with the rest (a renamed or disabled module, one only a feature branch builds), and nothing that does not hold lunatik is touched. Discovery is a directory listing plus `refcnt`, not a `/proc/modules` parse, and needs no Makefile/autogen change.

**Test:** `lunatik unload` brings the loaded set to zero with no rmmod errors; with `luadarken` loaded but removed from `config.modules`, `reload` still unloads it (the configured-set unload left it as a leftover).

### DIFF
--- a/bin/lunatik
+++ b/bin/lunatik
@@ -25,7 +25,7 @@ end
 
 local function sh(command, module, loaded)
 	local op = loaded and "&&" or "||"
-	os.execute(string.format("MODULE=%s;grep -wq $MODULE /proc/modules %s%s", module, op, command))
+	return os.execute(string.format("MODULE=%s;grep -wq $MODULE /proc/modules %s%s", module, op, command))
 end
 
 local function load_modules()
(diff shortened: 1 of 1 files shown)


bin/lunatik
```diff
@@ -43,12 +43,33 @@ local function load_modules()
 	if result ~= "" then error(result) end
 end
 
+local function loaded_modules()
+	local names = {}
+	for line in io.lines("/proc/modules") do
+		local name = line:match("^%S+")
+		if name:match("^lua") or name:match("^lunatik") then
```
lneto: We can remove legitm drivers if we match "^lua", I think

lneto: It's probably better to use autogen to generate the available module list

PR #741, opened 2026-08-23
Fix ranged linux.random using unsigned modulo

### LINKED ISSUE #740: linux.random(m, n) can return values outside [m, n] (passwd example breaks)
## Environment
Ubuntu 24.04 Server in VirtualBox
kernel 6.8.0-137-generic
1 vCPU, Lunatik built from current master

## What I did
I ran the README example that reads random characters from /dev/passwd using linux.random(32, 126).

## What happened
```
head: error reading '/dev/passwd': Operation canceled
[   95.714212] luadevice: /lib/modules/lua/passwd.lua:8: bad argument #1 to 'char' (value out of range): read
```
## Reproduction
```bash
sudo tee /lib/modules/lua/passwd.lua >/dev/null <<'EOF'
local device = require("device")
local stat = require("linux.stat")
local driver = {name = "passwd", mode = stat.IRUGO}
function driver:read()
return string.char(linux.random(32, 126))
end
device.new(driver)
EOF
sudo lunatik stop passwd 2>/dev/null
sudo dmesg -C
sudo lunatik run passwd
sudo head -c 16 /dev/passwd; echo
sudo dmesg | tail -20
sudo lunatik stop passwd
```

Related: #374

Harshdeep-creator:
Fixes #740.

linux.random(m, n) returned values outside the [m, n] range because get_random_u64() was being used with signed modulo. This also caused the string.char call in the passwd example to fail.
This change uses unsigned u64 values for the span and offset calculations, ensuring the result stays within the requested range.
It also adds a regression test under tests/linux/random and documents the test suite in tests/README.md.
I verified this with sudo lunatik test linux and passwd smoke test.

### DIFF
--- a/lib/lualinux.c
+++ b/lib/lualinux.c
@@ -67,7 +67,7 @@ static int lualinux_random(lua_State *L)
 	luaL_argcheck(L, low <= up, 1, "interval is empty");
 	luaL_argcheck(L, low >= 0 || up <= LUA_MAXINTEGER + low, 1, "interval too large");
 
-	rand = low + ((lua_Integer)get_random_u64()) % (up - low + 1);
+	rand = low + (lua_Integer)(get_random_u64() % ((lua_Unsigned)up - (lua_Unsigned)low + 1));
 	lua_pushinteger(L, rand);
 	return 1;
 }
--- a/tests/README.md
+++ b/tests/README.md
@@ -75,6 +75,11 @@ Covers the `crypto` module: `shash`, `skcipher`, `aead`, `rng`, `hkdf`,
 - **test**: kernel `io` library (open/read/write/seek/lines/type); also
   asserts `io` is absent from softirq runtimes.
 
+### linux
+
+- **random**: `linux.random` ranged draws stay within `[m, n]`, covering
+  the two-argument, one-argument and negative-range forms.
+
 ### monitor
 
 Regression tests for `lunatik_monitor` (spinlock + GC interaction).
--- a/tests/linux/random.lua
+++ b/tests/linux/random.lua
@@ -1,0 +1,32 @@
+--
+-- SPDX-FileCopyrightText: (c) 2026 Harshdeep Singh
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+-- Kernel-side script for the linux.random test (see run.sh).
+--
+
+local linux = require("linux")
+local test = require("util").test
+
+test("linux.random(m, n) stays within [m, n]", function()
+	for i = 1, 10000 do
+		local r = linux.random(32, 126)
+		assert(type(r) == "number", "expected number")
+		assert(r >= 32 and r <= 126, "out of range: " .. r)
+	end
+end)
+
+test("linux.random(n) stays within [1, n]", function()
+	for i = 1, 1000 do
+		local r = linux.random(1000)
+		assert(r >= 1 and r <= 1000, "out of range: " .. r)
+	end
+end)
+
+test("linux.random(m, n) stays within a negative range", function()
+	for i = 1, 1000 do
+		local r = linux.random(-126, -32)
+		assert(r >= -126 and r <= -32, "out of range: " .. r)
+	end
+end)
+
--- a/tests/linux/run.sh
+++ b/tests/linux/run.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+#
+# SPDX-FileCopyrightText: (c) 2026 Harshdeep Singh
+# SPDX-License-Identifier: MIT OR GPL-2.0-only
+#
+# Runs all linux tests and reports aggregated KTAP results.
+#
+# Usage: sudo bash tests/linux/run.sh
+
+DIR="$(dirname "$(readlink -f "$0")")"
+
+source "$DIR/../lib.sh"
+
+TESTS="random"
+TOTAL=$(echo $TESTS | wc -w)
+
+ktap_header
+ktap_plan $TOTAL
+
+for t in $TESTS; do
+	if run_test "tests/linux/$t"; then
+		ktap_pass "linux/$t"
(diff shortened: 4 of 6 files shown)


lib/lualinux.c
```diff
@@ -67,7 +67,11 @@ static int lualinux_random(lua_State *L)
 	luaL_argcheck(L, low <= up, 1, "interval is empty");
 	luaL_argcheck(L, low >= 0 || up <= LUA_MAXINTEGER + low, 1, "interval too large");
 
-	rand = low + ((lua_Integer)get_random_u64()) % (up - low + 1);
+	{
+		u64 span = (u64)(up - low + 1);
+		u64 offset = get_random_u64() % span;
+		rand = low + (lua_Integer)offset;
+	}
```
lneto: The fix is correct, but the inner block with two temporaries is foreign to this file. `math_random()` in `lua/lmathlib.c` — which the comment above this function cites — does the subtraction in unsigned arithmetic inline. The same form here is one line, and it also removes the signed overflow `up - low + 1` still has for `random(0, math.maxinteger)`, which today is defined only because the kernel builds with `-fno-strict-overflow`: https://github.com/luainkernel/lunatik/commit/5a127dc2645a688237c0c3e087dd2d1324846b55

tests/linux/run.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+#
+# SPDX-FileCopyrightText: (c) 2026 Harshdeep Singh
+# SPDX-License-Identifier: MIT OR GPL-2.0-only
+#
+# Runs linux module regression tests and reports aggregated KTAP results.
+#
+# Usage: sudo bash tests/linux/run.sh
+DIR="$(dirname "$(readlink -f "$0")")"
```
lneto, line 10: This is `tests/crypto/run.sh` with the blank lines dropped; files that mirror a sibling stay diffable against it. The same fixup restores the header-block format in `random.lua` (see `tests/monitor/gc.lua`) and the blank line at EOF that the tree keeps in every file: https://github.com/luainkernel/lunatik/commit/2d3af55999ac8bba5b69c65f93b96605b61e01a6

tests/linux/random.lua
```diff
@@ -0,0 +1,26 @@
...
+		local r = linux.random(1000)
+		assert(r >= 1 and r <= 1000, "out of range: " .. r)
+	end
+end)
+test("linux.random(m, n) fits string.char for ASCII printable", function()
+	for i = 1, 1000 do
+		local c = string.char(linux.random(32, 126))
+		assert(#c == 1, "expected one character")
+	end
+end)
```
lneto: The `string.char` test is implied by the first one: if every draw stays in `[32, 126]`, `string.char` cannot fail. https://github.com/luainkernel/lunatik/commit/4df00cd2210aabef7192523019de1e6d0ff927c6 drops it.

tests/linux/random.lua
```diff
@@ -0,0 +1,26 @@
...
+		assert(type(r) == "number", "expected number")
+		assert(r >= 32 and r <= 126, "out of range: " .. r)
+	end
+end)
+test("linux.random(n) stays within [1, n]", function()
+	for i = 1, 1000 do
+		local r = linux.random(1000)
+		assert(r >= 1 and r <= 1000, "out of range: " .. r)
+	end
+end)
```
lneto, line 24: A negative lower bound is the other case the signed modulo broke, and nothing covers it: https://github.com/luainkernel/lunatik/commit/252dc145301ea02c96545ac0ca56b6c1d6e60049

tests/README.md
```diff
@@ -75,6 +75,12 @@ Covers the `crypto` module: `shash`, `skcipher`, `aead`, `rng`, `hkdf`,
 - **test**: kernel `io` library (open/read/write/seek/lines/type); also
   asserts `io` is absent from softirq runtimes.
 
+
+### linux
+
+- **random**: `linux.random` ranged draws stay within `[m, n]` and fit
+  `string.char` for the README passwd ASCII range.
```
lneto: Double blank line before the section, and the entry describes the `string.char` test removed above. https://github.com/luainkernel/lunatik/commit/9597ce73b914613dcad0d48419e0d488c40b5410 rewrites it for the final test set.

lneto (changes requested):
The fix itself is correct: the old expression leaves `[m, n]` on every draw with the top bit set, and the new arithmetic holds across the edges, `random(0, math.maxinteger)` included. Built and ran the full suite on top of these changes: pass:110 fail:0 skip:0. The inline comments and the two notes below must be folded before merge; `review/741` carries one fixup per finding.

A new suite is also listed in the top-level README (AGENTS.md: "A test is not done until `tests/README.md` describes it, the suite's `run.sh` runs it, and, for a new suite, the top level `README.md` lists it"): https://github.com/luainkernel/lunatik/commit/d43e5fa9793bef4a578d4504454a1248c2598b7f

Test documentation lands in the same commit as the test it documents (AGENTS.md, same rule as above): fold "Document linux test suite in tests/README.md" into the first commit when applying the fixups. The fixups on `review/741` target the commit each one amends, so `git rebase -i --autosquash` folds everything in place.

lneto:
82ba4a98 folds everything but one finding, from the review body: the top-level `README.md` suite list still misses `linux` (AGENTS.md: for a new suite, the top level `README.md` lists it). The fixup applies cleanly as-is on this head: https://github.com/luainkernel/lunatik/commit/d43e5fa9793bef4a578d4504454a1248c2598b7f

The rest of the tree is identical to `review/741`, which built and ran the full suite here: pass:110 fail:0 skip:0. Fold that last one and this merges.

lneto (approved):
9ee30e0a matches `review/741` exactly; every finding is folded. Built and ran the full suite on this tree: pass:110 fail:0 skip:0.

PR #742, opened 2026-08-23
lunatikc: host bytecode compiler for kernel Lua scripts

lneto:
Adds `lunatikc`, a userspace compiler that turns kernel Lua scripts into binary chunks the kernel accepts, plus the wiring to ship scripts as bytecode.

The kernel already loads binary chunks everywhere (`lunatik_loadfile` runs with mode `"bt"`, `require` included); what was missing is a way to produce one. The distribution `luac` cannot: under `_KERNEL` the fork drops the float opcodes (renumbering everything after `OP_LOADI`), compiles `/` to integer division, and checks an integer `lua_Number` in the chunk header. So `lunatikc` is built with the host compiler from the same `lua/` sources and `_KERNEL` configuration as `lunatik.ko`, which makes its output match the kernel by construction.

* `lunatik_conf: fence the kernel-only part of the configuration`: the number configuration and limits stay shared with the host build; the kernel-only material goes under `__KERNEL__`. No change for the kernel build (same `lunatik.ko`).
* `lunatikc: add the host bytecode compiler`: `lunatikc [-s] [-o output] [-n chunkname] input.lua ...`, parsing text only and dumping with optional strip; built as part of `make`, installed next to `lunatik`. `BYTECODE=1 make install` ships the kernel Lua libraries and examples as stripped chunks under their usual `.lua` names (the loader detects the signature, so nothing changes on the kernel side); the full suite passes with source and with bytecode installs, `driver.lua` included. The kernel libraries go from 12.8 KB of source to 4.0 KB of stripped chunks.
* `runner: trim only a trailing .lua`: found by the new suite; `trim()` mangled any script path containing a name, one character and `lua` (`tests/luac/x` became `testsc/x`).
* `tests: add the luac suite`: compiled chunks run via `run` and `require`, full and stripped; error messages carry the `-n` chunk name, or `?:?:` when stripped (now recognized by `check_dmesg`); a stock number format is rejected by the header; `load(..., "t")` refuses a chunk in the kernel.
* `docs: add the bytecode compiler design notes`: `doc/design/luac/` with the format analysis (what is byte-order dependent in a 5.5 chunk), bytecode trust notes, precedents, and the plan for the follow-ups (cross-endian, parser-less kernel, fixed-buffer loading).

Cross-compilation to a different byte order is the next step, on top of this branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,25 @@ MKDIR = mkdir -p -m 0755
 LN = ln -sf
 INSTALL = install -o root -g root
 
+# host compiler for kernel Lua scripts; same lua/ sources and _KERNEL configuration as lunatik.ko
+HOSTCC ?= cc
+LUNATIKC := bin/lunatikc
+LUNATIKC_CORE := lapi lcode lctype ldebug ldo ldump lfunc lgc llex lmem lobject lopcodes \
+	lparser lstate lstring ltable ltm lundump lvm lzio lauxlib
+LUNATIKC_SRCS := ${LUNATIKC}.c $(addprefix lua/,$(addsuffix .c,$(LUNATIKC_CORE)))
+LUNATIKC_CFLAGS := -std=gnu99 -O2 -Wall -D_KERNEL -DLUA_USE_LINUX -I. -Ilua
+
+# BYTECODE=1 installs kernel Lua scripts as stripped chunks, under their .lua names
+ifeq ($(BYTECODE),1)
+define INSTALL_LUA
+	for f in $(1); do ${LUNATIKC} -s -o $(2)/$$(basename $$f) $$f || exit 1; done
+endef
+else
+define INSTALL_LUA
+	${INSTALL} -m 0644 $(1) $(2)
+endef
+endif
+
 CONFIG_LUNATIK ?= m
 CONFIG_LUNATIK_RUNTIME ?= y
 CONFIG_LUNATIK_RUN ?= m
@@ -62,14 +81,17 @@ AUTOGEN_KEY := $(KERNEL_RELEASE)|$(LUNATIK_MODULES)
 	tests_install tests_uninstall \
 	ebpf ebpf_install ebpf_uninstall
 
-all: lunatik_sym.h autogen
+all: lunatik_sym.h autogen ${LUNATIKC}
 	${MAKE} -C ${MODULES_BUILD_PATH} M=${PWD} $(LUNATIK_CONFIG_FLAGS)
+
+${LUNATIKC}: ${LUNATIKC_SRCS} lunatik_conf.h
+	${HOSTCC} ${LUNATIKC_CFLAGS} -o $@ ${LUNATIKC_SRCS} -lm
 
 clean:
 	${MAKE} -C ${MODULES_BUILD_PATH} M=${PWD} clean
 	${MAKE} -C ${MODULES_BUILD_PATH} M=${PWD}/autogen clean
 	${MAKE} -C examples/filter clean
-	${RM} lunatik_sym.h
+	${RM} lunatik_sym.h ${LUNATIKC}
 	${RM} -r lib/linux
 	${RM} autogen/lunatik/*.lua autogen/linux/*.lua \
 		autogen/dump_*.c autogen/dump_*.pp \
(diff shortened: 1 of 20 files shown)


lneto (commented):

@lneto request changes: one defect in the driver, the rest shape and harness. A fixup for each finding is on `review/742` (https://github.com/luainkernel/lunatik/tree/review/742), `--fixup` onto the commit it belongs to; each is linked from its line.

What ran, on 6.12.88 x86_64: the head builds with one new warning, `lua/llex.c:250: 'expo' set but not used`, from the `-Wall` host build (kbuild hides it with `-Wno-unused-but-set-variable`); the fix is a fork change, luainkernel/lua#9, to take as a submodule bump. The suite passed 117/117 with a source install and 117/117 with `BYTECODE=1`; with the fixups, 121/121 both ways. The examples `examples-touched.sh` names for `lib/lunatik/runner.lua` ran through `tools/watchdog.sh`: systrack fired and both runtimes are listed under their names; spyglass loaded and armed (no keyboard on the host); ifquarantine fired on the existing interfaces and took the host off the network, the watchdog stopped it and connectivity returned, which is this base's example and not the trim change.

Against #743: built from its fork branch, its `luac` writes the same 264 chunks byte for byte. What it adds is `-l` with the kernel's opcodes; what it costs is four commits in the fork and a third build flavour of `lauxlib` named after this tree's tool, re-merged on every upstream `luac.c` change. This shape is the one to keep, as phase 0 records (#874).

Not folded, on purpose: `int i;` in `main` (the loop head sets it), the `LUNATIKC_CORE` list beside `Kbuild`'s (sharing it costs a third file both build systems include), and the strip default of `BYTECODE=1`, which is a contract; the README fixup names its cost.

Before merging: squash the twelve fixups; rebase (`Makefile`, `README.md`, `tests/README.md`, `tests/lib.sh` and `tests/run.sh` conflict; `claude_luaebpf_host` carries the same content rebased and merges clean); the body through `tools/checks/pr-body.sh`; and `runner: trim only a trailing .lua` as its own pull request first, since master still has the bug and the luaebpf suites need it (`tests/luaebpf/x` trims to `testsebpf/x`).

bin/lunatikc.c
```diff
@@ -0,0 +1,166 @@
...
+	if (n > 4 && strcmp(base + n - 4, ".lua") == 0)
+		n -= 4;
+
+	const char *dir = output != NULL ? output : "";
+	char *name = malloc(strlen(dir) + 1 + n + sizeof(LUNATIKC_EXT));
+	if (name == NULL)
+		fail(strerror(ENOMEM));
+
+	if (output == NULL) /* next to the input */
+		sprintf(name, "%.*s%.*s" LUNATIKC_EXT, (int)(base - input), input, (int)n, base);
```
lneto, line 89: (posted by an agent, not by @lneto)

`outname` sizes `name` with `strlen(dir)` where `dir` is `""` when there is no `-o`, then writes the input's directory prefix into it. `lunatikc examples/echod/daemon.lua`, the first example in `api.md`, is a 27-byte write into a 13-byte allocation under ASan; on a path the length of `/lib/modules/lua/tests/luac/`, glibc aborts with `malloc(): corrupted top size`. The suite never reaches this path, since every case passes `-o`. https://github.com/luainkernel/lunatik/commit/23e6a37f8e6ed8e27ddf5f7d0443b15d05c84604 sizes the buffer from the same parts the `snprintf` writes; https://github.com/luainkernel/lunatik/commit/dc71c4d5b98df79c89dbaab66d3f0417308b8fed adds the case, which aborts on this head and passes with the fix.

lunatik_conf.h
```diff
@@ -6,30 +6,37 @@
...
 
-#include <linux/random.h>
-#define luai_makeseed(L)		get_random_u32()
-
 #undef lua_getlocaledecpoint
 #define lua_getlocaledecpoint()		('.')
 
+/* frame size shouldn't be larger than 1024 bytes; thus, LUAL_BUFFERSIZE
+ * must be adjusted for the stack of functions that use luaL_Buffer */
+#undef LUAL_BUFFERSIZE
```
lneto, line 21: (posted by an agent, not by @lneto)

The comment above says the shared region is everything that shapes bytecode, but `LUAL_BUFFERSIZE`, `LUAI_MAXSTACK` and `LUNATIK_GCCOUNT` never reach a chunk: `LUAI_MAXSTACK` is read in `ldo.c` only, `LUAL_BUFFERSIZE` by `lauxlib.c`'s buffers, `LUNATIK_GCCOUNT` by `lbaselib.c`, which the tool does not link. Every `.lua` under `lib/`, `examples/` and `tests/`, with and without `-s` (264 chunks), compiles byte for byte the same with a header that only fences from `<linux/random.h>` down and moves nothing; and `LUAI_MAXSTACK 200` on the host is a ceiling the state that will host the translator (#875) would inherit. https://github.com/luainkernel/lunatik/commit/daaddc5ff985da0e97d1c293e279cb9ee1d3f77a is that header: three lines against the base instead of 16+/13-, and `lua_getlocaledecpoint` stays kernel-only, as `kernel-notes.md:159` says it must.

Makefile
```diff
@@ -26,6 +26,25 @@ MKDIR = mkdir -p -m 0755
 LN = ln -sf
 INSTALL = install -o root -g root
 
+# host compiler for kernel Lua scripts; same lua/ sources and _KERNEL configuration as lunatik.ko
+HOSTCC ?= cc
+LUNATIKC := bin/lunatikc
+LUNATIKC_CORE := lapi lcode lctype ldebug ldo ldump lfunc lgc llex lmem lobject lopcodes \
+	lparser lstate lstring ltable ltm lundump lvm lzio lauxlib
+LUNATIKC_SRCS := ${LUNATIKC}.c $(addprefix lua/,$(addsuffix .c,$(LUNATIKC_CORE)))
+LUNATIKC_CFLAGS := -std=gnu99 -O2 -Wall -D_KERNEL -DLUA_USE_LINUX -I. -Ilua
```
lneto, line 35: (posted by an agent, not by @lneto)

`-DLUA_USE_LINUX` and `-lm` buy nothing here: without them the tool links, builds and writes the same 264 chunks. `LUA_USE_LINUX` only turns on `LUA_USE_POSIX` (`_setjmp`, and `dlopen` for `loadlib`, which is not linked), and no host object under `_KERNEL` calls libm. https://github.com/luainkernel/lunatik/commit/5f74bc5ee048a0df4fbb7b0d8442299ceb2e67cd drops both.

Makefile
```diff
@@ -62,14 +81,17 @@ AUTOGEN_KEY := $(KERNEL_RELEASE)|$(LUNATIK_MODULES)
 	tests_install tests_uninstall \
 	ebpf ebpf_install ebpf_uninstall
 
-all: lunatik_sym.h autogen
+all: lunatik_sym.h autogen ${LUNATIKC}
 	${MAKE} -C ${MODULES_BUILD_PATH} M=${PWD} $(LUNATIK_CONFIG_FLAGS)
 
+${LUNATIKC}: ${LUNATIKC_SRCS} lunatik_conf.h
```
lneto, line 87: (posted by an agent, not by @lneto)

The tool includes `lua/*.h` but depends only on `lunatik_conf.h` and the `.c` files, so a submodule bump that changes `lopcodes.h` alone leaves a stale `bin/lunatikc` writing the old opcode numbers. https://github.com/luainkernel/lunatik/commit/20b99e69cd43f3e762f25b36d35917905a054c93 adds `$(wildcard lua/*.h)`.

Makefile
```diff
@@ -26,6 +26,25 @@ MKDIR = mkdir -p -m 0755
...
+LUNATIKC := bin/lunatikc
+LUNATIKC_CORE := lapi lcode lctype ldebug ldo ldump lfunc lgc llex lmem lobject lopcodes \
+	lparser lstate lstring ltable ltm lundump lvm lzio lauxlib
+LUNATIKC_SRCS := ${LUNATIKC}.c $(addprefix lua/,$(addsuffix .c,$(LUNATIKC_CORE)))
+LUNATIKC_CFLAGS := -std=gnu99 -O2 -Wall -D_KERNEL -DLUA_USE_LINUX -I. -Ilua
+
+# BYTECODE=1 installs kernel Lua scripts as stripped chunks, under their .lua names
+ifeq ($(BYTECODE),1)
+define INSTALL_LUA
+	for f in $(1); do ${LUNATIKC} -s -o $(2)/$$(basename $$f) $$f || exit 1; done
```
lneto, line 40: (posted by an agent, not by @lneto)

This arm writes the chunk with `lunatikc`'s `fopen`, so the installed file's mode comes from the umask, while the other arm and every other install line go through `${INSTALL} -m 0644`. https://github.com/luainkernel/lunatik/commit/932c109a129bb0edf02c1e062c576c820aff41d0 compiles into a `mktemp` file and installs that, so both arms end in the same `${INSTALL}`; the `BYTECODE=1` suite stays green (121/121 with the new cases) and no temporary is left behind.

README.md
```diff
@@ -100,16 +100,41 @@ usage: lunatik [load|unload|reload|status|test|list] [run|spawn|stop <script>] [
...
+and integer-only number format; chunks from the distribution `luac` are rejected by the kernel.
+
+* `-s`: strip debug information (line numbers, local and upvalue names)
+* `-o output`: output file, or a directory when compiling several inputs (default: `<input>.luac`)
+* `-n chunkname`: chunk name recorded in the dump, as in `load` (default: `@<input>`)
+
+A chunk is installed and run under the usual `.lua` name; the kernel detects it by its signature:
+
+```Shell
+lunatikc -s -o hello.lua hello.lua    # overwrite with the stripped chunk
```
lneto, line 120: (posted by an agent, not by @lneto)

The first thing the README shows is overwriting the source with its chunk, and the `BYTECODE=1` paragraph does not say that `-s` drops the source name and the line numbers, so every kernel error raised from an installed library reads `?:?: <message>`. https://github.com/luainkernel/lunatik/commit/e1b25d0ff802a61944b65ca312c9e6f37d1cc4ff writes the chunk next to the source and states the cost. Whether `BYTECODE=1` should strip at all is yours to decide: a `STRIP` knob and `-n @<installed path>` in the recipe would keep the diagnostics for the 9 KB the strip saves.

tests/lib.sh
```diff
@@ -22,7 +22,7 @@ mark_dmesg() { dmesg -C 2>/dev/null; }
 dmesg_since() { dmesg; }
 check_dmesg() {
 	local errs
-	errs=$(dmesg_since | grep -E "\.lua:[0-9]+:" || true)
+	errs=$(dmesg_since | grep -E "(\.lua:[0-9]+|\?:\?):" || true)
```
lneto, line 25: (posted by an agent, not by @lneto)

The `?:?:` pattern reaches `check_dmesg` but not the twin pattern in `run_test` below, so a suite that reads through `run_test` still passes over a failing stripped chunk. On master both readers already share one `KTAP_ERRORS`, so https://github.com/luainkernel/lunatik/commit/857340021f78c62acab46b57e855325d9eae1eb5 gives this branch the same variable and the rebase lands the change once.

tests/luac/run.sh
```diff
@@ -0,0 +1,92 @@
...
+cleanup() {
+	rm -f "$SRC"/*_bc.lua "$SRC"/*_s.lua "$SRC"/stock.lua
+}
+trap cleanup EXIT
+cleanup
+
+ktap_header
+ktap_plan 8
+
+if ! command -v lunatikc >/dev/null; then
```
lneto, line 31: (posted by an agent, not by @lneto)

Eight `ktap_skip` after `plan 8`, where the tree skips a whole suite with `skip()` and `plan 1` (`tests/bpf/run.sh:30`); and nothing exercises the tool's own paths: a syntax error, a binary input, several inputs with `-n`, the default output next to the input. https://github.com/luainkernel/lunatik/commit/dc71c4d5b98df79c89dbaab66d3f0417308b8fed takes the idiom and adds those four cases (plan 12); the default-output one aborts on this head and passes with the `outname` fix.

tests/luac/run.sh
```diff
@@ -0,0 +1,92 @@
...
+#
+# Usage: sudo bash tests/luac/run.sh
+
+SCRIPT="tests/luac/hello_bc"
+SCRIPTS_PATH="/lib/modules/lua"
+SRC="$SCRIPTS_PATH/tests/luac"
+
+source "$(dirname "$(readlink -f "$0")")/../lib.sh"
+
+cleanup() {
```
lneto, line 22: (posted by an agent, not by @lneto)

`cleanup` removes the chunks but leaves the runtimes the cases register: `bash tests/luac/run.sh` twice after one reload fails the second run with `tests/luac/hello_bc is already running`, and `lunatik list` shows `hello_bc`, `hello_s`, `user` and `textmode`. https://github.com/luainkernel/lunatik/commit/f0c1b1c0e8692fb20953c75d06382b554f320321 stops the four in `cleanup`; two runs after one reload are then 12/12 each and the list is empty.

doc/design/luac/kernel-notes.md
```diff
@@ -0,0 +1,161 @@
...
+struct) are fenced with `#ifdef __KERNEL__`, the result is a ~200 KB static binary whose output the
+kernel runs (verified: full and stripped chunks, `7/2 == 3`).
+
+Details that bit during the prototype:
+
+* `lua_dump`'s writer is called with `size == 0` blocks; `fwrite(p, 0, 1, f)` returns 0 and looks
+  like a failure. Treat `size == 0` as success.
+* `luaL_loadbufferx(..., "t")` on the host so that the tool never re-dumps a chunk it did not
+  compile (a stock chunk fed by mistake is rejected at the parse step, not silently copied).
+* The `lunatik_conf.h` redefinition of `lua_getlocaledecpoint` must stay under `__KERNEL__` (the
```
lneto, line 159: (posted by an agent, not by @lneto)

This says the `lua_getlocaledecpoint` redefinition must stay under `__KERNEL__`, while `plan.md` lists it as shared and the header keeps it shared behind an `#undef`. The host core never reaches it (every use in `lobject.c` sits under `#ifndef _KERNEL`), so the note is the right one; with the fence-only header it holds, and https://github.com/luainkernel/lunatik/commit/6fb4a71b75de3116e5bd2c8d215cab7612334fbc updates `plan.md` and the build flags the notes quote.

PR #743, opened 2026-08-24
lunatikc: luac built against the kernel configuration

lneto:
Adds `lunatikc`, the `luac` bytecode compiler built against the kernel configuration, plus the wiring to ship scripts as bytecode. Alternative to #742: instead of a new driver, this reuses the upstream `luac.c`, imported into the `lua/` submodule and fenced with `_KERNEL` like the rest of the patch (branch `claude_lunatikc` of luainkernel/lua).

The kernel already loads binary chunks everywhere (`lunatik_loadfile` runs with mode `"bt"`, `require` included); what was missing is a way to produce one. The distribution `luac` cannot: under `_KERNEL` the fork drops the float opcodes (renumbering everything after `OP_LOADI`), compiles `/` to integer division, and checks an integer `lua_Number` in the chunk header. Building `luac` itself with the host compiler from the same `lua/` sources and `_KERNEL` configuration makes its output match the kernel by construction, and brings the standard CLI (`-l` listing of kernel opcodes included) for free.

* `lunatik_conf: fence the kernel-only part of the configuration`: the number configuration and limits stay shared with the host build; the kernel-only material goes under `__KERNEL__`. No change for the kernel build.
* `lua: add luac, buildable against the kernel configuration`: submodule bump; `luac.c` imported verbatim from the 5.5.0 distribution, float constants and opcodes in the listing fenced under `_KERNEL`, stock file loading kept for the `LUNATIKC` (userspace compiler) build, and the one `-Wall` warning in `llex.c` fenced.
* `lunatikc: build luac against the kernel configuration`: built as part of `make`, installed next to `lunatik` as `lunatikc`. `BYTECODE=1 make install` ships the kernel Lua libraries and examples as stripped chunks under their usual `.lua` names (the loader detects the signature, so nothing changes on the kernel side); the full suite passes with source and with bytecode installs, `driver.lua` included.
* `runner: trim only a trailing .lua`: found by the new suite; `trim()` mangled any script path containing a name, one character and `lua` (`tests/luac/x` became `testsc/x`).
* `tests: add the luac suite`: compiled chunks run via `run` and `require`, full and stripped; error messages carry the source path and line, or `?:?:` when stripped (now recognized by the harness); `-l` lists a compiled chunk; a stock number format is rejected by the header; `load(..., "t")` refuses a chunk in the kernel.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,25 @@ MKDIR = mkdir -p -m 0755
 LN = ln -sf
 INSTALL = install -o root -g root
 
+# luac built with the host compiler from the same lua/ sources and _KERNEL configuration as lunatik.ko
+HOSTCC ?= cc
+LUNATIKC := bin/lunatikc
+LUNATIKC_CORE := luac lapi lcode lctype ldebug ldo ldump lfunc lgc llex lmem lobject lopcodes \
+	lparser lstate lstring ltable ltm lundump lvm lzio lauxlib
+LUNATIKC_SRCS := $(addprefix lua/,$(addsuffix .c,$(LUNATIKC_CORE)))
+LUNATIKC_CFLAGS := -std=gnu99 -O2 -Wall -D_KERNEL -DLUNATIKC -DLUA_USE_LINUX -I. -Ilua
+
+# BYTECODE=1 installs kernel Lua scripts as stripped chunks, under their .lua names
+ifeq ($(BYTECODE),1)
+define INSTALL_LUA
+	for f in $(1); do ${LUNATIKC} -s -o $(2)/$$(basename $$f) $$f || exit 1; done
+endef
+else
+define INSTALL_LUA
+	${INSTALL} -m 0644 $(1) $(2)
+endef
+endif
+
 CONFIG_LUNATIK ?= m
 CONFIG_LUNATIK_RUNTIME ?= y
 CONFIG_LUNATIK_RUN ?= m
@@ -62,14 +81,17 @@ AUTOGEN_KEY := $(KERNEL_RELEASE)|$(LUNATIK_MODULES)
 	tests_install tests_uninstall \
 	ebpf ebpf_install ebpf_uninstall
 
-all: lunatik_sym.h autogen
+all: lunatik_sym.h autogen ${LUNATIKC}
 	${MAKE} -C ${MODULES_BUILD_PATH} M=${PWD} $(LUNATIK_CONFIG_FLAGS)
+
+${LUNATIKC}: ${LUNATIKC_SRCS} lunatik_conf.h
+	${HOSTCC} ${LUNATIKC_CFLAGS} -o $@ ${LUNATIKC_SRCS} -lm
 
 clean:
 	${MAKE} -C ${MODULES_BUILD_PATH} M=${PWD} clean
 	${MAKE} -C ${MODULES_BUILD_PATH} M=${PWD}/autogen clean
 	${MAKE} -C examples/filter clean
-	${RM} lunatik_sym.h
+	${RM} lunatik_sym.h ${LUNATIKC}
 	${RM} -r lib/linux
 	${RM} autogen/lunatik/*.lua autogen/linux/*.lua \
 		autogen/dump_*.c autogen/dump_*.pp \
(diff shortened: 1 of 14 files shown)


lunatik_conf.h
```diff
@@ -6,30 +6,37 @@
 #ifndef lunatik_conf_h
 #define lunatik_conf_h
 
+/* shared by the kernel and the host compiler (bin/lunatikc): everything that shapes bytecode */
```
lneto: (posted by an agent, not by @lneto)

The comment says the shared region is everything that shapes bytecode, but `LUAL_BUFFERSIZE`, `LUAI_MAXSTACK` and `LUNATIK_GCCOUNT` never reach a chunk: `LUAI_MAXSTACK` is read by `ldo.c` only, `LUAL_BUFFERSIZE` by `lauxlib.c`'s buffers, `LUNATIK_GCCOUNT` by `lbaselib.c`, which the tool does not link. `LUAI_MAXSTACK 200` on the host does show: `lunatikc -p` over 250 files stops with `too many input files`, since `luac.c` checks the stack for one slot per input. https://github.com/luainkernel/lunatik/commit/94655b58c4efb7c2c87965a7450dbcb21b625368 fences from `<linux/random.h>` down and moves nothing, three lines against master; the 516 chunks of `lib/`, `examples/` and `tests/`, with and without `-s`, come out byte for byte the same, and 250 inputs compile. The commit body's first sentence goes with it.

Makefile
```diff
@@ -62,14 +81,17 @@ AUTOGEN_KEY := $(KERNEL_RELEASE)|$(LUNATIK_MODULES)
 	tests_install tests_uninstall \
 	ebpf ebpf_install ebpf_uninstall
 
-all: lunatik_sym.h autogen
+all: lunatik_sym.h autogen ${LUNATIKC}
 	${MAKE} -C ${MODULES_BUILD_PATH} M=${PWD} $(LUNATIK_CONFIG_FLAGS)
 
+${LUNATIKC}: ${LUNATIKC_SRCS} lunatik_conf.h
```
lneto: (posted by an agent, not by @lneto)

The rule depends on the `.c` files and `lunatik_conf.h`, not on `lua/*.h`, where the opcode numbering lives: a submodule bump that changes `lopcodes.h` alone leaves a stale `bin/lunatikc` writing the old numbers. https://github.com/luainkernel/lunatik/commit/daf49bbea78c7e832501a9382c9b86b8cc2eac1d adds `$(wildcard lua/*.h)`.

Makefile
```diff
@@ -26,6 +26,25 @@ MKDIR = mkdir -p -m 0755
 LN = ln -sf
 INSTALL = install -o root -g root
 
+# luac built with the host compiler from the same lua/ sources and _KERNEL configuration as lunatik.ko
+HOSTCC ?= cc
+LUNATIKC := bin/lunatikc
+LUNATIKC_CORE := luac lapi lcode lctype ldebug ldo ldump lfunc lgc llex lmem lobject lopcodes \
+	lparser lstate lstring ltable ltm lundump lvm lzio lauxlib
+LUNATIKC_SRCS := $(addprefix lua/,$(addsuffix .c,$(LUNATIKC_CORE)))
+LUNATIKC_CFLAGS := -std=gnu99 -O2 -Wall -D_KERNEL -DLUNATIKC -DLUA_USE_LINUX -I. -Ilua
```
lneto: (posted by an agent, not by @lneto)

`-DLUA_USE_LINUX` and `-lm` buy nothing here: `LUA_USE_LINUX` only turns on `LUA_USE_POSIX` and the `dlopen` of `loadlib`, which is not linked, and no host object under `_KERNEL` calls libm. Built without them the tool writes the same chunk; https://github.com/luainkernel/lunatik/commit/981c7458a422e5d32c7cffba8c15531b16cf4c87 drops both.

lneto (commented):

@lneto, verdict: the mechanism is the right one and the branch is a rebase and the fixups below from merge-ready; no gate is set, since this account authored the pull request. `luac.c` built with the kernel configuration is the compiler this tree wants: `lua.c` on the host and `string.dump` in the kernel write the same chunk but have no listing, and a driver of our own is #742.

Verified on 6.8.0-138 aarch64. The head builds clean and passes 118/118 with the source install and with `BYTECODE=1`. Rebased onto master with the fixups, `review/743` builds clean, `make doc-site` exits zero, the suite passes on the source install (367, with the 11 skips master has on this kernel: the fsnotify mount kind and sched_ext), and the luac suite passes twice after one reload. On the `BYTECODE=1` install every suite passes except one wifi test per run, `nl80211_station` once and `nl80211_ap` once: in both the journal has NetworkManager taking `lunatikap0` within 3 ms of the test's own action ("unmanaged -> unavailable, reason managed"), networkd logging `Link DOWN` on it, and the next nl80211 call answering `ENETDOWN`, which is the race of #1010 and hits the source install the same way when the timing lands; the netlink suite alone on the bytecode install passed 44/44 once and lost the station test to the same race once. systrack ran as bytecode through the watchdog and its kprobe fired; spyglass loaded as bytecode with no key event on the host to fire it; ifquarantine was not run, since the head predates dde4531d0 and its load quarantines the uplink of the host the review runs on.

`review/743` is master plus the five commits rebased and the fixups, one per comment below, plus two you asked for: `lunatik compile` runs `lunatikc` with the same arguments and returns its exit status, https://github.com/luainkernel/lunatik/commit/aab3f3a0bcd53dd3d1abe63187d5257188dbcc01 (the CLI and the README, whose example now reads `lunatik compile -o hello.luac hello.lua`) and https://github.com/luainkernel/lunatik/commit/de8998903d101d3528f5828a882a843d7e30ac18 (the case, plan 14). The rebase conflicted in `Makefile`, `README.md`, `tests/lib.sh`, `tests/README.md` and `tests/run.sh`, all mechanical (the `lib/skb` install line, the percpu section, `KTAP_ERRORS`). Taking it is a force-push of `claude_lunatikc`, which is yours; the autosquash folds every fixup, including the one on the install arm that a later one supersedes. Two notes that are not fixups: the OpenWrt feed copies `Kbuild` over `Makefile` and runs `modules` only, so `bin/lunatikc` is never built there and packaging bytecode for a router needs a host build in the feed; and #995 moves the fork to `v5.5.1-kernel` without the four `luac` commits, which rebase cleanly onto it (the `llex.c` and `lauxlib.c` hunks do not overlap), so whichever lands second rebases its fork branch.

At the squash, two bodies follow their fixups: `lunatik_conf:` no longer moves the limits, so its first sentence goes; `tests:` says "their patterns now match it too", and on master that is the one `KTAP_ERRORS`.

Makefile
```diff
@@ -26,6 +26,25 @@ MKDIR = mkdir -p -m 0755
...
+
+# BYTECODE=1 installs kernel Lua scripts as stripped chunks, under their .lua names
+ifeq ($(BYTECODE),1)
+define INSTALL_LUA
+	for f in $(1); do ${LUNATIKC} -s -o $(2)/$$(basename $$f) $$f || exit 1; done
```
lneto: (posted by an agent, not by @lneto)

This arm writes the chunk with `lunatikc`'s `fopen`, so a fresh file's mode comes from the umask and an existing one keeps whatever it had, while the other arm and every other install line go through `${INSTALL} -m 0644`; and the two arms say the same install twice. https://github.com/luainkernel/lunatik/commit/1f5c409dda30ee3b248a0e194b3793c0b588714d makes the macro one arm: `${INSTALL} -m 0644` as before, and with `BYTECODE=1` the files just installed are compiled in place, so mode and owner come from `install` and `luac` overwrites without changing them. The line stays one logical line, since the examples loop calls the macro inside a `\`-continued `for`. Without `BYTECODE` the expansion is master's lines; with it, under `umask 077`, every one of the 79 installed scripts is a 0644 chunk and `BYTECODE=1` passes the suite. An earlier fixup on this line compiled into a `mktemp` file; this one supersedes it and the autosquash folds both.

README.md
```diff
@@ -100,16 +100,38 @@ usage: lunatik [load|unload|reload|status|test|list] [run|spawn|stop <script>] [
...
+
+`lunatikc` is `luac` built with the host compiler from the same `lua/` sources and configuration
+as `lunatik.ko`, so its chunks match the kernel's opcode set and integer-only number format;
+chunks from the distribution `luac` are rejected by the kernel. The options are `luac`'s
+(`-l` list, `-o` output, `-p` parse only, `-s` strip debug information, `-v` version).
+
+A chunk is installed and run under the usual `.lua` name; the kernel detects it by its signature:
+
+```Shell
+lunatikc -s -o hello.lua hello.lua    # overwrite with the stripped chunk
```
lneto: (posted by an agent, not by @lneto)

The first thing the README shows is overwriting the source with its stripped chunk, and from then on every error from it reads `?:?:`; and several inputs make one chunk that runs them in order, as with `luac`, which a reader used to a compiler per file does not expect. https://github.com/luainkernel/lunatik/commit/44bd2928f9e35f7eecce97b5d46860af9f5b3fb8 compiles into a chunk beside the source, installs it under the `.lua` name, and says what `-s` costs and that it is one file per call.

tests/luac/run.sh
```diff
@@ -0,0 +1,95 @@
...
+#
+# Usage: sudo bash tests/luac/run.sh
+
+SCRIPT="tests/luac/hello_bc"
+SCRIPTS_PATH="/lib/modules/lua"
+SRC="$SCRIPTS_PATH/tests/luac"
+
+source "$(dirname "$(readlink -f "$0")")/../lib.sh"
+
+cleanup() {
```
lneto, line 27: (posted by an agent, not by @lneto)

`cleanup` removes the chunks but leaves the runtimes the cases register: `bash tests/luac/run.sh` twice after one reload fails the second run with `tests/luac/hello_bc is already running`, and `lunatik list` shows `hello_s`, `user`, `textmode` and `hello_bc`. https://github.com/luainkernel/lunatik/commit/677d0309fad6aac653ec06953c14e56703954e91 stops the four in `cleanup`, as `tests/rcu/run.sh` does; two runs after one reload then pass and the list is empty.

tests/luac/run.sh
```diff
@@ -0,0 +1,95 @@
...
+	rm -f "$SRC"/*_bc.lua "$SRC"/*_s.lua "$SRC"/stock.lua
+}
+trap cleanup EXIT
+cleanup
+
+ktap_header
+ktap_plan 9
+
+if ! command -v lunatikc >/dev/null; then
+	for i in $(seq 9); do ktap_skip "luac: lunatikc not installed"; done
```
lneto: (posted by an agent, not by @lneto)

Nine `ktap_skip` after `plan 9`, where the tree skips a whole suite with `skip()` and `plan 1` (`tests/bpf/run.sh`). https://github.com/luainkernel/lunatik/commit/ccd7692f47841412586b6fbb529a18b8a3502a89 takes the idiom.

tests/luac/run.sh
```diff
@@ -0,0 +1,95 @@
...
+source "$(dirname "$(readlink -f "$0")")/../lib.sh"
+
+cleanup() {
+	rm -f "$SRC"/*_bc.lua "$SRC"/*_s.lua "$SRC"/stock.lua
+}
+trap cleanup EXIT
+cleanup
+
+ktap_header
+ktap_plan 9
```
lneto: (posted by an agent, not by @lneto)

Nothing exercises the tool's own paths: a syntax error, `-p`, a foreign chunk as input, and the input count the shared `LUAI_MAXSTACK` capped. https://github.com/luainkernel/lunatik/commit/7e6df44d4209fabcdbeb849752745de421543a41 adds those four cases; the 250-inputs one fails on this head with `too many input files` and passes with the header fixup, and the stock-chunk one reuses the header the kernel-side case crafts. https://github.com/luainkernel/lunatik/commit/de8998903d101d3528f5828a882a843d7e30ac18 adds the `lunatik compile` case on top (plan 14).

lua
```diff
@@ -1 +1 @@
-Subproject commit d7ca49b3a9c44278078b4e12a7cb26bf4c9e30fd
+Subproject commit 058f618496840c647b435db30d12a7c068e13e57
```
lneto: (posted by an agent, not by @lneto)

`luac.c` is not in the GitHub mirror of Lua, only in the release tarball, so a merge of the mirror never touches it and a bump has to re-import it (empty between 5.5.0 and 5.5.1) before the listing fence is reapplied. The tree says nothing about the fork today; https://github.com/luainkernel/lunatik/commit/260304c6ba0df34498b686a83f655c792111244c adds the `lua/` row to the layout table of `AGENTS.md` with that step.

PR #744, opened 2026-08-24
tc: forward an argument buffer to the Lua callback

lneto:
`bpf_luatc_run` now takes an `(arg, arg_sz)` pair, exposed to the Lua callback as `tc_ctx:argument()` (a `data` buffer) — mirroring the existing `xdp` mechanism.

**1. The channel** (`lib/luatc.c`): the argument object lifecycle (attach/detach/release/reset), the `tc_ctx:argument()` accessor, and the extended kfunc signature. The tc suite's `pass` case now writes a magic word from eBPF and reads it back through `ctx:argument()`.

**2. A real consumer** (`examples/sniclassify`): `classify.c` already computed the TCP payload offset and discarded it (it passed `NULL, 0`); `sni.lua`'s `tcp_payload` re-derived the identical value. Now `classify` hands the offset through the argument and `sni.lua` reads `ctx:argument():getuint32(0)`, dropping the re-parse — a function and seven constants gone. This is the redundancy the feature exists to remove, in the example that motivated it.

## Test
- `sudo lunatik test tc` → 6/6 (pass verifies the argument round-trip); `sudo lunatik test xdp` → 5/5.
- End-to-end (veth + a real TLS ClientHello): `SNI=netflix.com` is classified (`sniclassify: netflix.com 65584`); `SNI=example.com`, not in the policy, is not — so `sni.lua` reads the SNI at the handed-over offset, not a fixed path.

### DIFF
--- a/examples/sniclassify/classify.c
+++ b/examples/sniclassify/classify.c
@@ -53,7 +53,8 @@ int classify(struct __sk_buff *skb)
 	if (payload > data_end)
 		goto pass;
 
-	int action = bpf_luatc_run(runtime, sizeof(runtime), skb, NULL, 0);
+	__u32 offset = payload - data;
+	int action = bpf_luatc_run(runtime, sizeof(runtime), skb, &offset, sizeof(offset));
 	if (action < 0)
 		return TC_ACT_OK;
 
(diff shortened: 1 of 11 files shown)


sneaky-potato (approved):
Looks good

PR #748, opened 2026-08-27
lunatik: add luatask binding

sneaky-potato:
This patch will add `luatask` binding to be used by `luasched` as a context member (`ctx:task()`)


lib/luatask.h
```diff
@@ -0,0 +1,14 @@
...
+* SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+* SPDX-License-Identifier: MIT OR GPL-2.0-only
+*/
+
+#ifndef luatask_h
+#define luatask_h
+
+#include <lunatik.h>
+
+lunatik_object_t *luatask_new(lua_State *L, struct task_struct *task);
```
lneto, line 33: where are you using it?

sneaky-potato: > where are you using it?

This is to be used in luasched, specifically here: https://github.com/luainkernel/lunatik/pull/749/changes#diff-9d83ecf51adbfa68fbcfd3dad948941a43fef27425b1b578b92c8fdf3f4717f8R61
The plan is to use it in a way similar to how `skb` is used in `luatc` @lneto

lneto (changes requested):
@sneaky-potato this needs changes before it can merge: three correctness issues in the object model (one of which frees memory the binding does not own) and a missing LDoc `@type` anchor, plus a naming point on `cpu()`. Details are inline; a fixup for each is on `review/748`. I built the branch and ran the `task` suite (2/2, dmesg clean through the collectgarbage path).

lib/luatask.c
```diff
@@ -0,0 +1,171 @@
...
+	{NULL, NULL}
+};
+
+LUNATIK_OPENER(task);
+static const lunatik_class_t luatask_class = {
+	.name    = "task",
+	.methods = luatask_mt,
+	.release = luatask_release,
+	.opener  = luaopen_task,
+	.opt     = LUNATIK_OPT_SOFTIRQ,
```
lneto: The task object frees a `task_struct` it does not own. `luatask_class` is declared with `.opt = LUNATIK_OPT_SOFTIRQ` only, and `luatask_new` allocates `sizeof(struct task_struct *)` of private storage that it immediately overwrites with `object->private = task`. The allocation is leaked (private now points at the kernel's `task_struct`), and at release `lunatik_releaseprivate` runs `lunatik_free` (kvfree) on that borrowed pointer. Allocate zero private bytes and mark the class `LUNATIK_OPT_EXTERNAL`, so release only drops the reference.
https://github.com/luainkernel/lunatik/commit/3bc5770fba6123cf536bd769f9bbc4699e4abbc1

lib/luatask.h
```diff
@@ -0,0 +1,36 @@
...
+#include <lunatik.h>
+
+static inline void luatask_reset(lunatik_object_t *object, struct task_struct *task)
+{
+	struct task_struct *ltask = (struct task_struct *)object->private;
+	if (task != NULL)
+		get_task_struct(task);
+	if (ltask != NULL)
+		put_task_struct(ltask);
+	ltask = task;
```
lneto: `luatask_reset` never updates the object. It reads `object->private` into a local, adjusts the refcounts, then does `ltask = task`, which writes the local and leaves `object->private` pointing at the old task. Any caller that reuses a task object across resets keeps the stale task. Assign to `object->private`.
https://github.com/luainkernel/lunatik/commit/6bee8436afb3838a8eae849b462eba5f8e9c0ab5

lib/luatask.c
```diff
@@ -0,0 +1,171 @@
...
+};
+
+lunatik_object_t *luatask_new(lua_State *L, struct task_struct *task)
+{
+	lunatik_require(L, &luatask_class);
+	lunatik_object_t *object = lunatik_newobject(L, &luatask_class, sizeof(struct task_struct *), LUNATIK_OPT_NONE);
+	if (task)
+		get_task_struct(task);
+	object->private = task;
+	lunatik_getobject(object);
```
lneto: `luatask_new` leaks a reference and is not stack-neutral. It takes an extra `lunatik_getobject` and pops the object, returning with refcount 2 and nothing on the stack; `luatask_current` then re-pushes it. The object never reaches refcount 0, so every `task.current()` result and its task reference leak. Drop the internal `getobject`/`pop` so `luatask_new` returns a fresh Lua-facing object on the stack (as `luaskb_new` does), and drop the compensating push in `luatask_current`.
https://github.com/luainkernel/lunatik/commit/5a2463d8f6820f2487f1e9527dce2aedc5572814

lib/luatask.c
```diff
@@ -0,0 +1,171 @@
...
+#include <linux/kthread.h>
+
+#include "luatask.h"
+
+LUNATIK_PRIVATECHECKER(luatask_check, struct task_struct *,
+	luaL_argcheck(L, private != NULL, ix, "task is not set");
+);
+
+/***
+* Returns the command name (comm) of the task, i.e., the executable.
```
lneto, line 32: Methods are not grouped under a `@type`. The class documents `comm`/`pid`/`tgid`/`prio`/`cpu` with `@function` but there is no `@type task` anchor, so LDoc attaches them to the preceding block. Add the anchor before this block, after the checker.
https://github.com/luainkernel/lunatik/commit/9ab9a3ef32a1f2c84c2a79df6c05834d5678eb95

lib/luatask.c
```diff
@@ -0,0 +1,171 @@
...
+* Returns whether the task is currently running on a CPU.
+* Only available on SMP systems.
+* @function cpu
+* @treturn integer 1 if a task is actively occupying a CPU core, else 0.
+*/
+#ifdef CONFIG_SMP
+static int luatask_cpu(lua_State *L)
+{
+	struct task_struct *task = luatask_check(L, 1);
+	lua_pushinteger(L, task->on_cpu);
```
lneto: `cpu()` returns a flag, not the CPU. It returns `task->on_cpu`, a 0/1 flag for whether the task is currently executing on a CPU, not the CPU it runs on. The name gives the impression of a CPU index rather than a flag. If the CPU index is the intent, `task_cpu(task)` gives it: a `static inline` in `<linux/sched.h>` defined on UP too (returning 0), so the `#ifdef CONFIG_SMP` guard goes and the method becomes unconditional.
https://github.com/luainkernel/lunatik/commit/7f802486ffb0d1b25c211617a41679d5d95f2dbf
The test asserted `cpu == 0 or cpu == 1`, which only held because the value was a flag; it now asserts the index is in `[0, cpu.num_online())`.
https://github.com/luainkernel/lunatik/commit/13d7be76256b515980b6525b8cadb1ac4f8f30ba

lib/luatask.c
```diff
@@ -0,0 +1,165 @@
...
+/***
+* Returns the command name (comm) of the task, i.e., the executable.
+* This is truncated to TASK_COMM_LEN (16) characters by the kernel.
+* @function comm
+* @treturn string command name of the task
+*/
+static int luatask_comm(lua_State *L)
+{
+	struct task_struct *task = luatask_check(L, 1);
+	lua_pushstring(L, task->comm);
```
lneto: @sneaky-potato

`luatask_comm` reads `task->comm` without a lock. `comm` is mutable (prctl, execve), written under `task_lock`, so a read racing a rename can return a mix of old and new bytes — bounded and NUL-terminated, so no crash, but a wrong value if the caller compares on it. [`get_task_comm`](https://github.com/torvalds/linux/blob/v6.8/fs/exec.c#L1176) would lock the read, but its `alloc_lock` is process-context and this class runs in softirq and hardirq (the `task` is passed to `sched` callbacks), so that lock can deadlock. The fixup copies with `strscpy_pad` into a local, no lock — the read [`bpf_get_current_comm`](https://github.com/torvalds/linux/blob/v6.8/kernel/bpf/helpers.c#L265) does from any context. It does not remove the race; no lockless read can. `pid`/`tgid`/`prio`/`cpu` need no change: single aligned reads, and `cpu()` uses `task_cpu` (`READ_ONCE`).
https://github.com/luainkernel/lunatik/commit/2c3a81aa15060a6d106a653e97f6549322e6e0f9

lneto:
@sneaky-potato A new module needs a `config.ld` entry so LDoc builds its page — the README module row is here, but `config.ld` is untouched, so `task`'s API docs drop out of the generated set. Added in alphabetical order (between `luasyscall` and `luatc`).
https://github.com/luainkernel/lunatik/commit/3808263545186759e9e67d629d0659d1fed617b6

lneto (approved):
@sneaky-potato all findings are folded (object model, comm, config.ld). It builds clean and the `task` suite passes. Approving.

One optional idea, does not block the merge: a short comment on the getters saying why they read without a lock — a scalar read gets a full value, old or new, and comm is copied with `strscpy_pad`, so the worst case is a garbled name, never a crash. It stops someone from later adding a lock, which would deadlock in softirq/hardirq. The fixup is here: https://github.com/luainkernel/lunatik/commit/93e792a93f0ab3e422e19ffe7a3fd86135e8636d. What do you think?

PR #749, opened 2026-08-27
lunatik: add luasched binding

sneaky-potato:
`sched` lets a sched_ext scheduler hand its enqueue decision to Lua: a struct_ops program calls the `bpf_luasched_run` kfunc with the task, and the callback registered with `sched.attach()` sets the dispatch queue and slice through a `sched.ctx`, with the task exposed as a task object. The runtime must be hardirq, since enqueue runs with the rq lock held.

The binding needs a 6.12+ kernel with `CONFIG_SCHED_CLASS_EXT`; elsewhere the module builds as a stub and `linux.scx` is empty, so the tree keeps building on 5.15. The `sched` suite covers the attach guards and the dispatch path through a minimal scheduler, skipping where sched_ext is absent.


lib/luasched.c
```diff
@@ -0,0 +1,283 @@
...
+#include <linux/bpf.h>
+#include <linux/sched/ext.h>
+
+#include <lunatik.h>
+#include <lunatik_ebpf.h>
+
+#include "luarcu.h"
+#include "luatask.h"
+
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 4, 0))
```
lneto: This admits 6.4 and up, but sched_ext landed in 6.12; on 6.4 through 6.11 the block is enabled and `<linux/sched/ext.h>` does not exist, so the module fails to build (confirmed on 6.8: `fatal error: linux/sched/ext.h: No such file or directory`). Gate on 6.12, or on sched_ext's actual availability.

lib/luasched.c
```diff
@@ -0,0 +1,283 @@
...
+	get_task_struct(task);
+	luatask_reset(lctx->task_obj, task);
+
+	lctx->task = task;
+	lctx->dsq = ctx->dsq;
+	lctx->slice = ctx->slice;
+
+	ret = lunatik_ebpf_invoke(L, lctx->cb);
+	luasched_handler_cleanup(lctx);
+	return 0;
```
lneto: `ret` holds the callback result from `lunatik_ebpf_invoke` on the line above, but the handler returns 0, so a failing callback is reported as success and `ret` is dead. `luatc_handler` returns `ret`; do the same.

lib/luasched.c
```diff
@@ -0,0 +1,283 @@
...
+	luasched_ctx_t ctx = {
+		.task  = task,
+		.dsq   = &dsq,
+		.slice = &slice,
+	};
+
+	LUNATIK_EBPF_RUN(key, key__sz, luasched_handler, &ctx);
+	cls->dsq = dsq;
+	cls->slice = slice;
+	return dsq;
```
lneto: `bpf_luasched_run` is `int` but returns `dsq`, a `u64` whose values carry high bits (SCX_DSQ_* set bit 63), so this truncates; it also collides with the `-1` error convention the early returns use. The dispatch queue already leaves through `cls->dsq`; return 0 on success (or an explicit error code).

lib/luasched.c
```diff
@@ -0,0 +1,283 @@
...
+{
+	luasched_ctx_t *lctx = lunatik_ebpf_getctx(L);
+	int ret = 0;
+
+	if (lctx == NULL)
+		return -1;
+
+	struct task_struct *task = (struct task_struct *)ctx->task;
+
+	get_task_struct(task);
```
lneto: This pins the task, and `luatask_reset` on the next line pins it again (its own `get_task_struct`), so the ctx holds two references to the same task for the callback. `luatc` borrows its raw `skb` (no reference) and lets `skb_obj` own it. Borrowing `lctx->task` the same way drops this `get` and its matching `put` in cleanup, and removes the now-unreachable `put_task_struct(lctx->task)` in `luasched_release` (cleanup has already nulled `lctx->task`, and a hardirq callback cannot be collected mid-run).

lib/luasched.c
```diff
@@ -0,0 +1,283 @@
...
+
+	lunatik_object_t *object = lunatik_newobject(L, &luasched_class, sizeof(luasched_ctx_t), LUNATIK_OPT_NONE);
+	luasched_ctx_t *ctx = (luasched_ctx_t *)object->private;
+
+	ctx->task_obj = luatask_new(L, NULL);
+	lunatik_getobject(ctx->task_obj);
+	lunatik_register(L, -1, ctx->task_obj);
+	lua_pop(L, 1);
+
+	lunatik_ebpf_attach(L, 1, &ctx->cb);
```
lneto: On the branch's base these are the callback-ref `lunatik_ebpf_attach`/`detach`. Current master (after the eBPF macro rename) uses `lunatik_ebpf_attach`/`detach` for ctx objects and `lunatik_ebpf_bind`/`unbind` for the callback ref, so on rebase this and `luasched_detach`'s `lunatik_ebpf_detach(L, &lctx->cb)` become `bind`/`unbind`. Flagging for the rebase, not this base.

lib/luasched.c
```diff
@@ -0,0 +1,283 @@
+/*
+* SPDX-FileCopyrightText: (c) 2026 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
+* SPDX-License-Identifier: MIT OR GPL-2.0-only
+*/
+
+/***
+* Linux Exensible Scheduler (sched_ext) integration.
```
lneto: Doc text is carried over from the XDP binding: `Exensible` should be `Extensible` here; the `attach` doc says an `XDP program` calls the kfunc and mentions the `XDP verdict` (lines 202, 219); `dsq` and `slice` have `@tparam integer` with no parameter name (lines 69, 81); and `MODULE_AUTHOR` (line 282) has `ashwanikamal/im421@gmail.com`, a slash where the file header uses a dot.

lneto (commented):
@sneaky-potato early notes on the draft. The big ones: it does not build on the kernels the version guard admits, the callback's return value is dropped, and there is no test suite yet. The rest is inline.

It uses #748's `task` API in the shape my review of #748 asks for (a Lua-facing `luatask_new`, a `luatask_reset` that writes back to `object->private`), so it depends on #748 landing with those fixes; the natural order is #748 then this.

Tests: the binding has no suite yet. sched_ext needs a 6.12+ kernel, but the attach guards (refusing a non-hardirq runtime, re-attach replacing the callback) are testable without it; a `run.sh` and a `tests/README.md` entry would complete it.

I reviewed statically and built on top of #748. I could not run it: this kernel (6.8) has no sched_ext.

lib/luasched.c
```diff
@@ -0,0 +1,278 @@
...
+*
+* The primary mechanism involves an sched_ext program calling the `bpf_luasched_run`
+* kfunc, which in turn invokes a Lua callback function previously registered
+* using `sched.attach()`.
+* @module sched
+*/
+
+#define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
+#include <linux/bpf.h>
+#include <linux/sched/ext.h>
```
lneto: `<linux/sched/ext.h>` is included outside the guard, so on any kernel before 6.12 the file still fails at the include: the 5.15 CI job on this head stops at `linux/sched/ext.h: No such file or directory` (https://github.com/luainkernel/lunatik/actions/runs/33783167686/job/100741503104). And when the guard is false, `LUNATIK_EBPF_NEWLIB(sched, ..., &luasched_class)` and `LUNATIK_EBPF_KFUNC_INIT` still reference `luasched_class` and `bpf_luasched_kfunc_set`, which live inside it: that is the 6.12 job's `‘luasched_class’ undeclared` (https://github.com/luainkernel/lunatik/actions/runs/33783167686/job/100741503538). `luatc` gets away with the same layout because `lunatik_ebpf.h` carries stubs for its `< 6.4` branch; this binding's condition is a different one, so it needs its own `#else`.

https://github.com/luainkernel/lunatik/commit/80fae5bbf46de93a67a34be2389508b79785afc4 gates the whole implementation on `CONFIG_SCHED_CLASS_EXT` alone (the option exists only from 6.12, and it is the availability the round-one comment asked to key on), moves the include inside, and gives the `#else` an empty lib with stub init/exit, the shape `lunatik_ebpf.h` uses below 6.4. Built and loaded on 6.8: `require("sched")` yields a table with no `attach`.

.github/workflows/build.yaml
```diff
@@ -43,6 +43,9 @@ jobs:
         echo "CONFIG_STACK_VALIDATION=n" >> .config
         echo "CONFIG_WERROR=n" >> .config
         echo "CONFIG_BPF_SYSCALL=y" >> .config
+        if [[ "$SERIES" != "5."* ]]; then
+          echo "CONFIG_SCHED_CLASS_EXT=y" >> .config
```
lneto, line 50: Appending `CONFIG_SCHED_CLASS_EXT=y` to a defconfig does not enable it: the option `depends on BPF_SYSCALL && BPF_JIT && DEBUG_INFO_BTF` (https://github.com/torvalds/linux/blob/v6.12/kernel/Kconfig.preempt#L136-L138), and defconfig sets neither `BPF_JIT` nor `DEBUG_INFO_BTF`, so syncconfig drops it and the 6.12 job compiles the disabled path, which is why it hit the undeclared class rather than anything in the real code.

The fix is the two dependencies ahead of the option, with `DEBUG_INFO_DWARF_TOOLCHAIN_DEFAULT` to turn on DEBUG_INFO (pahole is already in the runner's deps). It is not on `review/749`: the token I push with cannot carry a commit that touches `.github/workflows`, so here it is as a patch:

```diff
@@ -44,6 +44,11 @@ jobs:
         echo "CONFIG_WERROR=n" >> .config
         echo "CONFIG_BPF_SYSCALL=y" >> .config
         if [[ "$SERIES" != "5."* ]]; then
+          # sched_ext depends on BPF_JIT and DEBUG_INFO_BTF; without them
+          # syncconfig drops SCHED_CLASS_EXT and luasched builds its stub
+          echo "CONFIG_BPF_JIT=y" >> .config
+          echo "CONFIG_DEBUG_INFO_DWARF_TOOLCHAIN_DEFAULT=y" >> .config
+          echo "CONFIG_DEBUG_INFO_BTF=y" >> .config
           echo "CONFIG_SCHED_CLASS_EXT=y" >> .config
         fi
```

Unverified: I have no way to run the workflow from here, so please watch that job's config step once it runs.

lib/luasched.c
```diff
@@ -0,0 +1,278 @@
...
+	{"slice", luasched_slice},
+	{NULL, NULL}
+};
+
+static void luasched_release(void *private)
+{
+	luasched_ctx_t *lctx = (luasched_ctx_t *)private;
+	if (lctx->task_obj)
+		luatask_close(lctx->task_obj);
+	if (lctx->task != NULL) {
```
lneto: With the task now borrowed, `lctx->task` never carries a reference, so this `put_task_struct` would drop one the ctx does not own if it ever ran; and it cannot run, because `cleanup` nulls `task` before the callback returns and a hardirq callback is not collected mid-run. It is the leftover the round-one comment mentioned. https://github.com/luainkernel/lunatik/commit/23c36bded3e30362469765a8feebb75e01022221 removes the block.

lib/luasched.c
```diff
@@ -0,0 +1,278 @@
...
+* @usage
+*   sched.detach()
+* @within sched
+*/
+static int luasched_detach(lua_State *L)
+{
+	luasched_ctx_t *lctx = lunatik_ebpf_getctx(L);
+
+	if (lctx == NULL)
+		return -1;
```
lneto: `luasched_detach` is a `lua_CFunction`, and its return is the number of results: `-1` here reaches `luaD_poscall` as a negative count. `luatc_detach` and `luaxdp_detach` return 0 on the same path. https://github.com/luainkernel/lunatik/commit/e2f6a213b4bd70462cfab0611ddf71d64e3aa623.

lib/luasched.c
```diff
@@ -0,0 +1,278 @@
...
+	}
+}
+
+LUNATIK_OPENER(sched);
+static const lunatik_class_t luasched_class = {
+	.name    = "sched.ctx",
+	.methods = luasched_mt,
+	.release = luasched_release,
+	.opener  = luaopen_sched,
+	.opt     = LUNATIK_OPT_HARDIRQ,
```
lneto: `tc.ctx` and `xdp.ctx` are `SINGLE`: `lunatik_cloneobject` refuses to copy such an object into another runtime, which is what keeps a ctx, valid only while its callback runs on this runtime, from being stored in an rcu table and read elsewhere with a dangling `task`. `sched.ctx` has the same lifetime and reads as the same kind of object; if that is the intent, https://github.com/luainkernel/lunatik/commit/0c6974ad812161c5b3e6dfa856b2d8759c60e5a6 adds the flag.

lib/luasched.c
```diff
@@ -0,0 +1,278 @@
...
+{
+	luasched_ctx_t *ctx = luasched_ctx_check(L, 1);
+	lunatik_getregistry(L, ctx->task_obj);
+	return 1;
+}
+
+/***
+* Sets the sched_ext dispatch queue for this task.
+* @function sched_ctx:dsq
+* @tparam integer dispatch queue to set for the task
```
lneto: `@tparam integer dispatch queue ...` makes LDoc read `dispatch` as the parameter name; same for `slice in ns`. https://github.com/luainkernel/lunatik/commit/d4bd487b4554dd0d05cceedefd245b10eab2fc48 names them `dsq` and `slice`.

Separately, the usage requires `linux.scx`, which does not exist: autogen has no `scx` table (`autogen/specs.lua`), and adding one is not a one-liner, because the header it would read is absent below 6.12 and the specs have no conditional entry yet. Either the example changes or the table comes with autogen support for an optional header; that is your call, so no fixup here.

Kconfig
```diff
@@ -162,6 +162,12 @@ config LUNATIK_CPU
 	help
 	  CPU management and information.
 
+config LUNATIK_SCHED
```
lneto, line 166: The binding calls `luatask_new`, so `luasched.ko` cannot load without `luatask.ko`; #757 adds `select LUNATIK_TASK` to `LUNATIK_THREAD` for the same dependency. https://github.com/luainkernel/lunatik/commit/b751346f01d04cedaf65b6074ad7c87e7d1a7c3f.

config.ld
```diff
@@ -61,6 +61,7 @@ file = {
 	'./lib/luanotifier.c',
 	'./lib/luaprobe.c',
 	'./lib/luarcu.c',
+	'./lib/luasched.c',
```
lneto, line 65: The module is in `config.ld` but not in the README's module table, which AGENTS.md pairs with it. https://github.com/luainkernel/lunatik/commit/142487e867563d271521f2a5b7924e225747627d adds the row.

lib/luasched.c
```diff
@@ -0,0 +1,278 @@
...
+*   end
+*   sched.attach(my_scheduler)
+*
+*   -- In eBPF C code, to call the above Lua function:
+*   -- char rt_key[] = "my_sched_handler.lua"; // Key matches the script name
+*   -- int ret = bpf_luasched_run(rt_key, sizeof(rt_key), p, cls);
+* @see task
+* @within sched
+*/
+static int luasched_attach(lua_State *L)
```
lneto, line 241: Still no suite. The guards need no scheduler program: https://github.com/luainkernel/lunatik/commit/aab0a45ebea77691258d536f7944ddec31ff5292 adds `tests/sched/` with `attach` refusing a sleepable runtime and a hardirq runtime attaching, re-attaching and detaching, skipped when `/sys/kernel/sched_ext` is absent, wired into `tests/run.sh` and described in `tests/README.md`. Here (6.8) only the skip path ran; the two cases are unverified until they run on a sched_ext kernel, so please run them before folding. The dispatch path, `bpf_luasched_run` reaching the callback from a struct_ops scheduler, stays uncovered; it needs a loadable scheduler, which is a larger piece than this round.

lneto (changes requested):
@sneaky-potato thanks for the rebase and the folds: the callback result now returns, the dsq no longer truncates through the kfunc's `int`, the task is borrowed instead of pinned twice, and the attach/detach macros follow the rename. Two of the round-one items are still open, and the build is the blocker: both CI jobs fail on this head, for different reasons, traced inline. The rest is inline too, each with a fixup on `review/749`.

Verified here: the disabled path builds and loads on 6.8, `require("sched")` returns the stub, the new `sched` suite skips cleanly on this kernel, and the full run is green with `luasched` in the module list. What I could not verify: the enabled path, which needs a 6.12 kernel with `CONFIG_SCHED_CLASS_EXT`, and the CI fixup, which only a run of the workflow can confirm.

config.ld
sneaky-potato: Ah, this reminds me, `tc` is also not there in the README module table. I have created this issue for it: #781

sneaky-potato:
Right now build on 5.15 is failing because autogen includes sched/ext.h. Can we add the macros conditionally in autogen? Since we'd want to include those headers only on machines with sched_ext support.

autogen/specs.lua
```diff
@@ -44,6 +44,8 @@ return {
 		desc = "TC verdicts and flags." },
 	{ header = "linux/sched.h", prefix = "TASK_", module = "task",
 		desc = "Task state flags." },
+	{ header = "linux/sched/ext.h", prefix = "SCX_", module = "scx",
```
lneto: Every dump is an unconditional `#include`, so this entry fails the preprocess step on any kernel without the header: the 5.15 job stops at `autogen/dump_10.c:5:10: fatal error: linux/sched/ext.h: No such file or directory` (https://github.com/luainkernel/lunatik/actions/runs/33783167686/job/101540998848), and so does this 6.8 host. https://github.com/luainkernel/lunatik/commit/b797474fb0ebcc0ce7a187d4014c7b16791073a0 adds `optional = true` to the spec and has autogen wrap the include in `__has_include` (GCC 5 and up, so 5.15's toolchain has it) in both the dump stub and `extract.c`. With the header absent the spec yields no candidates and `linux/scx.lua` is an empty table, so `require("linux.scx")` works everywhere and `scx.SLICE_DFL` is nil only where sched_ext is. Built on 6.8: `linux/scx.lua` empty, every other table unchanged.

autogen.lua
```diff
@@ -375,14 +375,27 @@ local function intermediate_paths(mods)
 	return util.sorted(needs)
 end
 
+local function to_lua_number(val)
```
lneto, line 389: `math.tointeger` already parses the decimal, sign included, and `%X` of a negative Lua integer prints its two's complement, so the match, the negation and the `mininteger` case collapse into one comparison: https://github.com/luainkernel/lunatik/commit/edad9f5e2f464ba7cda198ebfd7245a5fbc76abc. Same output for `-9223372036854775808`, `-9223372036854775807`, `-1` and the positives, checked with lua5.4.

tests/sched/sched_pass.bpf.c
```diff
@@ -0,0 +1,56 @@
...
+{
+	struct task_class cls;
+
+	int ret = bpf_luasched_run(runtime, sizeof(runtime), p, &cls);
+
+	if (ret) {
+		cls.dsq = DSQ_DEFAULT;
+		cls.slice_ns = SCX_SLICE_DFL;
+	}
+	scx_bpf_dsq_insert(p, cls.dsq, cls.slice_ns, enq_flags);
```
lneto: `scx_bpf_dsq_insert` and `scx_bpf_dsq_move_to_local` do not exist on 6.12, the binding's minimum: 6.12 has `scx_bpf_dispatch` and `scx_bpf_consume` (https://github.com/torvalds/linux/blob/v6.12/kernel/sched/ext.c#L6079, https://github.com/torvalds/linux/blob/v6.12/kernel/sched/ext.c#L6380), 6.13 added the new pair (https://github.com/torvalds/linux/blob/v6.13/kernel/sched/ext.c#L6484, https://github.com/torvalds/linux/blob/v6.13/kernel/sched/ext.c#L6798), and 6.17 dropped the old one (https://github.com/torvalds/linux/blob/v6.17/kernel/sched/ext.c#L6466-L6469 has only `dsq_insert`). Neither name loads across the range, and the program declares none of them, so it relies on a bpftool new enough to dump kfunc prototypes into `vmlinux.h` (this host's 7.4 does not). https://github.com/luainkernel/lunatik/commit/59e998e354b610feedab1720770dd8d84935a43b declares the four as weak ksyms in `scx_min.bpf.h` and picks with `bpf_ksym_exists`, the shape scx's own compat layer uses. Unverified: this host has neither sched_ext nor `struct sched_ext_ops` in its BTF, so I could not even compile the program.

tests/sched/run.sh
```diff
@@ -0,0 +1,72 @@
...
+	ktap_fail "sched attach: refuses a sleepable runtime"
+fi
+
+if run_test "$REATTACH" hardirq; then
+	ktap_pass "sched reattach: a hardirq runtime attaches, re-attaches and detaches"
+else
+	ktap_fail "sched reattach: a hardirq runtime attaches, re-attaches and detaches"
+fi
+
+if run_test "$PASS" hardirq && sudo bpftool struct_ops register "$BPF_OBJ" "$PIN"; then
```
lneto: As written the case passes once `register` returns, whether or not the callback ran, and the harness does not run where it is meant to: `sudo` inside a script the runner already executes as root (the other suites call `bpftool` directly), `BPF_OBJ` relative to the caller's cwd, no `make -C "$DIR"` to build it, a plan of 3 over a skip block of 2, no skip for bpftool, clang or the module's BTF, `rm -f` on the link directory bpftool creates (a `.struct_ops` map is not link-based, so the directory argument buys nothing), `pass.lua` printing on every enqueue on the host, and `scx_min.bpf.h` left out of `tests_install`, which copies only `*.bpf.c`. https://github.com/luainkernel/lunatik/commit/8327f65545ddc860234e2a7fca3dd8a6c31f3966 has the callback report once, asserts that line in dmesg with no Lua error after a second of registration, unregisters before the runtime stops, builds and skips the way `tc` does, installs the header, and updates the README entry, which still says the dispatch path is uncovered. Ran here only up to the skip; unverified on a sched_ext kernel.

.github/workflows/build.yaml
```diff
@@ -43,6 +43,13 @@ jobs:
         echo "CONFIG_STACK_VALIDATION=n" >> .config
         echo "CONFIG_WERROR=n" >> .config
         echo "CONFIG_BPF_SYSCALL=y" >> .config
+        if [[ "$SERIES" != "5."* ]]; then
+          # sched_ext depends on BPF_JIT and DEBUG_INFO_BTF; without them
+          # syncconfig drops SCHED_CLASS_EXT and luasched builds its stub
+          echo "CONFIG_BPF_JIT=y" >> .config
+          echo "CONFIG_DEBUG_INFO_DWARF_TOOLCHAIN_DEFAULT=y" >> .config
+          echo "CONFIG_DEBUG_INFO_BTF=y" >> .config
```
lneto, line 49: The block now sets the three dependencies but no longer `CONFIG_SCHED_CLASS_EXT=y` itself, so the 6.12 job still builds the stub: its log has `luasched.o` compiling with nothing from sched_ext in it, which is the green you see. The line goes after `DEBUG_INFO_BTF`; as last round, my token cannot push a workflow change, so as a patch:

```diff
           echo "CONFIG_DEBUG_INFO_BTF=y" >> .config
+          echo "CONFIG_SCHED_CLASS_EXT=y" >> .config
         fi
```

lneto (commented):
@sneaky-potato the round-two folds are all in, and the end-to-end case has the right shape: a struct_ops scheduler whose `enqueue` goes through the kfunc. Three things keep the request-for-changes standing, traced inline: the `linux.scx` table breaks every kernel without `<linux/sched/ext.h>` (the 5.15 job, and this 6.8 host), the CI commit no longer sets `CONFIG_SCHED_CLASS_EXT` itself, so the green 6.12 job compiles the stub, and the scheduler program calls kfuncs 6.12 does not have. The pass case's harness has a few more, also inline.

> Can we add the macros conditionally in autogen?

Yes: https://github.com/luainkernel/lunatik/commit/b797474fb0ebcc0ce7a187d4014c7b16791073a0 marks the spec `optional` and guards its include with `__has_include`, so the module comes out empty where the header is absent; details on the spec line.

Verified here, on 6.8: with that fixup the tree builds again, `linux/scx.lua` is the empty table, the stub module loads, the `sched` suite skips, and the full run is green with `luasched` in the module list. Not verifiable here: anything that needs sched_ext, so the pass case, the compat header and the CI change are unverified; please run `tests/sched/run.sh` on a 6.12+ box before folding.

sneaky-potato:
> please run tests/sched/run.sh on a 6.12+ box before folding.

Tested the latest push with 6.18 (uses scx_insert methods)
The build passes with linux 6.12 (uses scx_dispatch)

lib/luasched.c
```diff
@@ -0,0 +1,288 @@
...
+#include <lunatik.h>
+#include <lunatik_ebpf.h>
+
+#include "luarcu.h"
+#include "luatask.h"
+
+#ifdef CONFIG_SCHED_CLASS_EXT
+#include <linux/btf.h>
+#include <linux/btf_ids.h>
+#include <linux/sched.h>
```
lneto, line 30: The fold dropped `<linux/sched/ext.h>` altogether: `SCX_DSQ_GLOBAL` and `SCX_SLICE_DFL` resolve only because `linux/sched.h` includes it on 6.12 (https://github.com/torvalds/linux/blob/v6.12/include/linux/sched.h#L85). The file uses the header, so it names it; https://github.com/luainkernel/lunatik/commit/5c2133fedc3a339ac43e11080900fde59d6d2c6c puts the include inside the gate. Compiled against 6.12.109 with `CONFIG_SCHED_CLASS_EXT=y`.

.github/workflows/build.yaml
```diff
@@ -41,6 +41,14 @@ jobs:
         echo "CONFIG_STACK_VALIDATION=n" >> .config
         echo "CONFIG_WERROR=n" >> .config
         echo "CONFIG_BPF_SYSCALL=y" >> .config
+        if [[ "$SERIES" != "5."* ]]; then
+          # sched_ext depends on BPF_JIT and DEBUG_INFO_BTF; without them
+          # syncconfig drops SCHED_CLASS_EXT and luasched builds its stub
+          echo "CONFIG_BPF_JIT=y" >> .config
```
lneto, line 47: Reproducing this recipe on a 6.12.109 tree here dropped `SCHED_CLASS_EXT` again: the arm64 `defconfig` sets `DEBUG_INFO_REDUCED=y` (https://github.com/torvalds/linux/blob/v6.12/arch/arm64/configs/defconfig#L1713), which hides `DEBUG_INFO_BTF` (https://github.com/torvalds/linux/blob/v6.12/lib/Kconfig.debug#L379), and the sched_ext option goes with it. The x86_64 defconfig the runner uses has no such line, and the job's log shows `Skipping BTF generation ... due to unavailability of vmlinux`, which only `DEBUG_INFO_BTF_MODULES` prints, so on the runner the option most likely held. One line makes the recipe hold on any arch:

```diff
           echo "CONFIG_BPF_JIT=y" >> .config
+          echo "CONFIG_DEBUG_INFO_REDUCED=n" >> .config
           echo "CONFIG_DEBUG_INFO_DWARF_TOOLCHAIN_DEFAULT=y" >> .config
```

lneto (commented):
@sneaky-potato the round-three folds are in, byte for byte, and the CI line too. What keeps the request-for-changes standing is the history, not the code: the fixes to the binding landed in the tests commit. `tests: cover the sched module` carries the `#ifdef CONFIG_SCHED_CLASS_EXT` gate, `select LUNATIK_TASK`, the README row, `detach`'s `return 0`, the `SINGLE` flag and the `release` cleanup, so `lunatik: add luasched binding` on its own still includes `<linux/sched/ext.h>` unconditionally and does not build below 6.12; a bisect landing there breaks. `review/749` is your branch re-cut with those hunks in the binding commit: https://github.com/luainkernel/lunatik/commit/82f64c33af2e60f27db38f1abfdde75136d8113d, https://github.com/luainkernel/lunatik/commit/6946f876cbc3ba7c6bf05fc803895427d6fe549c and https://github.com/luainkernel/lunatik/commit/b51593751bd7c2b1bf93924e40e5f7c8e3ee0813, same final tree as your head, authorship kept; the CI commit rebases on top unchanged. One fixup on it, inline.

Verified here: on 6.8, build, install and the full run (146 pass, 3 skip, the `sched` suite). And the 6.12 recipe from the workflow, reproduced on a 6.12.109 tree: with the config honoured the enabled path compiles clean, `bpf_luasched_run` is in the module, and `linux/scx.lua` comes out with its 30 constants, `DSQ_FLAG_BUILTIN` as `0x8000000000000000`, so the u64 conversion holds on real input. Not run here: the scheduler program itself, which needs a running sched_ext kernel; your 6.18 run stands for that.

lib/luasched.c
sneaky-potato: Thanks for catching that!

sneaky-potato:
> What keeps the request-for-changes standing is the history, not the code: the fixes to the binding landed in the tests commit.

Fixed the order and individual commit contents. Thanks a lot.

lneto (approved):
@sneaky-potato this is ready. The history now reads as it should: the binding commit builds on its own (checked on 6.8, where it compiles to the stub), the tests and the CI change sit in their own commits, and the folds match `review/749` byte for byte. Verified here: on 6.8, build, install and the full run (146 pass, 3 skip); on a 6.12.109 tree configured with the workflow's recipe, the enabled path compiles clean, `bpf_luasched_run` is exported by the module and `linux/scx.lua` carries its 30 constants. The scheduler program itself I could not run here; your 6.18 run covers that.

Before merge, the body: it is still the one line, with `shed_ext`. What and why, in the tree's voice, would be something like:

> `sched` lets a sched_ext scheduler hand its enqueue decision to Lua: a struct_ops program calls the `bpf_luasched_run` kfunc with the task, and the callback registered with `sched.attach()` sets the dispatch queue and slice through a `sched.ctx`, with the task exposed as a `task` object. The runtime must be hardirq, since enqueue runs with the rq lock held.
>
> The binding needs a 6.12+ kernel with `CONFIG_SCHED_CLASS_EXT`; elsewhere the module builds as a stub and `linux.scx` is empty, so the tree keeps building on 5.15. The `sched` suite covers the attach guards and the dispatch path through a minimal scheduler, skipping where sched_ext is absent.

sneaky-potato:
Updated the PR body.

PR #757, opened 2026-08-31
luathread: return a task object from task()

lneto:
`thread:task()` built an ad-hoc table `{cpu, command, pid, tgid}`, duplicating what the `task` module exposes. It now returns a `task` object, so task introspection lives in one place and `thread:task()` gains `:prio()` and the rest for free.

It also drops a latent NULL dereference: the old code read `thread->task->comm` and friends without checking `thread->task`, which is NULL once the thread has exited. `luatask_new(NULL)` yields an object whose methods raise instead. And the old `cpu` field read `task->on_cpu`, the running flag, not a CPU index; `task:cpu()` reads `task_cpu()`.

No table consumers exist in the tree, so nothing else changes. The thread suite covers the current task, a running spawned thread (reached through `lunatik._ENV.threads`) and an exited one.

### DIFF
--- a/lib/luathread.c
+++ b/lib/luathread.c
@@ -13,6 +13,8 @@
 
 #include <lunatik.h>
 
+#include "luatask.h"
+
 /***
 * Represents a kernel thread object.
 * @type thread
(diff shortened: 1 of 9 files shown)


lib/luathread.c
```diff
@@ -113,35 +115,20 @@ static int luathread_stop(lua_State *L)
...
-	lua_createtable(L, 0, 4);
-	int table = lua_gettop(L);
-
-#ifdef CONFIG_SMP
-	lua_pushinteger(L, task->on_cpu);
-	lua_setfield(L, table, "cpu");
-#endif
-
-	lua_pushstring(L, task->comm);
-	lua_setfield(L, table, "command");
```
sneaky-potato, line 138: One thing that crossed my mind, I think task->comm and task->prio are mutable, which means another thread can modify its value while Lua does pushstring on task->comm. This does not affect this PR but I think some synchronization needs to be added to luatask PR.

lneto: @sneaky-potato
> I think task->comm and task->prio are mutable, which means another thread can modify its value while Lua does pushstring on task->comm

Right about `comm`: the read is unlocked and the buffer mutable, so a rename racing it can return a mix of old and new bytes. [`get_task_comm`](https://github.com/torvalds/linux/blob/v6.8/fs/exec.c#L1176) would lock the read, but its `alloc_lock` is process-context and this class runs in softirq and hardirq, so it can't be taken here — #748 uses `strscpy_pad` into a local instead, no lock, the read [`bpf_get_current_comm`](https://github.com/torvalds/linux/blob/v6.8/kernel/bpf/helpers.c#L265) does from any context ([fixup](https://github.com/luainkernel/lunatik/commit/2c3a81aa15060a6d106a653e97f6549322e6e0f9)). `prio` reads the field directly — which is what the kernel does for reporting: [`/proc/pid/stat`](https://github.com/torvalds/linux/blob/v6.8/fs/proc/array.c#L556) reads it through `task_prio()`, a plain read of `p->prio`, no lock and no `READ_ONCE`. On an aligned word that is a whole old-or-new value, never torn, so the plain read is right.

PR #762, opened 2026-09-02
lunatik: register a percpu script as a single object

lneto:
On top of #772, merged: the runner registers a percpu script as that object.

`env.percpu` and the `<script>:<cpu>` key convention are gone: one table, one key, one lookup. The runner loses the per-CPU loops and the two-phase shutdown that existed only because stopping a percpu script deleted N entries under an iteration that allows deleting only the entry it hands out; the eBPF kfuncs lose the per-packet `scnprintf` and second hash lookup. `lunatik.runtime` loses its cpu argument: the runner was its only caller, and a script claiming an instance id is the trust-based declaration the registration checks exist to replace.

`spawn percpu` stays refused; #686 was built on the per-instance registry this removes and is to be redone over the object, one bound thread per instance.

This subsumes the registry split of 9b900f5e, which existed to separate two key shapes in one table; the object removes the second shape.

Tests: `tests/runtime/percpu` covers the runner path (one instance per CPU id, listed once, stop and rerun, spawn refused, rollback of a failing instance); the xdp and tc drop cases and the receiving-CPU case cover the dispatch.

Tested on 6.8.0-138 (aarch64), on this head over master: full suite 123/0/0, clean dmesg.


lunatik_obj.c
```diff
@@ -89,6 +88,12 @@ int lunatik_closeobject(lua_State *L)
 
 	if (private != NULL)
 		lunatik_releaseprivate(object->class, private);
+}
+EXPORT_SYMBOL(lunatik_closeprivate);
+
+int lunatik_closeobject(lua_State *L)
+{
+	lunatik_closeprivate(lunatik_checkobject(L, 1));
```
lneto, line 96: why this indirection?

PR #792, opened 2026-09-06
lunatik: percpu:resume delivers to every runtime

lneto:
A percpu script has no way to be handed an object by the runtime that started it, the quarantine set of the `ifquarantine` example: `resume` belongs to the plain runtime alone, so a script that takes its state that way cannot be one runtime per CPU.

`percpu:resume` copies the objects into each runtime and resumes it under the runtime's lock, where it also reads the state, since a runtime dispatches as soon as its script registers a hook and a stopped one has no state left. What a runtime yields is dropped, a broadcast having no single set of values to return, and the error of the first runtime that refuses a value or raises comes back naming it. Delivering the objects and resuming become one entry point, `lunatik_resume`, so that sequence is written once for both callers. Test: `resume_percpu`.

Based on #803, which fixes what `resume` does with what the script leaves behind.

### DIFF
--- a/lunatik_percpu.c
+++ b/lunatik_percpu.c
@@ -100,6 +100,54 @@ static void lunatik_releasepercpu(void *private)
 	free_percpu(percpu->runtimes);
 }
 
+static int lunatik_resumeruntime(lua_State *L, lunatik_object_t *runtime, int nargs, char *error)
+{
+	int nresults = -ENXIO;
+
+	lunatik_lock(runtime); /* a runtime dispatches as soon as its script registers a hook */
+	if (likely(lunatik_isready(runtime))) {
+		lua_State *Lto = lunatik_getstate(runtime);
+
+		nresults = lunatik_resume(Lto, L, 2, nargs);
+		if (nresults < 0) /* the caller raises: a longjmp here would skip the unlock */
+			strscpy(error, lua_tostring(Lto, -1) ?: "error object is not a string", LUAL_BUFFERSIZE);
+		lua_pop(Lto, nresults < 0 ? 1 : nresults); /* the message, or the yield a broadcast drops */
+	}
+	lunatik_unlock(runtime);
+	return nresults;
+}
+
+/***
+* Resumes every runtime, as `runtime:resume` does, delivering the same objects to each. Nothing
+* comes back: a broadcast has no single set of values to return, so what a runtime yields is
+* dropped.
+* @function resume
+* @param ... objects delivered to each runtime as the return values of its `coroutine.yield()`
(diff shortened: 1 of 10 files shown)


lunatik_percpu.c
```diff
@@ -100,6 +100,66 @@ static void lunatik_releasepercpu(void *private)
 	free_percpu(percpu->runtimes);
 }
 
+#define LUNATIK_ERRLEN	128
```
lneto: why this magic number?

lunatik_percpu.c
```diff
@@ -100,6 +100,66 @@ static void lunatik_releasepercpu(void *private)
...
+			/* the caller raises: a longjmp here would skip the unlock */
+			strscpy(error, message != NULL ? message : "failed to resume", LUNATIK_ERRLEN);
+			lua_pop(Lto, 1); /* error message */
+			status = nresults;
+		}
+		else {
+			lua_pop(Lto, nresults); /* a broadcast has no single caller to return them to */
+			status = 0;
+		}
+	}
```
lneto, line 115: I have an impression that it might be simplified

lunatik_percpu.c
```diff
@@ -100,6 +100,66 @@ static void lunatik_releasepercpu(void *private)
...
+	lunatik_object_t *runtime;
+	int cpu;
+
+	lunatik_foreachruntime(percpu, cpu, runtime) {
+		int status = lunatik_resumeruntime(L, runtime, nargs, error);
+
+		if (status == -ENXIO)
+			luaL_argerror(L, 1, LUNATIK_ERR_NULLPTR);
+		if (status != 0)
+			luaL_error(L, "cpu %d: %s", cpu, error);
```
lneto: why having a special case here?

lunatik_percpu.c
```diff
@@ -100,6 +100,66 @@ static void lunatik_releasepercpu(void *private)
...
+*   with the runtimes after it not resumed. A runtime that refuses a value stays where it yielded;
+*   one that raises is dead, so every later resume delivers to the CPUs before it again and fails
+*   on it again
+*/
+static int lunatik_resumepercpu(lua_State *L)
+{
+	lunatik_object_t *object = lunatik_checkobjectclass(L, 1, &lunatik_percpu_class);
+	lunatik_percpu_t *percpu = lunatik_topercpu(object);
+	int nargs = lua_gettop(L) - 1;
+	char error[LUNATIK_ERRLEN];
```
lneto: isn't possible to use pushvfstring somehow? or any other way to avoid this static buffer?

PR #795, opened 2026-09-06
probe: share one kprobe among the runtimes of a percpu script

lneto:
`probe.new` refused a percpu runtime: each one would have registered its own kprobe on the same address and dispatched it to itself, so a single call would have been handled once per runtime.

The runtimes now share one kprobe per symbol or address, kept in an object of the class `probe.kprobes` from `lunatik_percpudata`, the way `luanetfilter` shares its hooks. The class that owns such a list is what the new `LUNATIK_PERCPUDATA` writes, in `lunatik_percpu.h` and documented in `doc/capi.md`. A registration is refcounted, so the handle is the kprobe itself and the class is `LUNATIK_OPT_EXTERNAL`; `stop` and `enable` are refused where the set owns the kprobe, and a plain runtime probes as before.

Depends on #801, merged: the shared kprobe is armed from the first runtime's script body, and without that guard a call on a CPU whose runtime is not published yet reaches `lunatik_run` with a NULL runtime. #839 is stacked on this one for the macro.


lib/luaprobe.c
```diff
@@ -10,16 +10,28 @@
...
 
 #include <lunatik.h>
 
+typedef struct luaprobe_kprobe_s {
+	struct hlist_node node;
+	struct kprobe kp;
+	lunatik_object_t *runtime;
+} luaprobe_kprobe_t;
+
+typedef struct luaprobe_spec_s {
```
lneto: why naming this "spec"? this doesn't seem alike the rest of the framework

lib/luaprobe.c
```diff
@@ -10,16 +10,28 @@
...
+#include <linux/list.h>
 #include <linux/string.h>
 
 #include <lunatik.h>
 
+typedef struct luaprobe_kprobe_s {
+	struct hlist_node node;
+	struct kprobe kp;
+	lunatik_object_t *runtime;
+} luaprobe_kprobe_t;
```
lneto, line 23: why do we need both luaprobe_t and luaprobe_kprobe_t? and runtime on both structures?

lneto, line 23: can't we generalize this kind of struct? is it the only case in the framework?

lib/luaprobe.c
```diff
@@ -10,16 +10,28 @@
...
+
+typedef struct luaprobe_spec_s {
+	const char *symbol;	/* the Lua string, copied only when a kprobe is registered */
```
lneto: this comment looks poor

lib/luaprobe.c
lneto: (posted by an agent, not by @lneto)

> why do we need both luaprobe_t and luaprobe_kprobe_t? and runtime on both structures?

The two structures have two lifetimes: the registration is one per symbol or address and is shared by every runtime of the percpu set, while the handle is one per runtime, keyed into that runtime's registry and carrying `stop` and `enable`. `luanetfilter` splits the same way on master, `luanetfilter_hook_t` against `luanetfilter_t`: https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luanetfilter.c#L19-L35

`runtime` on both was not needed, and it is gone from the handle. On the registration it names what `lunatik_run` dispatches through: the percpu object when the kprobe is shared, the runtime itself when it is not. The handle used it only to put the reference in `release` and to compare against `lunatik_toruntime` in `stop`, and both reach it through `probe->kprobe->runtime`. The handle is now a single pointer, NULL when the object owns the kprobe, and that is the one test `stop`, `enable` and `release` share. `stop` disarms through master's idempotent `luaprobe_delete` instead of freeing, so the pointer stays valid for `release` to put the reference and `p:stop()` twice is still a no-op.

Rewritten and force-pushed onto current master: https://github.com/luainkernel/lunatik/commit/1eaf711586bb93a05954d1993bf27f4bdb77d10e. Full suite there is 152 pass, 0 fail, 0 skip, core 06D5CE1A9DCF5B0396E4FF6.

lneto: (posted by an agent, not by @lneto)

> can't we generalize this kind of struct? is it the only case in the framework?

It is not the only case. `luanetfilter` already does this on master and this branch is shaped after it: https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luanetfilter.c#L175-L196

After the rewrite the two files carry the same six functions, find, new, free, the list teardown, share and own, and differ in three points only: the match predicate, the register call and the unregister call. So the generalization is a header the registration embeds plus those three as callbacks:

```c
typedef struct lunatik_shared_s {
	struct hlist_node node;
	lunatik_object_t *runtime;
} lunatik_shared_t;
```

with a `lunatik_share(L, percpu, class, size, spec, match, arm)` and a release that walks the list through the disarm callback.

I did not add it here. Each binding still needs its own error path around its register call, so what the core absorbs is around fifteen lines per binding for a header type, three function pointers and an indirect call per registration, and a third case would say whether that callback set is the right one. It also has to change `luanetfilter`, which makes it a commit of its own rather than part of this one.



Deferred as #834, so the decision does not live only in this thread.

lib/luaprobe.c
lneto: (posted by an agent, not by @lneto)

> why naming this "spec"? this doesn't seem alike the rest of the framework

The type is gone. `spec` is in the tree's vocabulary, but only as the name of an unregistered instance of the registration type, never as a type of its own: `luanetfilter_checkspec` fills a stack `luanetfilter_hook_t` and `luanetfilter_newhook` copies it with `*hook = *spec`. https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luanetfilter.c#L125-L139

`luaprobe` does the same now. `struct kprobe` already carries `addr` and `symbol_name`, so the spec is a stack `luaprobe_kprobe_t` that `luaprobe_checkspec` fills and `luaprobe_newkprobe` copies with `*kprobe = *spec`; `luaprobe_spec_t` and the fields it duplicated are gone: https://github.com/luainkernel/lunatik/commit/1eaf711586bb93a05954d1993bf27f4bdb77d10e

lib/luaprobe.c
lneto: (posted by an agent, not by @lneto)

> this comment looks poor

It was propping up a type that should not have existed, and both went with the rewrite. What is left is the one line the code cannot say: the spec holds a Lua string it does not own, and that is safe because the argument stays anchored at its index until `luaprobe_newkprobe` copies it.

```c
	spec->kp.symbol_name = luaL_checkstring(L, ix); /* anchored at ix until luaprobe_newkprobe copies it */
```

https://github.com/luainkernel/lunatik/commit/1eaf711586bb93a05954d1993bf27f4bdb77d10e

lib/luaprobe.c
```diff
@@ -10,17 +10,23 @@
...
+} luaprobe_kprobe_t;
+
 /***
 * Represents a registered kprobe.
 * @type probe
 */
 typedef struct luaprobe_s {
-	struct kprobe kp;
-	lunatik_object_t *runtime;
+	luaprobe_kprobe_t *kprobe;	/* NULL when the percpu object owns the kprobe */
```
lneto, line 30: why keeping this struct only to wrap up luaprobe_kbrobe_t?

lib/luaprobe.c
```diff
@@ -35,12 +41,13 @@ static int luaprobe_dump(lua_State *L)
...
-static int luaprobe_handler(lua_State *L, luaprobe_t *probe, const char *handler, struct pt_regs *regs)
+static int luaprobe_handler(lua_State *L, luaprobe_kprobe_t *kprobe, const char *handler, struct pt_regs *regs)
 {
-	struct kprobe *kp = &probe->kp;
+	struct kprobe *kp = &kprobe->kp;
 	const char *symbol = kp->symbol_name;
+	lunatik_object_t *object = lunatik_getregistryobject(L, kprobe);
 
-	if (lunatik_getregistry(L, probe) != LUA_TTABLE) {
+	if (object == NULL || lunatik_getregistry(L, object->private) != LUA_TTABLE) {
```
lneto: it would be clearer if we move it to a macro

lib/luaprobe.c
```diff
@@ -104,28 +111,101 @@ static void luaprobe_delete(luaprobe_t *probe)
...
+static void luaprobe_stopkprobes(void *private)
+{
+	luaprobe_kprobe_t *kprobe;
+	struct hlist_node *next;
+
+	hlist_for_each_entry_safe(kprobe, next, (struct hlist_head *)private, node) {
+		hlist_del(&kprobe->node);
+		luaprobe_freekprobe(kprobe);
+	}
+}
```
lneto: can't we generalize this in a macro template? is this patter used anywhere else in the framework? please, take a deep look into this..

lib/luaprobe.c
```diff
@@ -104,28 +111,101 @@ static void luaprobe_delete(luaprobe_t *probe)
 	}
 }
 
+static bool luaprobe_matches(const struct kprobe *kp, const struct kprobe *spec)
```
lneto: why naming it "spec"?

lneto: why "matches" instead of "match"?

lib/luaprobe.c
```diff
@@ -104,28 +111,101 @@ static void luaprobe_delete(luaprobe_t *probe)
...
+{
+	if (spec->symbol_name == NULL)
+		return kp->addr == spec->addr;
+	return kp->symbol_name != NULL && strcmp(kp->symbol_name, spec->symbol_name) == 0;
+}
+
+static luaprobe_kprobe_t *luaprobe_findkprobe(struct hlist_head *kprobes, const luaprobe_kprobe_t *spec)
```
lneto: why not simply "luaprobe_find"?

lib/luaprobe.c
```diff
@@ -104,28 +111,101 @@ static void luaprobe_delete(luaprobe_t *probe)
...
+	luaprobe_kprobe_t *kprobe;
+
+	hlist_for_each_entry(kprobe, kprobes, node) {
+		if (luaprobe_matches(&kprobe->kp, &spec->kp))
+			return kprobe;
+	}
+	return NULL;
+}
+
+static luaprobe_kprobe_t *luaprobe_newkprobe(lua_State *L, lunatik_object_t *runtime, const luaprobe_kprobe_t *spec)
```
lneto: I would avoid this naming pattern "luaprobe_\<func\>probe", instead I would find better "\<func\>" names.

lib/luaprobe.c
lneto: (posted by an agent, not by @lneto)

> why keeping this struct only to wrap up luaprobe_kbrobe_t?

It stays, and what it buys is an address of its own for the handle. The object model has one way to make a private that is the raw pointer, `LUNATIK_OPT_EXTERNAL`: `lunatik_newobject` then allocates nothing and the binding assigns `object->private` itself, as `socket`, `task`, `rng`, `shash`, `comp` and `bpf` do (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lunatik_obj.c#L31). Here that pointer would be the kprobe, and for a shared probe it cannot be: `percpu:stop()` closes the data first, which frees every shared kprobe, and closes the runtimes after, which is when the handle's `release` runs (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lunatik_percpu.c#L68), so a private pointing at the shared kprobe is a freed pointer by the time `release` receives it. Leaving it NULL for the shared case does not work either: `lunatik_registerobject` files the handlers table under `object->private` (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lunatik.h#L476-L480), so every shared probe in a runtime would share the key NULL, and the checkers read a NULL private as a closed object (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lunatik.h#L387-L396).

So the private has to be a block the framework owns, holding a pointer that is NULL when the object owns nothing, and that is the one field `luaprobe_t` has. `luanetfilter_t` carries the same field for the same reason (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luanetfilter.c#L32); its two other fields are there because its callback needs the runtime and the skb.

No code change for this one.

lib/luaprobe.c
lneto: (posted by an agent, not by @lneto)

> it would be clearer if we move it to a macro

The second lookup was the problem, not the missing macro. The handler needs only this runtime's handlers table, and the branch registered the object under the kprobe key and then hopped through `object->private` to reach the table. Registering the table under the kprobe key instead, `lunatik_register(L, 2, kprobe)`, makes the guard the one lookup master's handler already had (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luaprobe.c#L43-L46), which is the same line `luadevice_fop` (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luadevice.c#L89-L92), `luahid_op` (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luahid.c#L167) and `luanetfilter_pushcb` (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luanetfilter.c#L41-L44) carry, so there is nothing left to wrap. `luanetfilter_hook_cb` keeps its object hop because its callback needs the private for the skb.

Fixup: https://github.com/luainkernel/lunatik/commit/c5527f20d0575918dd417db77fb531adf099fd75

lib/luaprobe.c
lneto: (posted by an agent, not by @lneto)

> can't we generalize this in a macro template? is this patter used anywhere else in the framework? please, take a deep look into this..

Yes. Every `release` in the tree was read for it: two walk a list of registrations and free each entry, `luanetfilter_stophooks` on master (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luanetfilter.c#L147-L161) and the one `luaprobe` would otherwise write here, and they are the same nine lines plus the same two-field class, modulo names. The other list users are not this pattern: `luadevice` unlinks one entry from a global list per object, `luarcu` is an RCU hash table.

Two identical sites with a class attached is what `LUNATIK_PRIVATECHECKER` covers for checkers, so this pull request adds `LUNATIK_PERCPUDATA(prefix, cname, T, free)` next to `lunatik_percpudata`, generating `prefix_release` and `prefix_class`, and the binding becomes one line:

```c
LUNATIK_PERCPUDATA(luaprobe_kprobes, "probe.kprobes", luaprobe_kprobe_t, luaprobe_disarm);
```

Unlike the `lunatik_shared_t` of #834 it adds no header type, no function pointers and no indirect call, which is why it lands now where that one did not. It is documented in `doc/capi.md`.

Correcting this reply, which said the `luanetfilter` conversion rode in the same fixup: it does not. That conversion is #839, commit https://github.com/luainkernel/lunatik/commit/46a536c421b4b7c1ba5d2e58bcd4c60ee7ce2b93, which is stacked on this pull request for the macro. The fixup SHA and the suite numbers this reply carried were from a lineage the branch no longer has; the current head is two commits, and the suite on them is 154 pass, 0 fail, 0 skip, core 8949A6D2586BD5C0A86A048.

lib/luaprobe.c
lneto: (posted by an agent, not by @lneto)

> why naming it "spec"?

The earlier round dropped `luaprobe_spec_t` because a type of that name was new to the tree; the parameter is not. On master `spec` is what `luanetfilter_findhook` and `luanetfilter_newhook` call this same parameter (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luanetfilter.c#L113 and https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luanetfilter.c#L125): the requested registration before it is registered, the entry type but on no list and not armed. That is the side `luaprobe_match(kp, spec)` compares against, a request rather than a second live kprobe, and the name says which side is which where a symmetric pair would not. It stays; no change.

lneto: (posted by an agent, not by @lneto)

> why "matches" instead of "match"?

No reason the tree supports: it spells the verb `match` (`luaset_match`, https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luaset.c#L238) and its C predicates are `is`/`has` forms; nothing is in the third person. Renamed to `luaprobe_match` in the family fixup: https://github.com/luainkernel/lunatik/commit/88967b5a27bb78a3ef09d91d6928dcc970b2f838

lib/luaprobe.c
lneto: (posted by an agent, not by @lneto)

> why not simply "luaprobe_find"?

Agreed: `luadevice_find` (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luadevice.c#L69) and `luaset_find` (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luaset.c#L55) are the bare form, and the noun only repeats the module here. Renamed with the rest of the family: https://github.com/luainkernel/lunatik/commit/88967b5a27bb78a3ef09d91d6928dcc970b2f838

lib/luaprobe.c
lneto: (posted by an agent, not by @lneto)

> I would avoid this naming pattern "luaprobe_<func>probe", instead I would find better "<func>" names.

The pattern came from `luanetfilter`, `findhook`, `newhook`, `freehook`, `sharehook`, `ownhook` (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luanetfilter.c#L113-L196), and from the core's `lunatik_finddata` and `lunatik_newdata`; there the noun differs from the module, here it repeats it. The family becomes the bare verbs, in one fixup:

| was | is |
|---|---|
| `luaprobe_matches` | `luaprobe_match` |
| `luaprobe_findkprobe` | `luaprobe_find` |
| `luaprobe_newkprobe` | `luaprobe_register` |
| `luaprobe_freekprobe` | `luaprobe_free` |
| `luaprobe_sharekprobe` | `luaprobe_share` |
| `luaprobe_ownkprobe` | `luaprobe_own` |

`register` is what the function does beyond allocating, `register_kprobe`, and the name is free on the C side because this file's Lua constructor is `new`, unlike `hid` and `netfilter` where `register` is the Lua entry point. `free` frees, after master's `luaprobe_delete` has disarmed, and `luaprobe_delete` and the Lua-facing `luaprobe_new` keep their names. `luaprobe_stopkprobes` was already gone, generated as `luaprobe_kprobes_release` by `LUNATIK_HLISTCLASS` (https://github.com/luainkernel/lunatik/commit/8b6a60e5184901323a0fe77857f0f6aacd39f132), and `luaprobe_kprobes_class` keeps its noun since it names what the list holds. Nothing in the tests or the docs spelled the old names except the `capi.md` example, updated in the same fixup. The two signature wraps stay: joined, they are 128 and 130 columns.

Fixup: https://github.com/luainkernel/lunatik/commit/88967b5a27bb78a3ef09d91d6928dcc970b2f838



The space this renamed into is not homogeneous itself: `luanetfilter` still carries the noun on the same six functions, and `luarcu` and `luadarken` spell `new` and `free` with theirs. Raised as #835, to sweep it and write the pattern down in `AGENTS.md`, preferring a template over a rename wherever the family is the same code.

lib/luaprobe.c
```diff
@@ -104,28 +110,87 @@ static void luaprobe_delete(luaprobe_t *probe)
...
+static luaprobe_kprobe_t *luaprobe_register(lua_State *L, lunatik_object_t *runtime, const luaprobe_kprobe_t *spec)
+{
+	luaprobe_kprobe_t *kprobe = lunatik_checkalloc(L, sizeof(luaprobe_kprobe_t));
+	struct kprobe *kp = &kprobe->kp;
+	int ret;
+
+	*kprobe = *spec;
+	kprobe->runtime = runtime;
+
+	if (kp->symbol_name != NULL && (kp->symbol_name = kstrdup(kp->symbol_name, lunatik_gfp(runtime))) == NULL) {
```
lneto: this predicate seems a bit too complex.. I'd move it to a macro

lib/luaprobe.c
```diff
@@ -104,28 +110,87 @@ static void luaprobe_delete(luaprobe_t *probe)
...
+	return kprobe;
+}
+
+static void luaprobe_free(luaprobe_kprobe_t *kprobe)
+{
+	luaprobe_delete(kprobe);
+	lunatik_free(kprobe);
+}
+
+LUNATIK_HLISTCLASS(luaprobe_kprobes, "probe.kprobes", luaprobe_kprobe_t, luaprobe_free);
```
lneto: is this name pattern "probe.kprobes" well-stablished in the framework?

lib/luaprobe.c
```diff
@@ -186,41 +257,54 @@ static const lunatik_class_t luaprobe_class = {
...
-	if ((ret = register_kprobe(kp)) != 0) {
-		kp->pre_handler = NULL; /* shouldn't unregister on release() */
-		lunatik_unregisterobject(L, object);
-		luaL_error(L, "failed to register probe (%d)", ret);
-	}
+	luaprobe_kprobe_t *kprobe = percpu != NULL ? luaprobe_share(L, percpu, &spec) :
+		luaprobe_own(L, probe, runtime, &spec);
 
+	lunatik_registerobject(L, 2, object);
+	lunatik_register(L, 2, kprobe); /* the handler finds this runtime's handlers by the kprobe they share */
```
lneto, line 315: don't we have a macro for this?

lunatik_percpu.h
```diff
@@ -51,6 +52,22 @@ extern const lunatik_class_t lunatik_percpu_class;
 int lunatik_percpu(lua_State *L);
 lunatik_object_t *lunatik_percpudata(lua_State *L, const lunatik_class_t *class, size_t size);
 
+#define LUNATIK_HLISTCLASS(prefix, cname, T, free)					\
```
lneto: does this make sense only on percpu? why not naming it LUNATIK_PERCPUCLASS or something alike?

lib/luaprobe.c
lneto: I think it's a poor name.. we should find a better name for this on both cases.. what do you think?

lunatik_percpu.h
lneto: (posted by an agent, not by @lneto)

> does this make sense only on percpu? why not naming it LUNATIK_PERCPUCLASS or something alike?

The lines of the macro are not percpu, a safe walk and a free; its use is. A registration owned by a list rather than by its own handle exists only as percpu data, by construction: both consumers take the list head from `lunatik_percpudata` and nowhere else (https://github.com/luainkernel/lunatik/blob/88967b5a27bb78a3ef09d91d6928dcc970b2f838/lib/luanetfilter.c#L163, https://github.com/luainkernel/lunatik/blob/88967b5a27bb78a3ef09d91d6928dcc970b2f838/lib/luaprobe.c#L270), because on a plain runtime the handle owns the registration and `release` frees it there. So the header is the right home, and the name now says what the macro defines: the class whose objects `lunatik_percpudata` creates, `LUNATIK_PERCPUDATA(prefix, cname, T, free)`, pairing with the function the way `LUNATIK_CLASSES` pairs with `LUNATIK_NEWLIB`. Not `PERCPUCLASS`, since `lunatik_percpu_class` is already the class of the percpu object itself, and this is the class of what it holds.

Fixup: https://github.com/luainkernel/lunatik/commit/1c8e32f4e1b1dea2ba21e767f1785651f3c0c6ea

lib/luaprobe.c
lneto: (posted by an agent, not by @lneto)

> is this name pattern "probe.kprobes" well-stablished in the framework?

The dotted half is; the plural half has one precedent. Every `.name` in the tree: the module name alone for a module's main class (`probe`, `netfilter`, `device`, `fifo`, ...), `module.thing` for a further class of a module (`netlink.channel`, `rcu.table`, `set.labeled`, `tc.ctx`, `xdp.ctx`), and an older underscore form in two files (`bpf_hash`, `bpf_queue`, `crypto_aead` and siblings). The only plural is `netfilter.hooks` on master (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luanetfilter.c#L159), the data class this one mirrors, and the plural is what the object is, a list of kprobes. This class never reaches a script: it is created by `lunatik_createobject` for the percpu object, so no metatable, type error or `lunatik_require` carries the name; it appears in nothing but the source. Kept as is.

lib/luaprobe.c
lneto: (posted by an agent, not by @lneto)

> don't we have a macro for this?

No. `lunatik_registerobject` alone is what `device`, `notifier` and `hid` do (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luadevice.c#L384, https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luanotifier.c#L245, https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luahid.c#L307); only `netfilter` follows it with a second `lunatik_register` (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luanetfilter.c#L222-L223), and there it registers the object, since its callback needs the private for the skb, while here it registers the handlers table. So the second call does not belong inside `lunatik_registerobject`, which the other callers would then pay for, and the two sites do not even pair the same values, so a wrapper of the pair has one caller each. The key by which a runtime finds its own registration is the part of the shared-registration idiom that #834 costs as `lunatik_share`, which is where a macro for it would come from, with the rest of that family.

lib/luaprobe.c
lneto: (posted by an agent, not by @lneto)

> this predicate seems a bit too complex.. I'd move it to a macro

Split rather than wrapped. The tree's shape for this is master's own line, one assignment tested once, `if ((kp->symbol_name = kstrndup(...)) == NULL) lunatik_enomem(L);` (https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luaprobe.c#L206-L207); what made the branch's line heavy was folding the address case, no symbol to copy, into the same condition. It is now an `if` on having a symbol around that same one-assignment test, with the free of the kprobe next to the raise it precedes. This is the only `kstrdup` in the tree, so a macro would have one site and would hide an assignment from the reader of the function.

Fixup: https://github.com/luainkernel/lunatik/commit/3f4ab74b584c5360cb81cfbbc51264f5de01bb8b

lib/luaprobe.c
lneto: (posted by an agent, not by @lneto)

> I think it's a poor name.. we should find a better name for this on both cases.. what do you think?

Agreed, and the honest finding is that the noun is the entry type itself; only the state differs. The tree has no word for a value in that state: elsewhere it is filled and registered in one object, so nothing has to name it (`luaprobe_new` on master fills `probe->kp` in place, https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luaprobe.c#L196-L207). What the value is doing for the reader: filled from the Lua arguments, on no list and not armed, compared against the live entries, and then copied whole into the one that gets registered, `*kprobe = *template`. That last step is what names it: it is a template, and the kernel's netfilter uses the word the same way, a conntrack template is one shaped like an entry, not in the table, that the real entry is created from. So the shape is `luaprobe_kprobe_t template` and `luanetfilter_hook_t template`, with `checkspec` becoming `checktemplate` in both files, and the entry type keeps its name.

Correcting this reply, which said the rename had landed: it did not. The branch was rebuilt after it was written and that fixup did not come with it, so `spec` and `checkspec` stand in `lib/luaprobe.c`, and `lib/luanetfilter.c` is untouched by this pull request; #839 carries that file's naming now and keeps `spec` as well. The argument above is unchanged, and the rename is one commit across both files. #839 is where the rest of that family is being settled, so that is where it should go if you want it. Nothing in this pull request depends on it.

lib/luaprobe.c
```diff
@@ -18,10 +19,13 @@
...
 */
-typedef struct luaprobe_s {
+typedef struct luaprobe_kprobe_s {
+	struct hlist_node node;
+	struct kref kref;	/* one reference per handle, plus the set's when the runtimes share it */
 	struct kprobe kp;
+	kprobe_opcode_t *addr;	/* the address asked for, NULL for a symbol: register_kprobe resolves kp.addr */
 	lunatik_object_t *runtime;
-} luaprobe_t;
+} luaprobe_kprobe_t;
```
lneto: why renaming this?

lneto: (posted by an agent, not by @lneto)

> why renaming this?

It should not be, and it is not any more. The name was chosen while the file had two types, `luaprobe_t` wrapping the handle and `luaprobe_kprobe_t` for the registration, and the noun told them apart, the way `luanetfilter_hook_t` sits beside `luanetfilter_t`. Once the handle became the registration itself, through the `kref` and `LUNATIK_OPT_EXTERNAL` in this pull request, one type was left and `kprobe` only repeated what `probe` already says.

Restored to master's names, type and variables both: `luaprobe_t` and `probe`, with `kp` still the `struct kprobe` inside it. Fixup: https://github.com/luainkernel/lunatik/commit/ccc5bfec4123d6d43c9aa838383b8221d73bc085. Suite on it: 158 pass, 0 fail, 0 skip.

PR #797, opened 2026-09-06
notifier: report the devices of the initial namespace, and classify the replay by the registering task

lneto:
`notifier.netdevice` hands the callback every namespace's devices by name, and a name resolves for the script only in the initial one: a new namespace's `lo` put the host's `lo` back into DROP in the `ifquarantine` example after an operator had allowed it. And the binding told a replayed event from a live one by reading `unregister` still NULL, which a live event arriving between the registration returning and that store also saw, running the callback unlocked.

The netdevice handler now declines a device of another namespace before anything reaches Lua, so the name the callback is given always means the device `linux.ifindex` resolves it to and a consumer such as `dome/vlan.lua` needs no change; the chain, its verdict semantics and the kernel range are master's. The second commit records the registering task across `register_fn`: an event is replayed when it arrives on that task in task context, and takes `lunatik_run` otherwise. A script tells the replay apart the same way, with a flag it clears once `notifier.netdevice` returns, since the replay is delivered inside the call and no live event reaches the callback before the script body ends; the doc block says so. Tests: `netns_scope`, which counts a homonym once where the unfiltered chain reports it three times on this host, and `replay`.

Replaces #826, which kept the names and appended `ifindex` and a namespace inode for each callback to compare. #831 is the example using the flag, stacked on this branch. Run on 6.8.0-138 aarch64 at the head: notifier suite 20/20, full suite pass:174 fail:0 skip:3.


doc/capi.md
```diff
@@ -242,6 +242,20 @@ defines `luaprobe_kprobes_class`.
 
 ---
 
+## Network Namespace
+
+### LUNATIK\_NETNS
+```C
+#define LUNATIK_NETNS (&init_net)
+```
+The network namespace every binding acts in. Device lookups, sockets, netfilter
```
lneto: _Agent review (Claude Code), posted from @lneto's account._

The list here names the netdevice notifier, which this commit does not convert; `luanotifier.c` moves to `LUNATIK_NETNS` in the next one. It also leaves out the generic netlink unicast, which this commit does convert (https://github.com/luainkernel/lunatik/blob/c96d20dafad8e8b7831f4f722b402d46be74429f/lib/luanetlink.c#L118). AGENTS.md asks for no such list at all, "Enumerate in one place only, and prefer none: a list of modules that adopt a rule rots on the next adoption", and the block reads the same without it since the sentence after already states the contract. If that is the intent, https://github.com/luainkernel/lunatik/commit/64788f9bb31cfb0e4e2086649f34684390cc4ef7 drops the enumeration and keeps the reason.

lib/luanotifier.c
```diff
@@ -113,38 +120,56 @@ static int luanotifier_stop(lua_State *L)
...
-*   is a `linux.netdev` code and `name` is the device name (e.g. `"eth0"`).
-*   Returns a `linux.notify` status code.
+* @tparam function callback invoked as `callback(event, name, replayed)` —
+*   `event` is a `linux.netdev` code, `name` is the device name (e.g. `"eth0"`)
+*   and `replayed` is `true` for an event the registration delivers itself:
+*   a `REGISTER`, and an `UP` for a device that is up, for each device the
+*   namespace already has. A live event carries `false`.
+*   Returns a `linux.notify` status code. `notify.BAD` vetoes `REGISTER` and
+*   the events the kernel lets a notifier veto, and any code carrying
+*   `notify.STOP_MASK` stops the event before every other notifier in the
```
lneto: _Agent review (Claude Code), posted from @lneto's account._

"before every other notifier in the kernel sees it" reads as stronger than what the chain does. `call_netdevice_notifiers_info` calls the per-netns chain first and returns early on a `NOTIFY_STOP_MASK`, so a stop skips the rest of the per-netns chain and all of the global one, while the per-netns blocks called before Lunatik's have already seen the event (https://github.com/torvalds/linux/blob/v6.8/net/core/dev.c#L1963, same at https://github.com/torvalds/linux/blob/v5.15/net/core/dev.c#L1993). The commit body already puts it exactly, "stops the event before every globally registered notifier". https://github.com/luainkernel/lunatik/commit/d75799260fde96d3f7c14e556e97bd2f178d0668 brings the doc block to that wording.

tests/notifier/replay.sh
```diff
@@ -0,0 +1,99 @@
...
+}
+
+reported()
+{
+	dmesg_since | grep -cF "replay: $1"
+}
+
+marked()
+{
+	dmesg_since | grep -cE "^.*replay: .* true$"
```
lneto: _Agent review (Claude Code), posted from @lneto's account._

`^.*` matches nothing grep does not already match unanchored, and `reported()` just above uses the plain form. https://github.com/luainkernel/lunatik/commit/69bd5d01a0333ca673bd96d1a3be0c64c56fa4e8 drops it; the suite is 20/20 with it applied, core EB548264F03F69BB848684A.

lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._

Verdict on #797: request changes.

@lneto Three fixups on `review/797`, all in text the branch adds. The C traces clean and I did not touch it.

Against the declared floor: `register_netdevice_notifier_net` is absent in v5.4 and present and exported in v5.15 and v6.8, so the v5.5 claim holds and nothing else the change uses is newer than 5.15. In both 5.15 and 6.8 the replay is synchronous inside the call, under RTNL, on the calling task, and the per-netns chain runs before the global one.

On the classification, `in_task() && registrant == current` cannot be true for a live event, since `registrant` is cleared the statement after `register_fn` returns and any other task compares against its own `current`, and cannot be false for a replayed one, since the replay runs in the caller's own frame. The `in_task()` half matters past the netdevice chain: `register_keyboard_notifier` and `register_vt_notifier` are atomic chains, so an interrupt on the registering task during registration is what the old `unregister == NULL` sentinel would have read as a replay.

On the error path, a `notify.BAD` from the replay makes the kernel roll the registration back and `register_fn` return `-EPERM`, so `unregister` is never stored and `release` skips a block the kernel already unregistered.

Two outcomes sit outside the new tests, a callback's verdict reaching the chain and a registration that fails because the callback vetoed its own replay. Both are master's code and both behave the same without this branch, so neither is this branch's to cover. Verified here: `make` clean, notifier suite 20/20 with the fixups applied, full suite pass:174 fail:0 skip:3 at the head, core EB548264F03F69BB848684A.

lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #797, second round: it can merge as it stands.

@lneto The third commit now keeps the classification by the registering task and drops the `replayed` argument; a script tells the replay apart with a flag it clears once `notifier.netdevice` returns. The claim that rests on is exact, and it is worth recording where. `lunatik_extra(L)->ready` is false for the whole of the body: `lunatik_newruntime` runs the script under `lua_pcall` and only then calls `lunatik_setready`, which publishes the store under the runtime lock, the same lock `lunatik_run` takes before reading it. So a live event during the body runs on another task, is classified live, takes `lunatik_run`, finds the runtime not ready and comes back `-ENXIO`, which `max(ret, NOTIFY_DONE)` turns into `NOTIFY_DONE` with the callback never entered. Inside `register_fn` no live event can be delivered at all, since `register_netdevice_notifier_net` holds RTNL and every netdevice notification asserts it (https://github.com/torvalds/linux/blob/v6.8/net/core/dev.c#L1957). The keyboard and vt chains reach the same place by a shorter road: they replay nothing, and a live event during the body meets the same ready check.

One finding from this round is folded already: the doc block had lost the qualifier that the replayed `UP` is only for a device that is up (`call_netdevice_register_notifiers` returns before it otherwise, https://github.com/torvalds/linux/blob/v6.8/net/core/dev.c#L1673); the block now says so and wraps with the rest. The three findings of the first round are folded and read right.

One thing the new shape costs, worth knowing rather than changing: `replay.sh` no longer discriminates this commit's C half. With the old `unregister == NULL` sentinel the replay still takes `lunatik_handle` inside `register_fn`, so the flag is still set when the callback runs and all seven assertions still pass. The race the classification closes needs a live event in a few-instruction window, or a hardirq during an atomic-chain registration, which a test cannot produce, so there is nothing to add. The test does discriminate what it now covers, the flag and the delivery inside the call.

Verified here: `make` clean, notifier suite 20/20, full suite pass:174 fail:0 skip:3 at 8f7a0ea46, and the doc fold changes no code.

lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #797, third round: it can merge as it stands.

@lneto The branch is now the smallest shape of the two fixes. The first commit keeps master's global `register_netdevice_notifier` and declines a device of another namespace in the netdevice handler, `net_eq(dev_net(dev), &init_net)`, before anything is pushed; a handler returning a negative count declines the event and `luanotifier_handler` answers `NOTIFY_DONE` without calling the script. That drops the per-netns chain of the earlier shape, its two wrappers and the `LUANOTIFIER_NEWCHAIN` signature change, the dependency on `register_netdevice_notifier_net` (v5.5), and the widened reach of a returned `notify.STOP`, which no longer needs a doc sentence: the chain, the verdict semantics and the kernel range are master's. Traced: the global replay walks `for_each_net` synchronously in the caller's frame under RTNL (https://github.com/torvalds/linux/blob/v6.8/net/core/dev.c#L1735), so the registering-task classification of the second commit holds unchanged; `dev_net` is valid at UNREGISTER too, and `net_eq` is `==` with `CONFIG_NET_NS` and constant true without it (https://github.com/torvalds/linux/blob/v6.8/include/net/net_namespace.h#L284), where `dev_net` always yields `init_net`. Cost, named rather than measured: one uncontended runtime lock per foreign live event before it is declined, and a replay that walks every namespace once at registration.

`netns_scope` passes unchanged and still discriminates: with the filter removed, its first assertion fails with "lo was reported 3 times on the replay" on this host. The second commit is as reviewed in the last round, with the doc block's replayed `UP` qualified to devices that are up. Verified here at the head: `make` clean, notifier suite 20/20, full suite pass:174 fail:0 skip:3, and `examples/ifquarantine` on the stack with a dummy device created in a second namespace, which the callback never saw while a local veth was quarantined.

lib/luanotifier.c
```diff
@@ -68,10 +72,9 @@ static int luanotifier_handler(lua_State *L, luanotifier_t *notifier, unsigned l
 static int luanotifier_call(struct notifier_block *nb, unsigned long event, void *data)
 {
 	luanotifier_t *notifier = container_of(nb, luanotifier_t, nb);
-	bool islocked = !notifier->unregister; /* still inside register_fn? */
 	int ret;
 
-	if (islocked)
+	if (in_task() && notifier->registrant == current) /* the replay: locked by this task, or the body */
```
lneto: This predicate can be moved out to a macro

lib/luanotifier.c
```diff
@@ -125,18 +128,27 @@ static int luanotifier_netdevice_handler(lua_State *L, void *data)
 {
 	struct net_device *dev = netdev_notifier_info_to_dev(data);
 
+	if (!net_eq(dev_net(dev), &init_net)) /* the name resolves to another device for the script */
```
lneto: Also this can be turned into a macro.. we should also reinforce our harness to create helpers when we need to explain that much on comments

lib/luanotifier.c
```diff
@@ -125,18 +128,27 @@ static int luanotifier_netdevice_handler(lua_State *L, void *data)
...
+		return -1;
```
lneto: Why -1?

lib/luanotifier.c
lneto: _Agent reply (Claude Code), posted from @lneto's account._

> This predicate can be moved out to a macro

Done, in the shape the tree's predicates take, one expression over one argument, with the reason on the definition and nothing on the use:

```c
/* the registering task holds the runtime lock, or is still running the script body */
#define luanotifier_isreplay(notifier)	(in_task() && (notifier)->registrant == current)
```

Folded into the commit that introduces it, https://github.com/luainkernel/lunatik/commit/74a3e0b5b. Notifier suite 20/20 and the full suite 174/0/3 at the head.

lib/luanotifier.c
lneto: _Agent reply (Claude Code), posted from @lneto's account._

> Also this can be turned into a macro.. we should also reinforce our harness to create helpers when we need to explain that much on comments

Done, above its only user:

```c
/* the callback is given a name, which would resolve to a different device for the script */
#define luanotifier_isinitnet(dev)	net_eq(dev_net(dev), &init_net)
```

Folded into the commit that introduces it, https://github.com/luainkernel/lunatik/commit/815f4ff5e. The harness gets the rule in #896, under "Comments and documentation", right after the sentence that already says reaching for a comment is a signal to reconsider the code: a condition that needs a comment to say what it tests is a predicate not yet named, given one in the `lunatik_isirq` family's shape, defined where its reader is, with the reason on the definition and nothing on the use. A mechanical nudge in `comment-style.sh` was measured and held back: a trailing comment on an `if` line hits 13 places in the tree today, of which the seven that survive filtering explain why a branch exists rather than what it tests, and seven standing annotations on untouched code would train readers to ignore the check.

lib/luanotifier.c
lneto: _Agent reply (Claude Code), posted from @lneto's account._

> Why -1?

A handler returns the count of values it pushed for the callback, so a negative count is outside the domain and says the handler declined the event. It appeared in two places, the return and the test, which is where AGENTS.md says a sentinel gets a name:

```c
#define LUANOTIFIER_DECLINED	(-1) /* a handler returns it instead of a count of pushed values */
```

on the `luanotifier_handler_t` typedef, where the contract's reader is, and the test is `nargs == LUANOTIFIER_DECLINED`. The shapes without a sentinel were weighed: a per-chain accept predicate brings back the per-chain function-pointer machinery the filter removed; a `bool` out-parameter changes the shared typedef and all three handlers; a handler returning `NOTIFY_DONE` itself collides with a legitimate count of zero, since `NOTIFY_DONE` is 0. Folded into https://github.com/luainkernel/lunatik/commit/815f4ff5e.

lib/luanotifier.c
```diff
@@ -25,6 +26,8 @@
 #include <lunatik.h>
 
 typedef int (*luanotifier_register_t)(struct notifier_block *nb);
+
+#define LUANOTIFIER_DECLINED	(-1) /* a handler returns it instead of a count of pushed values */
```
lneto: why not using a LUA_ define suited for this? isn't there none?

lneto: why adding this between these two typedefs?

lneto: this comment seems unnecessary

lib/luanotifier.c
```diff
@@ -121,22 +129,34 @@ static int luanotifier_##name(lua_State *L)					\
 		(class));							\
 }
 
+/* the callback is given a name, which would resolve to a different device for the script */
```
lneto: this comment seems unnecessary and it's not explaining what the macro actually does; it goes against our rules

lib/luanotifier.c
```diff
@@ -48,16 +52,21 @@ static const lunatik_class_t luanotifier_hardirq_class;
 LUNATIK_PRIVATECHECKERS(luanotifier_check, luanotifier_t *, "notifier", &luanotifier_process_class,
 	&luanotifier_hardirq_class);
 
+/* the registering task holds the runtime lock, or is still running the script body */
```
lneto: the comment needs to explain what the macro actually does; also, reinforce our harness regarding this.

lib/luanotifier.c
lneto: _Agent reply (Claude Code), posted from @lneto's account._

> why not using a LUA_ define suited for this? isn't there none?

There is none in this domain. The vendored Lua has three `(-1)` defines and each means something else: `LUA_MULTRET` (`lua/lua.h:35`) does share the domain, a count of stack values, and means the opposite of declining, "keep all results", so a handler returning it would read as "push everything"; `LUA_TNONE` (`lua/lua.h:62`) is a type tag from `lua_type`, so `nargs == LUA_TNONE` would compare a count against a tag; `LUA_REFNIL` and `LUA_NOREF` (`lua/lauxlib.h:100-101`) are `luaL_ref` registry references. Each would borrow a name whose meaning is something else.

Keeping it in the module's own namespace rather than the core's: `luanotifier_handler_t` is the only typedef in the tree whose return is a count of pushed values, and AGENTS.md puts a library's symbols under `LUA<LIBNAME>_`, so a `LUNATIK_ARGS_NONE` in `lunatik.h` would make core API out of a one-module constant. The tree's precedent for the shape itself is `LUNATIK_CPU_NONE`.

lneto: _Agent reply (Claude Code), posted from @lneto's account._

> why adding this between these two typedefs?

Right, the two function-pointer typedefs are a pair and nothing belongs between them. Moved above both, which is also where `lunatik.h` puts a constant relative to its readers (`LUNATIK_CPU_NONE` above `lunatik_getcpu` and `lunatik_ispercpu`, the `LUNATIK_OPT_*` above the `lunatik_is*` family):

```c
#define LUANOTIFIER_DECLINED	(-1)

typedef int (*luanotifier_register_t)(struct notifier_block *nb);
typedef int (*luanotifier_handler_t)(lua_State *L, void *data);
```

https://github.com/luainkernel/lunatik/commit/095bb44b3

lneto: _Agent reply (Claude Code), posted from @lneto's account._

> this comment seems unnecessary

Dropped. `LUNATIK_CPU_NONE` carries none either and its domain is just as implicit; the only thing the name does not say, that a handler otherwise returns a count, belongs to `luanotifier_handler_t` rather than to the constant, and both use sites already read as prose. https://github.com/luainkernel/lunatik/commit/095bb44b3

lib/luanotifier.c
lneto: _Agent reply (Claude Code), posted from @lneto's account._

> this comment seems unnecessary and it's not explaining what the macro actually does; it goes against our rules

Dropped, no comment at all. It stated what the call site concludes rather than what the macro tests, and what it tests is what the name says, so a comment naming the subject would only restate the name. Why a device of another namespace is declined is a contract for the script author and the doc block already carries it, "Only devices of the initial network namespace, the one `linux.ifindex` resolves a name in, are reported". https://github.com/luainkernel/lunatik/commit/ddc9a4ae2

lib/luanotifier.c
lneto: _Agent reply (Claude Code), posted from @lneto's account._

> the comment needs to explain what the macro actually does; also, reinforce our harness regarding this.

Rewritten to the test rather than its consequence:

```c
/* an event delivered inside register_fn, on the task that registered the block */
#define luanotifier_isreplay(notifier)	(in_task() && (notifier)->registrant == current)
```

It leaves the word "replay" out on purpose, since `module-conventions.sh` flags a comment that repeats an identifier used on the next line. What follows from the test, that the registering task holds the runtime lock or is still the body, is the reason `lunatik_handle` is safe there and stays in the commit body. https://github.com/luainkernel/lunatik/commit/095bb44b3

The harness: #896 gets the line in "Comments and documentation", right before the rule on an internal helper carrying no block comment, which it extends from "restates the code" to "describes the caller": a comment on a definition says what the definition is, not what its one caller concludes from it, and when the name already says the subject it takes no comment at all. A mechanical check was measured and held back, as with the `if` nudge last round: the tree has 166 function-like macros and exactly one carries a comment above it, `luarcu_mask`, which is a correct example that a token-overlap heuristic would flag, so the whole corpus is a counterexample and the annotation would train readers to ignore the check.

PR #802, opened 2026-09-07
thread: run the body with the objects given to thread.run

lneto:
`thread.run` had no way to give the body anything. The body is the function the script returns, so what it needed had to reach it inside a closure, which cannot cross between Lua states, or through values some earlier `resume` had left on the runtime's stack: a channel invisible at the call site, resting on an order of operations no contract states.

`thread.run(runtime, name, ...)` copies the Lunatik objects it is given onto the runtime under the runtime's lock, and the kthread resumes with them, so the body takes what it needs as arguments, over the transport `runtime:resume` already uses instead of a second, tacit one. Two defects on that path go with it: `lunatik_lcopyobjects` pushed one value per object into the `LUA_MINSTACK` slots a C function is entered with and nothing more, and a `thread.run` whose kthread failed to start dropped neither reference it had taken. Test: `run_args`.

This lands before #803, which stops `resume` leaving anything on that stack.

### DIFF
--- a/examples/echod/worker.lua
+++ b/examples/echod/worker.lua
@@ -23,18 +23,16 @@ local function echo(session)
 end
 
 local function worker(control, session)
-	return function ()
-		local id = control:getbyte(0)
+	local id = control:getbyte(0)
 
-		info(id, "started")
-		repeat
-			local ok, err = pcall(echo, session)
-				if not ok then
-				return info(id, "aborted")
-			end
-		until (not alive(control) or err or shouldstop())
-		info(id, "stopped")
-	end
+	info(id, "started")
+	repeat
+		local ok, err = pcall(echo, session)
+		if not ok then
+			return info(id, "aborted")
+		end
+	until (not alive(control) or err or shouldstop())
+	info(id, "stopped")
 end
 
 return worker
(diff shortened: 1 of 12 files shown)


lib/luathread.c
```diff
@@ -156,16 +159,52 @@ static const lunatik_class_t luathread_class = {
...
+		luaL_error(L, "couldn't pass the thread arguments");
+}
+
+static void luathread_popargs(lunatik_object_t *runtime, int nargs)
+{
+	lunatik_lock(runtime);
+	if (lunatik_isready(runtime))
+		lua_pop(lunatik_getstate(runtime), nargs);
+	lunatik_unlock(runtime);
+}
```
lneto, line 194: don't we have these patterns in other places? can't we use a template for this? any other suggestions on how to better modularize it?

examples/echod/worker.lua
```diff
@@ -23,18 +23,16 @@ local function echo(session)
...
+	local id = control:getbyte(0)
+
+	info(id, "started")
+	repeat
+		local ok, err = pcall(echo, session)
+		if not ok then
+			return info(id, "aborted")
+		end
+	until (not alive(control) or err or shouldstop())
+	info(id, "stopped")
```
lneto, line 35: do we need this PR only for changing this pattern? can you explain why is it better in the new way?

PR #803, opened 2026-09-07
lunatik: resume returns the objects the resumed script left

lneto:
`runtime:resume` promises a vararg of what the resumed script yields, and `<` binds tighter than `=`, so the count it copied back was the result of a comparison and zero on every success: the copy-back read nothing and resume returned nothing. The index it read from was wrong as well, naming the caller's argument count while indexing the resumed state.

The count is now the one `lua_resume` left and the values are the top of that stack, where a return and a yield both leave them, so the contract is what the code does; only Lunatik objects cross, as they already do on the way in. A runtime whose body has returned is dead to the next resume now that nothing of it is left on the stack. Test: `resume_results`.

Based on #802, whose `thread.run` arguments are how `echod` gets its worker body once resume stops leaving it on the stack. Supersedes #799, which wrote the defect down as the contract.

### DIFF
--- a/lunatik_core.c
+++ b/lunatik_core.c
@@ -128,10 +128,13 @@ static inline int lunatik_resume(lua_State *Lto, lua_State *Lfrom, int nargs)
 
 /***
 * Resumes a yielded runtime, analogous to `coroutine.resume`.
+* Only Lunatik objects cross between the runtimes, in either direction.
 * @function resume
-* @param ... values delivered to the script as return values of `coroutine.yield()`
-* @treturn vararg values passed to the next `coroutine.yield()`, or returned by the script
-* @raise if the runtime errors on resumption
+* @param ... objects delivered to the script as return values of `coroutine.yield()`
+* @treturn vararg objects passed to the next `coroutine.yield()`, or returned by the script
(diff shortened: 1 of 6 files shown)


lunatik_core.c
```diff
@@ -128,24 +128,27 @@ static inline int lunatik_resume(lua_State *Lto, lua_State *Lfrom, int nargs)
...
 
-	if (lunatik_copyobjects(Lto, L, 2, nargs) != LUA_OK || (nresults = lunatik_resume(Lto, L, nargs) < 0)) {
+	if (lunatik_copyobjects(Lto, L, 2, nargs) != LUA_OK || (nresults = lunatik_resume(Lto, L, nargs)) < 0) {
 		lua_pushfstring(L, "%s\n", lua_tostring(Lto, -1));
 		lua_pop(Lto, 1); /* error message */
 		lua_error(L);
 	}
 
-	int status = lunatik_copyobjects(L, Lto, nargs, nresults);
+	int status = lunatik_copyobjects(L, Lto, -nresults, nresults);
```
lneto, line 151: why `-nresults` here?

PR #809, opened 2026-09-08
lunatik: give lunatik_realloc the block's old size

lneto:
`lunatik_realloc` passed `LUA_TNONE` where `lua_Alloc` takes the block's old size, which is `SIZE_MAX` as a `size_t`, and the allocator reads that size twice: it returns the old pointer when `nsize <= osize`, so a failed reallocation on the kvmalloc/kmalloc branch returned the old pointer and passed `lunatik_checknull`, and it copies `min(osize, nsize)` bytes, so one that grows onto that branch reads past the end of the source block.

`data:resize()` carried both. A failed resize left `data->ptr` on the old buffer while `data->size` took the requested value, and `luadata_checkbounds()` then admitted every offset inside it. `LUA_TNONE` stays in `lunatik_malloc`, where `lua_Alloc` means it, since the pointer is NULL there.

`lunatik_alloc` also asks for `__GFP_NOWARN`: the sizes Lua asks it for come from the script and Lua reports the failure itself, so the `WARNING` the kernel raises first past `INT_MAX` or `MAX_PAGE_ORDER` only taints the log. It sits here rather than in #810 because `resize_atomic` forces a failure past `MAX_PAGE_ORDER` and would otherwise be the first run to leave one. Test: `resize_atomic`, which forces the failure on the branch that reads the old size.


lunatik.h
```diff
@@ -139,13 +139,13 @@ static inline const char *lunatik_pushstring(lua_State *L, char *s, size_t len)
 
 #define lunatik_pushoptinteger(L, cond, val)	((cond) ? lua_pushinteger((L), (val)) : lua_pushnil((L)))
 
-static inline void *lunatik_realloc(lua_State *L, void *ptr, size_t size)
+static inline void *lunatik_realloc(lua_State *L, void *ptr, size_t osize, size_t nsize)
```
lneto, line 144: is this change necessary because of realloc'ing with kvmalloc/kvfree?

lunatik_core.c
```diff
@@ -52,11 +52,12 @@ static void *lunatik_alloc(void *ud, void *optr, size_t osize, size_t nsize)
 
 	lunatik_object_t *runtime = (lunatik_object_t *)ud;
 	gfp_t gfp = lunatik_gfp(runtime);
+	gfp_t nowarn = gfp | __GFP_NOWARN; /* Lua raises on a NULL; a size a script chose is not the kernel's to warn about */
```
lneto: this comment can be simplified

lunatik_percpu.c
```diff
@@ -33,9 +33,10 @@ static lunatik_object_t *lunatik_finddata(lunatik_percpu_t *percpu, const lunati
 static lunatik_object_t *lunatik_newdata(lua_State *L, lunatik_object_t *object, const lunatik_class_t *class, size_t size)
 {
 	lunatik_percpu_t *percpu = lunatik_topercpu(object);
+	size_t osize = percpu->ndata * sizeof(lunatik_object_t *);
 	lunatik_object_t *data;
 
-	percpu->data = lunatik_checknull(L, lunatik_realloc(L, percpu->data, (percpu->ndata + 1) * sizeof(lunatik_object_t *)));
+	percpu->data = lunatik_checknull(L, lunatik_realloc(L, percpu->data, osize, osize + sizeof(lunatik_object_t *)));
```
lneto, line 39: can you explain this? why don't we need `ndata + 1` anymore?

PR #810, opened 2026-09-08
lib: bound the allocation sizes rcu, data, fifo and hid take from Lua

lneto:
A size a script chooses reaches the allocator, and the arithmetic on the way there is where it goes wrong: `roundup_pow_of_two(0)` is undefined, `luarcu_sizeoftable()` wraps above 2^60 buckets, `__kfifo_alloc()` truncates its size to an `unsigned int`, and `luaL_len()` honours a `__len` a script wrote.

The sizes `rcu.table()`, `data.new()`, `data:resize()`, `fifo.new()`, `fifo:pop()` and `hid.register()` take are bounded for their own reasons, and `check_dmesg` now fails a suite on a `WARNING` or a `UBSAN` line, not only on a Lua error. `data:resize(0)` goes with them, which also closes a double free: the allocator frees the block at size zero and the raise left the object holding the freed pointer. Tests: `rcu/bounds`, `data/bounds`, `fifo/bounds`, `hid/register`.

Based on #809.

PR #818, opened 2026-09-08
lib: hid owns its id_table before the walk that can raise

lneto:
`hid.register()` allocated the `id_table` and then walked a table the script owns. `lua_geti()` honours `__index`, and `lunatik_optinteger()` expands to `lua_getfield()`, which honours it too, so a metamethod written in Lua raises from inside the walk; the longjmp skips the `lunatik_free()` on the error path and the block stays allocated for the rest of the boot.

The table is handed to the driver before the walk, the way `luaset_build()` hands over its offsets, so `luahid_release()` frees it on every path out, and the explicit free and the NULL return go with it. `tests/hid` gains `idtable_leak`, which reads out of `SUnreclaim` the 64 MiB that 512 refusals of a full-length table strand, and `register` covers the entries the walk accepts and the ones it refuses.

Based on #810, which creates `tests/hid` and bounds the length this walk reads.

### DIFF
--- a/lib/luahid.c
+++ b/lib/luahid.c
@@ -79,24 +79,21 @@ static const lunatik_class_t luahid_class = {
 
 #define LUAHID_MAXIDS	(4096)
 
-static const struct hid_device_id *luahid_setidtable(lua_State *L, int idx)
+static void luahid_setidtable(lua_State *L, int idx, struct hid_driver *driver)
 {
-	size_t len = luaL_len(L, idx);
+	lunatik_checkfield(L, idx, "id_table", LUA_TTABLE);
+	size_t len = luaL_len(L, -1);
 
 	if (len > LUAHID_MAXIDS)
 		luaL_error(L, "'id_table' is too long");
 
-	struct hid_device_id *user_table = lunatik_checkalloc(L, sizeof(struct hid_device_id) * (len + 1));
-
-	struct hid_device_id *cur_id = user_table;
+	struct hid_device_id *id_table = lunatik_checkalloc(L, sizeof(struct hid_device_id) * (len + 1));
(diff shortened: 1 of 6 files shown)


lib/luahid.c
```diff
@@ -79,24 +79,20 @@ static const lunatik_class_t luahid_class = {
...
+
+	driver->id_table = id_table; /* own id_table before the walk can raise, so release frees it */
 	for (i = 0; i < len; i++, cur_id++) {
-		if (lua_geti(L, idx, i + 1) != LUA_TTABLE) { /* table entry */
-			lua_pop(L, 1); /* table entry */
-			lunatik_free(user_table);
-			user_table = NULL;
-			goto out;
-		}
+		luaL_argcheck(L, lua_geti(L, idx, i + 1) == LUA_TTABLE, 1, "invalid id_table"); /* table entry */
```
lneto: why hardcoding `1` here?

lneto: is it safe to raise an error here? no leaks?

lib/luahid.c
```diff
@@ -301,7 +295,7 @@ static int luahid_register(lua_State *L)
 	lunatik_setstring(L, 1, driver, name, NAME_MAX);
 
 	lunatik_checkfield(L, 1, "id_table", LUA_TTABLE);
-	luaL_argcheck(L, (driver->id_table = luahid_setidtable(L, -1)) != NULL, 1, "invalid id_table");
+	luahid_setidtable(L, -1, driver);
```
lneto: shouldn't we call it "checkidtable"? do our set's raise errors? shouldn't we pass the arg index here?

PR #820, opened 2026-09-08
thread: hold a reference to the task it may stop

lneto:
`luathread` kept the `task_struct` that `kthread_run` returned and called `kthread_stop` on it later without a reference of its own, which the kernel requires: `kthread_stop_put` documents it as "only when holding an extra task struct reference obtained by calling `get_task_struct()`", and `kthread_create` says a body may return on its own only for a thread nobody will stop, while `luathread_func` returns as soon as the runtime lock is free.

Between reading `thread->task` and calling `kthread_stop` on it, the body could return and the task be freed and recycled, so `kthread_stop` read `worker_private` off something that is no longer a kthread: the `WARN_ON` in `to_kthread` and then a NULL dereference. A `thread:stop()` on a thread that had just ended took this host down from inside a device write callback, with that exact pair in the log.

Holding the reference makes both halves safe, so the body no longer clears the pointer and an exited thread still reports its task. That is a contract change, and the doc block, the suite description and the exited-thread case say so.

Stacked on #825, which creates the task before waking it: the reference is taken in between, before the body can run.


lib/luathread.c
```diff
@@ -102,8 +100,9 @@ static int luathread_stop(lua_State *L)
 	else if (task != NULL) {
 		int result = kthread_stop(task);
```
lneto, line 101: _Agent review (Claude Code), posted from @lneto's account._ On the base this stacks on the window the body describes does not open. `stop` is a method of a `LUNATIK_OPT_MONITOR` class, so `lunatik_monitor` holds the object's mutex across the whole call (https://github.com/luainkernel/lunatik/blob/201cf500d757c46615381c092168dcbf7a26fa8a/lunatik_obj.c#L146), and `luathread_func` cannot pass its `lunatik_trylock(object)` loop (https://github.com/luainkernel/lunatik/blob/201cf500d757c46615381c092168dcbf7a26fa8a/lib/luathread.c#L55) while that mutex is held unless `KTHREAD_SHOULD_STOP` is set, which only `kthread_stop` does, below this line. Checked on the running module rather than by reading: `debug.getinfo(getmetatable(thread.current()).stop, "u").nups` is 2 and upvalue 2 is `"stop"`, so the method really is the monitor closure.

The window that is open, on master before #825, is a different one: `thread->task = kthread_run(...)` (https://github.com/luainkernel/lunatik/blob/924d472a9af0181ae1bfc1eb21d7d2b028011b01/lib/luathread.c#L230) stores after the wake inside the macro, so a body that returns at once can finish and be reaped before the store lands, and the store then puts a dead pointer back where `stop()` will read it later. That is `tests/thread/exit` exactly, and it matches the pair in the log. Not reproduced, since proving it means running the crash on this host.

None of that touches the fix: the requirement is real, quoted from `kthread_stop_put` (https://github.com/torvalds/linux/blob/v6.8/kernel/kthread.c#L737), and `to_kthread` (https://github.com/torvalds/linux/blob/v6.8/kernel/kthread.c#L76) is where it bites. It is only the account of the cause that wants softening, since the body states it as traced. Reword: https://github.com/luainkernel/lunatik/commit/852a2b66918c7aad5112c6abe6aa402b184444af

lib/luathread.c
```diff
@@ -119,7 +118,8 @@ static int luathread_stop(lua_State *L)
 
 /***
 * Returns a task object for the kernel task associated with the thread.
-* Once the thread has exited the object has no task and its methods raise.
+* The thread holds a reference to it, so the object stays readable after the
```
lneto, line 121: _Agent review (Claude Code), posted from @lneto's account._ `stop` puts the reference and clears the pointer three lines above, so `task()` after a stop returns an object with no task and its methods raise `null pointer dereference`. As it stands the block reads as though `task()` always reports the task, which is the half a reader will lean on. One line: https://github.com/luainkernel/lunatik/commit/3c8242cd800540a0ce70ccf6c1e3afa375c6dd5b

tests/thread/task_exited.lua
```diff
@@ -10,7 +10,6 @@ local lunatik = require("lunatik")
 local SCRIPT <const> = "tests/thread/exit"
 
 local t = lunatik._ENV.threads[SCRIPT]:task()
```
lneto, line 12: _Agent review (Claude Code), posted from @lneto's account._ The stop side of the change has no case. The body no longer clears the pointer, so `stop()` now runs `kthread_stop` on an exited task, which is the thing the reference is held for, and this rewrite also dropped the only assertion that a task object with no task raises. Both fit as a fourth case on the thread case 3 leaves exited, so the two differ only by the stop: case 3 reads its pid, case 4 finds none, and the stop itself sits under `mark_dmesg`/`check_dmesg`, where a `WARN_ON` from `to_kthread` would land. It passes three runs here and takes the suite to pass:164. https://github.com/luainkernel/lunatik/commit/8d256b369b1f6b0f48d19a9ac9fad8df2cc82aaf

lib/luathread.c
```diff
@@ -151,9 +151,19 @@ static const luaL_Reg luathread_mt[] = {
 };
 
 LUNATIK_OPENER(thread);
```
lneto, line 153: _Agent review (Claude Code), posted from @lneto's account._ The release landed between `LUNATIK_OPENER(thread)` and the class that names it, which splits the declaration from its use. Every module in the tree with a release defines it above the method tables (luatask.c:114, luahid.c:42, luafifo.c:75, luadata.c:183, luadevice.c:228, luaprobe.c:214, luanotifier.c:82, luarcu.c:196, luaset.c:87, luasocket.c:440). Moved: https://github.com/luainkernel/lunatik/commit/9bb0d408b6743e49ad0650314a7c46b33c9d300a

lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #820: request changes, two small folds; the fix itself is right.

@lneto The reference is the right fix and the branch is green here: the thread suite passes, the examples that spawn threads (`shared`, `lldpd`, `echod/daemon`, `cpuexporter`) run and stop clean, and five spawn-then-stop cycles on `tests/thread/exit` leave dmesg empty, which is the new path `kthread_stop` takes on an already-exited task. Holding the reference is what makes that safe: `worker_private` is freed only from `free_task` (https://github.com/torvalds/linux/blob/v6.8/kernel/fork.c#L615), `wake_up_process` on a dead task matches no state, and the completion has already fired.

Two things to fold, both inline: the `task` doc block, and the commit body's account of the window. Two more inline that do not gate: the missing case on the stop side, and where the release sits. Fixups are on `review/820`, one per finding.

One supporting note: `kthread_stop_put` is absent before 6.6, so its documentation is a citation rather than an API the 5.15 floor offers, and the `kthread_stop` plus `put_task_struct` pair here is exactly what it does.

lneto:
This merged into `claude_alloc_order`, not `master`: the base was still the branch this had been stacked on, so the commit landed there and `master` does not contain it. Reopened as #934, the same commit cherry-picked onto `master`, byte for byte this one.

PR #821, opened 2026-09-08
examples: arm the sentinel that stops the child before resuming it

lneto:
The finalizer that stops the child script of `ifquarantine` and `systrack` is the only thing that stops it, and it was installed after the resume that hands the child its state, so a resume that raises closes the control runtime with no sentinel registered and leaves the child running, owned by nothing.

Both examples now register the sentinel between `runner.run` and the resume. Checked on `systrack` with a resume forced to raise: the probe runtime is gone after the failed load, where before it stayed in `lunatik list` and only a manual `lunatik stop` removed it.


lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #821: it can merge as it stands.

@lneto Reproduced rather than read. With the sentinel after the resume, a resume forced to raise leaves `examples/systrack/probe` in `lunatik list` and only a `lunatik stop` by hand clears it; with it before, the list is empty after the same failed load. That was on scratch copies under `/lib/modules/lua`, removed afterwards, not on the examples themselves.

Both examples run on this host. `systrack` loads, `/dev/systrack` reports syscall counts, and stopping the control script takes the probe with it. `ifquarantine` through `tools/watchdog.sh` does what master does without #831: it quarantines `lo` and the uplink on the notifier replay, the watchdog stops it, connectivity comes back, and the list is empty afterwards, so the sentinel stopped the filter too. No kernel complaints from either. A raising resume was not forced on `ifquarantine`: its filter hook is registered before the resume, so the reproduction takes the host off the network for what is the same three lines as `systrack`.

Composes with #831: a real merge of the two branches is clean, and the merged `control.lua` keeps the sentinel between `runner.run` and the resume with #831's flag intact. Rebases onto current master without conflict.

PR #823, opened 2026-09-08
lib: thread names its task with the string, not with a format

lneto:
`thread.run()`'s second argument reached `kthread_run()` as the `namefmt` itself, so a name a script wrote made the kernel format arguments that were never passed: `__kthread_create_on_node()` hands it to `kvasprintf()` with an empty `va_list`, and `vsnprintf()` reads past the varargs and prints whatever the address holds. `thread.run(runtime, "%s%p thread")` names the task `(null)000000000`.

The name goes as an argument to a literal format. Test: `thread/name`, which reads the task's `comm` back through `thread:task()` and fails on the old call.


lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #823: it can merge as it stands.

@lneto Traced and green. `kthread_run`'s `namefmt` reaches `kvasprintf` with the `va_list` of the create call (https://github.com/torvalds/linux/blob/v6.8/kernel/kthread.c#L444), and what comes out of it becomes the task's `comm` through https://github.com/torvalds/linux/blob/v6.8/kernel/fork.c#L2357. On the 5.15 floor the same argument reaches `vsnprintf` into a `TASK_COMM_LEN` buffer before `set_task_comm` (https://github.com/torvalds/linux/blob/v5.15/kernel/kthread.c#L407), so a name carrying a conversion is formatted against varargs that were never passed across the whole supported range. Passing it as an argument to a literal is right.

Two notes, neither a change. `runner.spawn` builds its name with `string.match(script, "(%w*/*%w*)$")` (https://github.com/luainkernel/lunatik/blob/07ef16458eaae85cd86f475cb034683f7165c08f/lib/lunatik/runner.lua#L81), which cannot produce a `%`, so the bug is only reachable from a script calling `thread.run` directly, which is why the test cannot go through `runner.spawn` and does not. And the kernel builds with `-Wno-format-security`, which is why a non-literal format never showed at build time.

The A/B on `name` was not run: on the old call `%s` makes `vsnprintf` dereference whatever the vararg slot holds, and `check_pointer` only rejects NULL, sub-`PAGE_SIZE` and `ERR_PTR` values, so a fault there takes this host down; the `(null)000000000` in the description is the evidence, and the path was read rather than run. Everything else is green: thread suite 9/9 with `name`, full suite pass:162 fail:0 skip:3, and the branch rebases cleanly onto current master.

PR #825, opened 2026-09-08
lunatik: validate before acquiring, so a raise strands nothing

lneto:
Seven entry points acquired something and only then reached a raise, so the raise left the acquisition behind: `lunatik_newobject` loses the `lunatik_object_t` behind the userdata when `lunatik_checkclass` refuses the context, and again when `lunatik_setclass` finds no metatable; `crypto.shash()` and `crypto.comp()` strand a tfm holding a reference on the algorithm's module (sha2_ce's refcount went from 7 to 8 on one refused call here and stayed); `lunatik_loadfile()` leaves an open file and a page-sized buffer when pushing the chunk name raises; `hid.register()` stores a borrowed runtime pointer that `luahid_release()` then puts without anyone having taken it; and `thread.run()` takes its references before `kthread_run()` and has to put them back when it fails.

Each one validates before it acquires, or creates the object that owns the resource first. `thread.run()` creates the task, takes the references and only then wakes it, so the put pair on the failure path goes. The metatable lookup moves ahead of the allocation as `lunatik_checkmetatable`, while `lunatik_setclass` keeps its own raise for `lunatik_cloneobject` and `lunatik_lruntime`, which reach it without that check. Test: `crypto/context`, which counts what a leaked tfm holds, the reference on the module implementing the algorithm, and checks that a refusal leaves it where it found it, for `shash` and for `comp`.

#820 takes its task reference between the creation and the wake, so it stacks on this one.


lib/luacrypto_shash.c
```diff
@@ -221,10 +221,10 @@ const lunatik_class_t luacrypto_shash_class = {
 int luacrypto_shash_new(lua_State *L)
 {
 	const char *algname = luaL_checkstring(L, 1);
+	lunatik_object_t *object = lunatik_newobject(L, &luacrypto_shash_class, 0, LUNATIK_OPT_NONE);
 	struct crypto_shash *tfm = crypto_alloc_shash(algname, 0, 0);
 	size_t desc_size;
```
lneto, line 226: _Agent review (Claude Code), posted from @lneto's account._ Offered rather than asked for, since it does not gate. The commit lifted `object` into its declaration and left `desc_size` and `sdesc` as bare ones; with the `IS_ERR` check now above them, both can be initialized where they are declared, which is the C99 rule in AGENTS.md. https://github.com/luainkernel/lunatik/commit/1f7651082923ff1b200bb633a329361e61b074e1

lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #825: it can merge as it stands; one fixup is offered, not asked for.

@lneto All six traced against 6.8 and green here, and the crypto case measured rather than read: with the branch's order `tests/crypto/context` passes, and with master's order restored it fails naming `sha2_ce` and the count stays up. The test discriminates.

Three things worth recording. `lunatik_newobject`: one raise still sits after the allocation. `lunatik_setclass` raises `metatable not found` (https://github.com/luainkernel/lunatik/blob/7328b519b8b5c395cff9c1d4c6fe7189ed740cdf/lunatik.h#L238) after `*pobject = object` (https://github.com/luainkernel/lunatik/blob/7328b519b8b5c395cff9c1d4c6fe7189ed740cdf/lunatik_obj.c#L28) and before `lua_setmetatable`, so on that path the block is still lost. It is master's shape, and reachable only from a caller that skips `lunatik_require`, so not this branch's to fix; it is only the one case the title does not cover.

The thread commit undersells itself. On master `thread->task = kthread_run(...)` (https://github.com/luainkernel/lunatik/blob/924d472a9af0181ae1bfc1eb21d7d2b028011b01/lib/luathread.c#L230) stores after the wake inside the macro, so a body that returns at once can finish and be reaped before the store lands, and the store then puts a dead pointer where a later `stop()` reads it. Creating the task, taking the references and only then waking closes that by construction, which is the reason #820 stacks here and might be worth a line in the body.

Two `Checks` annotations to expect, neither needing action. `luacrypto_shash.c` "raises after acquiring" is the heuristic reading `lunatik_enomem` after `crypto_alloc_shash`; `crypto_free_shash` runs first, which is the commit. `lunatik_aux.c` gains "36 lines of body" (master sat at the threshold, 30); the growth is the error-path split itself and there is no step in it wanting a name. Its "comment right after a preprocessor branch" is master's, at the `#if BITS_PER_LONG == 32` block.

Green here: crypto 6/6 plus `context`, thread 9/9, hid 4/4, full suite pass:163 fail:0 skip:3, and the branch rebases onto current master without conflict.

PR #826, opened 2026-09-08
notifier: report the index and the namespace of the device an event is about

lneto:
Superseded by #797, which fixes the same defect the other way round. Kept open until #797 merges, as the record of the alternative that was weighed.

The defect: `notifier.netdevice` hands the callback a device name, and the chain is global, so a device created in another network namespace arrives under a name that means a different device in the initial one. Bringing up a namespace put the host's `lo` back into DROP in the `ifquarantine` example after an operator had allowed it. This branch keeps the global chain and appends `dev->ifindex` and `dev_net(dev)->ns.inum`, so a callback reads `(event, name, ifindex, netns)` and compares against a new `linux.netns()`, deciding for itself which devices it means.

What decided against it: nothing else in Lunatik can act outside the initial namespace, since `lualinux`, `luanetfilter`, `luasocket` and `luanetlink` all pass `&init_net`, so the generality this preserves is that of a script which can only print foreign names, while the cost is that the fix becomes opt-in and every script must remember it. The out-of-tree `dome/vlan.lua` would keep running unchanged and keep the bug. What was worth keeping here was carried over: the PID-1 ground-truth precondition into `netns_scope`, and the example's replay policy as #831.

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,7 +342,9 @@ sudo lunatik stop examples/ifquarantine/control    # stops both runtimes
 Pre-existing interfaces are covered as well: `register_netdevice_notifier`
 synchronously replays `NETDEV_REGISTER` (and `NETDEV_UP`) for each existing
 netdev when the notifier block is registered, so they enter quarantine at
-script start too.
+script start too. The replay spans every network namespace, and the example
+keeps the devices of the one its netfilter hook acts on, comparing the
+namespace each event carries against `linux.netns()`.
 
 ### filter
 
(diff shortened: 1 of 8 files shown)


lneto:

Superseded by #797, merged. The defect this reported, a device of another namespace reaching the callback under a name that resolves to a different device in the initial one, is closed there by declining such an event in the netdevice handler before anything reaches Lua, with `tests/notifier/netns_scope` covering the replay and the live events both ways. The alternative weighed here, keeping every namespace and appending `ifindex` and the namespace inode for each callback to compare, left the fix to every script and would not have let one act outside the initial namespace anyway, since every other binding passes `&init_net`. Closing.

PR #830, opened 2026-09-10
notifier: return NOTIFY_DONE when the runtime does not take the event

lneto:
A netdevice event reaching a Lunatik block whose runtime is not ready, during the script body past `notifier.netdevice()` or during teardown, was answered with `lunatik_run`'s `-ENXIO` as the verdict. Every negative errno has bit 15 set, which is `NOTIFY_STOP_MASK`, so `notifier_call_chain` stopped there and every notifier registered after Lunatik's block missed the event, while `notifier_to_errno` reported no error and the caller went on as if all had seen it.

`luanotifier_call` now returns `notify.DONE` for anything negative, and the module doc says so. Test: `chain_continues`, on the teardown window, which a finalizer that sleeps holds open: a second runtime registered after the first must see a device created while the first is being stopped, and the test checks the stop was still running when the device was created, so it cannot pass with the window already closed.

Independent of #797, which registers the block on the per-netns chain and would widen the stopped chain from the blocks after Lunatik's to every global notifier. Not run on a device this round: the tree builds clean and the test has not been executed.

### DIFF
--- a/lib/luanotifier.c
+++ b/lib/luanotifier.c
@@ -8,6 +8,8 @@
 * This library allows Lua scripts to register callback functions that are
 * invoked when specific kernel events occur, such as keyboard input,
 * network device status changes, or virtual terminal events.
+* A callback returns a `linux.notify` code; anything else, and an event that
+* reaches a runtime not ready to take it, counts as `notify.DONE`.
 *
 * @module notifier
 */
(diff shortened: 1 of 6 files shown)


lib/luanotifier.c
```diff
@@ -74,7 +76,7 @@ static int luanotifier_call(struct notifier_block *nb, unsigned long event, void
 	else
 		lunatik_run(notifier->runtime, luanotifier_handler, ret, notifier, event, data);
 
-	return ret;
+	return ret < 0 ? NOTIFY_DONE : ret; /* -ENXIO from lunatik_run has NOTIFY_STOP_MASK set */
```
lneto: why hardcoding 0? isn't it better to use an actual define here? also, it seems that the case of using min()

PR #831, opened 2026-09-10
examples: ifquarantine denies what appears after it loads, not what it found

lneto:
`register_netdevice_notifier_net` replays `NETDEV_REGISTER` for every device that already exists, so loading `examples/ifquarantine/control` quarantined `lo` and the uplink at once and its filter dropped everything on them: the machine went off the network, which has cost this host reboots.

The replay is delivered inside `notifier.netdevice`, and no live event reaches the callback before the script body ends (#797), so the script clears a flag once that call returns: the devices reported before it are recorded and left allowed, listed by `cat /dev/ifquarantine` and reachable with `deny=<name>`; only a REGISTER that arrives live is quarantined. The README paragraph and the usage order say so.

Stacked on #797; merges after it. Run through `tools/watchdog.sh` on 6.8.0-138: the host kept its connectivity, a veth created afterwards was reported and quarantined while `lo` and the uplink stayed ALLOW.


lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #831: it can merge as it stands, after #797.

@lneto Reads clean and matches what the machine shows. `record` leaves the replayed devices in `known` with no entry in `quarantined`, so they list as ALLOW and `deny=<name>` reaches them, which is what the README paragraph now says. A device replayed while up has its REGISTER recorded and its UP ignored, as on master, and a recorded device deleted and recreated goes back to being quarantined, since UNREGISTER clears `known[name]` before the live REGISTER arrives.

It composes with #821 rather than conflicting: the two touch `control.lua` at opposite ends, and `git merge-tree` over 924d472a9 merges them with no markers, leaving the sentinel armed before the resume and the replay still recorded.

Run through `tools/watchdog.sh` at 928e74b11: the replay recorded `lo` and `enp0s1` as present, the host kept its connectivity, a veth pair created afterwards was quarantined and listed by `/dev/ifquarantine`, and the stop left nothing behind.

lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #831, second round: it can merge as it stands, after #797.

@lneto Clean. `loading` is declared beside `known` and `quarantined` with the same trailing note, `callback` branches on it, and it is cleared on the line after the registration, which carries the one line that says why. The flag is exact for the reason #797's doc block gives, so a device that already exists is recorded and left allowed and only a live REGISTER is quarantined.

Run through `tools/watchdog.sh` at the previous head (the rebase changes nothing in the example): the replay recorded `lo` and `enp0s1` as present, the host kept its connectivity, a veth pair created afterwards was quarantined and listed by `/dev/ifquarantine`, and the stop left nothing behind. Approving on the stack, it merges after #797.

lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #831, third round: it can merge as it stands, after #797.

@lneto Rebased on the final #797; the example is unchanged and the README paragraph now names `register_netdevice_notifier`, the chain the binding keeps. Run through `tools/watchdog.sh` at this head: `lo` and `enp0s1` recorded as present, gateway reachable, a veth pair created afterwards quarantined, a dummy device created in a second namespace never reported, clean stop and dmesg.

examples/ifquarantine/control.lua
```diff
@@ -87,5 +97,6 @@ driver.sentinel = setmetatable({}, {__gc = function()
 	runner.stop(filter)
 end})
 
-notifier.netdevice(callback)
+notifier.netdevice(callback) -- replays a REGISTER for each device that already exists
```
lneto: why this comment here? it seems you are explaining the underlying API.. not good.. it needs to be fixed in the harness as well

lneto: _Agent reply (Claude Code), posted from @lneto's account._

> why this comment here? it seems you are explaining the underlying API.. not good.. it needs to be fixed in the harness as well

Right, and the same was true of the note beside the flag two screens up. Both are gone: `notifier.netdevice`'s own doc block says the registration delivers a `REGISTER` for each device the namespace already has, and this example's README paragraph says it again for the reader who is here for the policy. The script now reads `notifier.netdevice(callback)` and `loading = false`, with the flag declared bare. https://github.com/luainkernel/lunatik/commit/638a38bce

The harness: #907 puts it in "Comments and documentation", next to the rule on restating API usage, as a consumer does not document the API it calls, since a copy in the caller rots on the next change to the binding. No mechanical check there: the pattern is a trailing comment on a call, and every example has legitimate ones (`control:setbyte(1, 1) -- alive`, `pcall(log.push, log, char) -- drop silently if full`), so only the reader can tell whose behaviour is being described.

PR #837, opened 2026-09-11
probe: pass an argument closure to handlers

lneto:
A probe handler could only dump the registers to the log; nothing let it read a value the probed function was called with.

Handlers now receive a third value, `argument(n)`, returning the n-th argument of the probed function counting from zero, read with `regs_get_kernel_argument()`. Like `dump`, it raises once the handler that received it returns. It exists where the architecture selects `HAVE_FUNCTION_ARG_ACCESS_API`; elsewhere the handler receives `dump` alone.


lib/luaprobe.c
```diff
@@ -51,24 +80,26 @@ static int luaprobe_handler(lua_State *L, luaprobe_kprobe_t *kprobe, const char
 		pr_err_ratelimited("couldn't find probe table\n");
 		goto out;
 	}
+	int base = lua_gettop(L);
```
lneto, line 85: _Agent review comment, written by Claude Code at @lneto's request._

`int base;` was hoisted uninitialized above the `goto out`. AGENTS asks for C99 initialize-at-declaration, and the nearest sibling handler, [`luanetfilter_hook_cb`](https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/luanetfilter.c#L66), declares each local at first use with its initializer and jumps `out` over the later ones.

Folded as https://github.com/luainkernel/lunatik/commit/39587944bed1058c8632deebdcb7521637511059

lib/luaprobe.c
```diff
@@ -51,24 +80,26 @@ static int luaprobe_handler(lua_State *L, luaprobe_kprobe_t *kprobe, const char
...
+
+	luaprobe_pushregs(L, luaprobe_dump, regs);	/* base + 1 */
+	luaprobe_pushregs(L, luaprobe_argument, regs);	/* base + 2 */
```
lneto: _Agent review comment, written by Claude Code at @lneto's request._

Deferred, not folded, because the fix is master's code rather than this branch's.

`probe.new` installs both handlers on every kprobe, so a script that defines only `pre` still enters here for `post`: the registry lookup runs, the symbol and both closures are pushed under `GFP_ATOMIC`, and only then does `lunatik_optcfunction` resolve the handler to `lunatik_nop`. Master builds one closure per hit and this builds two, which is what brought it into view. Resolving the handler first and returning when it is the nop is the fix.

Raised as https://github.com/luainkernel/lunatik/issues/844. Nothing here timed it, so what I have is an extra allocation per hit, not a hot path.

lib/luaprobe.c
```diff
@@ -239,7 +270,13 @@ static int luaprobe_new(lua_State *L);
...
 * @tparam string|lightuserdata symbol kernel symbol name or address
 * @tparam table handlers table with optional `pre` and `post` callback functions;
-*   each receives the symbol (string or lightuserdata) and a `dump` closure
+*   each receives the symbol (string or lightuserdata), a `dump` closure and an
+*   `argument` closure, and both closures raise once the callback returns.
+*   `argument(n)` reads the n-th argument of the probed function, counting from zero,
+*   through `regs_get_kernel_argument()`, which guesses the register mapping: it raises
+*   where the architecture has no `CONFIG_HAVE_FUNCTION_ARG_ACCESS_API`, and otherwise
+*   returns whatever that accessor gives, which is not the argument past the registers
+*   the architecture passes arguments in, nor after a parameter 16 bytes or larger
```
lneto: _Agent review comment, written by Claude Code at @lneto's request._

`argument(n)` takes `lunatik_checkinteger(L, 1, 0, UINT_MAX)`, so an index past the argument registers the architecture defines is accepted, and what came back was undocumented.

Not a safety hole. On x86-64 an `n` at or above 6 goes to [`regs_get_kernel_stack_nth`](https://github.com/torvalds/linux/blob/v6.8/arch/x86/include/asm/ptrace.h#L319), which reaches the word through [`regs_within_kernel_stack`](https://github.com/torvalds/linux/blob/v6.8/arch/x86/include/asm/ptrace.h#L281), masking with `~(THREAD_SIZE - 1)` against `regs->sp`, and reads it with `copy_from_kernel_nofault`: only the trapped task's own stack block is reachable, and a bad address returns 0.

It was a contract hole. Past the registers the result is silently architecture defined: [arm64 returns 0](https://github.com/torvalds/linux/blob/v6.8/arch/arm64/include/asm/ptrace.h#L354-L361), [x86 returns a stack word](https://github.com/torvalds/linux/blob/v6.8/arch/x86/include/asm/ptrace.h#L346-L371). Measured here on 6.8.0-138 aarch64: `argument(12)` on `vfs_read` returned 0 with no error.

Documented rather than refused, in https://github.com/luainkernel/lunatik/commit/624f6e7c4b114ffa688819519ef17d54bb86f4a5. Refusing would need `NR_REG_ARGUMENTS`, which each architecture defines inside its own header body rather than exporting, and on x86 it would reject legitimate stack arguments.

The block was then taken further than I reported, in https://github.com/luainkernel/lunatik/commit/05fcb1490cc6060f378f0a3fcac72f39487a7a9e, to cover the second half of the same contract: `regs_get_kernel_argument()` opens by saying it picks the most likely register mapping, and that a parameter 16 bytes or larger shifts every argument after it. An index inside the register count can return a value that is not the argument either.

tests/probe/argument.sh
```diff
@@ -0,0 +1,64 @@
...
+# use, and both argument and dump stop reaching the registers once the handler
+# that received them returned.
+#
+# The probe is on vfs_read(file, buf, count, pos), so argument(2) is the byte count
+# the caller asked for. The shell reads exactly COUNT bytes from /dev/zero, a size
+# nothing else on the host is likely to ask for, and the script reports only when it
+# sees that size: another reader cannot make the case pass for the wrong reason. On
+# that same call it also checks that a negative index raises, then keeps both closures
+# for the next call to prove the handler that received them no longer reaches the
+# registers: argument raises, and dump is read through debug.getupvalue rather than
```
lneto, line 17: _Agent review comment, written by Claude Code at @lneto's request._

The `dump` drop had no test, and this commit is what restructured it. Master cleared one upvalue with `lua_setupvalue(L, -2, 1)`; here that becomes two indexed `luaprobe_dropregs` calls, and a wrong index fails silently, since `lua_setupvalue` on a non-function returns NULL and does not even pop. Only the `argument` half was asserted.

Folded as https://github.com/luainkernel/lunatik/commit/21fd617bccd28feedf5c1d40a4d19461cc88de4d: the stale case now checks `dump` as well, reading its upvalue with `debug.getupvalue` rather than calling it, so a build that stopped clearing it is reported instead of running `show_regs` on a `pt_regs` that is gone. `debug` is loaded in an IRQ runtime ([`luaL_openselectedlibs(L, ~LUA_IOLIBK, 0)`](https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lunatik_core.c#L228)) and [`lib/util.lua`](https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/util.lua#L47) already uses it. `tests/README.md` and both case texts moved in the same fixup.

lneto:
_Agent review comment, written by Claude Code at @lneto's request._

Reviewed at `05fcb149`, against its own base `claude_percpu_probe`. aarch64, kernel 6.8.0-138-generic, `CONFIG_HAVE_FUNCTION_ARG_ACCESS_API=y`. `make` clean with no new warnings; `sudo make install && sudo lunatik reload && sudo lunatik test` gives pass:156 fail:0 skip:0, probe suite 10/10 (percpu_probe 7, kprobe_concurrent 2, argument 3). Convention checks clean on the changeset.

Three findings, all folded on the branch; each is on its own line above. A fourth is deferred to #844.

What I drove to ground rather than read off the diff:

**Stack discipline, both paths.** `lunatik_handle` records `lua_gettop` before the call and `lua_settop` after, so the three values left on success and the four on the `lua_pcall` error path are all truncated. The two `luaprobe_dropregs` calls use absolute indices, so the error object `lua_pcall` pushes cannot shift them. Not taken on paper: a scratch probe whose handler raises uncaught showed the error logged by `pr_err_ratelimited` and, on the next hit, the saved closure raising `null pointer dereference`, so the drop does run on the error path.

**Lifetime.** Neither closure can read `regs` after the handler returns, on any path: a registry miss reaches `out` before a closure exists; success and pcall error nil both upvalues at absolute indices; a raise from `lunatik_optcfunction` or from the pushes leaves the closures only on the stack, where `lua_settop` drops them with nothing holding a reference. Building them fresh per hit is what buys that: a closure a handler stashes is dead permanently, not until the next hit. The registry reset shape `luanetfilter` uses for its `skb` removes the per hit allocation but would silently re-arm a stashed closure with another call's registers, so I left this as it is.

**The test discriminates.** I broke each mechanism and watched the matching case fail, then restored, with the full suite green after: an off-by-one in the index gives `not ok 1`; a bound widened to `INT_MIN` gives `not ok 2` with 1 still passing; the NULL-regs raise replaced by a push gives `not ok 3` with 1 and 2 still passing; removing `luaprobe_dropregs(L, base + 1)` gives `not ok 3`. For the third I disabled the raise rather than the drop, since removing the drop would have called `regs_get_kernel_argument` on a freed task stack, and proving a crash guard by removing it on this host is off limits.

**The `#ifdef`.** Right instrument, right place: it wraps the body, so the `#else` arm has no unused variables. I compiled that arm by forcing a bogus symbol and it builds with zero warnings. It is compile verified, not run, since I have one aarch64 host. Refusing at `probe.new` would reject scripts that never call `argument`, and a nil third value would surface as "attempt to call a nil value" instead of a named reason.

Cells the suite still does not cover, which I exercised by hand instead: in a `post` handler `argument(2)` returned the same value the `pre` handler saw; `argument("x")` and `argument()` both raise `bad argument #1 ... number expected`; `argument(12)` returned 0. I did not add a `post` case, because kprobe post-handler delivery differs between arm64 (breakpoint plus single-step) and x86 (`kprobe_ftrace_handler` with `FTRACE_OPS_FL_SAVE_REGS`), and I can run only one of them, so the case would be an architecture claim I cannot stand behind.

With the four fixups folded, this reads mergeable to me.

lib/luaprobe.c
```diff
@@ -39,6 +40,34 @@ static int luaprobe_dump(lua_State *L)
 	return 0;
 }
 
+static int luaprobe_argument(lua_State *L)
+{
+#ifdef CONFIG_HAVE_FUNCTION_ARG_ACCESS_API
```
lneto: isn't it better to register the method conditionally instead of giving this runtime error?

lib/luaprobe.c
```diff
@@ -39,6 +40,34 @@ static int luaprobe_dump(lua_State *L)
...
+#endif
+}
+
+static void luaprobe_pushregs(lua_State *L, lua_CFunction closure, struct pt_regs *regs)
+{
+	lua_pushlightuserdata(L, regs);
+	lua_pushcclosure(L, closure, 1);
+}
+
+static void luaprobe_dropregs(lua_State *L, int ix)
```
lneto: why not inline?

lib/luaprobe.c
```diff
@@ -39,6 +40,34 @@ static int luaprobe_dump(lua_State *L)
...
+{
+	lua_pushnil(L);
+	lua_setupvalue(L, ix, 1);
```
lneto, line 73: isn't this pattern used anywhere else?

lib/luaprobe.c
lneto: _Agent reply (Claude Code), posted from @lneto's account._

> isn't it better to register the method conditionally instead of giving this runtime error?

Yes, and it is what the tree does for `skb:connmark` under `CONFIG_NF_CONNTRACK_MARK` and `notifier.keyboard` under `CONFIG_VT`. `argument` is not an entry in a `luaL_Reg`, it is a closure the handler receives per hit, so the equivalent is the handler not receiving it: `luaprobe_argument` now exists only under `CONFIG_HAVE_FUNCTION_ARG_ACCESS_API`, the closures a handler gets come from a table with that entry gated, and the handler pushes, passes and drops the table in a loop. On a kernel without the API the handler receives `dump` alone and a script tests `if argument then`, as it does for `connmark`. The `@tparam` says so; `argument.sh` already skips there.

Folded in https://github.com/luainkernel/lunatik/commit/2d8005e91d2ff8dd2d6d7972ac0f7a6d733e67f9. Both arms build: the gated one is what the suite ran (probe 17/17, full pass:161 fail:0 skip:3); the other was compiled with the config undefined by hand and carries no `luaprobe_argument` symbol.

lib/luaprobe.c
lneto: _Agent reply (Claude Code), posted from @lneto's account._

> why not inline?

No reason: `luanetfilter_pushcb`, `_pushskb` and `_docall` are `static inline`, and these two are the same kind of helper. https://github.com/luainkernel/lunatik/commit/894ee063e77f597e80cb78f746613955c1142369

lib/luaprobe.c
lneto: _Agent reply (Claude Code), posted from @lneto's account._

> isn't this pattern used anywhere else?

No. Handing `regs` to a closure as its upvalue and clearing it once the handler returned exists only here: master's `dump` did it with `lua_setupvalue(L, -2, 1)`, and this branch generalizes it to every closure in the table. The only other `lua_pushcclosure` in the tree is `lunatik_monitor` in `lunatik_obj.c`, whose upvalues live as long as the metatable. The sibling shape, the registry reset `luanetfilter` uses for its `skb`, was considered and not taken: a closure a script stashed would come back armed with another call's registers, where clearing the upvalue leaves it dead for good. So there is nothing to consolidate yet; a second binding that hands `pt_regs` to a closure is what moves this into the core.

PR #838, opened 2026-09-11
examples: add dropreason, an skb drop monitor

lneto:
A dropped skb carries the reason the kernel dropped it, and nothing in the tree read it.

`linux.dropreason` names every core reason. `examples/dropreason` arms a kprobe on `kfree_skb_reason()`, counts the drops that reach it by name in an RCU table published on `lunatik._ENV`, and reads them live from the REPL; the first drop matching `WATCH` dumps its registers and call trace, which names the drop site. Reading the reason is what the `argument` closure is for.


autogen/specs.lua
```diff
@@ -78,6 +78,11 @@ return {
 	{ header = "uapi/linux/netfilter.h", prefix = "NF_", module = "nf.action",
 		desc = "Netfilter hook verdicts.",
 		include = { "DROP", "ACCEPT", "STOLEN", "QUEUE", "REPEAT", "STOP" } },
+	{ header = "linux/skbuff.h", prefix = "SKB_DROP_REASON_", module = "dropreason",
+		desc = "Socket buffer (skb) drop reasons, as `kfree_skb_reason()` receives them.",
+		exclude = { "SKB_DROP_REASON_MAX", "SKB_DROP_REASON_SUBSYS_" } },
```
lneto, line 85: _Agent review comment, written by Claude Code at @lneto's request._

`exclude = "SKB_DROP_REASON_MAX"` did not catch [`SKB_DROP_REASON_SUBSYS_MASK`](https://github.com/torvalds/linux/blob/v6.8/include/net/dropreason-core.h#L409), which is a member of `enum skb_drop_reason` itself, nor the [`SKB_DROP_REASON_SUBSYS_SHIFT`](https://github.com/torvalds/linux/blob/v6.8/include/net/dropreason-core.h#L412) define beside it. Generated on this kernel:

```
dropreason["SOCKET_RCVBUFF"] = 16
dropreason["SUBSYS_SHIFT"]   = 16
dropreason["SUBSYS_MASK"]    = -65536
```

`monitor.lua` builds `names[reason] = name` over that table, so `names[16]` was whichever key `pairs` reached last, reproducibly `SUBSYS_SHIFT` over three runs here. A receive buffer overflow, one of the commonest drops there is, got reported under the name of a bit shift constant. `SUBSYS_MASK` also sign extends to `-65536`, which the commit body's "plus the two sentinels" did not describe.

Folded as https://github.com/luainkernel/lunatik/commit/d7a956255ceabe73ef0c7db876b487b9b28612e4. Regenerated: 95 entries, no duplicate values, the `SUBSYS_*` pair gone, `CONSUMED = 1`, `NOT_DROPPED_YET = 0`, so the commit body's count is now true.

examples/dropreason/monitor.lua
```diff
@@ -0,0 +1,64 @@
+--
+-- SPDX-FileCopyrightText: (c) 2026 Ring Zero Desenvolvimento de Software LTDA
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+-- Answers "why is my packet dying?". A kprobe on kfree_skb_reason(), where every
+-- kfree_skb() lands, reads the reason off the probed function's arguments and
+-- counts the drops by name in an RCU table published on the shared environment,
+-- where any runtime, the REPL included, reads it live. A drop freed from hardirq
+-- (dev_kfree_skb_any()) or as a segment list (kfree_skb_list()) never reaches it
```
lneto, line 9: _Agent review comment, written by Claude Code at @lneto's request._

This header said `kfree_skb_reason()` is "the choke point every dropped skb goes through", and the README said the example "counts every skb drop". Two paths never reach it:

- [`dev_kfree_skb_any_reason()`](https://github.com/torvalds/linux/blob/v6.8/net/core/dev.c#L3163-L3169) takes `dev_kfree_skb_irq_reason()` when `in_hardirq() || irqs_disabled()`, which queues onto `softnet_data.completion_queue` for `net_tx_action` to free through `__kfree_skb`;
- [`kfree_skb_list_reason()`](https://github.com/torvalds/linux/blob/v6.8/net/core/skbuff.c#L1176-L1196) calls `__kfree_skb_reason()` and a bulk free directly.

For a tool whose pitch is "why is my packet dying?", a false "every" sends the reader hunting in the wrong place.

Folded as https://github.com/luainkernel/lunatik/commit/3361674602f4f9b6cafce74ad84020c3930b950f, in this header and in the README both.

examples/dropreason/report.lua
```diff
@@ -0,0 +1,38 @@
+--
+-- SPDX-FileCopyrightText: (c) 2026 Ring Zero Desenvolvimento de Software LTDA
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+-- Query API for the dropreason monitor (examples/dropreason/monitor.lua). Bind it
+-- in the REPL, read a count by reason name, or call report() for the full picture:
+--   > drops = require("examples.dropreason.report")
+--   > drops.NO_SOCKET
+--   1
+--   > drops.report()
```
lneto, line 10: _Agent review comment, written by Claude Code at @lneto's request._

The README and both `.lua` headers showed `> drops.NO_SOCKET` bare, then `> return drops.report()` on the next line, so the documented session taught a `return` that the REPL fix carried on this same branch had just made unnecessary.

Folded as https://github.com/luainkernel/lunatik/commit/8d20c847748e193de20ff91403e709cec68c2358

The fix that makes this work is no longer in this pull request: it is #841, off master. Without it, `drops.NO_SOCKET` typed on its own swallows the next line instead of evaluating.

examples/dropreason/report.lua
```diff
@@ -0,0 +1,38 @@
...
+local drops = {}
+
+local function track()
+	return lunatik._ENV.dropreason or error("the dropreason monitor is not running")
+end
+
+-- Every drop count, one reason per line, sorted by count.
+function drops.report()
+	local lines = {}
+	rcu.map(track(), function(reason, count)
```
lneto, line 24: _Agent review comment, written by Claude Code at @lneto's request._

Named because I raised it and it was deliberately not taken, so it does not go unsaid.

`rcu.map(track(), function(reason, count) ... end)` is a nested function definition, which AGENTS forbids outright. I did not ship a fixup: the only module level shape closes over a module level `lines` table, trading a one line closure for module level mutable state, and the tree's only other `rcu.map` caller, [`examples/systrack/device.lua`](https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/examples/systrack/device.lua#L23), is identical. If the rule binds here it binds in both places in one commit, which is a tree wide change and not this feature's business.

examples/dropreason/report.lua
```diff
@@ -0,0 +1,38 @@
...
+	return table.concat(lines, "\n")
+end
+
+local counts = {}
+
+function counts.__index(_, reason)
+	return track()[reason]
+end
+
+return setmetatable(drops, counts)
```
lneto, line 37: _Agent review comment, written by Claude Code at @lneto's request._

The proxy was an anonymous `{__index = count}` with the metamethod as a separately named local. AGENTS names `bpf.map`'s `view` as the measure for a proxy, and [`lib/bpf/map.lua`](https://github.com/luainkernel/lunatik/blob/c2ae028103c82712129dbd27a5bd43c1c013e67d/lib/bpf/map.lua#L154-L156) names the metatable, declares `function view.__index(proxy, key)` in place, and passes it to `setmetatable`.

Folded as https://github.com/luainkernel/lunatik/commit/afe91b086a633fcddcb6129754714c4899cb9d2f

lneto:
_Agent review comment, written by Claude Code at @lneto's request._

Reviewed at `afe91b08`, rebased onto the fixed #837. aarch64, kernel 6.8.0-138-generic. `make` clean with no new warnings; `sudo make install && sudo lunatik reload && sudo lunatik test` gives pass:156 fail:0 skip:0. Four findings, all folded on the branch; each is on its own line above.

**The REPL fix is no longer in this pull request.** It is a user visible parse bug that was gated behind the whole percpu stack for no reason, so it left for #841, rooted on master. The usage this PR documents needs it: without #841, `drops.NO_SOCKET` typed on its own swallows the next line instead of evaluating. Anyone reading this branch for that fix should look there.

I ran the example end to end four times, once per state of the branch: armed through `tools/watchdog.sh`, `echo x > /dev/udp/127.0.0.1/9999`, read from the REPL, stopped, confirmed teardown.

```
> drops = require("examples.dropreason.report")
> drops.NO_SOCKET
1
> drops.report()
      1  NO_SOCKET
     12  NOT_SPECIFIED
```

dmesg carried `dropreason: caught NO_SOCKET` with `x1 : 0000000000000003`, which is `SKB_DROP_REASON_NO_SOCKET`, and a call trace naming the drop site (`kfree_skb_reason` from `udp_rcv` from `ip_protocol_deliver_rcu`, up to `net_rx_action`). No WARNING, BUG or Oops on any run. After `lunatik stop`, `lunatik._ENV.dropreason` reads nil, so the `__gc` sentinel fires while the env object is still live. Known reason gives a count, unknown gives nil, monitor stopped gives the named error.

Two things I ran rather than reasoned about:

- `linux/skbuff.h` is the right header for 5.15 and up: the enum moved to `net/dropreason-core.h` in 6.4 and `skbuff.h` includes it, so the spec reaches it wherever it lives. I checked the "empty table instead of a build error" claim by pointing both specs at prefixes matching nothing and rebuilding: clean build, and the generated file is a two line empty table, because `extract.write` emits the module record outside the candidates loop.
- The `__gc` sentinel is the tree's idiom, not a trick. `examples/systrack/device.lua`, `examples/ifquarantine/control.lua`, `examples/spyglass/device.lua` and `tests/notifier/chain_continues_held.lua` all do it, and this is the only one that names the finalizer instead of inlining an anonymous function. Ordering is sound: the sentinel's metatable is set in the script body, after `lunatik._ENV` was pushed during `lunatik_runscript`, so it runs while the env object is still live.

No automated test. Not one of the roughly twenty examples in the tree has one, so this matches the tree rather than regressing it, but it does mean the run above is the evidence rather than a build. The CLI suite that would have caught the REPL bug is #845.

One thing worth knowing before merging, which does not gate this PR: the monitor runs `hardirq` and writes an `rcu.table`, which takes `spin_lock_bh` whatever context its runtime is in. That is #843. The mechanism is traced from source; reachability is not, since no run here produced the warning, and `examples/systrack/probe.lua` on master is the same shape.

With the four fixups folded, and #841 merged so the documented usage works, this reads mergeable to me.

PR #839, opened 2026-09-11
lunatik: name the helpers as the tree does, and write the rule down

### LINKED ISSUE #835: lunatik: homogenize the helper name space, extracting templates where the family repeats
The review of #795 renamed one binding's family of static helpers to match what the rest of the tree calls the same things: `matches` became `match`, `findkprobe` became `find`, `newkprobe` became `register`, and `freekprobe`, `sharekprobe`, `ownkprobe` lost the repeated noun. That settled one file by grepping for its peers, which is what `AGENTS.md` asks for, but the peers themselves are not homogeneous, and nothing in the tree writes the rule down.

A census of the verbs as they stand:

| verb | spellings in the tree |
|---|---|
| find | `luadevice_find`, `luaprobe_find`, `luaset_find`, `lunatik_finddata`, but `luanetfilter_findhook` |
| free | `luaprobe_free`, `luarcu_free`, `lunatik_free`, but `luanetfilter_freehook`, `luadarken_freerequest` |
| new | `luadata_new`, `luadevice_new`, `luafifo_new`, `luaprobe_new`, but `luanetfilter_newhook`, `luarcu_newentry`, `luarcu_newtable` |
| share, own | `luaprobe_share`, `luaprobe_own`, but `luanetfilter_sharehook`, `luanetfilter_ownhook` |

So the same function carries the noun in one binding and drops it in another, and a reader cannot tell from a name whether a suffix means something. Two things follow, and the second is the one that lasts:

**Homogenize the space.** Decide when the noun belongs — it does when a file has two of a kind, as `lunatik_finddata` does beside the runtime, and it does not when it merely repeats the module — and sweep the bindings to match. This is a rename across `luanetfilter`, `luarcu`, `luadarken` and whatever the sweep turns up, so it is a commit of its own and touches no behaviour.

**Extract the template wherever the family is the same code.** This is already happening one piece at a time: #795 added `LUNATIK_PERCPUDATA` for the list teardown that `luanetfilter` and `luaprobe` both write, and used it in `luaprobe` only, so converting `luanetfilter` to it belongs to this sweep; #834 costs the larger `lunatik_shared_t` that would absorb the rest of that family. Every such extraction removes a naming decision instead of settling it, which is the cheaper fix, so the sweep should look for the template first and rename only what is left.

**Write the pattern down.** `AGENTS.md` says a name matches what the tree already calls the same thing and that the answer is found by grepping for the peer, which works only while the peers agree. The outcome of this work is a short paragraph in *C style* naming the verbs and when a noun joins them, so the next binding is written right rather than corrected in review.

Raised from the naming thread of #795: https://github.com/luainkernel/lunatik/pull/795#discussion_r3988435195

lneto:
Closes #835. The same helper carried the noun in one binding and dropped it in another, so a name could not be settled by grepping for its peer, which is what `AGENTS.md` asks for.

`luanetfilter`'s shared-hook family becomes `find`, `register`, `free`, `share` and `own`, and `luasocket`'s allocator becomes `luasocket_new`; both Lua bindings take the `l` prefix reserved for the collision that creates, as `luadata_lnew` does. The list teardown `luanetfilter` wrote by hand is what `LUNATIK_PERCPUDATA` emits, so it uses the macro. `AGENTS.md` records when the noun joins the verb, and the other names #835's census listed, `luarcu_newentry`, `luacrypto_*_newrequest` and `luadarken_freerequest`, keep theirs under that rule.

Nothing Lua sees changes: `netfilter.register` and `socket.new` are the same names, every renamed symbol is `static`, and the out-of-tree consumers `tools/checks/consumers.sh` reports need no change.


AGENTS.md
```diff
@@ -293,6 +293,13 @@ afterwards) is used by `lib/luanetfilter.c` for its `skb`. Follow it rather than
 * A name matches what the tree already calls the same thing; grep for it before choosing. A Lua stack
   index is `ix`, a callback `cb`. Importing `arg` or `callback_ref` where the base settled on a
   shorter word is a deviation.
+* A helper's name is the verb the tree already uses for that step, `match`, `find`, `register`,
+  `free`, `share`, `own`, `attach` and `detach` being the ones in place. The noun joins the verb
+  when that verb has more than one kind to act on — `luarcu_newentry` beside `luarcu_newtable`,
+  `luacrypto_aead_newrequest` beside `luacrypto_aead_new` — or when the mirror it pairs with
+  carries one, as `luadarken_freerequest` does beside `luadarken_setrequest`; it does not when it
+  only repeats what the type already says, which is what `luanetfilter_findhook` did in a file
+  whose only list holds hooks.
```
lneto, line 312: Agent review (Claude Code), posted from @lneto's account.

The rule first landed as "the noun joins the verb only when the file has two kinds to tell apart", with `lunatik_finddata` and `luarcu_newentry` as its examples. Neither example instantiated it. `lib/luadarken.c` holds exactly one `free`, and `lunatik_finddata` is the only `lunatik_find*` in the tree (`grep -oE 'lunatik_find[a-z]*'` over the whole tree returns it alone), so the rule as written condemned a name the branch deliberately keeps, and cited an example that does not stand it up. What actually keeps `luadarken_freerequest`'s noun is the mirror rule already in this file: its partner `luadarken_setrequest` must carry the noun, because `luadarken_setkey` shares its verb.

Folded in https://github.com/luainkernel/lunatik/commit/40d33012ae1a7abb68c16ab8a3986e9263cb9837 . The unsupported example is replaced by `luacrypto_aead_newrequest` beside `luacrypto_aead_new`, which is a true instance (`LUACRYPTO_NEWCTX(aead, ...)` at `lib/luacrypto_aead.c:230` generates the sibling in the same file), and the mirror clause is what now covers `luadarken_freerequest`. The paragraph answers exactly the three names #835's census listed. The PR body's earlier claim that those names "each tell two kinds apart inside one file" was false for `luadarken` for the same reason, and the rewritten body no longer says it.

AGENTS.md
```diff
@@ -293,6 +293,13 @@ afterwards) is used by `lib/luanetfilter.c` for its `skb`. Follow it rather than
 * A name matches what the tree already calls the same thing; grep for it before choosing. A Lua stack
   index is `ix`, a callback `cb`. Importing `arg` or `callback_ref` where the base settled on a
   shorter word is a deviation.
+* A helper's name is the verb the tree already uses for that step, `match`, `find`, `register`,
```
lneto, line 306: Agent review (Claude Code), posted from @lneto's account.

One name in the tree this rule still does not cover, said here rather than left in a report: `lunatik_finddata`. It is the only `lunatik_find*` there is, so its verb has one kind to act on and no mirror carrying a noun, and by this paragraph it would be `lunatik_find`.

It is on master, in `lunatik_percpu.c`, so it is out of this branch's reach: a fixup touches only what the pull request introduced, and a rename would be its own branch. I dropped it as an *example* of the rule rather than endorsing the name, and left the name alone.

Which way it should go if anyone picks it up, plainly: a bare `lunatik_find` in a header the whole tree includes says less than the noun does. The honest reading is that the core prefix is the exception the rule does not reach, not that the rule is wrong.

lib/luasocket.c
```diff
@@ -519,12 +519,12 @@ static int luasocket_accept(lua_State *L)
 * @see linux.socket.ipproto
 * @within socket
 */
-static int luasocket_new(lua_State *L)
+static int luasocket_lnew(lua_State *L)
```
lneto, line 522: Agent review (Claude Code), posted from @lneto's account.

This rename arrived as a `fixup!` of the netfilter commit. A fixup's own subject is discarded at autosquash, so `lib/luasocket.c` would have landed under a `netfilter:` subject.

Folded by lifting it out of the fixup chain rather than by rewording anything: it is now its own commit, https://github.com/luainkernel/lunatik/commit/4b8e48ee4862e95f3a485adf77ef0e30ca263847 , "socket: name the socket allocator as the rest of the tree does", which autosquash cannot fold into the netfilter one. The title and body were widened to match, since the branch touches three files and the old title named only netfilter.

The change itself holds. `luasocket_newsocket` makes the one kind of object the module is named for, so the noun only repeats what the type says, and the binding that then collides with the plain name takes the `l` prefix. That is `luadata_new` and `luadata_lnew` exactly, and `luathread_new` is the same shape.

lib/luanetfilter.c
```diff
@@ -220,7 +206,7 @@ static const lunatik_class_t luanetfilter_class = {
 * @raise if the hook cannot be registered; in a percpu script, if this runtime already
 *   registered the same `pf`, `hooknum`, `priority` and `mark`, or if called after module load
 */
-static int luanetfilter_register(lua_State *L)
+static int luanetfilter_lregister(lua_State *L)
```
lneto, line 209: Agent review (Claude Code), posted from @lneto's account.

The `l` is correct as this branch stands, and only while #840 is held. Renaming the C helper `newhook` to `register` is what creates the collision, and `AGENTS.md` reserves the `l` prefix for exactly that, with `luadata_new` and `luadata_lnew` as the precedent.

It stops being correct the moment #840 lands. That branch deletes the C helper `luanetfilter_register`, leaving the binding with nothing to disambiguate from, and renames it back in https://github.com/luainkernel/lunatik/commit/c4bc737fb19e73528b6a3de0db48892f3b83f7ea .

Worth knowing before deciding the order of the two: master already spells the Lua binding `luanetfilter_register` (base `lib/luanetfilter.c:223`). So net of #840, exactly one of this commit's renames survives, `freehook` to `free`. `find`, `register`, `share` and `own` are renamed here and then deleted or renamed back there, which is a rename and a rename back of a name master already had right. If #840 is held, this commit stands whole. If #840 lands, this one is better reduced to the single surviving rename than landed and unwound one commit later.

lneto:
Agent review (Claude Code), posted from @lneto's account. Verdict on #839: it can merge as it stands.

#835 is right that the tree disagreed with itself, and this branch settles it. The four findings I raised are folded; each is on the line it applies to.

What I checked:

**The sweep is complete.** I enumerated every `(lua|lunatik)[a-z0-9_]*_<verb><noun>` definition and `#define` across `lunatik*.{c,h}` and `lib/*.{c,h}`, 444 names. The verb-plus-noun survivors are `luarcu_newentry`, `luacrypto_aead_newrequest`, `luacrypto_skcipher_newrequest`, `luadarken_freerequest`, `lunatik_newdata`, `lunatik_stopdata` and `lunatik_finddata`. Every one but the last is covered by the paragraph this branch lands; `lunatik_finddata` is on master and out of reach, which is the note on `AGENTS.md` above.

**It is pure naming.** `size(1)` on `luanetfilter.o`, `luaprobe.o` and `lunatik_percpu.o` is byte-identical between the base and this branch: 4929 / 5879 / 3947 text, 332 / 276 / 348 data. That is also the proof the second commit's claim needs, that `LUNATIK_PERCPUDATA` emits exactly the teardown `luanetfilter` wrote by hand.

**Nothing Lua sees changes, and the out-of-tree consumers are unaffected.** `tools/checks/consumers.sh` reports two consumers of netfilter and two of socket. The netfilter ones call `netfilter.register(handler)` and the lib table is still `{"register", ...}`; `luasocket_lib` is still `{"new", ...}`; every renamed symbol is `static`.

**Suite.** `make` (clean, no new warnings), `sudo make install` with the installed `.ko` verified against the built one, `sudo lunatik reload`, `sudo lunatik test`: pass 153, fail 0, skip 0. Operations serialized, no two overlapping. That run was on the head before the last rewrite; the only difference between the two trees is the `AGENTS.md` paragraph, no C or test file changed, so the result carries.

One thing I did not settle, and it is a call rather than a defect: whether the mirror clause is the rule wanted, or `luadarken_freerequest` should simply become `luadarken_free` and accept the asymmetry with `luadarken_setrequest`. I took the reading that preserves the pair, because `AGENTS.md` already protects mirrors.

Merge order: #840 is stacked on this branch and carries a recommendation to hold. If it is held, #839 stands alone with nothing to change.

PR #840, opened 2026-09-11
lunatik: one shared registration for the runtimes of a percpu set

### LINKED ISSUE #834: lunatik: one shared registration for the runtimes of a percpu set, generalized
`luaprobe` and `luanetfilter` now solve the same problem with the same shape: one registration owned by a percpu object and shared by the runtimes of the set, found by a match on the target, armed once and disarmed on release.

After #795 the two files carry the same five functions — find, new, free, share and own — plus the list teardown and its class, which `LUNATIK_PERCPUDATA` generates for `luaprobe` while `luanetfilter` still writes them out, its conversion being part of #835. They differ in three points: the match predicate, the register call and the unregister call. The generalization would be a header the registration embeds,

```c
typedef struct lunatik_shared_s {
	struct hlist_node node;
	lunatik_object_t *runtime;
} lunatik_shared_t;
```

plus a `lunatik_share(L, percpu, class, size, spec, match, arm)`; the release is the one `LUNATIK_PERCPUDATA` already generates.

What it costs, which is why #795 did not land it: each binding still needs its own error path around its register call, so the core absorbs roughly fifteen lines per binding for a header type, three function pointers and an indirect call per registration. It also has to change `luanetfilter`, so it is a commit of its own. And with two consumers it is not yet clear that this callback set is the right one — a third case is what would say.

So the decision is whether to wait for a third binding that shares a registration, or to fold the two now. Deferred from the review of #795: https://github.com/luainkernel/lunatik/pull/795#discussion_r3980598197

lneto:
Closes #834, with a recommendation to hold it. `luaprobe` and `luanetfilter` solved the same problem twice: one kernel registration owned by a percpu object and shared by the runtimes of the set, found by a match on the target, armed once and freed on release. A registration now begins with a `lunatik_shared_t`, each binding gives a `lunatik_sharing_t` naming the three points where they differ, and `find`, `share` and `own` move into the core.

The issue asked whether to fold the two consumers now or wait for a third, and the measurement answers it: on aarch64 the bindings lose 42 source lines and 584 bytes of text for 73 lines and 980 bytes in the core, so it breaks even at the fourth binding that shares a registration, not the third. It also trades two typed predicates for `lunatik_shared_t *` callbacks whose first-member cast and `.size` field the compiler cannot check, and `arm` frees an allocation the core made, which is a hazard for whoever writes the third consumer.

Landing it anyway needs two things this branch does not do. #839's `c2966501` reduces to the one rename that survives here, `freehook` to `free`, because the rest of it is undone by this commit; and `arm` should return `-errno` and release only what it took, leaving the free to the core, which changes the message `probe.new` raises and the test that asserts it.

### DIFF
--- a/doc/capi.md
+++ b/doc/capi.md
@@ -223,9 +223,9 @@ or `NULL`.
 ```
 Defines `static const lunatik_class_t prefix_class`, named `cname`, for the data a `percpu` object
 holds for the registrations its runtimes share, as [`lunatik_percpudata`](#lunatik_percpudata)
-creates it: a private that is a `struct hlist_head` of `T` entries linked through their
-`struct hlist_node node`, whose `release`, `prefix_release`, unlinks every entry and passes it to
-`free`. For example,
+creates it: a private that is a `struct hlist_head` of `T` entries, each beginning with the
+[`lunatik_shared_t`](#lunatik_share) that links them, whose `release`, `prefix_release`, unlinks
+every entry and passes it to `free`. For example,
 `LUNATIK_PERCPUDATA(luaprobe_kprobes, "probe.kprobes", luaprobe_kprobe_t, luaprobe_free)`
 defines `luaprobe_kprobes_class`.
 
--- a/doc/capi.md
+++ b/doc/capi.md
@@ -229,6 +229,39 @@ creates it: a private that is a `struct hlist_head` of `T` entries linked throug
 `LUNATIK_PERCPUDATA(luaprobe_kprobes, "probe.kprobes", luaprobe_kprobe_t, luaprobe_free)`
 defines `luaprobe_kprobes_class`.
 
+### lunatik\_share
+```C
+typedef struct lunatik_shared_s {
+	struct hlist_node node;
+	lunatik_object_t *runtime;
+} lunatik_shared_t;
+
+lunatik_shared_t *lunatik_share(lua_State *L, lunatik_object_t *percpu, const lunatik_sharing_t *sharing,
+	const lunatik_shared_t *spec);
+lunatik_shared_t *lunatik_own(lua_State *L, lunatik_object_t *runtime, const lunatik_sharing_t *sharing,
+	const lunatik_shared_t *spec);
+```
+One registration the runtimes of a `percpu` set share: the first call installs it, the others find
+it and attach to it, and a callback dispatched through `shared.runtime` reaches the runtime of the
+CPU it fired on. A registration begins with a `lunatik_shared_t`, and `sharing` says how to handle
+it:
+
+```C
+typedef struct lunatik_sharing_s {
+	const lunatik_class_t *class;	/* percpu data holding the list of registrations */
+	size_t size;			/* of the registration, which begins with lunatik_shared_t */
+	lunatik_match_t match;		/* bool (*)(const lunatik_shared_t *, const lunatik_shared_t *) */
+	lunatik_arm_t arm;		/* void (*)(lua_State *, lunatik_shared_t *) */
+	const char *registered;		/* raised when this runtime already registered the same target */
+} lunatik_sharing_t;
+```
+
+`lunatik_share` copies `size` bytes of `spec`, sets the runtime and calls `arm`, which registers
+with the kernel and, if it cannot, frees the registration before raising. `lunatik_own` does the
+same for a plain runtime, which holds a reference of its own; a `percpu` object is held by its
+data instead. `class` is what [`LUNATIK_PERCPUDATA`](#lunatik_percpudata) defines, and its
+`release` frees the list.
+
 ---
 
 ## Object Lifecycle
--- a/doc/capi.md
+++ b/doc/capi.md
@@ -217,6 +217,18 @@ afterwards. Returns `NULL` on a plain runtime, which has no runtimes to share wi
 `lunatik_getpercpu(L)` tells the two apart, returning the `percpu` object owning the runtime `L`,
 or `NULL`.
 
+### LUNATIK\_PERCPUDATA
+```C
+#define LUNATIK_PERCPUDATA(prefix, cname, T, free)
+```
+Defines `static const lunatik_class_t prefix_class`, named `cname`, for the data a `percpu` object
+holds for the registrations its runtimes share, as [`lunatik_percpudata`](#lunatik_percpudata)
+creates it: a private that is a `struct hlist_head` of `T` entries linked through their
+`struct hlist_node node`, whose `release`, `prefix_release`, unlinks every entry and passes it to
+`free`. For example,
+`LUNATIK_PERCPUDATA(luaprobe_kprobes, "probe.kprobes", luaprobe_kprobe_t, luaprobe_free)`
+defines `luaprobe_kprobes_class`.
+
 ---
 
 ## Object Lifecycle
--- a/doc/capi.md
+++ b/doc/capi.md
@@ -238,8 +238,6 @@ typedef struct lunatik_shared_s {
 
 lunatik_shared_t *lunatik_share(lua_State *L, lunatik_object_t *percpu, const lunatik_sharing_t *sharing,
 	const lunatik_shared_t *spec);
-lunatik_shared_t *lunatik_own(lua_State *L, lunatik_object_t *runtime, const lunatik_sharing_t *sharing,
-	const lunatik_shared_t *spec);
 ```
 One registration the runtimes of a `percpu` set share: the first call installs it, the others find
 it and attach to it, and a callback dispatched through `shared.runtime` reaches the runtime of the
@@ -257,10 +255,19 @@ typedef struct lunatik_sharing_s {
 ```
 
 `lunatik_share` copies `size` bytes of `spec`, sets the runtime and calls `arm`, which registers
-with the kernel and, if it cannot, frees the registration before raising. `lunatik_own` does the
-same for a plain runtime, which holds a reference of its own; a `percpu` object is held by its
-data instead. `class` is what [`LUNATIK_PERCPUDATA`](#lunatik_percpudata) defines, and its
-`release` frees the list.
+with the kernel and, if it cannot, frees the registration before raising. `class` is what
+[`LUNATIK_PERCPUDATA`](#lunatik_percpudata) defines, and its `release` frees the list, which is how
+a shared registration is freed; the `percpu` object is held by that data.
+
+### lunatik\_own
+```C
+lunatik_shared_t *lunatik_own(lua_State *L, lunatik_object_t *runtime, const lunatik_sharing_t *sharing,
+	const lunatik_shared_t *spec);
+```
+The same registration for a plain runtime, which has none to share with, so only `size` and `arm`
+are read of `sharing` and nothing is linked into a list: the runtime is held here rather than by a
+`percpu` object's data, and the caller keeps the pointer, freeing it and releasing the runtime when
+its own object is released.
 
 ---
 
--- a/lib/luanetfilter.c
+++ b/lib/luanetfilter.c
@@ -110,7 +110,7 @@ static unsigned int luanetfilter_hook(void *priv, struct sk_buff *skb, const str
 	return luanetfilter_docall((luanetfilter_hook_t *)priv, skb);
 }
 
-static luanetfilter_hook_t *luanetfilter_findhook(struct hlist_head *hooks, const luanetfilter_hook_t *spec)
+static luanetfilter_hook_t *luanetfilter_find(struct hlist_head *hooks, const luanetfilter_hook_t *spec)
 {
 	luanetfilter_hook_t *hook;
 
@@ -122,7 +122,7 @@ static luanetfilter_hook_t *luanetfilter_findhook(struct hlist_head *hooks, cons
 	return NULL;
 }
 
-static luanetfilter_hook_t *luanetfilter_newhook(lua_State *L, lunatik_object_t *runtime, const luanetfilter_hook_t *spec)
+static luanetfilter_hook_t *luanetfilter_register(lua_State *L, lunatik_object_t *runtime, const luanetfilter_hook_t *spec)
 {
 	luanetfilter_hook_t *hook = lunatik_checkalloc(L, sizeof(luanetfilter_hook_t));
 	int ret;
@@ -138,13 +138,13 @@ static luanetfilter_hook_t *luanetfilter_newhook(lua_State *L, lunatik_object_t 
 	return hook;
 }
 
-static void luanetfilter_freehook(luanetfilter_hook_t *hook)
+static void luanetfilter_free(luanetfilter_hook_t *hook)
 {
 	nf_unregister_net_hook(&init_net, &hook->nfops);
 	lunatik_free(hook);
 }
 
-LUNATIK_PERCPUDATA(luanetfilter_hooks, "netfilter.hooks", luanetfilter_hook_t, luanetfilter_freehook);
+LUNATIK_PERCPUDATA(luanetfilter_hooks, "netfilter.hooks", luanetfilter_hook_t, luanetfilter_free);
 
 static void luanetfilter_checkspec(lua_State *L, int ix, luanetfilter_hook_t *spec)
 {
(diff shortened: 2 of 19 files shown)


doc/capi.md
```diff
@@ -223,12 +223,52 @@ or `NULL`.
 ```
 Defines `static const lunatik_class_t prefix_class`, named `cname`, for the data a `percpu` object
 holds for the registrations its runtimes share, as [`lunatik_percpudata`](#lunatik_percpudata)
-creates it: a private that is a `struct hlist_head` of `T` entries linked through their
-`struct hlist_node node`, whose `release`, `prefix_release`, unlinks every entry and passes it to
-`free`. For example,
+creates it: a private that is a `struct hlist_head` of `T` entries, each beginning with the
```
lneto, line 226: Agent review (Claude Code), posted from @lneto's account.

This paragraph described the macro as it was before the commit that ships it: "a `struct hlist_head` of `T` entries linked through their `struct hlist_node node`". After this commit `T` has no `node`. It begins with the `lunatik_shared_t` that carries the link, and that is what the macro iterates.

Folded in https://github.com/luainkernel/lunatik/commit/6f9512c3caefdb7da81af684c1deb4e691a1baca .

lunatik_percpu.h
```diff
@@ -52,15 +52,37 @@ extern const lunatik_class_t lunatik_percpu_class;
...
+} lunatik_shared_t;
+
+typedef bool (*lunatik_match_t)(const lunatik_shared_t *shared, const lunatik_shared_t *spec);
+typedef void (*lunatik_arm_t)(lua_State *L, lunatik_shared_t *shared);
+
+typedef struct lunatik_sharing_s {
+	const lunatik_class_t *class;	/* percpu data holding the list of registrations */
+	size_t size;			/* of the registration, which begins with lunatik_shared_t */
+	lunatik_match_t match;
+	lunatik_arm_t arm;		/* registers with the kernel; frees the registration before raising */
```
lneto, line 68: Agent review (Claude Code), posted from @lneto's account.

This comment and `doc/capi.md` stated two different contracts. The header said "frees what it took", which reads as what `arm` itself allocated; `capi.md` said `arm` frees the registration before raising. The core is the one that allocates it, through `lunatik_checkalloc` in `lunatik_newshared`, and never gets the pointer back on the raising path, so `capi.md` has the contract. A consumer following the header instead leaks the registration on every failed arm.

Folded in https://github.com/luainkernel/lunatik/commit/e9bf697dcf6510d1a86edeed34fc613c9b887954 ; the header now says what `capi.md` says.

Which way the contract itself ought to go is a separate point, on `lunatik_percpu.c` below.

lib/luanetfilter.c
```diff
@@ -217,15 +195,21 @@ static int luanetfilter_lregister(lua_State *L)
...
+
 	luaskb_attach(L, nf, skb);
 	lunatik_registerobject(L, 1, object);
 	lunatik_register(L, -1, hook); /* the callback finds this runtime's registration by the hook they share */
 	return 1;
 }
 
 static const luaL_Reg luanetfilter_lib[] = {
-	{"register", luanetfilter_lregister},
+	{"register", luanetfilter_register},
```
lneto, line 212: Agent review (Claude Code), posted from @lneto's account.

The binding arrived here as `luanetfilter_lregister`, carried up from #839, with nothing left to disambiguate from: this commit deletes the C helper `luanetfilter_register` that the `l` was standing aside for. `grep -c luanetfilter_register lib/luanetfilter.c` on the commit before the fix returns 0. `AGENTS.md`: without that collision, the plain name is the right one.

Folded in https://github.com/luainkernel/lunatik/commit/c4bc737fb19e73528b6a3de0db48892f3b83f7ea .

Which also means a squash of #839 and this branch leaves no trace of that rename at all, and is the second reason it is the right fix rather than leaving the `l` in place. See the note on `luanetfilter_lregister` in #839.

tests/probe/percpu_probe.sh
```diff
@@ -134,6 +139,22 @@ case "$listed" in
...
+	fail "a probe on a symbol the kernel does not have was accepted: $output"
+dmesg_since | grep -qF "$ARMED_ONE" || fail "the set did not arm a kprobe before the one the kernel refuses"
+listed=$(lunatik list)
+case "$listed" in
+	*"$UNKNOWN"*) fail "the refused run left the script registered: $listed" ;;
+esac
+if [ -n "$idle" ]; then
+	[ "$left" = "$idle" ] || fail "the refused run left $((left - idle)) kprobes armed"
+fi
+ktap_pass "a probe the kernel refuses is freed, and the set unwinds the one it had armed"
```
lneto, line 156: Agent review (Claude Code), posted from @lneto's account.

`arm` frees the registration the core allocated and then raises, so nothing must have linked it yet. That path had no test in a percpu set: only the plain runtime's own probe walked it, through `lunatik_own`, and `lunatik_share`'s side was uncovered. It is the exact path the contract note on `lunatik_percpu.c` warns about, which is why it needed a case rather than a remark.

Added in https://github.com/luainkernel/lunatik/commit/e27f189921697f9edb3179a4717bcb54089decd3 : the set arms one kprobe, the next target is one the kernel refuses, and the run must leave no kprobe armed, no script registered, and the already-armed one unwound.

It was proven to discriminate, both ways. Pointing the script at `sys_ni_syscall` instead of the unknown symbol gives `not ok 4 a probe on a symbol the kernel does not have was accepted`; biasing `idle` by one gives `not ok 4 the refused run left 1 kprobes armed`. Restored, green.

The netfilter counterpart of the same gap, a percpu set holding a second, distinct target, is master-side code and so out of this branch's scope; it is #842.

doc/capi.md
```diff
@@ -223,12 +223,52 @@ or `NULL`.
...
+	const char *registered;		/* raised when this runtime already registered the same target */
+} lunatik_sharing_t;
+```
+
+`lunatik_share` copies `size` bytes of `spec`, sets the runtime and calls `arm`, which registers
+with the kernel and, if it cannot, frees the registration before raising. `class` is what
+[`LUNATIK_PERCPUDATA`](#lunatik_percpudata) defines, and its `release` frees the list, which is how
+a shared registration is freed; the `percpu` object is held by that data.
+
+### lunatik\_own
```
lneto, line 262: Agent review (Claude Code), posted from @lneto's account.

`lunatik_own` was declared inside `lunatik_share`'s code block, with no heading of its own, so the index does not reach it and what it does differently was one clause of another function's paragraph. `doc/capi.md` gives each of its other 37 public functions a `###`, and splits pairs: `lunatik_attach` and `lunatik_detach` have one each, as do `lunatik_getobject` and `lunatik_putobject`.

Folded in https://github.com/luainkernel/lunatik/commit/d402d9e3e814c7f77e9618f9cbf0cc265efd83a7 .

lib/luaprobe.c
```diff
@@ -308,8 +278,12 @@ static int luaprobe_new(lua_State *L)
...
 	luaprobe_t *probe = (luaprobe_t *)object->private;
 
-	luaprobe_kprobe_t *kprobe = percpu != NULL ? luaprobe_share(L, percpu, &spec) :
-		luaprobe_own(L, probe, runtime, &spec);
+	luaprobe_kprobe_t *kprobe = (luaprobe_kprobe_t *)(percpu != NULL ?
+		lunatik_share(L, percpu, &luaprobe_sharing, &spec.shared) :
+		lunatik_own(L, runtime, &luaprobe_sharing, &spec.shared));
+
+	if (percpu == NULL)
+		probe->kprobe = kprobe;
```
lneto, line 286: Agent review (Claude Code), posted from @lneto's account.

This assignment and its netfilter twin each carried `/* the percpu object owns the shared one */`, while both structs already say it on the field: `lib/luaprobe.c:29` is `luaprobe_kprobe_t *kprobe; /* NULL when the percpu object owns the kprobe */` and `lib/luanetfilter.c:31` the same for `hook`. That is one reason in four places, which `AGENTS.md` settles: one non-obvious reason, one line, at the site that needs it, not repeated wherever the feature is touched. As placed, the comment also described the branch not taken, since it sat on the `percpu == NULL` arm.

Folded in https://github.com/luainkernel/lunatik/commit/7f350dcd8e7e821f170521d19d7f4baf45b3b064 ; the field comments stay and the two on the assignments go.

lunatik_percpu.c
```diff
@@ -59,6 +59,57 @@ lunatik_object_t *lunatik_percpudata(lua_State *L, const lunatik_class_t *class,
...
+}
+
+static lunatik_shared_t *lunatik_newshared(lua_State *L, lunatik_object_t *runtime,
+	const lunatik_sharing_t *sharing, const lunatik_shared_t *spec)
+{
+	lunatik_shared_t *shared = lunatik_checkalloc(L, sharing->size);
+
+	memcpy(shared, spec, sharing->size);
+	shared->runtime = runtime;
+	sharing->arm(L, shared);
```
lneto, line 81: Agent review (Claude Code), posted from @lneto's account. This one was not applied, deliberately; it is the maintainer's call and it is the main thing standing between this branch and merge if it is folded anyway.

`lunatik_newshared` calls `lunatik_checkalloc` and then hands the block to `sharing->arm`, which must free it before raising. So the allocator and the releaser are different functions in different modules, against the shared-helper rule `AGENTS.md` states: a pair mirrors, whatever one half acquires its partner releases.

It is not a live bug. Both consumers get it right, and on every failure path the registration is freed before it is linked, since `hlist_add_head` runs after `lunatik_newshared` returns. It is a hazard for whoever writes the third consumer, and in the worst shape, an `arm` that frees and then returns rather than raising, the core links freed memory.

The clean alternative is `typedef int (*lunatik_arm_t)(lua_State *, lunatik_shared_t *)`, with `arm` releasing only what it acquired and returning `-errno`, and the core doing `lunatik_free(shared)` then `lunatik_throw(L, ret)`. `nf_register_net_hook` already returns 0 or `-errno`, so netfilter is a one-line change.

I did not ship it because it changes `luaprobe`'s user-visible message from `failed to register probe (%d)` to the symbolic errno, and `tests/probe/percpu_probe_plain.lua:29` asserts `err:match("failed to register probe")`. That is a behaviour change this PR did not intend, and the error-message convention is a separate call. The PR body now carries it as one of the two things landing this branch would need.

lneto:
Agent review (Claude Code), posted from @lneto's account. Verdict on #840.

**The recommendation is to hold or close this branch, and wait for the third consumer.** The code is correct. The reason not to land it is not in its own code.

**1. The measurement answers #834's own question, and it answers it the other way.** #834 asked whether to fold the two consumers now or wait for a third. Squashed against its base, this branch is 42 source lines out of the bindings for 73 into the core (`git diff --numstat`: `lib/` +53/-95, `lunatik_percpu.{c,h}` +77/-4), and by `size(1)` on aarch64: `luanetfilter.o` text 4929 to 4645, `luaprobe.o` 5879 to 5579, `lunatik_percpu.o` 3947 to 4927. So 584 bytes out of the bindings for 980 into the core, net +396. Break-even is 980/292 = 3.4 consumers by text and 73/21 = 3.5 by lines, that is the fourth binding that shares a registration, not the third. At three it is still +10 lines and +104 bytes. At two, which is today, the tree carries +31 lines and +396 bytes for one copy of a nine-line refusal rule.

**2. What it trades that for.** Two typed predicates, `bool luaprobe_match(const luaprobe_kprobe_t *, const luaprobe_kprobe_t *)` and its netfilter twin, become `lunatik_shared_t *` callbacks with three first-member casts per binding, a `memcpy` sized by a struct field, and a contract where the callback frees the caller's allocation. None of that is checkable by the compiler. A consumer that puts `lunatik_shared_t` anywhere but first, or sets `.size` wrong, gets silent corruption with no diagnostic, which is a mistake the old shape could not express. #834 says it itself: with two consumers it is not yet clear that this callback set is the right one, a third case is what would say. Nothing in the branch answers that, and the measurement argues the other way.

**3. It undoes most of the branch directly below it.** Master already spells the Lua binding `luanetfilter_register`. #839 renames the C helper `newhook` to `register`, which creates the collision, and resolves it by pushing the binding to `lregister`; this branch deletes the helper and renames the binding back. Net of #840, exactly one of #839's netfilter renames survives: `freehook` to `free`. `AGENTS.md`: do not land an implementation you already intend to replace; and reshaping or renaming an API means updating every consumer, including consumers in stacked pull requests that will rebase onto the change.

**If it is folded anyway,** it should land with the six fixups on this branch, and with the two things the PR body now names: #839's netfilter commit reduced to the one rename that survives here, and `arm` returning `-errno` and releasing only what it took, which is the comment on `lunatik_percpu.c` and does change the message `probe.new` raises and the test that asserts it.

**The branch is otherwise sound.** Correctness driven to ground, all clean: the `memcpy` over a stack `spec` is safe because both callers use a designated initializer, so the `lunatik_shared_t` head is zero-initialised and `node` is `{NULL,NULL}` before `hlist_add_head`; every first-member cast is valid and consistent, including `lunatik_register(L, ..., hook)` against `lunatik_getregistry(L, shared)`, which are the same address; `lunatik_percpudata` can raise before any allocation; `arm` raising leaves nothing linked and no runtime reference taken; percpu teardown ordering is unchanged. Full suite: pass 154, fail 0, skip 0, over `make` (no new warnings), `sudo make install` with the installed `.ko` verified, `sudo lunatik reload`, `sudo lunatik test`, serialized. `examples/dnsblock/nf_dnsblock` was run end to end in both plain and percpu modes, and `examples/systrack/device` on the probe side. That run was on the head before the last rewrite; the only tree difference since is the `AGENTS.md` paragraph, no C or test file changed, so the result carries.

**One correction to my own earlier report,** for the record. I had written that the netfilter arm-failure case is untestable because every rejecting branch of `nf_hook_entry_head` is a `WARN_ON_ONCE` and `check_dmesg` would match it. That is wrong: `tests/lib.sh` greps Lua error lines only, nothing in it matches `WARNING`, `BUG` or `Oops`. The case is also constructible, since `luanetfilter_checkspec` validates neither `pf` nor `hooknum`, so any integer from Lua reaches `nf_register_net_hook`. The real objection is that the case taints the host with `TAINT_WARN`, which is a maintainer call, not an impossibility. The probe-side equivalent needed no such trick and is covered.

lunatik_percpu.h
```diff
@@ -51,6 +52,44 @@ extern const lunatik_class_t lunatik_percpu_class;
...
+	const char *registered;		/* raised when this runtime already registered the same target */
+} lunatik_sharing_t;
```
lneto, line 70: (posted by an agent, not by @lneto)

Since this branch was cut, master's `luaprobe_share` gained two steps this struct has no slot for (https://github.com/luainkernel/lunatik/blob/f705c2fdc4ddb67b23af3cd91ae770cbdc880863/lib/luaprobe.c#L349-L354): it refuses a target another runtime of the set registered with a different post handler (https://github.com/luainkernel/lunatik/commit/2c7de524dc252d2ba83a4f8857aa27e2a1b2760a), and it takes a reference per handle, since a handle's private is now the registration itself (https://github.com/luainkernel/lunatik/commit/af1827434ed1113d9bb0e4daf4cb698fc264a0ed). `luanetfilter_share` does neither. A port that keeps this struct as it is would keep both in `luaprobe_new`, after `lunatik_share` returns, so the refusal rule this struct centralizes would be split between the core and the binding, and the reference would stay in the binding alone.

#834 said a third case would show whether `match`, `arm` and `registered` are the right callback set; the second case has already changed shape. If the intent is still to generalize, it may be easier to design the struct from all three consumers once the third exists.

lneto (commented):

**Verdict: close without merging.** @lneto, the branch can no longer merge as it stands, and #834's condition for folding the two consumers is still unmet.

Its base, #839, merged on 2026-09-13 (https://github.com/luainkernel/lunatik/commit/a9f671a82cd3518ab48ddd12702835882c1d8873 through https://github.com/luainkernel/lunatik/commit/9fc5fcded67c777ad16b622d871a2f334099ed97), and this pull request still targets `claude_naming_sweep`. It was also cut from a version of #795 and #839 that was rewritten before either merged: 14 of its 21 commits, 7c1dadbc5a1939171ec74e0ffbd5def3032271f4 through 0baf77a4b371e4a946af5a1361c822c15e52b6be, are pre-squash copies of commits master carries under other SHAs, the probe one in another shape. The branch's own commit, 1b19b2394d71ba26cf9a39d437bcd231bede6fc2, cherry-picked onto master at f705c2fdc4ddb67b23af3cd91ae770cbdc880863, applies cleanly to the core and to netfilter and conflicts in nine hunks of `lib/luaprobe.c`. `lunatik_percpudata` still has the same two users, so the break-even the last round measured, the fourth consumer, is no nearer; what changed is that the two diverged, which the inline comment on `lunatik_sharing_t` describes.

The case this branch added, a probe the kernel refuses in a percpu set, covers a path master still leaves open, so it moves to master on its own in `claude_refused_registration`, with its netfilter mirror in a set and in a plain runtime. The last round left the netfilter case out, believing it needed a `WARN`; it does not, since `__nf_register_net_hook` returns `-EINVAL` for a netdev hook with no device before reaching one (https://github.com/torvalds/linux/blob/v7.2/net/netfilter/core.c#L397-L409), and `netfilter.register` never sets a device. #834 stays open for the third consumer.

Issue #657, opened 2026-08-03
Epic: filesystem notification and permission binding


lneto:
Lunatik can see packets but not files. This epic tracks a binding for the kernel's filesystem
notification subsystem, covering both regimes it offers: asynchronous notification, the way `inotify`
and `fanotify` expose it to userspace, and synchronous **permission events**, where the handler's
answer decides whether the access happens.

## Where we are

Nothing in the tree touches `fs/notify`. A sweep of every local and remote branch, and of the closed
issues and pull requests, found no `fsnotify`, `fanotify` or `inotify` code. The structurally similar
things that do exist are `notifier` (callback plus verdict on a kernel notifier chain), `netfilter`
(callback per packet, registry-cached object, verdict back) and `probe` (observes, cannot decide).

So a Lua script today cannot tell that a file was opened, let alone refuse the open.

## What we are building

One kernel module, `fsnotify` (`lib/luafsnotify.c`), owning a **watch** (an `fsnotify_group` plus its
Lua callback), a **mark** (on an inode, a mount or a superblock) and the **event** handed to the
callback.

The design rests on one kernel fact worth stating up front:

> Permission events are delivered synchronously, in the context of the process performing the access,
> and the return value of the group's `handle_event` is the answer: non zero denies. The syscall is
> parked inside your handler. It may sleep, which is why a process context runtime is correct, and it
> re-enters if the handler touches a watched path, which is why the reentrancy guard is part of the
> first phase rather than a later fix.

A kernel module creating its own group is the ordinary case, not a trick: `dnotify`, `audit` and
`nfsd` all do it, and `fsnotify_alloc_group` and the mark API are exported. Permission events need
`CONFIG_FANOTIFY_ACCESS_PERMISSIONS`, which distributions ship enabled.

This is deliberately **not** an LSM. It sees file access only, and it can refuse but never grant. The
rest of the security surface is the subject of the LSM epic.

## Documents

Everything a contributor needs is in the repository, under review in #672:

- `doc/design/fsnotify/plan.md`, current state, gap analysis, phases, non goals, risks
- `doc/design/fsnotify/api.md`, the proposed Lua API with worked examples
- `doc/design/fsnotify/kernel-notes.md`, verified kernel reference: exported symbols, ops, priorities, version drift
- `doc/design/fsnotify/testing.md`, test strategy and matrix
- `AGENTS.md`, repository conventions for contributors and AI assistants

## Phases

[...]

PR #846, opened 2026-09-12
docs: correct the fsnotify notes on version drift and the permission config

lneto:
Bottom of a ten pull request stack implementing fsnotify, #657. The nine above rest on these notes, so the three statements they make that are wrong about the kernel are corrected first.


lneto (commented):

Verdict: merge as it stands; bottom of the stack, and the nine above rest on it.

@lneto docs only, so nothing to build or run; CI is green on both kernels and the docs job, and the branch merges onto master without conflict at 42 commits behind, since master has not touched doc/design/fsnotify since the base. Each of the three corrections checks out against the tagged headers: fsnotify_alloc_group takes flags from v5.19 (https://github.com/torvalds/linux/blob/v5.19/include/linux/fsnotify_backend.h#L597-L599; one argument at https://github.com/torvalds/linux/blob/v5.18/include/linux/fsnotify_backend.h#L544); the second parameter of fsnotify_add_mark is void *obj from v6.10 (https://github.com/torvalds/linux/blob/v6.10/include/linux/fsnotify_backend.h#L795-L796; connp at https://github.com/torvalds/linux/blob/v6.9/include/linux/fsnotify_backend.h#L764-L766) and so is the priority enum (https://github.com/torvalds/linux/blob/v6.10/include/linux/fsnotify_backend.h#L183); fsnotify_add_inode_mark keeps (mark, inode, int) from v5.15 (https://github.com/torvalds/linux/blob/v5.15/include/linux/fsnotify_backend.h#L537-L539) to v6.15 (https://github.com/torvalds/linux/blob/v6.15/include/linux/fsnotify_backend.h#L876-L878); FS_OPEN_PERM and ALL_FSNOTIFY_PERM_EVENTS are unconditional at v5.15, v6.8 and v6.15 (https://github.com/torvalds/linux/blob/v6.8/include/linux/fsnotify_backend.h#L55 and #L80), and the hooks' return-0 branch under CONFIG_FANOTIFY_ACCESS_PERMISSIONS appears at v6.8 (https://github.com/torvalds/linux/blob/v6.8/include/linux/fsnotify.h#L103), where v6.7's fsnotify_perm is unconditional (https://github.com/torvalds/linux/blob/v6.7/include/linux/fsnotify.h#L104).

Read against the stack: #852 detects the config in C with IS_ENABLED, as the note prescribes, and #855 extends the drift table rather than contradicting it. One thing found above this pull request and fixed there: #850 inserts the stable-backports paragraph inside the drift table, which splits it; the fixup is on review/850, https://github.com/luainkernel/lunatik/commit/e47558bbf3c9ac80f5110f33d021f2dafa7ae274.

Issue #659, opened 2026-08-03
linux.fs event mask constants


lneto:
Part of the fsnotify epic (#657). This is the warm up task: small, self contained, and it forces a
full pass through the build, the generated `linux.*` modules and the docs pipeline before touching
`fs/notify`.

## Context

Every later phase needs the `FS_*` event masks in Lua. They live in `linux/fsnotify_backend.h`, which
is not a uapi header — but autogen already reads non-uapi headers (`linux/notifier.h`,
`linux/sched.h`, `linux/netdevice.h`), so this is a spec entry, not a new mechanism.

`autogen/specs.lua` drives it: each entry names a header, a constant prefix and the resulting Lua
module; `autogen.lua` runs the headers through `cpp`, resolves the candidates to integers and emits
`autogen/linux/<module>.lua` plus an LDoc stub.

## Scope

Add to `autogen/specs.lua`:

| Header | Prefix | Module |
|--------|--------|--------|
| `linux/fsnotify_backend.h` | `FS_` | `fs` |

The wrinkles you will hit:

1. These are **bit masks**, not a dense enum. Confirm the generated values are the masks themselves
   (`FS_OPEN` is `0x00000020`), not ordinals.
2. The prefix is broad. `FS_` will also catch things that are not event masks. Use `include` to
   enumerate what belongs, the way the `bpf` entry already does, or `exclude` for the families that
   do not: the result should contain the event masks, `FS_EVENT_ON_CHILD`, `FS_ISDIR`, `FS_RENAME`.
3. The three `*_PERM` masks exist only under `CONFIG_FANOTIFY_ACCESS_PERMISSIONS`. Check what autogen
   does when a candidate does not resolve on the build machine, and make sure the outcome is a table
   without those keys rather than a build failure.

## Acceptance criteria

- [ ] `make` regenerates cleanly; `autogen/linux/fs.lua` contains the event masks
- [ ] a kernel script can `require("linux.fs")` and read `fs.OPEN`, `fs.MODIFY`, `fs.CREATE`,
      `fs.DELETE`, `fs.MOVED_FROM`, `fs.MOVED_TO`, `fs.ATTRIB`, `fs.CLOSE_WRITE`, `fs.OPEN_EXEC`,
      `fs.EVENT_ON_CHILD`, `fs.ISDIR`
[...]

PR #847, opened 2026-09-12
autogen: add fsnotify event mask spec (linux.fs)

lneto:
Merges after #846. The `linux.fs` table the binding's masks are written in, generated from the kernel headers the way the other `linux.*` tables are and usable before any of the module exists.

Part of the fsnotify stack, #657. Answers #659.


autogen/specs.lua
```diff
@@ -24,6 +24,13 @@ return {
 		desc = "Ethernet protocol IDs." },
 	{ header = "linux/stat.h", prefix = "S_", module = "stat",
 		desc = "File mode bits." },
+	{ header = "linux/fsnotify_backend.h", prefix = "FS_", module = "fs",
+		desc = "Filesystem notification event masks and flags.",
+		include = { "ACCESS", "MODIFY", "ATTRIB", "CLOSE_WRITE", "CLOSE_NOWRITE", "OPEN",
+			"MOVED_FROM", "MOVED_TO", "CREATE", "DELETE", "DELETE_SELF", "MOVE_SELF",
+			"OPEN_EXEC", "UNMOUNT", "Q_OVERFLOW", "ERROR",
+			"OPEN_PERM", "ACCESS_PERM", "OPEN_EXEC_PERM",
+			"EVENT_ON_CHILD", "RENAME", "ISDIR" } },
```
lneto, line 33: we need to check if any of our examples can leverage this and see if we can clean up some code adding this

tests/linux/fs.lua
```diff
@@ -0,0 +1,51 @@
...
+
+-- composites and private names the include list leaves out
+local dropped = { "MOVE", "IN_IGNORED", "DN_MULTISHOT", "EVENTS_POSS_ON_CHILD" }
+
+test("linux.fs carries every event mask at its uapi value", function()
+	for name, value in pairs(masks) do
+		assert(fs[name] == value, name .. ": " .. tostring(fs[name]))
+	end
+	for name, value in pairs(recent) do
+		assert(fs[name] == nil or fs[name] == value, name .. ": " .. tostring(fs[name]))
```
lneto, line 33: This tolerance passes whether or not the constant is there: `fs[name] == nil or fs[name] == value` pins a value only when there is one, so ERROR and RENAME are the two of the 22 the test does not require. That matters because a missing name is invisible, not loud: `include` deletes from `seen` everything not listed (autogen.lua:193-198) and nothing checks that a listed name matched, so a typo in either suffix at autogen/specs.lua:31,33 yields a 21-entry `linux.fs` and a green run on 6.8.

Since this epic is new 6.x work, the version story the tolerance carries does not apply here. Both names predate 6.0 and are unchanged through the range: FS_ERROR at [v6.0](https://github.com/torvalds/linux/blob/v6.0/include/linux/fsnotify_backend.h#L47) and FS_RENAME at [v6.0](https://github.com/torvalds/linux/blob/v6.0/include/linux/fsnotify_backend.h#L66), both still at the same values at [v7.0](https://github.com/torvalds/linux/blob/v7.0/include/linux/fsnotify_backend.h#L73). Folding `recent` into `masks` and dropping the comment at :22 would pin all 22 and take two lines off. Nothing in the stack consumes either name, so this is coverage rather than a defect.

One adjacent gap stays open either way: a 23rd name added to `include` and not to `masks` still passes. A count assertion would close that one, if you think it is worth the line.

lneto (commented):
@lneto

This can merge as it stands; nothing I found gates it. Built and run here, through the host lock: `make` clean, `autogen/linux/fs.lua` generated with all 22 entries, and `tests/linux/run.sh` green, 2 pass 0 fail, dmesg clean. The 22 values were checked one by one against `#define FS_*` in 6.8's `include/linux/fsnotify_backend.h` rather than against the table itself, and none diverges. No fixup branch: every comment below is a wording or a coverage call, not a change I could make for you.

It is stacked on #846, so I measured it against that base and not master. It is also 35 commits behind origin/master: the four files it touches are unchanged there, so the rebase is textually clean, but the PRs above it touch lunatik.h and the Makefile, so it is cheaper now than at the top of the stack.

On versions, scoped to this epic (#657) and nothing wider: the 22 names are stable across 6.x. The set of `FS_` hex defines at [v6.0](https://github.com/torvalds/linux/blob/v6.0/include/linux/fsnotify_backend.h#L31) is identical to 6.8's, and [v7.0](https://github.com/torvalds/linux/blob/v7.0/include/linux/fsnotify_backend.h#L31) differs only by adding FS_PRE_ACCESS and FS_MNT_ATTACH/FS_MNT_DETACH, which the list does not carry. So the entry pins no minimum above 6.0, and nothing in it depends on a 5.x kernel.

One mechanism is worth naming, because two of the inline comments rest on it: `include` is a filter, not an assertion. autogen.lua:193-198 builds `keep` and deletes from `seen` everything not in it, and nothing checks that a `keep` entry matched. Measured rather than read: spelling one entry `RENAEM` and rebuilding gives rc=0, a 21 entry table with no `RENAME`, and not one line of diagnostic. So a typo there yields a silently smaller table, and until the watch PR lands tests/linux/fs.lua is the entry's only guard.

Five non-gating comments inline, none of them a defect in what ships.

One item belongs to #846 rather than here. kernel-notes.md:236 says the composites drop out on their own because autogen keeps only integer expressions; is_integer_expr (autogen.lua:145-147) rejects only a function-like macro call or a lowercase letter outside a hex literal, so `(FS_MOVED_FROM | FS_MOVED_TO)` passes it. The `include` list is what drops them, which this PR's own test at fs.lua:26 asserts. A fixup on https://github.com/luainkernel/lunatik/commit/f8cdad6b2b411dbae8fb0004a5073959718cbbf8 would land the correction.

Your comment on autogen/specs.lua:33 is answered in a separate reply.

tests/linux/fs.lua
```diff
@@ -0,0 +1,51 @@
...
+	OPEN_EXEC = 0x00001000, UNMOUNT = 0x00002000, Q_OVERFLOW = 0x00004000,
+	OPEN_PERM = 0x00010000, ACCESS_PERM = 0x00020000, OPEN_EXEC_PERM = 0x00040000,
+	EVENT_ON_CHILD = 0x08000000, ISDIR = 0x40000000,
+}
+
+-- absent from the header before 5.16 (ERROR) and 5.17 (RENAME)
+local recent = { ERROR = 0x00008000, RENAME = 0x10000000 }
+
+-- composites and private names the include list leaves out
+local dropped = { "MOVE", "IN_IGNORED", "DN_MULTISHOT", "EVENTS_POSS_ON_CHILD" }
```
lneto, line 26: `dropped` covers MOVE, IN_IGNORED, DN_MULTISHOT and EVENTS_POSS_ON_CHILD, but not the names the commit body leads with. FS_PRIO_0/1/2 are plain defines at 6.8 (include/linux/fsnotify_backend.h:208-210, values 0, 1 and 2, indented inside the struct, which `cpp -dD` re-emits at column zero), and they become [`enum fsnotify_group_prio`](https://github.com/torvalds/linux/blob/v6.10/include/linux/fsnotify_backend.h#L183) in 6.10, so on a 6.8 build they are live candidates that only `include` keeps out. FS_EVENTS_POSS_TO_PARENT is the second composite.

Nothing is unguarded today: the single-bit case at :40 rejects 0, and 1 and 2 collide with ACCESS and MODIFY. Adding the four names would make the test assert what the body claims, on the kernels where the claim has teeth; the comment at :25 would need a word, since a priority is neither a composite nor a private name.

If you would rather not keep a list that goes stale as the header moves, a closed-world check on the key set of `fs` subsumes `dropped` entirely. I have not checked the precondition for that across the whole range, so it is an option to weigh rather than a proven replacement.

tests/linux/run.sh
```diff
@@ -11,7 +11,7 @@ DIR="$(dirname "$(readlink -f "$0")")"
 
 source "$DIR/../lib.sh"
 
-TESTS="random"
+TESTS="random fs"
```
lneto, line 14: The suite has no `cleanup()` and no trap, so nothing stops what the cases register. lib/lunatik/runner.lua keys `env.runtimes[script]` on the script path and raises "%s is already running" when it is set, and a one-shot that registers no hook still leaves the entry, so `sudo bash tests/linux/run.sh`, the invocation the Usage line documents, passes once and fails the second time on both cases. `sudo lunatik test` masks it, since that reloads first.

This is traced from source, not observed; I ran nothing this session. It is also pre-existing, `random` leaks the same way. The reason to raise it on this line is that AGENTS asks a new case to extend the cleanup in the commit that adds it, and there is none here to extend. tests/rcu/run.sh:23-29 drops in with the path changed. Worth saying that the loop form necessarily also covers `random`, which is already on master, so it reaches one case past what this PR introduces.

tests/linux/fs.lua
```diff
@@ -0,0 +1,51 @@
+--
+-- SPDX-FileCopyrightText: (c) 2026 Ring Zero Desenvolvimento de Software LTDA
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+-- Kernel-side script for the linux.fs test (see run.sh).
```
lneto, line 5: The pointer lands on a file that describes no case: tests/linux/run.sh carries a one-line header and a Usage line, and the description is in tests/README.md:120-124. tests/rcu/run.sh:8-12 is the shape the tree uses, where a `(see run.sh)` pointer resolves to a per-case paragraph. `random` has the same gap today, so this is the suite's habit and not something the commit introduces; a line or two in run.sh's header, saying what the case pins and that the values are the uapi's, would make the pointer true.

tests/README.md
```diff
@@ -117,6 +117,11 @@ Covers the `crypto` module: `shash`, `skcipher`, `aead`, `rng`, `hkdf`,
 
 - **random**: `linux.random` ranged draws stay within `[m, n]`, covering
   the two-argument, one-argument and negative-range forms.
+- **fs**: `linux.fs` carries every fsnotify event mask at the value the
```
lneto, line 129: "carries every fsnotify event mask" reads as unconditional, while the test exempts ERROR and RENAME at tests/linux/fs.lua:33. If the fold suggested there happens, this sentence becomes accurate as written and nothing is needed. If the tolerance stays, "every fsnotify event mask the build kernel defines" is the smallest qualifier, and it keeps the version detail in the test comment rather than in two places.

autogen/specs.lua
lneto: @lneto, on

> we need to check if any of our examples can leverage this and see if we can clean up some code adding this

Nothing can use `linux.fs` yet: there is no fsnotify, inotify or fanotify code in examples/, lib/, bin/ or tests/ at this head, and the only consumer is the test this PR adds. The consumers are further up the stack, the watch PR's tests/fsnotify/ scripts and then examples/fsmonitor.lua and examples/execguard.lua; at the tip 12 of the 22 names are used and none is missing.

The sweep of examples/ and lib/ for kernel literals a `linux.*` table would replace found four sites:

1. `local udp = 0x11`, examples/dnsblock/common.lua:14 and examples/dnsdoctor/nf_dnsdoctor.lua:17, is IPPROTO_UDP ([v6.8 in.h:44](https://github.com/torvalds/linux/blob/v6.8/include/uapi/linux/in.h#L44)). Already generated as socket.ipproto at autogen/specs.lua:62, idiom at lib/socket/inet.lua:142,178. A substitution today, no new spec.
2. examples/keylocker.lua:9, the konami codes, are KEY_UP/DOWN/LEFT/RIGHT/LEFTCTRL/LEFTALT ([v6.8 input-event-codes.h:178](https://github.com/torvalds/linux/blob/v6.8/include/uapi/linux/input-event-codes.h#L178) and :183, :180, :181, :104, :131), decoded only by the comment above them. Needs a new spec in the shape this PR just used: 521 KEY_ defines, so a curated include list.
3. examples/xiaomi.lua:11, `bus = 0x05`, is BUS_BLUETOOTH ([v6.8 input.h:258](https://github.com/torvalds/linux/blob/v6.8/include/uapi/linux/input.h#L258)). One literal at one site, and tests/hid/register.lua:11 already names it as a local `<const>` with the kernel name in a comment, which is smaller than a BUS_ spec.
4. examples/tcpreject/nf_tcpreject.lua:17-36, biggest block, half fixable: a *.layout spec gives most offsets, but lib/struct.lua exposes no per-field offset and doff/flags are bitfields. Its RST/ACK are TCPHDR_*, plain hex at 6.8 and [BIT(n) from v6.15](https://github.com/torvalds/linux/blob/v6.15/include/net/tcp.h#L977), which is_integer_expr rejects, so that spec would emit an empty table with no build error.

lib/ came back clean. All four are files this PR does not touch, so they are their own PRs against master.

lneto:
Retargeted to `master` and rebased onto it, since #846 landed as a different commit and `claude_fsnotify_notes` is not an ancestor of `master`: merging this while it pointed there would have put the commit on that branch and not on `master`, which is what happened to #820 earlier today.

The commit is unchanged. Its diff against the new base is identical to the one reviewed at a2076174d apart from one hunk line number in `tests/README.md`, which moved because `master` gained lines above it. The five inline comments above are anchored to the old head and GitHub now shows them as outdated; they still apply line for line.

Every remaining PR in this stack, #848 through #855, has the same shape: each points at the branch below it, and each will need the same retarget as the one below it merges.

Issue #660, opened 2026-08-03
fsnotify module: watch, inode marks, notification events


lneto:
Part of the fsnotify epic (#657). This is the phase that creates the module and gets the first
event into Lua.

## Context

Read [`doc/design/fsnotify/kernel-notes.md`](https://github.com/luainkernel/lunatik/blob/master/doc/design/fsnotify/kernel-notes.md)
before writing code, and read `fs/notify/dnotify/dnotify.c` in the kernel tree: it is about 400 lines
and it is the smallest complete example of exactly what this module does — allocate a group, place
marks, receive events, tear down.

`lib/luanotifier.c` is the closest thing in this repository: a callback registered from Lua, invoked
from a kernel event, returning a status code.

## Scope

A new module, `fsnotify` (`lib/luafsnotify.c`), delivering notification events for inode marks:

    local fsnotify = require("fsnotify")
    local fs       = require("linux.fs")

    local function handler(mask, event)
        print("event", mask)
    end

    local watch = fsnotify.watch(handler)
    watch:mark("/tmp/scratch/file", fs.OPEN | fs.MODIFY)
    -- later
    watch:stop()

In this phase the `event` object may be a stub; identity accessors are the next phase. What must be
real here is the group, the mark, the delivery, the teardown and the reentrancy guard.

## Kernel APIs

All `EXPORT_SYMBOL_GPL`, verified in `Module.symvers` for 6.8.0-124-generic:

    fsnotify_alloc_group(const struct fsnotify_ops *ops, int flags)
    fsnotify_init_mark(struct fsnotify_mark *mark, struct fsnotify_group *group)
    fsnotify_add_mark(struct fsnotify_mark *mark, <object>, unsigned int obj_type, int add_flags)
    fsnotify_destroy_mark(struct fsnotify_mark *mark, struct fsnotify_group *group)
    fsnotify_put_group(struct fsnotify_group *group)

Plus `kern_path` and `path_put` (plain `EXPORT_SYMBOL`) to turn a Lua string into an object.

Two traps:

- the second parameter of `fsnotify_add_mark` changed between 6.8 (`fsnotify_connp_t *connp`, so
  `&inode->i_fsnotify_marks`) and 6.15 (`void *obj`, the inode itself). Write both under a
  `LINUX_VERSION_CODE` guard, in the style `lib/luaxdp.c` already uses. Do not paper over it with a
  cast;
[...]

PR #848, opened 2026-09-12
fsnotify: watch filesystem events from Lua

lneto:
Merges after #847. The module itself. It also changes the core: `lunatik_lock` records the lock's owner and `lunatik_isowner` reads it, which every binding's runtime path goes through, so this one is worth reading beyond `lib/luafsnotify.c`.

Part of the fsnotify stack, #657. Answers #660.


lunatik.h
```diff
@@ -50,8 +51,19 @@ do {									\
...
 #define lunatik_newlock(o)   lunatik_locker((o), mutex_init, spin_lock_init, spin_lock_init);
 #define lunatik_freelock(o)  lunatik_locker((o), mutex_destroy, (void), (void));
-#define lunatik_lock(o)      lunatik_locker((o), mutex_lock, spin_lock_bh, spin_lock_irqsave, (o)->flags)
-#define lunatik_unlock(o)    lunatik_locker((o), mutex_unlock, spin_unlock_bh, spin_unlock_irqrestore, (o)->flags)
+#define lunatik_setowner(o, task)	WRITE_ONCE((o)->owner, (task))
+
+#define lunatik_lock(o)								\
+do {										\
+	lunatik_locker((o), mutex_lock, spin_lock_bh, spin_lock_irqsave, (o)->flags);	\
+	lunatik_setowner((o), current);						\
```
lneto: why do we need this?

lunatik.h
```diff
@@ -102,6 +114,7 @@ typedef struct lunatik_object_s {
 		struct mutex mutex;
 		spinlock_t spin;
 	};
+	struct task_struct *owner; /* only the holder writes it, so an unlocked read never mistakes a stranger for itself */
```
lneto: did we borrow this pattern from somewhere in the kernel? please, double check this whole owner scheme

lunatik.h
```diff
@@ -137,13 +150,24 @@ extern const lunatik_class_t lunatik_class;
...
 	if (lunatik_issoftirq(object->opt))
-		return spin_trylock(&object->spin);
-	if (lunatik_isirq(object->opt))
-		return spin_trylock_irqsave(&object->spin, object->flags);
-	return mutex_trylock(&object->mutex);
+		locked = spin_trylock(&object->spin);
+	else if (lunatik_isirq(object->opt))
+		locked = spin_trylock_irqsave(&object->spin, object->flags);
+	else
+		locked = mutex_trylock(&object->mutex);
```
lneto: can't we use `?`? won't it be better/cleaner?

doc/capi.md
```diff
@@ -176,6 +176,16 @@ Returns `true` if the script associated with `L` has finished loading (i.e., the
...
 
+### lunatik\_isowner
+```C
+bool lunatik_isowner(lunatik_object_t *object);
+```
+Returns `true` if the calling task holds `object`'s lock. `lunatik_lock` records the owner
+and `lunatik_unlock` clears it, so every holder is seen, whichever route it took into the
+lock. Use it in a kernel callback that can fire on a task already running Lua under the
+runtime it would dispatch to, where `lunatik_run` would deadlock rather than block: skip the
+event instead.
```
lneto, line 201: this whole "owner" addition should be introduced as a separate commit in this PR and we need reinforce our harness to prevent this mistake again

lib/luafsnotify.c
```diff
@@ -0,0 +1,293 @@
...
+
+typedef struct luafsnotify_mark_s {
+	struct fsnotify_mark mark;
+	struct list_head entry;
+} luafsnotify_mark_t;
+
+static const lunatik_class_t luafsnotify_class;
+
+LUNATIK_PRIVATECHECKER(luafsnotify_check, luafsnotify_t *,
+	lunatik_argcheckclass(L, ix, &luafsnotify_class, "fsnotify");
```
lneto: is it really necessary?

lib/luafsnotify.c
```diff
@@ -0,0 +1,293 @@
...
+static int luafsnotify_callback(lua_State *L, luafsnotify_t *watch, __u32 mask)
+{
+	if (lunatik_getregistry(L, watch) != LUA_TFUNCTION)
+		return 0; /* callback removed by stop() */
+
+	lua_pushinteger(L, (lua_Integer)mask);
+
+	if (lua_pcall(L, 1, 0, 0) != LUA_OK) /* callback(mask) */
+		pr_err("%s\n", lua_tostring(L, -1));
+
```
lneto:
```suggestion
```

lib/luafsnotify.c
```diff
@@ -0,0 +1,293 @@
...
+static const lunatik_class_t luafsnotify_class;
+
+LUNATIK_PRIVATECHECKER(luafsnotify_check, luafsnotify_t *,
+	lunatik_argcheckclass(L, ix, &luafsnotify_class, "fsnotify");
+);
+
+static int luafsnotify_callback(lua_State *L, luafsnotify_t *watch, __u32 mask)
+{
+	if (lunatik_getregistry(L, watch) != LUA_TFUNCTION)
+		return 0; /* callback removed by stop() */
```
lneto, line 51: shouldn't we log anything here? nor returning any error?

lib/luafsnotify.c
```diff
@@ -0,0 +1,293 @@
...
+
+	lua_pushinteger(L, (lua_Integer)mask);
+
```
lneto:
```suggestion
```

lib/luafsnotify.c
```diff
@@ -0,0 +1,293 @@
...
+	lunatik_run(watch->runtime, luafsnotify_callback, ret, watch, mask);
+	(void)ret;
+	return 0; /* fsnotify only reads this for permission events */
+}
+
+static void luafsnotify_freemark(struct fsnotify_mark *mark)
+{
+	lunatik_free(container_of(mark, luafsnotify_mark_t, mark));
+}
```
lneto, line 77: why a function wrapper here? can't it be simply a macro?

lib/luafsnotify.c
```diff
@@ -0,0 +1,293 @@
...
+
+	fsnotify_init_mark(&mark->mark, watch->group);
+	mark->mark.mask = mask;
+
+	ret = fsnotify_add_inode_mark(&mark->mark, d_inode(path.dentry), 0);
+	path_put(&path);
+
+	if (ret != 0) {
+		fsnotify_put_mark(&mark->mark); /* frees through free_mark */
+		lunatik_throw(L, ret);
```
lneto, line 158: won't be cleaner to use goto for error handling?

lunatik.h
lneto: (posted by an agent, not by @lneto)

> why do we need this?

Because the guard has to see every holder of the object's lock, and `lunatik_lock` is the one place all of them pass. Traced at 024e1163, the routes into a `lunatik_object_t`'s lock are `lunatik_run`, taken by every binding's callback including this one (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lunatik.h#L88-L99); `lunatik_monitor`, which wraps every method of a MONITOR class (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lunatik_obj.c#L137-L156), and the runtime class is one (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lunatik_core.c#L194-L200), so `runtime:resume()` runs another runtime's Lua under that runtime's own lock (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lunatik_core.c#L134-L151); `lunatik_setready`; `lunatik_closeprivate`, that is `runtime:stop()` and `__gc` (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lunatik_obj.c#L80-L91); and `lunatik_trylock` on luathread's exit path (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lib/luathread.c#L44-L63). A tree-wide grep for `mutex_lock|spin_lock|mutex_unlock|spin_unlock` finds no site taking an object's lock outside these macros; the only other hits are `luadevice_lock`/`luadevice_unlock` on a module-global mutex, not on an object. The store at this line is what makes capi.md's "every holder is seen, whichever route it took into the lock" true by construction rather than by inspection.

Recording it in the binding instead, around its own `lunatik_run`, covers the fsnotify callback and leaves the thread body, a device file operation and another module's callback in the same runtime to hang. The tree already has that weaker shape: `luanotifier_call` infers "we hold the lock" from `notifier->unregister == NULL`, meaning still inside `register_fn` (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lib/luanotifier.c#L60-L71). That is a proxy that correlates; the owner field is the property itself.

Cost is one `WRITE_ONCE` on each side of a lock that is already being acquired. I did not measure it and claim nothing beyond what it is: two marked stores in a path that already takes a mutex or a spinlock.

One asymmetry, since it is where "every holder is recorded" stops being true by construction: `lunatik_trylock` returns 1 without recording an owner when the object is not a MONITOR (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lunatik.h#L151-L166), while `lunatik_unlock` clears unconditionally. The sole caller today is `luathread_func` on a MONITOR class, so the pair is balanced and the gap is latent; it can erase a record and never invent one, so it is false-negative-only. It is carried in https://github.com/luainkernel/lunatik/issues/862 with the rest of the guarantee's boundary.

The line is unchanged. It moved, with the rest of the owner scheme, into a commit of its own: https://github.com/luainkernel/lunatik/commit/1406a7660e3c9b5f9dfa20e0b3b0e0e0d9bb4f36. Push and suite are reported on the `doc/capi.md:187` thread.

lunatik.h
lneto: (posted by an agent, not by @lneto)

> did we borrow this pattern from somewhere in the kernel? please, double check this whole owner scheme

Audited, and the four answers are: the pattern is borrowed, no existing facility makes the field redundant for a module, the annotation is the right one, and one answer does not cover all three lock classes. The last is a documentation gap and it is now closed.

**Borrowed, in two shapes.** `struct regulator_dev` carries `struct task_struct *mutex_owner` beside its `ww_mutex` (https://github.com/torvalds/linux/blob/v6.12/include/linux/regulator/driver.h#L627); `regulator_lock_nested` compares it to `current` and `regulator_unlock` clears it before the unlock (https://github.com/torvalds/linux/blob/v6.12/drivers/regulator/core.c#L145-L176 , https://github.com/torvalds/linux/blob/v6.12/drivers/regulator/core.c#L200-L210). `local_lock_t` carries `struct task_struct *owner` under CONFIG_DEBUG_LOCK_ALLOC, set after the acquire and cleared before the release, the exact ordering of `lunatik_lock`/`lunatik_unlock` (https://github.com/torvalds/linux/blob/v6.12/include/linux/local_lock_internal.h#L11-L38). Same family elsewhere: `mq_freeze_owner`, `cpufreq_policy.transition_task`, libata's `eh_owner`. One difference to state plainly: both precedents serialize the read, the regulator under `regulator_nesting_mutex` and `local_lock` in same-CPU preemption-disabled context. I found no in-tree site that reads a stored-`current` owner unlocked, so the soundness of the unlocked read is argued below from the memory model, not from a kernel precedent for that read.

**Not redundant, for a module.** `mutex_is_locked()` is exported but answers "is anyone holding it", `__mutex_owner(lock) != NULL` (https://github.com/torvalds/linux/blob/v6.12/kernel/locking/mutex.c#L76-L93), which would drop every event delivered while any other task was in a Lua step. `struct mutex` does record the owner unconditionally in the non-RT build, correcting the premise in the commit body, but the only accessor is that same `static inline __mutex_owner`, under a comment reading "DO NOT USE (outside of mutex code)", not exported and in no header a module includes; and under CONFIG_PREEMPT_RT the field does not exist at all, `struct mutex` becoming a wrapper over `rt_mutex_base` (https://github.com/torvalds/linux/blob/v6.12/include/linux/mutex_types.h#L62-L67). lockdep answers exactly this question, `lockdep_is_held()` is `lock_is_held(&(lock)->dep_map)` (https://github.com/torvalds/linux/blob/v6.12/include/linux/lockdep.h#L237-L253), but only under CONFIG_LOCKDEP: without it the kernel declares `lock_is_held`/`lockdep_is_held` with no definition on purpose, so a real call is a link error (https://github.com/torvalds/linux/blob/v6.12/include/linux/lockdep.h#L381-L393), and neither symbol is in this host's Module.symvers. Spinlocks keep `void *owner` only under CONFIG_DEBUG_SPINLOCK, and `arch_spin_is_locked` is `((void)(lock), 0)` on UP (https://github.com/torvalds/linux/blob/v6.12/include/linux/spinlock_types_raw.h#L14-L23 , https://github.com/torvalds/linux/blob/v6.12/include/linux/spinlock_up.h#L62). The PREEMPT_RT, LOCKDEP and DEBUG_SPINLOCK claims are read from the headers and from the absence of the symbols here; all three configs are off on this host, so none was exercised.

**The annotation is exactly right, neither weaker nor stronger.** The field is written under the lock and read without it, an intentional data race the kernel's own rules say must be marked. `data_race()` is the weaker option and is wrong here because a wrong value deadlocks rather than degrades: "if the heuristics might be fatally confused by one or more of the possible erroneous values, use READ_ONCE() instead of data_race()" (https://github.com/torvalds/linux/blob/v6.12/tools/memory-model/Documentation/access-marking.txt#L141-L144 , https://github.com/torvalds/linux/blob/v6.12/tools/memory-model/Documentation/access-marking.txt#L41-L48). Stronger buys nothing, and the argument is identity rather than ordering: every writer stores either `current` or NULL, so the only value that can make the test true is the reader's own pointer and the only task that ever writes it is the reader. A true answer means "I wrote this and have not overwritten it", which single-location coherence guarantees on every architecture the kernel supports (https://github.com/torvalds/linux/blob/v6.12/include/asm-generic/rwonce.h#L47-L62). Nothing is published through the field, so there is nothing for an acquire or release to order; on the false branch the code goes on to take the lock, which supplies its own ordering. The premise that must keep holding is that no writer ever stores a task pointer other than its own, and at 024e1163 the three writers are the two in `lunatik_lock`/`lunatik_unlock` plus `lunatik_trylock`, with the init at object creation.

**One answer does not cover all three lock classes.** For the mutex, process class, it is exact: a mutex is task-owned, only the owner may unlock, and it is not recursive (https://github.com/torvalds/linux/blob/v6.12/include/linux/mutex_types.h#L13-L54), so `owner == current` is precisely "this task holds it". A spinlock is owned by a CPU and an execution context, not a task: when the acquirer is a softirq or hardirq, `current` is whatever task happened to be interrupted and the field records an identity with nothing to do with the lock. It is not a live false positive, since during such a holder's window that task cannot run and no other CPU can be running it, so nothing reads a wrong answer today; but the contract capi.md stated, "the calling task holds `object`'s lock", was true only for the mutex class. The second restriction is the same kind: `lunatik_run` locks `lunatik_pin(o)`, which for a percpu object is a different `lunatik_object_t`, while `lunatik_isowner(o)` reads `o` itself (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lunatik.h#L88-L99 , https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lunatik.h#L125-L135 , https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lunatik.h#L168-L171), so on a percpu runtime the guard is always false. This PR was unaffected only by accident of the binding: `luafsnotify_watch` calls `lunatik_checkpercpu` and `lunatik_checkruntime` first (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lib/luafsnotify.c#L241-L252 , https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lunatik.h#L250-L262), so `watch->runtime` always carries the mutex and is never percpu. Nothing said the next binding must do the same. Both restrictions are now stated in doc/capi.md under `lunatik_isowner`, with the third boundary, that a wait routed through a second task is a cycle an owner check cannot see: https://github.com/luainkernel/lunatik/commit/1406a7660e3c9b5f9dfa20e0b3b0e0e0d9bb4f36.

Two things the audit found beyond the question.

The owner field subsumes an existing mechanism. `luanotifier_call` answers the same question with the `notifier->unregister == NULL` proxy (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lib/luanotifier.c#L60-L71); now that the core records the fact, keeping the proxy beside it is two mechanisms doing one job. It is not a mechanical swap, because notifier runs the nested handler on the held lock through `lunatik_handle` while fsnotify skips, so what happens on nesting is a decision rather than a refactor. That makes it a change of its own, on its own branch, not a fixup here; it is the related item on https://github.com/luainkernel/lunatik/issues/862 so it does not die with this thread.

The guarantee is opt-in, not structural. `lunatik_run` still takes the lock unconditionally and each binding must remember to ask first, which is the shape AGENTS.md refuses: "a guarantee that holds only on some paths is not a guarantee". The structural alternative is to fold the test into `lunatik_run` and return `-EDEADLK`, at the cost of changing every binding's dispatch, netfilter, xdp, probe, notifier and device all seeing a new error code, and against `luanotifier`, which wants the opposite behaviour. That is your call rather than mine, so it is written out with its trade in https://github.com/luainkernel/lunatik/issues/862 rather than left in this thread.

Push and suite are reported on the `doc/capi.md:187` thread.

lunatik.h
lneto: (posted by an agent, not by @lneto)

> can't we use `?`? won't it be better/cleaner?

Not on this line, and yes on the three above it.

On this line there is no spelling of it. `lunatik_setowner` expands to `WRITE_ONCE`, which is a `do { ... } while (0)` statement (https://github.com/torvalds/linux/blob/v6.12/include/asm-generic/rwonce.h#L58-L62), and a statement cannot be a ternary operand; compiled against a stand-in for the macro, `locked ? setowner(o, cur) : (void)0;` gives "expected expression before 'do'". The rule it would fall under is about two short mutually exclusive calls, and this is single-armed: there is no second call, and the alternative is not a void arm but nothing at all, which is what a single-statement `if` without braces is for.

The three lines above are the four-line if/else the rule does address. The diff turned three early returns into an if/else-if/else assigning `locked` (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lunatik.h#L151-L166), three mutually exclusive calls each yielding a value. They are now a chain, and the body goes from ten lines to six:

```c
	locked = lunatik_issoftirq(object->opt) ? spin_trylock(&object->spin) :
		lunatik_isirq(object->opt) ? spin_trylock_irqsave(&object->spin, object->flags) :
		mutex_trylock(&object->mutex);
	if (locked)
		lunatik_setowner(object, current);
	return locked;
```

`int locked` stays correct: all three yield an int (https://github.com/torvalds/linux/blob/v6.12/include/linux/mutex.h#L191-L196 , https://github.com/torvalds/linux/blob/v6.12/include/linux/spinlock.h#L419-L422). It is three arms and not two, which is the edge of that rule, said here because the chain is what changed.

https://github.com/luainkernel/lunatik/commit/1406a7660e3c9b5f9dfa20e0b3b0e0e0d9bb4f36. Push and suite are reported on the `doc/capi.md:187` thread.

doc/capi.md
lneto: (posted by an agent, not by @lneto)

> this whole "owner" addition should be introduced as a separate commit in this PR and we need reinforce our harness to prevent this mistake again

Both halves done.

**The split.** The owner addition is a core change with one consumer, not a part of the binding: it changes `lunatik_lock`, `lunatik_unlock`, `lunatik_trylock`, the object layout and the C API reference, and every binding's runtime path goes through it. The head is now two commits:

- `lunatik: record the task holding an object's lock`, lunatik.h and doc/capi.md, carrying the reentrancy rationale, why the record lives at the lock and not in the binding, why the unlocked read is sound, and the two restrictions from the audit on the lunatik.h:117 thread: https://github.com/luainkernel/lunatik/commit/1406a7660e3c9b5f9dfa20e0b3b0e0e0d9bb4f36
- `fsnotify: watch filesystem events from Lua`, everything else, with the body trimmed to what the binding decides, skip rather than run on the held lock: https://github.com/luainkernel/lunatik/commit/0d60bb2126eeadc557f03a8c98b8152fc8678cb3

Rewritten and force-pushed to `claude_fsnotify_watch`: https://github.com/luainkernel/lunatik/commit/0d60bb2126eeadc557f03a8c98b8152fc8678cb3. PRs 849 through 855 were rebased onto it, each ref moved with its content unchanged. Suite on the new head: pass:147 fail:0 skip:0.

**The cost of the split**, said here rather than left to be found. 849 through 855 are a strict linear stack, each based on the previous branch, so splitting one commit into two forces a `git rebase --onto` and a force-push on all seven, in order, and seven PR diffs to re-read against their bodies. An inline comment already anchored to a line in 849 through 855 would move off its line; I checked, and none of 846, 847 or 849 through 855 carries one, so nothing was displaced. Doing it now is cheaper than after those branches grow further commits.

**The harness.** The mistake is commit-scoped, not file-scoped, so it takes a commit-scoped check: `tools/checks/core-subject.sh`, in the shape the other checks use, file paths as arguments and the subject on stdin, the way `review-post-guard.sh` already reads its input (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/tools/checks/review-post-guard.sh). It is its own mechanism and stands apart from fsnotify, so it is its own pull request: https://github.com/luainkernel/lunatik/pull/861

What it flags: a commit whose file list carries at least one core file and at least one file outside that set, unless the subject names the core. Core set named explicitly, `lunatik.h`, `lunatik_core.c`, `lunatik_obj.c`, `lunatik_val.[ch]`, `lunatik_aux.c`, `lunatik_run.c`, `lunatik_percpu.[ch]`, `doc/capi.md`, not globbed as `lunatik_*`, since `lunatik_ebpf.h`, `lunatik_cfi.h` and `lunatik_conf.h` are module and config headers and globbing them in cost 5 extra false flags in the trial; `lunatik_percpu.h` is in because it is the pin path every `lunatik_run` goes through, and adding it changed no flag in the measurement. "Names the core" is an area prefix of `lunatik*:`, `core:`, `capi:` or `api:`, or a `lunatik_`/`LUNATIK_` identifier, `capi.md`, or "C API" anywhere in the subject. It fires on 024e1163 and names lunatik.h and doc/capi.md.

What it lets through: a pure core commit, since the mixed condition never engages, so commit A above is never nagged; commit B, which has no core file; and the mixed commit AGENTS.md sanctions, a core change adopted by four bindings, because the subject names the core. The rule is not "do not mix", it is "say so in the subject".

False positives, measured rather than guessed: over the last 400 commits of origin/master it fires on 9, of which 4 are the real thing, including `thread: run the body with the objects given to thread.run` and `luaskb: new module wrapping struct sk_buff for Lua`, and 5 are noise, a docs sweep, two autogen migrations whose only core file is doc/capi.md, a treewide include sweep and a version bump. That is 1.25% noise on this history. Dropping doc/capi.md from the core set removes two of the five and blinds the check to a new C API entry, which is the line this comment is anchored to, so I kept it and took the noise. The measurement is on this repository's history only: it says how often the shape has recurred, not how often it will.

**One wart it runs into**, also in that PR. `tools/checks/pre-commit` does not take file paths; it reads `git diff --cached --name-only` and ignores `$@` (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/tools/checks/pre-commit#L1-L19), which is why the workflow has to `git reset -q --soft origin/$base` to fake an index before calling it (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/.github/workflows/checks.yml#L15-L33). It also means the workflow sees a pull request as one squashed change, so a per-commit rule cannot ride that path: the harness PR loops over `git rev-list origin/$base..HEAD` in the advisory step, so each commit is checked against its own subject and file list, and one sentence in AGENTS.md's Checks paragraph states the stdin contract. A `commit-msg` hook for the same check, and teaching `pre-commit` to take paths, are left for whoever wants them; the sentence makes the hook a one-liner.

lib/luafsnotify.c
lneto: (posted by an agent, not by @lneto)

> is it really necessary?

The check is necessary; the line is not.

The check is load-bearing. On this branch's base, `LUNATIK_PRIVATECHECKER` is a raw read, `T private = (T)lunatik_toobject(L, ix)->private;` plus a null test and nothing else, so without the `lunatik_argcheckclass` in its body `getmetatable(w).mark(other, "/tmp/x", 1)` reads another class's private as a `luafsnotify_t *` and hands whatever sits at the `group` offset to `fsnotify_init_mark`. That is the crash af3f84e1 was written to close (https://github.com/luainkernel/lunatik/commit/af3f84e1a988e2829605d8e79c0cb7b75e894468), and it is why the three-line form is what this base allows: the macro takes no class argument there.

The line itself goes on the rebase. Master's `LUNATIK_PRIVATECHECKER` takes the class as its third argument and expands through `LUNATIK_CHECKER`, which does `lunatik_checkobject`, then `lunatik_argcheckclass`, then the private-null test, and only then the vararg body: https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lunatik.h#L387-L399 . `lunatik_argcheckclass` changed shape with it, from `(L, ix, cls, tname)` to `(L, ix, object, cls)`, taking the type name from `cls->name`: https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lunatik.h#L259-L260 . So on master the whole checker is one line, the bare form `luaskel.c` uses (https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luaskel.c#L25-L25):

```c
LUNATIK_PRIVATECHECKER(luafsnotify_check, luafsnotify_t *, &luafsnotify_class);
```

All 16 `LUNATIK_PRIVATECHECKER` users on master pass the class and none writes `lunatik_argcheckclass` by hand; the vararg body is for a secondary check that must always follow, such as `luaskb_check`'s `private->skb != NULL` (https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luaskb.c#L27-L29), and fsnotify has no such condition.

The rebase onto master will do that, and it was not done in this round because the branch is 63 commits behind and two of them land on this file, so it is not mechanical. af3f84e1 is one: the current call site expands under master's macro to `(&luafsnotify_class)->class` and `("fsnotify")->name`, two hard errors. The other is 7ac925bd, which added `lunatik_checkobjectclass` and collapsed the checkobject/argcheckclass pair in seven methods (https://github.com/luainkernel/lunatik/commit/7ac925bd5d952f89fd960a6d49f8efa5bb238ad4 , https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lunatik.h#L344-L349); `luafsnotify_stop` writes that pair by hand with the old four arguments and will not compile either (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lib/luafsnotify.c#L178-L191). Both collapse in the same rebase, which is the next round on this branch; the two-site list is what I traced from the two macros that changed, not from all 63 commits.

lib/luafsnotify.c
lneto: (posted by an agent, not by @lneto)

> ```suggestion
> ```

The empty suggestion deletes line 59, the blank between `pr_err(...)` and the function's final `return 0;`.

It matches the tree decisively. Counting every function in master's `lib/*.c` and `lunatik_*.c` at 22c5afe2 whose closing brace is preceded by a one-line `return`: 230 have no blank before that return, 6 do. All 6 separate a multi-line block from the return, a braced `if` with its own return, an if/else pair, a push sequence feeding the returned call: https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luanotifier.c#L57-L66 , https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luaprobe.c#L147-L155 , https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luarcu.c#L221-L229 , https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luasocket.c#L231-L239 , https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lunatik_core.c#L112-L120 , https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lunatik_run.c#L22-L30 . None is this shape, a single-line `if` with a single-line body above a bare `return 0;`. The count was run over those two file sets only, keying on a one-line `return` directly above a column-0 closing brace.

Applied. With the deletion at 56 and the ratelimit from the line-53 thread, the function is:

```c
static int luafsnotify_callback(lua_State *L, luafsnotify_t *watch, __u32 mask)
{
	if (lunatik_getregistry(L, watch) != LUA_TFUNCTION)
		return 0; /* callback removed by stop() */

	lua_pushinteger(L, (lua_Integer)mask);
	if (lua_pcall(L, 1, 0, 0) != LUA_OK) /* callback(mask) */
		pr_err_ratelimited("%s\n", lua_tostring(L, -1));
	return 0;
}
```

https://github.com/luainkernel/lunatik/commit/0d60bb2126eeadc557f03a8c98b8152fc8678cb3. Push and suite are reported on the `doc/capi.md:187` thread.

lib/luafsnotify.c
lneto: (posted by an agent, not by @lneto)

> shouldn't we log anything here? nor returning any error?

The line is reached by design, not by accident, so neither: there is nothing to report and no consumer for an error.

What reaches it. `luafsnotify_stop` runs holding the runtime lock; it calls `lunatik_unregisterobject`, clearing the callback at the private key, then `luafsnotify_detach`, which destroys the marks without waiting for events in flight, because waiting flushes a worker that calls `synchronize_srcu` and the event it would wait on is itself blocked on the lock `stop` holds (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lib/luafsnotify.c#L98-L116). The kernel side says the window is real: `fsnotify_destroy_mark`/`fsnotify_detach_mark` clear ATTACHED and ALIVE but never `mark->mask` (https://github.com/torvalds/linux/blob/v6.12/fs/notify/mark.c#L579-L586 , https://github.com/torvalds/linux/blob/v6.12/fs/notify/mark.c#L528-L546), and the dispatching task already holds a mark reference across `fsnotify_mark_srcu`, so the interest re-check still passes and `->handle_inode_event` still fires (https://github.com/torvalds/linux/blob/v6.12/fs/notify/fsnotify.c#L274-L299). That event takes the lock `stop` just released and finds nil, once per in-flight event.

Whether to log. The tree has both idioms and the discriminator is whether a miss means a defect. `luaprobe`, `luadevice` and `luanetfilter` log `couldn't find ...` because their stop unregisters from the kernel synchronously before clearing the registry, so a later miss should not happen (https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luaprobe.c#L121-L131). `luanotifier` returns silently, with "callback removed by stop() - silent no-op", because its stop deliberately leaves the notifier_block on the chain (https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luanotifier.c#L51-L66 , https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luanotifier.c#L94-L111). fsnotify is structurally the notifier case: post-stop delivery is what makes the non-waiting detach safe. A log here would fire on every ordinary `stop` that raced an unrelated process's syscall on a watched path.

Whether to return an error. There is no consumer. `handle_inode_event`'s value is read exactly once, `if (ret && (mask & ALL_FSNOTIFY_PERM_EVENTS)) goto out;`, identical in 5.15 and 6.12 (https://github.com/torvalds/linux/blob/v6.12/fs/notify/fsnotify.c#L606-L613 , https://github.com/torvalds/linux/blob/v5.15/fs/notify/fsnotify.c#L540-L547), and `luafsnotify_mark` refuses permission masks, so nothing ever looks. A nonzero return is not inert either: in `fsnotify_handle_event` a nonzero from the parent-mark call returns early and suppresses the inode-mark delivery in the same group (https://github.com/torvalds/linux/blob/v6.12/fs/notify/fsnotify.c#L326-L332), so it would drop a sibling mark's event instead of reporting anything.

This is traced through kernel source across 5.15 and 6.12 and through the binding's own detach comment; it was not reproduced on the machine. The claim is that the path exists by design, not that a run observed it. Line 53 stands as written.

One finding in the same function, its own thing and not an answer to this question: line 58 logged the pcall failure with plain `pr_err`. It was the only unratelimited per-event log in the tree. Every callback path elsewhere ratelimits (luanotifier, luaprobe, luadevice, luanetfilter), and master's four plain `pr_err` sites are all once-per-lifetime, module init, driver-runtime creation, and a thread body that then dies. `luafsnotify_callback` runs inside the syscall of any process touching a watched path, so a Lua callback that raises raises at that rate. It is now `pr_err_ratelimited`: https://github.com/luainkernel/lunatik/commit/0d60bb2126eeadc557f03a8c98b8152fc8678cb3. Push and suite are reported on the `doc/capi.md:187` thread.

lib/luafsnotify.c
lneto: (posted by an agent, not by @lneto)

> ```suggestion
> ```

The empty suggestion deletes line 56, the blank between `lua_pushinteger(L, (lua_Integer)mask);` and the `lua_pcall` that consumes what it pushed. (The blank at 54, after the registry guard, is not targeted and stays.)

It matches the tree. Across master's `lib/*.c` and `lunatik_*.c` at 22c5afe2 there are 16 `lua_call`/`lua_pcall` sites: 9 contiguous with the preceding statement, 7 with a blank before them, and the split is not arbitrary. Every site whose immediately preceding line is a lone `lua_push*` is contiguous: https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luacpu.c#L90-L91 , https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luaset.c#L120-L121 , https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lunatik_val.c#L51-L52 . The 7 with a blank all follow a multi-line stack shuffle, as in luaprobe's push/insert ladder: https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luaprobe.c#L55-L64

The closest sibling settles it: `luanotifier_handler` has the same guard-push-pcall body and runs the push straight into the call, https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luanotifier.c#L51-L66

Applied, in the rewritten binding commit: https://github.com/luainkernel/lunatik/commit/0d60bb2126eeadc557f03a8c98b8152fc8678cb3. Push and suite are reported on the `doc/capi.md:187` thread.

lib/luafsnotify.c
lneto: (posted by an agent, not by @lneto)

> why a function wrapper here? can't it be simply a macro?

It cannot, because its address is taken. The only reference to `luafsnotify_freemark` besides its definition is `.free_mark = luafsnotify_freemark,` in `static const struct fsnotify_ops luafsnotify_ops` (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lib/luafsnotify.c#L79-L96). `free_mark` is a function-pointer member of `struct fsnotify_ops` and the kernel calls it indirectly, `group->ops->free_mark(mark)` from `fsnotify_final_mark_destroy` when the mark's refcount reaches zero, the same in 5.15 and 6.12: https://github.com/torvalds/linux/blob/v6.12/include/linux/fsnotify_backend.h#L155-L168 , https://github.com/torvalds/linux/blob/v5.15/include/linux/fsnotify_backend.h#L148-L161 , https://github.com/torvalds/linux/blob/v6.12/fs/notify/mark.c#L353-L361 , https://github.com/torvalds/linux/blob/v5.15/fs/notify/mark.c#L230-L238 . A macro has no address, so there is no macro form of this.

There is also no direct call to fold away: grep gives exactly two hits, the definition and the ops initialiser. `luafsnotify_freegroup` on the next line is there for the same reason.

On the body, since it is the other thing a reader may ask about: `container_of` resolves to a zero offset here, `struct fsnotify_mark mark` being the first member of `luafsnotify_mark_t`. It survives a reorder of the struct and names what the pointer is, so it stays.

No change.

lib/luafsnotify.c
lneto: (posted by an agent, not by @lneto)

> won't be cleaner to use goto for error handling?

No: the two error paths release different things, so they cannot share a label, and a two-label ladder adds a fallthrough double free that no compiler here will warn about.

The function whole (https://github.com/luainkernel/lunatik/blob/024e1163666e88e42dd936c0453aacc6d8b57c10/lib/luafsnotify.c#L136-L168). Four raises happen before anything is held: `luafsnotify_check`, `luaL_checkstring`, `luaL_checkinteger` and the permission-mask `luaL_argcheck`; `lunatik_checkzalloc` raises with nothing held. That leaves two unwinds, and they are not two stages of one cascade but two ownership states of the same object. `kern_path` fails: `fsnotify_init_mark` has not run, the mark is a plain allocation with no refcount and no group reference, so `lunatik_free(mark)`. `fsnotify_add_inode_mark` fails: `fsnotify_init_mark` has run and done `refcount_set(&mark->refcnt, 1)` plus `fsnotify_get_group(group)` (https://github.com/torvalds/linux/blob/v6.12/fs/notify/mark.c#L956-L965), so the only correct release is `fsnotify_put_mark`, which reaches `fsnotify_final_mark_destroy` and from there `ops->free_mark` and `fsnotify_put_group` (https://github.com/torvalds/linux/blob/v6.12/fs/notify/mark.c#L353-L361). `lunatik_free` on the second path leaks the group reference; `fsnotify_put_mark` on the first touches an uninitialised refcount.

The ordering is deliberate in the other direction too: no raise sits between `kern_path` succeeding and `path_put`, and allocating the mark after `kern_path`, which would remove the first path entirely, would move the OOM raise inside the window where `path` is held.

What goto buys in this tree is N failures sharing one unwind, as in `luadarken_setrequest`, whose three `goto err` all land on the same release (https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luadarken.c#L57-L86). Where the unwinds differ, the tree releases and throws at the site, which is what this function does: https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luadarken.c#L43-L55 , https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lib/luacrypto_aead.c#L139-L149

And the hazard. `lunatik_throw` calls `lua_error` (https://github.com/luainkernel/lunatik/blob/22c5afe26049b3adf4da7b05a512411ef262e29c/lunatik.h#L168-L172), declared `LUA_API int (lua_error) (lua_State *L);`, not noreturn (https://github.com/luainkernel/lua/blob/d7ca49b3a9c44278078b4e12a7cb26bf4c9e30fd/lua.h#L386). So a ladder

```c
put:
	fsnotify_put_mark(&mark->mark);
	lunatik_throw(L, ret);
free:
	lunatik_free(mark);
	lunatik_throw(L, ret);
```

falls through `put:` into `free:` as far as the compiler is concerned, nothing but a `return` it will never demand prevents the double free, and it cannot warn about the fallthrough because it does not know the throw does not return. The straight-line form cannot have that bug and is two lines shorter than the ladder, which needs two labels, two throws and a trailing unreachable return.

No change.

doc/capi.md
lneto: (posted by an agent, not by @lneto)

The rebase onto master that the reply above deferred has landed. The stack sits on master 22c5afe2 now, ten branches, each moved with a lease against the head it had:

- claude_fsnotify_notes: https://github.com/luainkernel/lunatik/commit/e1cc0d2d1ebaf82ebefed4046a73e03bba71684e
- claude_fsnotify_constants: https://github.com/luainkernel/lunatik/commit/2436662a6d3ab97b5801b9fec80be0ef3f5a4ffc
- claude_fsnotify_watch: https://github.com/luainkernel/lunatik/commit/653c70db284aea98e26546d0595368314502d855, with the owner commit below it at https://github.com/luainkernel/lunatik/commit/50f60effe7fb5e6f966af425b3fdd28a9f631acf
- claude_fsnotify_identity: https://github.com/luainkernel/lunatik/commit/e0bf15b8b27748d307ffe5e8695da640aae00b98
- claude_fsnotify_marks: https://github.com/luainkernel/lunatik/commit/4d4d46b969dc7f86ecbb3257e44301af046e013f
- claude_fsnotify_lookup: https://github.com/luainkernel/lunatik/commit/204bce3bec18457d6c888c2af055cf27dee99a8c
- claude_fsnotify_permission: https://github.com/luainkernel/lunatik/commit/2b5878c608776c85717bfff401456c62c0fa4a9a
- claude_fsnotify_fsmonitor: https://github.com/luainkernel/lunatik/commit/2dfa0891c35895093a16b3d8987faf073e57281e
- claude_fsnotify_execguard: https://github.com/luainkernel/lunatik/commit/d9fb92945372d144be57aec3413a85779a9695b0
- claude_fsnotify_api_note: https://github.com/luainkernel/lunatik/commit/d107a2f0f47403aacdb86a28c8358bf16b7d1aca

What the rebase changed in this pull request, beyond master's own commits: the two things the line-47 reply said it would. All three checkers are the one-line form master's macro takes, luafsnotify_check, luafsnotify_checkevent and luafsnotify_checkmark, each class's .name equal to the tname the branch passed so the error text is unchanged; and luafsnotify_stop uses lunatik_checkobjectclass. Two prose conflicts in README.md and tests/README.md were resolved by keeping master's new suites and placing fsnotify in order. No commit body described the old shape, so none was amended.

Every level was built alone. Suite at this branch: pass:170 fail:0 skip:0. Suite at the top of the stack: pass:263 fail:0 skip:0. The review threads here now show as outdated against the moved lines; the answers stand.

lneto: (posted by an agent, not by @lneto)

Build 5.15 failed on this pull request and on every one stacked above it once the runner's kernel fetch stopped flaking: too few arguments to fsnotify_alloc_group, and no member ignored_mask. The cause is not the tag the design notes were traced against. The workflow resolves 5.15 to the newest 5.15.y from kernel.org, 5.15.220 today, and that stable branch carries both the flags argument of fsnotify_alloc_group and the ignore_mask rename as backports: its include/linux/fsnotify_backend.h has the two-argument declaration, the ignore_mask field, and the macros FSNOTIFY_GROUP_USER and FSNOTIFY_MARK_FLAG_HAS_IGNORE_FLAGS, none of which v5.15.0 has. A LINUX_VERSION_CODE guard therefore picks the old form on a kernel that no longer has it.

Both guards now test the macro that arrived with each change instead of the version: FSNOTIFY_GROUP_USER for the argument, in this commit, and FSNOTIFY_MARK_FLAG_HAS_IGNORE_FLAGS for the field, in the marks commit, with the reason on the line and in each body. The other version guards in the file, for fsnotify_add_mark's fsid and connp parameters, fsnotify_find_mark, the mount connector and the group priority, were checked against the same 5.15.220 header and take the right arm there. kernel-notes.md gained a paragraph on stable backports beside the drift table, in the marks commit.

Rewritten and force-pushed to `claude_fsnotify_watch`: https://github.com/luainkernel/lunatik/commit/6e6cab9f1ab0a5d13c1d5320f11b91514416997e . `claude_fsnotify_marks`: https://github.com/luainkernel/lunatik/commit/c7497f43e5cdcb27777aeb623a8d9f6effaef375 . PRs 849 through 855 were rebased onto them, each ref moved with its content unchanged; 846 and 847 were not touched. On 6.12 every changed level compiles and the top of the stack gives pass:263 fail:0 skip:0. The 5.15 build is CI's to confirm; it cannot be run here.

lib/luafsnotify.c
```diff
@@ -0,0 +1,288 @@
...
+
+/* the group outlives this call: a mark can still be running an event under
+ * fsnotify's SRCU, and waiting for it here would deadlock against the runtime
+ * lock that event blocks on. free_group_priv frees the watch instead. */
+static void luafsnotify_detach(luafsnotify_t *watch)
+{
+	struct fsnotify_group *group = watch->group;
+	luafsnotify_mark_t *mark, *next;
+
+	if (group == NULL)
```

No caller reaches this with `group` NULL: stop nulls `object->private` before calling detach, so the release path cannot run detach a second time (https://github.com/luainkernel/lunatik/blob/fc2a9ddd61e9becc2e2a0786144280ba2218549d/lunatik_obj.c#L101-L111 releases a non-NULL private only), and newwatch throws before returning a watch without a group. The guard and the `watch->group = NULL` beside it read as a second idempotency mechanism next to the nulling in stop. If stop's nulling is the intended one, this drops the guard: https://github.com/luainkernel/lunatik/commit/d3b21abe9e7d914ed4405593a59fa8245f333f52

lneto (commented):

Verdict: changes to fold, five fixups pushed onto this branch on top of fc2a9ddd6, and merge after #847. A comment rather than a request for changes because the posting account authored the pull request.

@lneto built at fc2a9ddd6 and run on 6.8.0-138 aarch64: full suite pass:185 fail:0 skip:3, fsnotify 23/23. The mechanism holds under trace: the group owns the watch, every mark's put reaches the reaper after the SRCU period, an event blocked on the runtime lock finds the callback unregistered or the runtime not ready, and the module exit flushes the reaper before the text goes.

What I ask to fold is convention level, one fixup per finding, each anchored inline: a guard in detach no path can take, the block comment above it, the 5.15 arm of fsnotify_alloc_group, the suite's own error grep where lib.sh already carries KTAP_ERRORS, and the thread test's body. The fixups build clean, pass the checks, and ran on the same host at 1496df5fe: fsnotify 26/26 with the three added cases, full suite pass:188 fail:0 skip:3.

Cells the suite does not cover, left as #956: a second mark on the same inode by the same watch (the kernel answers EEXIST without FSNOTIFY_GROUP_DUPS), a callback that raises and the event after it, stop() called from inside the callback, and the percpu refusal of fsnotify.watch.

lib/luafsnotify.c
```diff
@@ -0,0 +1,288 @@
...
+	lunatik_free(watch);
+}
+
+static const struct fsnotify_ops luafsnotify_ops = {
+	.handle_inode_event = luafsnotify_event,
+	.free_mark = luafsnotify_freemark,
+	.free_group_priv = luafsnotify_freegroup,
+};
+
+/* the group outlives this call: a mark can still be running an event under
```

The block carries three sentences above the signature, and the one thing the code cannot say is why the group is put rather than destroyed. The comment rules keep that to one line on the call it explains (https://github.com/luainkernel/lunatik/blob/fc2a9ddd61e9becc2e2a0786144280ba2218549d/AGENTS.md#L427-L429). The fixup moves it onto fsnotify_put_group: https://github.com/luainkernel/lunatik/commit/828846e199b2c957d8495e04002edf2efd4a4f66

The trace it rests on: fsnotify_destroy_group waits in fsnotify_wait_marks_destroyed (https://github.com/torvalds/linux/blob/v6.8/fs/notify/group.c#L76), which flushes the reaper whose work runs synchronize_srcu (https://github.com/torvalds/linux/blob/v6.8/fs/notify/mark.c#L862), while the event that would block on the runtime lock sits inside fsnotify()'s SRCU read section (https://github.com/torvalds/linux/blob/v6.8/fs/notify/fsnotify.c#L547-L580).

lib/luafsnotify.c
```diff
@@ -0,0 +1,288 @@
...
+
+	lunatik_unregisterobject(L, object);
+	object->private = NULL; /* the group frees the watch; release must not run on it */
+	luafsnotify_detach(watch);
+	return 0;
+}
+
+static inline struct fsnotify_group *luafsnotify_allocgroup(void)
+{
+/* the flags came with the macro, backported into 5.15.y past a version guard; GROUP_USER accounts a userspace fd */
```

This binding belongs to the fsnotify epic, which targets 6.x, so the #else arm has no kernel left to serve: fsnotify_alloc_group takes flags from v5.19 (https://github.com/torvalds/linux/blob/v5.19/include/linux/fsnotify_backend.h#L597-L599; the one-argument form is https://github.com/torvalds/linux/blob/v5.18/include/linux/fsnotify_backend.h#L544). The fixup calls it directly and drops the wrapper and this line, which carried two reasons: https://github.com/luainkernel/lunatik/commit/1496df5febb79464facf275ef71b8dbe3db8606b. The commit body's paragraph on 5.15.220 goes with it.

tests/fsnotify/open.sh
```diff
@@ -0,0 +1,61 @@
...
+
+echo "$marked" | grep -qF "fsnotify open test pass" || \
+	fail "no FS_OPEN for the marked file: $(echo "$marked" | grep -F 'fsnotify open test')"
+ktap_pass "an inode mark reports FS_OPEN for the file it marks"
+
+echo "$neighbour" | grep -qF "fsnotify open test" && \
+	fail "the callback ran for an unmarked neighbour"
+ktap_pass "an unmarked neighbour reports nothing"
+
+errs=$(printf '%s\n%s\n' "$marked" "$neighbour" | grep -E "\.lua:[0-9]+:" || true)
```
lneto: (posted by an agent, not by @lneto)

The suite greps `\.lua:[0-9]+:` by hand here and in nomask, child and lifetime, and lifetime adds x86 oops signatures on a host where an arm64 oops prints `Internal error:`. tests/lib.sh already defines KTAP_ERRORS with the Lua error, WARNING, UBSAN and both arm64 signatures (https://github.com/luainkernel/lunatik/blob/fc2a9ddd61e9becc2e2a0786144280ba2218549d/tests/lib.sh#L25), which is what check_dmesg reads, and the test rules ask for that check at the end of every test (https://github.com/luainkernel/lunatik/blob/fc2a9ddd61e9becc2e2a0786144280ba2218549d/AGENTS.md#L470). Since each test stashes several dmesg windows, the fixup greps the stashed windows with `$KTAP_ERRORS` and adds the case to reentrancy, thread and context, which had none: https://github.com/luainkernel/lunatik/commit/bf9bdb5a0b2d47a397c4a390fec0e751cdc2f7e9

`BUG:` is not in KTAP_ERRORS; widening it is a lib.sh change, not this suite's.

tests/fsnotify/thread.lua
```diff
@@ -0,0 +1,31 @@
...
+
+local function report(mask)
+	print(string.format("fsnotify thread test fail: the callback ran, mask %x", mask))
+end
+
+local function body()
+	local file = assert(io.open(WATCHED))
+	file:close()
+	print("fsnotify thread test pass: the open returned inside the thread body")
+	while not thread.shouldstop() do
```
lneto: (posted by an agent, not by @lneto)

The body loops on shouldstop after the open it exists to make. The build section asks a test that spawns a thread to give its body work that ends rather than a loop waiting to be stopped (https://github.com/luainkernel/lunatik/blob/fc2a9ddd61e9becc2e2a0786144280ba2218549d/AGENTS.md#L58-L59), the shape tests/thread/exit.lua uses, and the assertion here needs only the open and the print. The fixup returns after the print and drops the two requires: https://github.com/luainkernel/lunatik/commit/e1ce3f4a805268dec8f4242e55364649fc87f806

lunatik stop on the exited thread then logs "thread has already stopped", as task.sh's exited case does. Not run here.

lunatik.h
```diff
@@ -84,6 +85,8 @@ typedef struct lunatik_object_s {
 		struct mutex mutex;
 		spinlock_t spin;
 	};
+	/* only the holder writes it, so an unlocked read never mistakes a stranger for itself */
```
lneto: why this field is special? why do you need a comment like this only for this particular one? I think a good idea is to create a new issue to document all these fields in the style of other kernel structs (e.g., skb)

lunatik_lock.h
```diff
@@ -20,6 +20,16 @@ static inline void lunatik_freelock(lunatik_object_t *object)
 		mutex_destroy(&object->mutex);
 }
 
+static inline void lunatik_setowner(lunatik_object_t *object, struct task_struct *task)
+{
+	WRITE_ONCE(object->owner, task);
+}
```
lneto: why not a macro?

lib/luafsnotify.c
```diff
@@ -0,0 +1,271 @@
...
+typedef struct luafsnotify_s {
+	struct fsnotify_group *group;
+	lunatik_object_t *runtime;
+	struct list_head marks;
+} luafsnotify_t;
+
+typedef struct luafsnotify_mark_s {
+	struct fsnotify_mark mark;
+	struct list_head entry;
+} luafsnotify_mark_t;
```
lneto, line 42: isn't this a common pattern that we could provide framework facilities to? I think we have PR/issue that were waiting for a third client to be brought in. Please check.

lib/luafsnotify.c
```diff
@@ -0,0 +1,271 @@
...
+		fsnotify_destroy_mark(&mark->mark, group);
+		fsnotify_put_mark(&mark->mark); /* the reference fsnotify_init_mark left us */
+	}
+	fsnotify_put_group(group); /* not destroy_group: it waits for an event blocked on the runtime lock */
+}
+
+static void luafsnotify_release(void *private)
+{
+	luafsnotify_detach((luafsnotify_t *)private);
+}
```
lneto: mixed feelings regarding this indirectio

doc/capi.md
```diff
@@ -190,6 +190,24 @@ Returns `true` if the script associated with `L` has finished loading (i.e., the
 chunk has returned). Use this to guard operations that must not run during module
 initialization — for example, spawning a kernel thread from a `runner.spawn` callback.
 
+### lunatik\_isowner
+```C
+bool lunatik_isowner(lunatik_object_t *object);
```
lneto, line 195: shouldn't this be applied to other modules as well?

lunatik.h
lneto: (posted by an agent, not by @lneto)

> why this field is special? why do you need a comment like this only for this particular one?

It is the one field read without the lock, and the line stated the invariant that read rests on; the others are read under the lock or set once at creation. That belongs with the reader of lunatik_isowner, not on the field, and the tree's structs carry no member comments, so the fixup drops it and states the invariant in capi.md's lunatik_isowner entry: https://github.com/luainkernel/lunatik/commit/7b178e4f1dfbb74494ebccfb529b7e7c51c3a634. The field-by-field documentation, skb style, is #958.

lunatik_lock.h
lneto: (posted by an agent, not by @lneto)

> why not a macro?

No reason a function buys here: each is one expression, and the file's own predicate, lunatik_isirqsave, and lunatik.h's lunatik_is*, lunatik_get* and lunatik_setruntime are macros. Both are macros now: https://github.com/luainkernel/lunatik/commit/d35e3feb30451bbf6c29e25ad6c143ba582f19a1. capi.md keeps the prototype form it already uses for lunatik_isready, a macro too.

lib/luafsnotify.c
lneto: (posted by an agent, not by @lneto)

> isn't this a common pattern that we could provide framework facilities to? I think we have PR/issue that were waiting for a third client to be brought in. Please check.

Checked. The pull request on hold is #840, and it is a different pattern: one kernel registration shared by the runtimes of a percpu set, with find, share and own lifted out of probe and netfilter. A watch's marks are never shared; the list exists because a module cannot enumerate a group's marks, fsnotify_clear_marks_by_group is not exported, so the watch keeps what it has to destroy. The sibling of this shape in the tree is luadevice's list of devices (https://github.com/luainkernel/lunatik/blob/97469261693265c0a60aa283c9074283911f62ab/lib/luadevice.c#L39 and https://github.com/luainkernel/lunatik/blob/97469261693265c0a60aa283c9074283911f62ab/lib/luadevice.c#L49-L65), a module-wide list kept for lookup: two consumers of an embedded kernel object plus a list link, each walked for a different reason. Nothing to fold now; a third is the moment #840's measurement asked for.

lib/luafsnotify.c
lneto: (posted by an agent, not by @lneto)

> mixed feelings regarding this indirectio

Gone. release carries the teardown, and stop reaches it through lunatik_closeprivate, the core's own way to close a private now, under the object's lock and once, so stop no longer nulls object->private by hand either: https://github.com/luainkernel/lunatik/commit/13512ae337a79a62e7fe4e07bfbd6740465fc00d. What it buys is one teardown body and no hand-written copy of what closeprivate does; what it costs is that stop's cleanup is a call into release rather than a named detach. Built and checked; not run here, the host's crash guard blocks an install while two foreign worktrees carry uncommitted checker removals.

doc/capi.md
lneto: (posted by an agent, not by @lneto)

> shouldn't this be applied to other modules as well?

One of them, yes. device dispatches its file operations with lunatik_run on the runtime that made the device (https://github.com/luainkernel/lunatik/blob/97469261693265c0a60aa283c9074283911f62ab/lib/luadevice.c#L165), so Lua under that lock, a thread body say, that opens its own /dev/<name> waits on the mutex it holds. Traced, not reproduced, and filed as #959 with the two shapes the fix can take. The others do not have the shape: netfilter and xdp hooks cannot be re-entered from Lua, since sockets are process context only; notifier runs its nested call on purpose through lunatik_handle and keeps its own registrant check; a kprobe hit inside a kprobe handler is not run by the kprobes core. Whether the check goes into each binding or into lunatik_run once is #862's question, and #959 is the second consumer it was waiting for.

Issue #661, opened 2026-08-03
fsnotify: event identity (name, inode, directory, pid)


lneto:
Part of the fsnotify epic (#657). Depends on the module landing first.

## Context

Phase 1 delivers events with a mask. That is enough to know something happened and not enough to act:
a policy needs to know *what* it happened to. This phase fills in the event object.

What is available depends on the data type the kernel attached to the event:

    FSNOTIFY_EVENT_PATH     struct path *   (has mnt and dentry)
    FSNOTIFY_EVENT_INODE    struct inode *
    FSNOTIFY_EVENT_DENTRY   struct dentry *
    FSNOTIFY_EVENT_ERROR    struct fs_error_report *

with `fsnotify_data_inode`, `fsnotify_data_path`, `fsnotify_data_dentry` and `fsnotify_data_sb` as the
accessors. `fsnotify_data_path` returns `NULL` for anything but `FSNOTIFY_EVENT_PATH`, which is why
these are methods returning `nil` rather than fields.

## Scope

    event:name()     -- dirent name, when the event carries one
    event:ino()      -- inode number of the object
    event:dir()      -- inode number of the parent directory, for dirent events
    event:isdir()    -- FS_ISDIR
    event:pid()      -- the process performing the access
    event:path()     -- full path, only when the event carries a struct path

`event:pid()` is meaningful because delivery is synchronous: `current` is the process doing the thing.
Say so in the LDoc, because it is the kind of fact a reader will otherwise assume is a lie.

`event:path()` needs `d_path` and a page sized buffer. Keep it a method so a handler matching on the
name never pays for it, and document the cost.

The event object follows the registry pattern `lib/luanetfilter.c` uses for its `skb`: allocated
once, reset per event, cleared when the handler returns. A script that stashes the event and uses it
later must get an error, not a dangling pointer — test that.

## Acceptance criteria

- [ ] each accessor returns the right value for a file event, a dirent event and a directory event
- [ ] each returns `nil` rather than raising when the event does not carry the field
- [ ] `event:name()` matches the file created, deleted or moved, for all three
- [ ] `event:ino()` matches `stat -c %i` from the shell
- [ ] `event:pid()` matches `$$` of the process that triggered it
- [ ] `event:path()` returns the full path when available; the buffer is freed on every path,
      including the error one
- [ ] an event object kept past the handler raises when used
- [ ] tests cover the identity matrix (event kind × accessor), described in `tests/README.md`
[...]

PR #849, opened 2026-09-12
fsnotify: hand the callback the event's identity

lneto:
Merges after #848.

Part of the fsnotify stack, #657. Answers #661.


lib/luafsnotify.c
```diff
@@ -41,32 +44,74 @@ typedef struct luafsnotify_mark_s {
...
-	luafsnotify_t *watch = (luafsnotify_t *)mark->group->private;
+	luafsnotify_t *watch = (luafsnotify_t *)group->private;
+	luafsnotify_event_t event = {
+		.data = data,
+		.name = name,
+		.dir = dir,
+		.data_type = data_type,
+		.pid = task_pid_nr(current), /* delivery is synchronous: this is the accessing task */
+		.mask = mask,
+	};
```
lneto, line 105: should this be created on the stack?

lneto (commented):

Verdict: merge with the five fixups on this branch folded; the base is master now.

@lneto rebased onto master after #848 merged: the branch carried the pre-squash copies of #846 to #848, and the one commit that is this pull request replays on b20968d78 with its added and removed lines unchanged. The two conflicts sat where #848's fixups reshaped the file, release and the dropped alloc wrapper, and took master's shape. The body still opens with "Merges after #848", which is on master now.

Read against the code: the event object follows the netfilter registry pattern, the frame it borrows is cleared on every exit of the callback, and handle_event is the op the data behind event:path() needs. What comes with it, and neither the LDoc nor the note said: FS_RENAME is delivered too, across directories, with name and dir naming the old entry. Five fixups, one per finding, anchored inline. Built and checked; not run here, the host's crash guard blocks an install while two foreign worktrees carry uncommitted checker removals. Cells the suite leaves uncovered are #960: FS_RENAME itself, event:path() on an unlinked file, and the raise of event:path().

lib/luafsnotify.c
```diff
@@ -41,32 +44,74 @@ typedef struct luafsnotify_mark_s {
...
-static int luafsnotify_callback(lua_State *L, luafsnotify_t *watch, __u32 mask)
+LUNATIK_PRIVATECHECKER(luafsnotify_checkevent, luafsnotify_event_t *, &luafsnotify_event_class);
+
+static inline lunatik_object_t *luafsnotify_pushevent(lua_State *L, luafsnotify_t *watch,
+	luafsnotify_event_t *event)
 {
+	lunatik_object_t *object = watch->event;
+
+	if (lunatik_getregistry(L, object) != LUA_TUSERDATA) {
+		pr_err("couldn't find event\n");
```

A pr_err on a path every event reaches; the rule and luanetfilter's "couldn't find skb" ratelimit it. The branch cannot run today, the event object is registered at attach and unregistered only in release, after which the callback check above returns first, so it stays as the mold's own guard. Fixup: https://github.com/luainkernel/lunatik/commit/32880321f7f9f4f57c34f8bb2dc6ad8d7fa5aae8

tests/fsnotify/identity.sh
```diff
@@ -0,0 +1,134 @@
...
+check "$line" ino "$createdino" "a delete"
+check "$line" dir "$dirino" "a delete"
+check "$line" isdir false "a delete"
+check "$line" path nil "a delete"
+ktap_pass "a delete still names the unlinked entry and its inode"
+
+lunatik stop "$SCRIPT" 2>/dev/null
+
+errs=$(printf '%s\n' "$opened" "$listed" "$created" "$moves" "$deleted" | \
+	grep -E "\.lua:[0-9]+:" || true)
```
lneto: (posted by an agent, not by @lneto)

The three new tests grep `\.lua:[0-9]+:` by hand where tests/lib.sh's KTAP_ERRORS carries the Lua error, WARNING, UBSAN and both arm64 oops signatures, the finding already folded into #848's suite. Fixup: https://github.com/luainkernel/lunatik/commit/49cd25b8eaf391fc2d68a222ceedb6f56b2bc1a0

lib/luafsnotify.c
```diff
@@ -171,6 +219,152 @@ static int luafsnotify_stop(lua_State *L)
...
+* What an accessor can answer depends on what the kernel attached to the event:
+* an event on an open file carries a `struct path` and answers everything, while
+* a directory entry event carries only the entry's dentry or inode and has no
+* path. Each accessor returns `nil` for a field the event does not carry.
+* @type fsnotify_event
+*/
+
+/***
+* Returns the directory entry name the event is about.
+* Carried by the directory entry events (`CREATE`, `DELETE`, `MOVED_FROM`,
```
lneto, line 236: (posted by an agent, not by @lneto)

With handle_event the group is called for FS_RENAME too, which fsnotify_handle_event filtered to same-parent renames for handle_inode_event users (https://github.com/torvalds/linux/blob/v6.8/fs/notify/fsnotify.c#L285-L290). fsnotify_move raises it with the old directory, the old name and the moved dentry as data (https://github.com/torvalds/linux/blob/v6.8/include/linux/fsnotify.h#L178-L199), so name() and dir() answer the old side and ino() the moved inode. The LDoc lists the four dirent events and not this one; fixup: https://github.com/luainkernel/lunatik/commit/4a03fd9f3259fa4f77f4ce81550077ff62004f82. Its test is in #960.

lib/luafsnotify.c
```diff
@@ -171,6 +219,152 @@ static int luafsnotify_stop(lua_State *L)
...
+	if (IS_ERR(resolved))
+		lunatik_throw(L, (int)PTR_ERR(resolved));
+
+	len = strlen(resolved);
+	memmove(buffer, resolved, len); /* d_path fills from the end of the buffer */
+	luaL_pushresultsize(&B, len);
+	return 1;
+}
+
+static inline lunatik_object_t *luafsnotify_newevent(lua_State *L)
```
lneto, line 362: (posted by an agent, not by @lneto)

luafsnotify_newevent exists to fit lunatik_attach's new_fn(L) shape, and the macro takes varargs for exactly this: lunatik_attach(L, watch, event, lunatik_newobject, &luafsnotify_event_class, 0, LUNATIK_OPT_NONE). Fixup: https://github.com/luainkernel/lunatik/commit/80150ac38c8febcfedd0e7661a5f930fe0bedcf2

lib/luafsnotify.c
```diff
@@ -41,32 +44,74 @@ typedef struct luafsnotify_mark_s {
 	struct list_head entry;
 } luafsnotify_mark_t;
 
+/* the dispatcher's frame, borrowed by the event object for the length of one call */
+typedef struct luafsnotify_event_s {
+	const void *data;
+	const struct qstr *name;
+	struct inode *dir;
+	int data_type;
+	pid_t pid;
```

The frame carries the pid the dispatcher read, and the accessor that reports it can only run inside the callback, which runs in the accessing task: event:pid() reads task_pid_nr(current) itself and the field goes, with the comment that explained it. Fixup: https://github.com/luainkernel/lunatik/commit/b9fb711f3d2c7183ec6353c709a6ab7ffe50818d. Whether isdir() stays is yours: it is a second spelling of mask & fs.ISDIR, which api.md's own rule declines, but #661 lists it and the mask field it reads stays for #852's verdict either way.

lib/luafsnotify.c
lneto: (posted by an agent, not by @lneto)

> should this be created on the stack?

Yes, on purpose. What the frame points at lives only for this call whatever the frame's own storage: name is a snapshot the caller takes before fsnotify() and releases when it returns (https://github.com/torvalds/linux/blob/v6.8/fs/notify/fsnotify.c#L223 and https://github.com/torvalds/linux/blob/v6.8/fs/notify/fsnotify.c#L233), and data is the caller's own path or dentry (https://github.com/torvalds/linux/blob/v6.8/include/linux/fsnotify.h#L42 and https://github.com/torvalds/linux/blob/v6.8/include/linux/fsnotify.h#L100). So the event object cannot outlive the callback anywhere; a kmalloc per event would buy nothing and cost a kfree, and a frame embedded in the watch, one per watch since the runtime lock serializes events, would cost a field and still need the same clearing, since the NULL private is what makes a kept event raise. Five words on the kernel stack, next to a Lua call, is the minimal form. The pid it carried is gone with https://github.com/luainkernel/lunatik/commit/b9fb711f3d2c7183ec6353c709a6ab7ffe50818d: the accessor runs inside the callback, in the accessing task, so it reads current itself.

Issue #662, opened 2026-08-03
fsnotify: mark objects, mount and superblock marks


lneto:
Part of the fsnotify epic (#657). Depends on the module and the event object.

## Context

Phase 1 marks an inode. The kernel marks three kinds of object with the same call, distinguished by
`obj_type`:

    FSNOTIFY_OBJ_TYPE_INODE      one file or directory
    FSNOTIFY_OBJ_TYPE_VFSMOUNT   everything under a mount
    FSNOTIFY_OBJ_TYPE_SB         everything on a filesystem

A directory mark reports events on the directory itself; adding `FS_EVENT_ON_CHILD` to the mask is
what makes it report events on the files inside it. That distinction surprises everyone once.

## Scope

Mount and superblock marks:

    watch:mark("/tmp/scratch", fs.OPEN, "mount")
    watch:mark("/tmp/scratch", fs.CREATE, "sb")

and the mark object's own lifecycle:

    mark:mask([mask])     -- read, or set and recalculate
    mark:ignore([mask])   -- the ignore mask
    mark:remove()         -- fsnotify_destroy_mark
    watch:find(path[, kind])

Marks are objects rather than a table keyed by path inside the watch: the kernel already keys them
(`fsnotify_find_mark` on the connector), and a second mapping in C would be a second source of truth.
`watch:find` exists so a script does not have to keep its own.

Changing a mark's mask after it is added requires recalculating the connector's mask
(`fsnotify_recalc_mask`); setting `mark->mask` alone silently does nothing on some paths. Verify with
a test that adds an event to a live mark and sees it arrive.

## Acceptance criteria

- [ ] a mount mark reports events for files under that mount and not for files outside it
- [ ] a superblock mark reports events across the filesystem; the test creates its own (tmpfs) and
      skips if it cannot
- [ ] an invalid kind string raises with a message naming the valid ones
- [ ] `mark:mask(new)` changes what arrives, both adding and removing an event
- [ ] `mark:ignore(mask)` suppresses those events and nothing else
- [ ] `watch:find(path)` returns the installed mark, `nil` when there is none
- [ ] after `mark:remove()` no further events arrive
- [ ] `watch:stop()` with live marks destroys all of them; running the same test twice in a row
      passes both times (no leaked mark, no leaked reference)
- [ ] collecting a watch that still has marks does not oops
- [ ] tests wired into the suite and described in `tests/README.md`

## Notes

[...]

PR #850, opened 2026-09-12
fsnotify: mark objects, and marks on mounts and superblocks

lneto:
Part of the fsnotify stack, #657. Answers #662.


tests/fsnotify/kinds.sh
```diff
@@ -0,0 +1,128 @@
...
+
+ktap_header
+ktap_plan 9
+
+mkdir -p -m 0700 "$MOUNT" "$BIND"
+: > "$SCRATCH/outside"
+
+if ! mount -t tmpfs -o size=1M,mode=0700 lunatik-fsnotify "$MOUNT" 2>/dev/null || \
+	! mountpoint -q "$MOUNT"; then
+	for _ in $(seq 9); do ktap_skip "fsnotify/kinds: no tmpfs to mark"; done
```
lneto: (posted by an agent, not by @lneto)

Before 6.10 the binding has no mount kind, so mount.lua's first mark raises "mount" needs a 6.10 kernel, the script does not load, and the test fails where the rule says it skips; this host, 6.8, is one, and the run is quoted in the review body. The fixup skips the test there, the way it already skips without a tmpfs, and says so in the header and in tests/README.md. Fixup: https://github.com/luainkernel/lunatik/commit/ae5644fa1e18f435583cf5b8f9f45b8a06640c81

tests/fsnotify/kinds.sh
```diff
@@ -0,0 +1,128 @@
...
+	fail "the invalid kind was not refused: $(echo "$mounted" | grep -F 'fsnotify kinds test')"
+ktap_pass "an invalid kind raises naming the valid ones"
+
+found=$(echo "$unmounted" | grep -cF "fsnotify kinds test pass: find")
+[ "$found" -eq 4 ] || \
+	fail "find by kind: $(echo "$unmounted" | grep -F 'fsnotify kinds test' | tr '\n' ';')"
+ktap_pass "find returns the mark of each kind, and nil for a kind not marked"
+
+oops=$(printf '%s\n%s\n%s\n' "$mounted" "$super" "$unmounted" | \
+	grep -E "Oops:|BUG:|kernel BUG at|NULL pointer dereference|general protection" || true)
```
lneto: (posted by an agent, not by @lneto)

The four new scripts grep patterns of their own, this one an x86 oops set the arm64 host never prints, where tests/lib.sh has $KTAP_ERRORS, the pattern the merged fsnotify tests read since #848. With the shared pattern the two closing cases here read the same text, so the fixup keeps one and the plan drops to 8; marks.sh, mask.sh and inside.sh take the pattern too. Fixup: https://github.com/luainkernel/lunatik/commit/8b6e9b9361d58ef88b60cd5b5f4cde480793d063

lneto (commented):

Verdict: merge with the six fixups on this branch folded, and #851 right behind it.

@lneto rebased onto master after #849 merged: the one commit that is this pull request replays on 8f7a44f49 with its added and removed lines unchanged; the conflicts sat where #848's fixups reshaped release and stop, and took master's shape. The body still opens with "Merges after #849", which is on master now.

Read against the code: the mark record embeds the kernel's mark and sits on the watch's list, the registry holds its handle for as long as the mark lives, and remove and mask from inside the callback take the path inotify's one-shot takes from the same place. The signatures hold as the guards say: fsnotify_add_mark lost fsid in 6.8 (https://github.com/torvalds/linux/blob/v6.8/include/linux/fsnotify_backend.h#L764-L766), and add and find take the object itself from 6.10 (https://github.com/torvalds/linux/blob/v6.10/include/linux/fsnotify_backend.h#L792-L796). What #851 closes is reachable here on its own: mark and find resolve their path with kern_path (https://github.com/torvalds/linux/blob/v6.8/fs/namei.c#L2620), and a callback for an event the kernel delivers under the parent's i_rwsem, vfs_create's fsnotify_create (https://github.com/torvalds/linux/blob/v6.8/fs/namei.c#L3204) inside filename_create's lock (https://github.com/torvalds/linux/blob/v6.8/fs/namei.c#L3895), that resolves a name not in the dcache reaches lookup_slow and takes that lock shared in its own task (https://github.com/torvalds/linux/blob/v6.8/fs/namei.c#L1709). Nothing in this suite goes there, inside.lua resolves a cached name on FS_OPEN, so it gates the pair, not this commit.

Five findings, one fixup each, anchored inline; the sixth fixup is the rebase's. Built clean on 6.8 and the fsnotify suite ran here: 14 tests, 67 pass, no failure, kinds skipping on this kernel as its fixup says. The author's kinds.sh, run once on the same host, is the failure that fixup answers: not ok 1 Lua error in script, mount.lua:17: bad argument #3 to 'mark' ("mount" needs a 6.10 kernel). The 6.10 side of kinds did not run, there is no such kernel here. The whole suite ran after it: 245 pass, 11 skip, no failure. One shape note, not a fixup: checkkind walks a name table where luadata, luasignal and luaskb take luaL_checkoption; the walk buys the message that names the kinds and the 6.10 refusal, which checkoption's "invalid option" would not. Cells the suite leaves uncovered are #968: the mount refusal before 6.10, permission bits on mask(), a mask() whose re-add fails, a method on a removed handle, and find on a missing path or an invalid kind.

lib/luafsnotify.c
```diff
@@ -217,6 +378,101 @@ static int luafsnotify_stop(lua_State *L)
...
+/***
+* A mark on a filesystem object.
+* A userdata `watch:mark` returns and `watch:find` hands back. The watch owns
+* the mark: dropping the handle leaves the mark in place, and `watch:stop`
+* removes every mark the watch placed. A handle whose mark is gone raises.
+* @type fsnotify_mark
+*/
+
+/* the flag came with the rename, which 5.15.y backported: a version guard does not see it */
+#ifdef FSNOTIFY_MARK_FLAG_HAS_IGNORE_FLAGS
```

The #ifdef selects ignored_mask only on a 5.15.y point before the backport: 5.15.220, the 5.15 the tree builds against, has ignore_mask and the flag (https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/linux/fsnotify_backend.h?h=v5.15.220#n522), and every 6.x has them since the rename (https://github.com/torvalds/linux/blob/v6.0/include/linux/fsnotify_backend.h#L522). #848 dropped the FSNOTIFY_GROUP_USER guard on the same ground. The fixup reads the field directly and cuts the sentence in kernel-notes.md that describes the detection, whose alloc half describes what master no longer does. The commit body's last paragraph is about this guard and goes at the squash; body-identifiers.sh will name the macro. Fixup: https://github.com/luainkernel/lunatik/commit/ffab56d33e160ec3c7c8eadd85fbc8506a5e0f65

lib/luafsnotify.c
```diff
@@ -134,6 +152,33 @@ static const struct fsnotify_ops luafsnotify_ops = {
 	.free_group_priv = luafsnotify_freegroup,
 };
 
+/* the registry holds the handle while the watch holds the mark, so a script may
```

A two-line block above a static inline, where the file keeps its helpers bare and the reason belongs on the call it explains: the register keeps the handle alive as long as the mark, which is what lets find() hand it back. Fixup: https://github.com/luainkernel/lunatik/commit/dea5232e3bdad23b465d264209018afb1ac83b5a

lib/luafsnotify.c
```diff
@@ -142,59 +187,175 @@ static void luafsnotify_release(void *private)
...
-		fsnotify_put_mark(&mark->mark); /* the reference fsnotify_init_mark left us */
-	}
+
+	lua_State *L = lunatik_getstate(watch->runtime); /* NULL once the runtime is closing */
+	list_for_each_entry_safe(mark, next, &watch->marks, entry)
+		luafsnotify_removemark(L, mark);
 	fsnotify_put_group(group); /* not destroy_group: it waits for an event blocked on the runtime lock */
 }
 
+/* from 6.10 the mark API takes the object itself; before that the object's
```

Three lines for one reason: the #if below already says the API takes the object from 6.10, and what the code cannot say is why there is no mount kind before it. The fixup keeps that in one line. Fixup: https://github.com/luainkernel/lunatik/commit/12257c397df05024a219f384fb785e93758030f1

doc/design/fsnotify/kernel-notes.md
```diff
@@ -182,25 +184,69 @@ called. Set it explicitly on every kernel; it costs nothing on 6.8 and is requir
...
 | Change | before | from |
 |--------|--------|------|
 | `fsnotify_alloc_group` | `(ops)` | `(ops, flags)`, 5.19 |
+| The mark's ignore mask | `mark->ignored_mask` | `mark->ignore_mask`, 6.0 |
+
+Stable branches carry these changes back: 5.15.220, the 5.15 the tree builds against, has both the
+`flags` argument and `ignore_mask`, so `LINUX_VERSION_CODE` cannot tell the two forms apart. The module
+detects each by the macro that arrived with it, `FSNOTIFY_GROUP_USER` for the argument and
+`FSNOTIFY_MARK_FLAG_HAS_IGNORE_FLAGS` for the field, which are present on every kernel that has the change.
+| `fsnotify_add_mark` | `(mark, connp, obj_type, add_flags, fsid)` | `(mark, connp, obj_type, add_flags)`, 6.8 |
```
lneto, line 228: (posted by an agent, not by @lneto)

The rebase's fixup: the three rows on add and find landed below the paragraph that follows the drift table; this puts them back in the table, in version order. Fixup: https://github.com/luainkernel/lunatik/commit/85eeb23d7712b141129f8c4f4306aa03894fccba

lib/luafsnotify.c
```diff
@@ -142,59 +185,173 @@ static void luafsnotify_release(void *private)
...
+{
+	const char *kind = luaL_optstring(L, ix, "inode");
+	size_t i;
+
+	for (i = 0; i < ARRAY_SIZE(luafsnotify_kinds); i++)
+		if (strcmp(kind, luafsnotify_kinds[i].name) == 0)
+			return luafsnotify_kinds[i].type;
+#if (LINUX_VERSION_CODE < KERNEL_VERSION(6, 10, 0))
+	if (strcmp(kind, "mount") == 0)
+		return luaL_argerror(L, ix, "\"mount\" needs a 6.10 kernel");
```
lneto: why not argcheck?

lneto: do you mean superior to that version?

lib/luafsnotify.c
```diff
@@ -142,59 +185,173 @@ static void luafsnotify_release(void *private)
...
+	__u32 mask, unsigned int type)
+{
+	size_t len = strlen(pathname);
+	luafsnotify_mark_t *mark = (luafsnotify_mark_t *)lunatik_checkzalloc(L, sizeof(luafsnotify_mark_t) + len + 1);
+	struct path path;
+	int ret;
+
+	if ((ret = kern_path(pathname, LOOKUP_FOLLOW, &path)) != 0) {
+		lunatik_free(mark);
+		lunatik_throw(L, ret);
```
lneto, line 252: wondering how much are we repeating this pattern and if it won't be worth to have helper in the core

lib/luafsnotify.c
lneto: @lneto
> why not argcheck?

No reason: a wrong value at an index is what argcheck is for, and the check is one line. Fixup, together with the wording below: https://github.com/luainkernel/lunatik/commit/ae0285af3115ea1521fb3b77a437e2d8c960db34

lneto: @lneto
> do you mean superior to that version?

Yes, 6.10 and everything after it; "needs a 6.10 kernel" reads as that one release. The message and the two LDoc lines now say "needs kernel 6.10 or later", in the same fixup: https://github.com/luainkernel/lunatik/commit/ae0285af3115ea1521fb3b77a437e2d8c960db34

lib/luafsnotify.c
lneto: @lneto
> wondering how much are we repeating this pattern and if it won't be worth to have helper in the core

The core has it: lunatik_try and lunatik_tryret (https://github.com/luainkernel/lunatik/blob/8f7a44f49b44305c4395f6223b5caf4d7debee60/lunatik.h#L151-L160), 31 call sites across nine bindings. find holds nothing at its kern_path, so it takes lunatik_try: https://github.com/luainkernel/lunatik/commit/814fd0229ee36c37915cb0d288ede5eb45d80934. This site cannot: the record is allocated before the path resolves, so the failure frees it first, and the raise below it puts the mark instead, which after fsnotify_init_mark is free_mark's to free. Resolving before allocating moves the problem rather than removing it, since the allocation's raise would then leak the path reference. Every other explicit throw left in lib is of that shape, a release first (lunatik_free, crypto_free_skcipher, fsnotify_put_mark, a request) or a filtered errno as in luabpf_map_checkret, four different releases across seven sites; a helper for it would take the release as an argument and hide the line the rule asks to keep in sight, what is held and who releases it.

Two more fixups from the same pass, not from a question: the permission argcheck mark and mask both carried is one helper, https://github.com/luainkernel/lunatik/commit/87787ee0eb26b550f95243a16569698477611e4b, and checkkind takes luaL_checkoption as luadata, luasignal and luaskb do, https://github.com/luainkernel/lunatik/commit/971a9ae576d8fbf882bb4101cf4c9ad3a011752c, which drops the kind table, its typedef and the message macro; an invalid kind now reads "invalid option 'x'" rather than naming the valid ones, and mount.lua asserts that.

PR #851, opened 2026-09-12
fsnotify: a path resolved from a callback stays in the directory cache

lneto:
Its own pull request rather than part of either neighbour: the hazard it closes is reachable from every entry point that resolves a path, which is three of them by the time this lands.

Part of the fsnotify stack, #657.


lib/luafsnotify.c
```diff
@@ -73,6 +75,28 @@ LUNATIK_PRIVATECHECKER(luafsnotify_checkmark, luafsnotify_mark_t *, &luafsnotify
...
+/* set while a callback is on this state's stack */
+static const char luafsnotify_dispatching;
+
+static inline void luafsnotify_setdispatching(lua_State *L, bool on)
+{
+	lua_pushboolean(L, on);
+	lua_rawsetp(L, LUA_REGISTRYINDEX, &luafsnotify_dispatching);
+}
+
+/* the event's own task may hold the directory's i_rwsem, and a walk that leaves
```

The right fact in the right store: a callback on this runtime's stack, per state rather than per watch, in the registry every coroutine shares, under a static key as lunatik_ebpf_env_key is. Two things change. The name: incallback, the reader's view at luafsnotify_kernpath, with the setter in the tree's set/is shape; this file's dispatcher is the kernel task delivering the event, so "dispatching" said the opposite. The seed: the callback writes the slot from lunatik_handle, outside any protected call, and a first insertion that rehashes and fails to allocate is luaD_throw with no handler, abort, BUG; watch() now writes the slot once under the pcall every Lua C function runs in, so the dispatch path stores a boolean on an existing node, which cannot allocate. Fixups: https://github.com/luainkernel/lunatik/commit/26fa79b557bce186e969ecc72a4e13db8e16ec59 (the three @raise name EAGAIN and the condition the kernel answers it for), https://github.com/luainkernel/lunatik/commit/a6658917a665557432bc5c14b5ab50ac522e1ecc (static inline, as the file's helpers), https://github.com/luainkernel/lunatik/commit/2269efd353c0aeda8920ffa61434d230ef902f03 (the test case under the directory lock), https://github.com/luainkernel/lunatik/commit/db440c3a61fd1622db9037248cfb9c0c7edbf6cf (the pair renamed, the predicate, the seed, the reason on the flag line).

lneto (commented):

Verdict: merge with the four fixups on this branch folded; the base is master now, and the flag stays in the binding.

@lneto rebased onto master after #850 merged: the one commit that is this pull request replays on bc7ae3485 with its added and removed lines unchanged but one, find's kern_path, which master had turned into lunatik_try and now reads lunatik_try(L, luafsnotify_kernpath, L, pathname, &path); the test's error grep took master's $KTAP_ERRORS with the new cases above it.

Read against fs/namei.c: LOOKUP_CACHED does what the commit says at v5.15, v6.8 and v6.12, the same three sites in each: legitimize_links fails under the flag, which is the first call of try_to_unlazy (https://github.com/torvalds/linux/blob/v6.8/fs/namei.c#L727-L731), complete_walk drops it before the final unlazy (https://github.com/torvalds/linux/blob/v6.8/fs/namei.c#L890), and path_init refuses it outside RCU mode (https://github.com/torvalds/linux/blob/v6.8/fs/namei.c#L2358-L2359); filename_lookup's ref-walk retry, and its ESTALE retry, land in that refusal, which is where the EAGAIN the test asserts comes from. The open case at v6.12 is a direct fsnotify_open in open_last_lookups under the directory lock (https://github.com/torvalds/linux/blob/v6.12/fs/namei.c#L3690-L3699), which the commit body attributes to do_dentry_open, the 6.8 route. That a callback is on the stack is a fact per runtime, not per watch, since a second watch's mark called from the first's callback runs in the same task; a boolean in the registry under a static key is that fact's home, as lunatik_ebpf_env_key is the eBPF environment's: the registry belongs to the state and every coroutine shares it, and the core neither sets nor reads the flag, so it stays in the binding. What the original shape lacked is a seed: the callback writes the slot from lunatik_handle, outside any protected call, and a first insertion that rehashes and fails to allocate aborts the state, so watch() writes the slot once under the pcall a Lua C function runs in, and the dispatch path's boolean stores land on an existing node, where they cannot allocate.

Four fixups, anchored inline. Built clean on 6.8, the core's srcversion read live after each reload, and the fsnotify suite ran on every head: 14 tests, 69 pass, no failure. The open events the test stimulates arrive with no directory lock held, so one fixup adds the cell this pull request exists for, a callback of FS_CREATE on the scratch directory, delivered inside the parent's i_rwsem, resolving the uncached name and answering EAGAIN; a tree without the flag wedges the host there, so its discrimination is the message. Two independent agent rounds reviewed this branch against the same sources and their findings are the fixups; the pre-existing defects found on the way are #990, #991 and #1001. Cells still uncovered, on #968: watch:mark from a callback, cached and not, and a second watch of the same runtime resolving from the first's callback.

Issue #663, opened 2026-08-03
fsnotify: permission events and the verdict path


lneto:
Part of the fsnotify epic (#657). This is the phase that turns the binding from an observer into
an enforcer, and the one that can lock a machine out of its own files. It lands last on purpose:
after the guard, the identity and the lifecycle are already reviewed.

## Context

Permission events are not notifications. The dispatcher parks the syscall inside the handler and takes
the return value as the answer:

    while (fsnotify_iter_select_report_types(&iter_info)) {
            ret = send_to_group(...);
            if (ret && (mask & ALL_FSNOTIFY_PERM_EVENTS))
                    goto out;
            fsnotify_iter_next(&iter_info);
    }

`fs/notify/fsnotify.c`, 6.8. Non zero denies, and it propagates out of the syscall as that errno.
The hooks live at `security_file_open` (`fsnotify_open_perm`) and in the read path
(`fsnotify_file_perm`), all under `CONFIG_FANOTIFY_ACCESS_PERMISSIONS`.

## Scope

    FS_OPEN_PERM        gate an open
    FS_ACCESS_PERM      gate a read
    FS_OPEN_EXEC_PERM   gate an exec

    local function guard(mask, event)
        if not allowed:get(event:name()) then
            return fsnotify.action.DENY
        end
        return fsnotify.action.ALLOW
    end

`fsnotify.action.ALLOW` is 0 and `DENY` is `-EPERM`. Named constants rather than raw numbers: a bare
`-1` returned by accident is a denial, and a boolean would invert badly, because a handler that
forgets to return must allow.

## The version trap

On 6.8 the dispatcher does not gate permission delivery on the group's priority. On 6.14 and newer the
open path consults `fsnotify_sb_has_priority_watchers(sb, FSNOTIFY_PRIO_CONTENT)` and marks the file
`FMODE_NONOTIFY_PERM` when no such watcher exists — so a group that does not set
`FSNOTIFY_PRIO_CONTENT` before its marks are added simply never gets called there.

Set `group->priority = FSNOTIFY_PRIO_CONTENT` on every kernel: it costs nothing on 6.8 and is required
later. A wrong priority shows up as "works fine" on the old kernel and "no events at all" on the new
one, which is the worst possible failure shape.

6.14+ also adds `fsnotify_mmap_perm` and `fsnotify_truncate_perm`, and pre-content events. Guard the
new masks by version; do not try to emulate them on older kernels.

## Acceptance criteria

[...]

PR #852, opened 2026-09-12
fsnotify: permission events and the verdict path

lneto:
The enforcement boundary the epic is for: a callback's return value becomes the syscall's answer, so the default when anything goes wrong has to be allow.

Part of the fsnotify stack, #657. Answers #663.


lib/luafsnotify.c
```diff
@@ -115,7 +127,16 @@ static inline lunatik_object_t *luafsnotify_pushevent(lua_State *L, luafsnotify_
 	return object;
 }
 
-static int luafsnotify_callback(lua_State *L, luafsnotify_t *watch, luafsnotify_event_t *event)
+/* the caller reads a non-negative return as success and an out-of-range one as
+ * a pointer, so only a negative errno denies; lua_tointeger would take "-1" as one */
+static inline int luafsnotify_toverdict(lua_State *L)
```
lneto, line 132: (posted by an agent, not by @lneto)

A two-line block above a static inline helper, where the file keeps one line of reason: the caller takes anything else for success or a pointer, and "-1" is a string. Fixup: https://github.com/luainkernel/lunatik/commit/ca30807369de740b49f8246bbf7367833b401c6c

lib/luafsnotify.c
```diff
@@ -625,6 +652,13 @@ static inline lunatik_object_t *luafsnotify_newevent(lua_State *L)
 	return lunatik_newobject(L, &luafsnotify_event_class, 0, LUNATIK_OPT_NONE);
 }
 
+/* 6.10 made the three priorities an enum; the values did not change */
```

The comment says what 6.10 did; the rule keeps comments in the present, and the guard below already puts the current kernel first. Fixup: https://github.com/luainkernel/lunatik/commit/78267b4019c68d497f4bf4fb73b61d127d672609

lib/luafsnotify.c
```diff
@@ -19,11 +19,20 @@
 * directory the event is about, so `watch:mark`, `watch:find` and `mark:mask`
 * called from a callback resolve their path from the directory cache alone.
 *
+* A permission event parks the syscall in the callback and takes its return
+* value as the answer: `fsnotify.action.ALLOW` lets the access happen,
+* `fsnotify.action.DENY` fails it with `EPERM`, and any other negative errno
+* fails it with that errno. Anything else allows, a callback that returns
+* nothing or raises included, so a rule that fails, or forgets to answer,
+* takes nothing away.
```

fsnotify_open_perm asks twice for an exec, OPEN_EXEC_PERM under __FMODE_EXEC and then OPEN_PERM, each through the marks that carry it (https://github.com/torvalds/linux/blob/v6.8/include/linux/fsnotify.h#L136-L147), so a mark with both bits sees the callback twice and a denial on the first ends it; the paragraph now says so. Fixup: https://github.com/luainkernel/lunatik/commit/e3d2f9f7f607f7853f89683a06f4f09decc71a3b

tests/fsnotify/default.sh
```diff
@@ -0,0 +1,69 @@
...
+# SPDX-FileCopyrightText: (c) 2026 Ring Zero Desenvolvimento de Software LTDA
+# SPDX-License-Identifier: MIT OR GPL-2.0-only
+#
+# Everything that is not a deliberate denial allows.
+#
+# A callback that forgets to return has to allow, or the first rule anyone
+# writes wrong takes a machine's files away. The other three answers are the
+# ones the kernel could not use: security_file_open's caller turns a non-zero
+# return into an error pointer and dereferences it, so a 1 or a -1000000 arrives
+# as a value IS_ERR does not recognise, and a string arrives as no number at
```
lneto: (posted by an agent, not by @lneto)

A 1 never reaches ERR_PTR: do_dentry_open's cleanup turns a positive error into a WARN_ON_ONCE and EINVAL (https://github.com/torvalds/linux/blob/v6.8/fs/open.c#L1012-L1014), and on the read path vfs_read would return it as a count (https://github.com/torvalds/linux/blob/v6.8/fs/read_write.c#L467-L469); the value below -MAX_ERRNO is the pointer IS_ERR misses. The clamp stays; the header and the README entry say the traced mechanism. Fixup: https://github.com/luainkernel/lunatik/commit/0856684cb65425c4b1a1c8009b4a03e851282d3e

tests/fsnotify/deny.sh
```diff
@@ -0,0 +1,72 @@
...
+
+perm_begin 6 "fsnotify/deny"
+
+echo classified > "$MOUNT/secret"
+echo readable > "$MOUNT/public"
+echo boundary > "$MOUNT/edge"
+
+mark_dmesg
+run_script "$SCRIPT"
+denied=$(cat "$MOUNT/secret" 2>&1)
```
lneto: (posted by an agent, not by @lneto)

tools/checks/machine-leak.sh, the commit gate, reads a cat of a file named secret as a credential read out of a file and refuses the commit, so the file is gated here, as in the other tests; deny.sh's grep takes $KTAP_ERRORS in the same fixup because the gate refuses a commit of this file before the rename. Fixup: https://github.com/luainkernel/lunatik/commit/dc907a204fa34da79cefd5e58a491d741408b7d9

tests/fsnotify/error.sh
```diff
@@ -0,0 +1,62 @@
...
+ktap_pass "a callback that raises allows the access"
+
+echo "$raised" | grep -qE "error\.lua:[0-9]+: the guard raised" || \
+	fail "the raise was not logged: $(echo "$raised" | grep -F 'error.lua')"
+ktap_pass "the error the callback raised is logged"
+
+[ "$second" = content ] || fail "the second open failed: $second"
+echo "$again" | grep -qF "error test: $MOUNT/gated mask 10000" || \
+	fail "the watch stopped delivering after the raise: $(echo "$again" | grep -F 'fsnotify error test')"
+ktap_pass "the watch keeps delivering and allowing after a raise"
```
lneto, line 58: (posted by an agent, not by @lneto)

The one test that reads its log without $KTAP_ERRORS, since the Lua error is the point, so a verdict regression on the raise path that returned the pcall status, a positive 2, would trip do_dentry_open's warning and the test would report only a failed open; a case greps the other alternatives. Fixup: https://github.com/luainkernel/lunatik/commit/64b5a88923d11ae3311cdcfb7d0d4d6a15cc930a

tests/fsnotify/run.sh
```diff
@@ -8,7 +8,8 @@
 # Usage: sudo bash tests/fsnotify/run.sh
 
 DIR="$(dirname "$(readlink -f "$0")")"
-TESTS="open nomask child kinds mask marks inside identity overlap expired reentrancy thread lifetime context"
+TESTS="open nomask child kinds mask marks inside identity overlap expired reentrancy thread lifetime context\
+	allow deny default exec access error sleep"
```
lneto: (posted by an agent, not by @lneto)

Two cells the matrix left open. upgrade: a mark placed for FS_OPEN and set to FS_OPEN_PERM through mark:mask, the reason the body gives for the priority on every group, which on 6.10 and later is where a priority taken on the first permission mask would deliver nothing; https://github.com/luainkernel/lunatik/commit/4c562155084c82c3bcc21bf05560009676bb3bb5. directory: OPEN_PERM with EVENT_ON_CHILD on a directory denying a file inside through the parent iterator while the directory's own open stays allowed, and ACCESS_PERM on a directory denying its listing, which iterate_dir asks for (https://github.com/torvalds/linux/blob/v6.8/fs/readdir.c#L99), while a file inside still reads; https://github.com/luainkernel/lunatik/commit/b54d0458db3bb178e38fba3f4018d040618532b2. The five new scripts also read the log with $KTAP_ERRORS: https://github.com/luainkernel/lunatik/commit/249d284cc022859cb62b9fdd50ebeb457f2343ad

tests/README.md
```diff
@@ -103,8 +103,9 @@ Covers the `crypto` module: `shash`, `skcipher`, `aead`, `rng`, `hkdf`,
 
 Every mark in this suite goes on `/tmp/lunatik-fsnotify`, created and removed
 by the test: a mark outside a scratch subtree is what makes a machine unable to
-read its own files. A mount or superblock mark reaches every file it covers, so
-the test that places one mounts its own tmpfs there and marks that.
+read its own files. A mount or superblock mark reaches every file it covers, and
+a permission mark decides whether an access happens at all, so the tests that
+place either mount their own tmpfs there and mark that.
```
lneto: (posted by an agent, not by @lneto)

context.lua places a permission mark on the scratch directory under /tmp, not on a tmpfs, and allows; the rule as written said every permission mark is on a tmpfs the test mounts, so it names the exception. Fixup: https://github.com/luainkernel/lunatik/commit/78aa791717f8155c024344672ee87c06099c2753

doc/design/fsnotify/testing.md
```diff
@@ -117,7 +117,19 @@ that denies everything: a test suite that only asserts denials passes on a modul
...
 The suite runs on one kernel. The version drift table in `kernel-notes.md` covers three interfaces
-that changed inside 6.x, so phase 4 is not done until permission events have been seen working on
-both a 6.8 kernel and a 6.14 or newer one — the priority gating means a wrong `group->priority` shows
-up as "no events at all" on the newer kernel and as "works fine" on the older one.
+that changed inside 6.x, and the priority gating means a wrong `group->priority` shows up as "no
+events at all" on a 6.10 or newer kernel and as "works fine" on an older one.
+
+What has been seen working: 6.12 (`6.12.88+deb13-amd64`), which is on the gated side of that change,
+so the allow, deny, exec, access, default, error and sleep cases all exercise the priority as well as
+the verdict. Not yet run: a 6.8 kernel, where the gate does not exist and the same code has to keep
```
lneto: (posted by an agent, not by @lneto)

The 6.8 run is done, on this host; the note records it and keeps 6.14 as the unverified side. Fixup: https://github.com/luainkernel/lunatik/commit/be0886a35236d8008bf2dd6af76bf9b3f1d1fb53

lneto (commented):

Verdict: merge with the eleven fixups on this branch folded; the base is master now.

@lneto rebased onto master after #851 merged: the one commit that is this pull request replays on 336e9414f with its tests and notes unchanged and its C resolved against master's shape, the incallback registry flag, luaL_checkoption in checkkind, a checkmask that returns the mask, and fsnotify_alloc_group(ops, 0) with no GROUP_USER guard, so the priority is set in newwatch right after the group exists. The body still opens with "Merges after #851", which is on master now.

Read against the kernel: fsnotify() stops at the first group that answers non-zero for a permission event and propagates it (https://github.com/torvalds/linux/blob/v6.12/fs/notify/fsnotify.c#L606-L616); fsnotify_update_sb_watchers reads group->priority when a mark is added and never again, so the priority has to precede the first mark and cannot change under it (https://github.com/torvalds/linux/blob/v6.12/fs/notify/mark.c#L161-L186), and from 6.10 fsnotify_file() drops a permission mask unless a content-priority group marked the object (https://github.com/torvalds/linux/blob/v6.10/include/linux/fsnotify.h#L121-L129), the gate the notes placed at 6.14; the priorities are defines before 6.10 and an enum from it at the same values (https://github.com/torvalds/linux/blob/v6.6/include/linux/fsnotify_backend.h#L208-L211, https://github.com/torvalds/linux/blob/v6.10/include/linux/fsnotify_backend.h#L183-L188); the hooks sit under CONFIG_FANOTIFY_ACCESS_PERMISSIONS from 6.8 and are unconditional at 6.6 (https://github.com/torvalds/linux/blob/v6.8/include/linux/fsnotify.h#L103, https://github.com/torvalds/linux/blob/v6.6/include/linux/fsnotify.h#L105), so the refusal is a policy on 6.6 and 6.7 and a necessity after; the open verdict returns through security_file_open (https://github.com/torvalds/linux/blob/v6.12/security/security.c#L3113) to path_openat's ERR_PTR, and MAX_ERRNO bounds what IS_ERR recognises (https://github.com/torvalds/linux/blob/v6.8/include/linux/err.h#L18-L28). One claim in the body does not hold: a positive return never reaches ERR_PTR as itself; do_dentry_open turns it into a WARN_ON_ONCE and EINVAL (https://github.com/torvalds/linux/blob/v6.8/fs/open.c#L1012-L1014) and vfs_read would return it as a byte count (https://github.com/torvalds/linux/blob/v6.8/fs/read_write.c#L467-L469); the clamp is right, the stated reason is right only for the value below -MAX_ERRNO, and the prose in default.sh and tests/README.md now says so. The lua_pcall error path leaves the error object where lunatik_handle's settop pops it, a yield inside the callback lands in the error arm with the verdict still ALLOW, and a permission event in flight while the watch stops cannot reach a freed watch, since every mark holds the group and the reaper drops it after the SRCU grace period.

Eleven fixups, anchored inline. Built clean on 6.8, make doc-site exits zero with the action table rendered, and the fsnotify suite ran on the final head: 23 tests, 116 pass, no failure, 8 skips where kinds needs 6.10, with CONFIG_FANOTIFY_ACCESS_PERMISSIONS=y, which is the 6.8 run testing.md had listed as not yet done. An independent agent reviewed this branch against the same sources; its findings are the last seven fixups, two of them tests the matrix had left open: a mark whose mask gains a permission bit later, the design's own reason for setting the priority on every group, and a directory mark reaching a child's open and its own listing. The reentrant permission event, the runtime closing during one, and a permission mark on a mount or superblock are covered by construction, the first two by the isowner drop and the preset ALLOW, the third by the same send_to_group merge kinds proves for FS_OPEN. Not verified here: a 6.14 or newer kernel, where the gate's answer is cached per open, as testing.md says.

PR #853, opened 2026-09-12
examples: log what changes in a directory with fsnotify

lneto:
Part of the fsnotify stack, #657. One of the two examples #664 asks for.

Issue #664, opened 2026-08-03
fsnotify: examples and documentation pass


lneto:
Part of the fsnotify epic (#657). Closes it out.

## Scope

Two examples, one per regime, in `examples/`:

**Integrity monitor** (notification): watch a directory subtree, log writes and attribute changes with
the file name, inode and the pid that made them. This is the "what happened" use, and the one that
needs no special kernel configuration.

**Exec allowlist** (permission): deny `FS_OPEN_EXEC_PERM` for anything not in a `set`, over a scratch
subtree. This is the "what may happen" use. It ships as an allowlist over a confined scope, never as a
system wide default deny, and the script says so in a comment at the top.

Then a documentation pass: the module in the README table, LDoc reviewed as a whole rather than per
patch, and the api document in `doc/design/fsnotify/` updated to describe what was actually built
where it diverged from the proposal.

## Acceptance criteria

- [ ] both examples run from `lunatik run`, do what they claim, and stop cleanly
- [ ] the enforcement example cannot deny anything outside its scratch scope
- [ ] `examples_install` installs them; the README describes both with the command to run them
- [ ] a test per example, asserting the observable behaviour, wired into the suite
- [ ] the LDoc for the module reads as one document: no method documented in terms of an earlier
      phase's design that changed on the way
- [ ] `doc/design/fsnotify/api.md` matches the shipped API, and its open questions section is
      resolved rather than left hanging
- [ ] the whole suite passes: `sudo lunatik test`

## Notes

Write the examples early enough that what they teach can still change the API. That is what happened
in the conntrack project: the demo changed the shape of the API once, and it was cheap because it was
written before the API was frozen.

Effort: small to medium.

PR #854, opened 2026-09-12
examples: an fsnotify allowlist for exec over one directory

lneto:
Merges after #853. The other of the two examples #664 asks for, and the one that exercises the verdict path end to end.

Part of the fsnotify stack, #657.


lneto (commented):

@lneto, #854 after the rebase onto master, one commit of its own. Built, installed and run on the head three times, the last with every fixup: fsnotify suite 124 pass, 0 fail, 8 skip (the mount kind before 6.10), execguard 8 of 8, the example running from where examples_install puts it. A second reader reviewed the head independently, tried to refute the two kernel traces below and could not, and found the case the last fixup adds.

Five fixups, one per finding, each on its line. Mergeable once they are folded, with three things at the squash: the body's second paragraph ("Everything the rule cannot match allows, including an event carrying no entry name ...") describes the arm the first fixup removes and, read against the code, states the inverse of an allowlist; the fourth paragraph can add the copy one level below the scope; and 2507926bd carries the gmail.com author where the tree uses ringzero.com.br, which an autosquash keeps, so `--reset-author` there.

Not covered, and why: a symlink or hard link into the scope, since the verdict keys on the final dentry's name, the fsnotify model rather than the example's contract; and a disconnected dentry, open_by_handle_at then execveat with AT_EMPTY_PATH, which fsnotify_parent reports to the inode, superblock and mount marks alone (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/fsnotify.h?h=v6.8#n74), a limit of every EVENT_ON_CHILD mark, traced and not reproduced, filed as #1021.

examples/execguard.lua
```diff
@@ -0,0 +1,32 @@
...
+local format = string.format
+
+local SCOPE <const> = "/tmp/lunatik-execguard"
+
+local allowed = set.new({"true", "false"})
+
+local function guard(_, event)
+	local name = event:name()
+
+	if allowed:has(name) then
```
lneto, line 22: (posted by an agent, not by @lneto)

The `name == nil` arm cannot run for this mark. An exec never opens the marked directory itself: may_open refuses S_IFDIR with MAY_EXEC (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/namei.c?h=v6.8#n3252) before do_open reaches vfs_open (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/namei.c?h=v6.8#n3643) and, through security_file_open, the permission hook (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/security/security.c?h=v6.8#n2937); and a child's event carries the child's name through __fsnotify_parent (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/notify/fsnotify.c?h=v6.8#n224). On 6.6 uselib is a second FS_OPEN_EXEC_PERM producer and takes the same do_filp_open path. `set:has(nil)` would raise and the module would answer ALLOW; neither happens. Fixup: https://github.com/luainkernel/lunatik/commit/411de89ae

examples/execguard.lua
```diff
@@ -0,0 +1,32 @@
+--
+-- SPDX-FileCopyrightText: (c) 2026 Ring Zero Desenvolvimento de Software LTDA
+-- SPDX-License-Identifier: MIT OR GPL-2.0-only
+--
+-- An allowlist for exec over one directory: the mark is on that directory's
```
lneto, line 5: (posted by an agent, not by @lneto)

The header repeated the README's paragraph on the mark kinds; what stays is the one reason the mark is on the inode. Fixup: https://github.com/luainkernel/lunatik/commit/36a7363b7

tests/fsnotify/execguard.sh
```diff
@@ -0,0 +1,100 @@
...
+ktap_pass "the same name one level below the marked directory still runs"
+
+grep -qF "execguard: denied blocked to pid $pid" <<< "$output" || \
+	fail "the rule did not name what it refused to $pid: $(grep -F 'execguard:' <<< "$output")"
+ktap_pass "the refusal names the entry and the pid it was refused to"
+
+[ "$after" -eq 0 ] || fail "the refused program did not run after the example was stopped (exit $after)"
+ktap_pass "the rule ends with the example that made it"
+
+errs=$(grep -E "$KTAP_ERRORS" <<< "$output" || true)
```
lneto, line 94: (posted by an agent, not by @lneto)

exec.sh, deny.sh and fsmonitor.sh grep `$KTAP_ERRORS`, which also names an oops and a WARNING; this test grepped the Lua pattern alone. Fixup: https://github.com/luainkernel/lunatik/commit/46bb8bd50

tests/fsnotify/execguard.sh
```diff
@@ -0,0 +1,100 @@
...
+cp /bin/true "$OUTSIDE/blocked" || fail "could not copy a program outside the scope"
+cp /bin/true "$SCOPE/sub/blocked" || fail "could not copy a program below the scope"
+
+"$SCOPE/blocked" || fail "the copied program does not run before anything marks it"
+
+mark_dmesg
+run_script "$SCRIPT"
+"$SCOPE/true"
+listed=$?
+refused=$( { echo "$BASHPID"; exec "$SCOPE/blocked"; } 2>&1 )
```
lneto, line 59: (posted by an agent, not by @lneto)

The refusal's pid was asserted as any number. The exec in a subshell after its own `echo $BASHPID` is the task the callback sees, since luafsnotify_pid reads task_pid_nr(current) inside the execve, so the log line is asserted with that pid; a failed exec in a non-interactive subshell prints the error and exits 126, which the two cases before it keep matching. Fixup: https://github.com/luainkernel/lunatik/commit/47c8190f6

tests/fsnotify/execguard.sh
```diff
@@ -0,0 +1,100 @@
...
+status=$?
+pid=${refused%%$'\n'*}
+outside=$("$OUTSIDE/blocked" 2>&1)
+outsidestatus=$?
+below=$("$SCOPE/sub/blocked" 2>&1)
```
lneto, line 64: (posted by an agent, not by @lneto)

The README and the body say the rule reaches only an entry directly inside the directory, and the test proved a sibling directory, not one level below. A copy at `$SCOPE/sub/blocked` runs: `sub` carries no mark, so __fsnotify_parent finds no watched parent (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/notify/fsnotify.c?h=v6.8#n185) and SCOPE's marks never reach the iterator (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/notify/fsnotify.c?h=v6.8#n302). The second reader's finding and fixup: https://github.com/luainkernel/lunatik/commit/c16301439

PR #861, opened 2026-09-12
checks: a core change says so in its subject

lneto:
A core change that rides inside a binding's commit, under a subject naming only the binding, is what a review of #848 found: the lock-owner scheme in lunatik.h and doc/capi.md shipped as part of "fsnotify: watch filesystem events from Lua". The rule is not that core and binding never mix; AGENTS.md sanctions a core check adopted by four bindings as one commit. The rule is that the subject says so.

core-subject.sh takes a commit's file list as arguments and its subject on stdin, and fails when the list carries a core file and a file outside the core while the subject names neither. The Checks workflow runs it per commit over the pull request's range, since the squashed diff it already stages cannot see commit boundaries. Over the last 400 commits of master it fires on 9, 4 of them the shape it is for; the noise is docs sweeps and autogen migrations whose only core file is doc/capi.md, kept in because dropping it blinds the check to a new C API entry.


lneto:

PR #871, opened 2026-09-12
examples: add workflow scheduling

sneaky-potato:
Add a `workload` example showing a `sched_ext` scheduler whose task classification is delegated to Lua. The eBPF struct_ops program creates dispatch queues and caches each task's `(dsq, slice)` decision in a BPF hash map; on a cache miss it calls the `bpf_luasched_run` kfunc, which invokes the policy registered with `sched.attach()`. The Lua policy matches task commands and sets the dispatch queue and slice through `sched.ctx`.

The example includes a minimal helper header to avoid depending on the full `scx/common.bpf.h` compatibility layer.


examples/workload/scx_min.bpf.h
```diff
@@ -0,0 +1,44 @@
+/*
```
lneto, line 1: isn't this an autogenerated file?

examples/workload/scheduler.c
```diff
@@ -0,0 +1,89 @@
...
+	struct task_class received_cls = { .dsq = DSQ_DEFAULT, .slice_ns = SCX_SLICE_DFL };
+
+	int ret = bpf_luasched_run(runtime, sizeof(runtime), p, &received_cls);
+
+	if (ret) {
+		received_cls.dsq = DSQ_DEFAULT;
+		received_cls.slice_ns = SCX_SLICE_DFL;
+	}
+
+	bpf_map_update_elem(&task_classes, &pid, &received_cls, BPF_ANY);
```
lneto, line 69: are we always updating the map? isn't there any advantage on updating the map from the Lua script only for the matched patterns?

sneaky-potato: Today every miss gets cached, including the default fallback. It caches negative results, so unmatched tasks don't call into Lua on every enqueue. If you only cached matches, every non-matching task would hit Lua each time it's enqueued, which defeats the cache.

examples/workload/scx_min.bpf.h
sneaky-potato: Not exactly, this header is hand-written. It holds the `BPF_STRUCT_OPS` macros and the kfunc prototypes that upstream ships in `scx/common.bpf.h`

We could make this common as `tests/sched` also uses it.

PR #872, opened 2026-09-12
docs: design notes for compiling Lua to eBPF

lneto:
Lunatik reaches eBPF through a trampoline: a C stub calls a kfunc that runs a Lua callback in the kernel interpreter, and the verifier sees twelve lines of C. This adds `doc/design/luaebpf/`, the design for the other direction: a Lua function compiled into an eBPF program that the verifier checks and the kernel runs natively, with the trampoline kept as an explicit escape hatch the program calls where it chooses. The notes settle where the loader lives (the `lunatik` CLI over `bpf(2)`, since a module-side load through `kern_sys_bpf` returns no verifier log, cannot resolve the tree's own kfuncs or a pinned map by path, and imports a symbol namespaced against the use), which Lua runs the compiler (the kernel-configured 5.5 that `lunatikc` builds, so what it compiles is what the kernel would have interpreted), that a precompiled BPF runtime library is in while a second back end and a second loader are out, and what the compiler displaces: the C stubs that only call the kfunc, their Makefiles, and the `bpftool` and `tc` steps in the README and the suites. Every kernel claim is cited at v7.2, and the phases are tracked on the Lua to eBPF board.


lneto:

PR #890, opened 2026-09-13
checks: list the examples a change to a binding has to run

lneto:
#795 and #837 passed the probe suite, kprobe_concurrent included, and neither review ran systrack; its first run on the merged code took the host down with an unexpected single-step brk (traced separately). The suite covers what a test author thought of; an example is the binding at the rate a user drives it.

`tools/checks/examples-touched.sh` maps each changed binding to the examples that require it and prints what to run. The review skill takes that list as `args.examples`, and its build phase runs whichever of them `args.validated` does not carry, so a head validated by the suite alone no longer skips the examples. `AGENTS.md` lists the check.

The `Checks` workflow line is a separate push, since the token cannot push a workflow file.


lneto:

PR #894, opened 2026-09-14
checks: list what a changed lunatik.h primitive reaches

lneto:
#891's first shape fixed a kprobe handler writing an `rcu.table` by moving every softirq hook, netfilter and XDP included, from bottom halves off to interrupts off, and said so in a footnote of the pull request. A primitive every class inherits is changed for the caller that needs it, in the arm that caller takes, and what else moves is the first line of the commit body.

`tools/checks/blast-radius.sh` maps each primitive a diff touches in `lunatik.h` to the classes whose opt flags select an arm in it and to the files that call it: on the first shape it lists 9 softirq and 4 hardirq classes under `lunatik_locker`. `AGENTS.md` carries the rule, with a second reviewer before the pull request opens, and the review skill's hunt names the arm the bug takes against the arms the fix moves.

The `Checks` workflow line is a separate push, since the token cannot push a workflow file.


lneto:

PR #895, opened 2026-09-14
lunatik: the lock operations in lunatik_lock.h

lneto:
The five lock operations and the predicate that picks the arm are the one place the execution contexts meet, and they read as a unit, as `lunatik_val.h` does for the values. `lunatik.h` includes the new header after the object struct, which is all it needs, and before `lunatik_setobject`, which calls `lunatik_newlock`, so it cannot sit at the end with the value and percpu headers. The move is byte-identical; no `make` target installs a header and no C consumer in or out of tree names the functions.

Verified: `make` clean, full suite pass:161 fail:0 skip:3 with `probe/kprobe_concurrent` included, and every runnable example loaded, run and stopped on the build with a clean dmesg.


lunatik_lock.h
```diff
@@ -0,0 +1,58 @@
+/*
+* SPDX-FileCopyrightText: (c) 2026 Ring Zero Desenvolvimento de Software LTDA
```
lneto: wrong copyright date

lneto: _Agent reply (Claude Code), posted from @lneto's account._

> wrong copyright date

Right: the code moves out of `lunatik.h`, whose header reads `(c) 2023-2026`, and `lunatik_trylock` predates this year, so the range travels with it. Amended in place, 165364dd9bf7362841c978e61e232a7dfbba8c4d.

PR #897, opened 2026-09-15
probe: a hit resolves the handler before it builds the closures

### LINKED ISSUE #844: probe: a hit builds the closures for the handler the script did not define
`probe.new` installs both `luaprobe_pre_handler` and `luaprobe_post_handler` on every kprobe, so both fire even when the script defines only `pre`. `luaprobe_handler` then looks the handler table up, pushes the symbol and builds the closures, and only afterwards resolves the handler, which for the one the script did not define is `lunatik_nop`. A `pre`-only probe pays a registry lookup and the closure allocations per hit, under `GFP_ATOMIC`, for a call that does nothing.

Resolving the handler first and returning when it is the nop is the fix, and it is master's code, not a feature branch's. Master builds one closure per hit; #837 makes it two, which is what brought it into view.

What this is not is a measurement: nothing here timed it. The claim is an extra allocation per hit on a hardirq path, not a hot path.

Raised in the review of #837.

lneto:
A kprobe hit built one closure per entry of `luaprobe_closures`, each a `CClosure` through the runtime allocator, before it read the handler out of the table and fell back to `lunatik_nop`. A script with only a `pre` handler paid that on every `post` hit.

The handler is now read first, and a hit leaves when the script did not define one. `tests/probe/handlers` covers the four handlers tables `probe.new` takes, one script per row; the `post` half of a hit had no coverage before, since nothing in the tree registered a `post` handler.

Closes #844


lib/luaprobe.c
```diff
@@ -86,24 +86,27 @@ static int luaprobe_handler(lua_State *L, luaprobe_t *probe, const char *handler
 	int nclosures = ARRAY_SIZE(luaprobe_closures);
 	int i;
 
+	if (lua_getfield(L, base, handler) != LUA_TFUNCTION) /* base + 1 */
```
lneto, line 89: _Agent review (Claude Code), posted from @lneto's account._ The early return short-circuits the cost; what would make it unreachable is not installing
`post_handler` at all when the table has no `post`, since
[line 359](https://github.com/luainkernel/lunatik/blob/4fdd5e21031b5e1113ff131d0f7b4eaf12d5a2d1/lib/luaprobe.c#L359)
sets it unconditionally and that is what gives a `pre`-only script post hits in the first place. I do not
think it belongs here: `kp->pre_handler != NULL` is the "still registered" sentinel at
[line 139](https://github.com/luainkernel/lunatik/blob/4fdd5e21031b5e1113ff131d0f7b4eaf12d5a2d1/lib/luaprobe.c#L139)
and [line 268](https://github.com/luainkernel/lunatik/blob/4fdd5e21031b5e1113ff131d0f7b4eaf12d5a2d1/lib/luaprobe.c#L268),
so `pre_handler` cannot become conditional without replacing it, a `post`-only script would still pay the
lookup on every `pre` hit, and reading the table once at `probe.new` narrows the contract, since today a
`post` added afterwards starts firing. So this is the right shape for this issue, and the other one is a
change of its own — worth an issue if you want it kept in view, since it is also what
[kernel/kprobes.c:685](https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L685) and
[:1104](https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L1104) charge for a `post_handler`,
neither of which this host can show (arm64 selects neither `KPROBES_ON_FTRACE` nor `OPTPROBES`).

lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #897: it can merge as it stands.

@lneto This one reads clean. I checked it out at `4fdd5e210`, built, installed and ran the suites and
the examples on 6.8.0-138 aarch64.

The index shift is right in all three places that name the closures. After `lua_pcall(L, 1 + nclosures,
0, 0)` the top is back at `base + 1 + nclosures`, so `luaprobe_dropregs(L, base + 2 + i)` still addresses
the template closures, and it does on the error path too, where `pcall` leaves the message above them —
the indices are absolute and `lua_pushnil` precedes `lua_setupvalue`. The early `goto out` leaves the
field value on the stack and `lunatik_handle`'s `lua_settop(L, n)` drops it, as it already did for the
registry check above. `lunatik_optcfunction` and `lunatik_nop` still have callers in
[luadevice](https://github.com/luainkernel/lunatik/blob/4fdd5e21031b5e1113ff131d0f7b4eaf12d5a2d1/lib/luadevice.c#L92)
and [luahid](https://github.com/luainkernel/lunatik/blob/4fdd5e21031b5e1113ff131d0f7b4eaf12d5a2d1/lib/luahid.c#L192),
so nothing in `lunatik.h` went dead.

Suites: probe 21 pass, full run pass:165 fail:0 skip:3, the three skips being sched_ext on a kernel
without it. Examples run, not only loaded: `examples/dropreason/monitor` caught a NO_SOCKET drop with the
register trace from `dump()`, and the REPL read `drops.NO_SOCKET` back as 1; `examples/systrack/device`
armed 306 kprobes over 0 idle and gave them all back. Both go through the new lookup, and dropreason's
handlers table carries a non-function field, so the table with no `post` is exercised by the example too.

One note rather than a request, on the inline comment below.

PR #898, opened 2026-09-15
probe: a repeated target is aggregated, and systrack no longer repeats one

### LINKED ISSUE #836: probe: a percpu script cannot probe two targets that resolve to the same address
`examples/systrack/probe.lua` probes every entry of `syscall.table`. On 6.8.0-138 aarch64 two of its 306 slots hold one address, `nfsservctl` and `lookup_dcookie`, both left pointing at the not-implemented stub.

As a plain hardirq runtime that is 306 kprobes, two of them at the same address, and both handlers run on a hit:

```
# lunatik run examples/systrack/probe hardirq
# grep -c "" /sys/kernel/debug/kprobes/list
306
```

As a percpu script, after #795, the second `probe.new` on that address finds the registration the first one added to the set's list, sees that this runtime already filed handlers under it, and raises, so the script does not load:

```
# lunatik run examples/systrack/probe hardirq percpu
/lib/modules/lua/examples/systrack/probe.lua:15: probe already registered
```

Nothing is left armed and the script is not registered, so the refusal itself is clean. What it costs is that the example cannot move to percpu as written, and neither can any script that probes a list of targets it has not deduplicated.

The refusal is the one #795 documents on `probe.new`, and it follows from the key a runtime finds its handlers by: they are filed under the shared kprobe, https://github.com/luainkernel/lunatik/blob/bcc01f5282a4add36941870452e1995f1f9cbc36/lib/luaprobe.c#L315, so a runtime can hold one handler table per shared registration and a second one on the same target is refused, https://github.com/luainkernel/lunatik/blob/bcc01f5282a4add36941870452e1995f1f9cbc36/lib/luaprobe.c#L285-L286.

Lifting it means letting a runtime attach several handler sets to one shared registration and dispatching to each. That is the same question `luanetfilter`'s `hook already registered` raises, so it belongs with #834 rather than with #795.

Raised from the third review round of #795.

### LINKED ISSUE #892: probe: two entries of syscall.table share an address, and probe.new aggregates them silently
`examples/systrack` registers one kprobe per entry of `syscall.table`, keyed by address, and at least two of those entries carry the same address. On arm64 the syscall table is filled with `__arm64_sys_ni_syscall` for every number with no implementation ([`arch/arm64/kernel/sys.c`](https://github.com/torvalds/linux/blob/v6.8/arch/arm64/kernel/sys.c#L59)), and the generic table maps [`__NR_lookup_dcookie`](https://github.com/torvalds/linux/blob/v6.8/include/uapi/asm-generic/unistd.h#L74) and [`__NR_nfsservctl`](https://github.com/torvalds/linux/blob/v6.8/include/uapi/asm-generic/unistd.h#L127) to `sys_ni_syscall` explicitly, so `syscall.table.lookup_dcookie` and `syscall.table.nfsservctl` resolve to the same pointer.

A second `register_kprobe` at an address that already has one does not fail. It goes to `register_aggr_kprobe` and [skips `prepare_kprobe`](https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L1645), so the second probe gets no slot of its own, and [`init_aggr_kprobe`](https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L1274) copies the first probe's slot into an aggregate that becomes what `current_kprobe` points at. Both Lua `probe` objects then share one kprobe, and every hit calls both handlers.

Nothing is known to break: the aggregate is freed by `free_aggr_kprobe` under the same `stop_machine` and grace period as a plain probe, and systrack counting one hit under two names is arguably what a reader of that table would expect. What looks worth deciding is the inconsistency: in a percpu script `probe.new` raises "probe already registered" for a repeated address ([`luaprobe_share`](https://github.com/luainkernel/lunatik/blob/924d472a9/lib/luaprobe.c#L337)), while a plain runtime aggregates silently, and the `probe.new` documentation does not mention either. Either behaviour seems defensible; they should probably not differ, and whichever is chosen belongs in the doc block.

Found while tracing the crash on #843.

lneto:
`register_kprobe` aggregates a probe on an address another kprobe already holds: it succeeds with no diagnostic, and every handler attached to that address then runs on a hit. `probe.new` documented only the percpu refusal, which is the narrower case, and `examples/systrack` relied on neither answer: it probed all 306 entries of the syscall table, two of which resolve to one entry point on arm64, so the counts it reported for those two names were each other's.

The doc block now states both answers, and the example probes only single-named entry points, which is also what it took for it to load as a percpu set.

Closes #892
Closes #836


examples/systrack/probe.lua
```diff
@@ -7,12 +7,26 @@ local systab = require("syscall.table")
...
+end
+
+local function count(address)
+	local symbol = names[address]
+	track[symbol] = (track[symbol] or 0) + 1
+end
+
+for address, symbol in pairs(names) do
+	if symbol then
+		probe.new(address, {pre = count})
```
lneto: _Agent review (Claude Code), posted from @lneto's account._ Now that the handler is constant, the handlers table is too, and it can come out of the loop: master had
to build one per syscall because the closure differed, this one does not, and as written 304 one-field
tables stay anchored in the registry for the life of the script, times one runtime per CPU on the percpu
run. Sharing one is safe — `lunatik_registerobject` files it under each probe's own key
([lunatik.h:458](https://github.com/luainkernel/lunatik/blob/7621becdd4b8443e514e8e2f03a6f42ab4997371/lunatik.h#L458))
and nothing writes to it. Fixup:
https://github.com/luainkernel/lunatik/commit/597c9cd160465be905bef3c01260353baff7c29b

I ran both modes against it: 304 armed and released, counts by name, neither collided name reported, the
percpu set loads and stops clean, probe suite green.

lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #898: request changes, one fixup on the second commit.

@lneto One fixup on this one, and the rest holds.

I did not take the doc block on trust. A scratch script registered two probes on
`systab["personality"]` in one plain hardirq runtime: `/sys/kernel/debug/kprobes/list` showed two entries
at the same address, one `setarch` printed both handlers, and the stop left zero. That is
[`old_p = get_kprobe(p->addr)` into `register_aggr_kprobe`](https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L1657)
with [`aggr_pre_handler` walking `p->list`](https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L1198),
and the same shape is at v5.15, so the block holds to the floor. Worth recording why only the percpu path
has anything to refuse: the plain path escapes
[`warn_kprobe_rereg`](https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L1639) because
`luaprobe_own` allocates a fresh `luaprobe_t` per call, so the duplicate is a different `struct kprobe`.

The example: 304 kprobes armed and released in both modes, counts by name, neither `nfsservctl` nor
`lookup_dcookie` reported, and the percpu run loads, which is #836's report. I reproduced the before
state rather than assume it — master's `probe.lua` copied over the installed one and run as a set printed
`probe already registered` at line 15 and armed nothing, and the branch's file was restored and diffed
afterwards. Full suite pass:161 fail:0 skip:3.

The fixup is on the second commit, inline below. There is also a coverage gap I am deliberately leaving
for a follow-up rather than fixing here: nothing asserts the plain half of the contract the block now
states. `tests/probe/percpu_probe.sh` case 3 holds the percpu refusal, and
[`kprobe_concurrent.lua`](https://github.com/luainkernel/lunatik/blob/7621becdd4b8443e514e8e2f03a6f42ab4997371/tests/probe/kprobe_concurrent.lua#L16)
does register 306 probes on 305 addresses, so the aggregate registration and teardown are exercised, but
it asserts only that the stop returns and dmesg is clean. A case for it needs its own `.sh` and a line in
`tests/probe/run.sh`, and that line is the one #897 also edits, so adding it now would put a conflict
between two open PRs for a guard on kernel behaviour the binding cannot easily break. Happy to open the
issue if you want it tracked.

examples/systrack/probe.lua
```diff
@@ -7,12 +7,26 @@ local systab = require("syscall.table")
 local rcu    = require("rcu")
 
 local track = rcu.table()
+local names = {}
 
+-- syscall numbers the kernel does not implement can share one entry point, where a hit belongs to no name
```
lneto: why adding this comment?

examples/systrack/probe.lua
```diff
@@ -7,12 +7,26 @@ local systab = require("syscall.table")
...
 for symbol, address in pairs(systab) do
-	local function handler()
-		track[symbol] = (track[symbol] or 0) + 1
+	if names[address] == nil then
+		names[address] = symbol
+	else
+		names[address] = false
+	end
```
lneto: why not a ternary?

lib/luaprobe.c
```diff
@@ -280,6 +280,11 @@ static int luaprobe_new(lua_State *L);
 * In a percpu script the runtimes share one kprobe per symbol or address: the first
 * registration installs it, the others attach their handlers, and a call reaches the
 * runtime of the CPU it ran on.
+* A target another kprobe already holds is aggregated by the kernel: the registration
+* succeeds and every handler attached to that address runs on a hit, so two symbols the
+* kernel resolves to one address each count the other's calls. The exception is a percpu
+* script repeating a target of its own: its runtimes file one handlers table per shared
+* kprobe, so the second registration is refused with `probe already registered`.
```
lneto: why changing this desc if we are not changing the lib implementation?

examples/systrack/probe.lua
lneto: _Agent reply (Claude Code), posted from @lneto's account._

> why adding this comment?

Dropped. It explained the kernel's own table rather than this script, which is the rule that landed in #907, and where the fact belongs is `probe.new`'s block, which the first commit here already gives it. https://github.com/luainkernel/lunatik/commit/1aa2f43cfae31d52e4b5bc7fed765483592a0ce9

examples/systrack/probe.lua
lneto: _Agent reply (Claude Code), posted from @lneto's account._

> why not a ternary?

Done: `names[address] = names[address] == nil and symbol or false`. It is safe in the form AGENTS.md allows, since the true arm is a `syscall.table` key and never false or nil. The test stays `== nil` rather than `not names[address]`, or an address already marked would take its second name. https://github.com/luainkernel/lunatik/commit/1aa2f43cfae31d52e4b5bc7fed765483592a0ce9

lib/luaprobe.c
lneto: _Agent reply (Claude Code), posted from @lneto's account._

> why changing this desc if we are not changing the lib implementation?

The block states the contract rather than the implementation, and that contract was undocumented: calling `probe.new` twice on one address does not fail, and both handlers then run on a hit, which is what #892 reported as happening silently. Nothing in the binding changes, and the paragraph is now the observable behaviour in three lines instead of the kernel's mechanics:

```
* Probing a target another probe already holds does not fail: both handlers run on a hit,
* so two symbols the kernel resolves to one address each count the other's calls. In a
* percpu script, repeating a target of the set is refused instead.
```

If you would rather this pull request stay the example alone, the commit drops cleanly and #892 stays open for the doc. https://github.com/luainkernel/lunatik/commit/1aa2f43cfae31d52e4b5bc7fed765483592a0ce9

lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #898, second round: it can merge as it stands.

@lneto Your three points are folded, each in the commit that owns it, and the review's own fixup (the constant handlers table out of the loop) with them. Rebased on master with #897 and #899 in.

Run through the new `tools/lunatik-host` lock, so the cycle is not sharing the device or the installed tree with anything: probe suite 21/21 (kprobe_concurrent 2, armed 4, percpu_probe 8, argument 3, handlers 4), and `examples/systrack/device` armed its kprobes over an idle list, counted by name, and gave them all back with a clean dmesg.

PR #899, opened 2026-09-15
probe: record why the handlers table is registered after the kprobe

### LINKED ISSUE #893: probe: the handlers table is registered after register_kprobe, so ordering no longer guards the first hit
Before #795, `luaprobe_new` registered the handlers table and only then armed the probe, with the reason in a comment on the call: "must precede register_kprobe: kprobe may fire immediately". After #795 the registration happens inside [`luaprobe_register`](https://github.com/luainkernel/lunatik/blob/924d472a9/lib/luaprobe.c#L168), which is called from `luaprobe_own`/`luaprobe_share`, and [`lunatik_registerobject`](https://github.com/luainkernel/lunatik/blob/924d472a9/lib/luaprobe.c#L366) runs after it. The comment moved to the new call site and now reads "keyed by the kprobe, which is what the handler looks up", so the old reason is no longer recorded anywhere.

As far as I can trace nothing breaks, because a hit in that gap does not reach `luaprobe_handler` at all: `probe.new` runs from the script body, which finishes before [`lunatik_setready`](https://github.com/luainkernel/lunatik/blob/924d472a9/lunatik_core.c#L287), and [`lunatik_run`](https://github.com/luainkernel/lunatik/blob/924d472a9/lunatik.h#L78) skips the handler while the runtime is not ready. So the guarantee the old ordering gave is now given by readiness instead.

That is a fine answer, but it is not written down, and the next reader of `luaprobe_new` has no way to tell that the order is now free to be either. Two options seem reasonable: restore the old order, which costs nothing, or keep it and say on the call why a hit before the table exists cannot reach the handler.

Found while tracing the crash on #843.

lneto:
`register_kprobe` arms the kprobe before `lunatik_registerobject` files the handlers table, and what makes that window safe is that a hit finds the runtime not ready and is dropped. The reason was recorded only in `af1827434`'s commit body, so the next reader of `luaprobe_new` does not have it.

Restoring the old order is not what this does, and the commit body says why: in a percpu set the later runtimes file their handlers under a kprobe the first one already armed, so no ordering inside `luaprobe_new` reaches that path.

Closes #893


lib/luaprobe.c
```diff
@@ -363,6 +363,7 @@ static int luaprobe_new(lua_State *L)
 
 	object->private = percpu != NULL ? luaprobe_share(L, percpu, &spec) : luaprobe_own(L, runtime, &spec);
 
+	/* register_kprobe already ran: lunatik_run drops every hit until the runtime is ready */
```
lneto, line 366: _Agent review (Claude Code), posted from @lneto's account._ `percpu_probe.sh`'s second case drops its early hits through `lunatik_pin` returning NULL, which is the
publication half. The `lunatik_isready` half — a hit on a plain runtime between `register_kprobe` and
`lunatik_setready` — is exercised by no case, though
[`percpu_probe_early.lua`](https://github.com/luainkernel/lunatik/blob/e0dce2ee6b0dd223eb0548cff10a5a4d0f9892d6/tests/probe/percpu_probe_early.lua#L23)
already spins to widen exactly that window and would only need running as a plain hardirq runtime. The
gap is `af1827434`'s and already on master, so not this PR's to close; flagging it so it is not lost.

lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #899: it can merge as it stands.

@lneto Nothing to change here. I re-traced the invariant in the worktree rather than reading it off the
commit body: `register_kprobe` at
[luaprobe.c:186](https://github.com/luainkernel/lunatik/blob/e0dce2ee6b0dd223eb0548cff10a5a4d0f9892d6/lib/luaprobe.c#L186),
`lunatik_registerobject` at
[:367](https://github.com/luainkernel/lunatik/blob/e0dce2ee6b0dd223eb0548cff10a5a4d0f9892d6/lib/luaprobe.c#L367),
and between them `lunatik_run` reaches the handler only under `lunatik_isready`, which
[lunatik_core.c:265](https://github.com/luainkernel/lunatik/blob/e0dce2ee6b0dd223eb0548cff10a5a4d0f9892d6/lunatik_core.c#L265)
sets false and only
[:287](https://github.com/luainkernel/lunatik/blob/e0dce2ee6b0dd223eb0548cff10a5a4d0f9892d6/lunatik_core.c#L287)
sets true, after the body's `lua_pcall`. For a set, `lunatik_pin` reads a pointer the
[`smp_store_release` at :290](https://github.com/luainkernel/lunatik/blob/e0dce2ee6b0dd223eb0548cff10a5a4d0f9892d6/lunatik_core.c#L290)
publishes afterwards, so "until the runtime is ready" is true on both paths and one word covers both.

The part that would bite hardest if it were false also holds: the body runs under `lua_pcall` with no
runtime lock held, so a hit on the same CPU takes the lock rather than spinning on one the body owns with
interrupts off. And `luaprobe_register` sets `probe->runtime` before it arms, so there is no narrower
window either.

Two comments on one call reads a little heavy, but they carry different reasons and merging them runs
past `comment-style.sh`'s 110 columns, so leaving the existing trailing comment where it is looks right.
Suites green, `.ko` byte-identical to the built one.

One thing to know rather than fix, in the inline below.

PR #900, opened 2026-09-15
probe: say what a handler does instead of stopping the probe

### LINKED ISSUE #869: probe: a handler cannot disarm its own probe, and the tree has two answers to that
#860 makes `probe:stop()` raise from a handler, because `unregister_kprobe` waits on `synchronize_rcu` with interrupts off there. Before it, a handler calling `stop` worked only by stopping the machine, so nothing regresses; but nothing replaces it either, and a probe script now has no supported way to disarm itself from its own handler.

The tree already holds the other answer for the same problem. `luanotifier_stop` does not unregister from the handler: it clears only `registry[private]`, so the handler becomes a silent no-op, and the chain is unregistered in `release` at `lua_close`, in process context. Porting that shape to probe is not free: `luaprobe_handler` treats a missing registry table as an error (`couldn't find probe table`) rather than a silent stop, and the kprobe stays armed, paying a hardirq Lua dispatch per hit until the runtime stops.

Which of the two the probe binding should follow is a design call: refuse, as #860 does and `AGENTS.md` sanctions, or defer as notifier does.

Raised in the review of #860.

lneto:
`probe:stop` raises from a handler, where `unregister_kprobe` would sleep, and its doc block said only that it raises. A script already has a supported way to stop delivering there, and the block now says it: return early on a flag of its own, with the kprobe unregistered when the runtime stops.

This settles the question on refusing rather than deferring, and the commit body carries the reasoning and the constraint that binds any later attempt: a deferred stop must keep the object registered, since Lua's collector runs inside a handler and unanchoring the object there would let `__gc` reach `unregister_kprobe` in hardirq.

Closes #869


lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #900: it can merge as it stands.

@lneto This one is right as it stands, and the call it settles looks right to me too.

"From a handler this raises" is unconditional and traces: `luaprobe_stop` opens with
[`lunatik_checkarmed`](https://github.com/luainkernel/lunatik/blob/13a2b0a09f9b9c703be41cda541321b75c2d2fa2/lib/luaprobe.c#L239),
which raises once the runtime is ready and is an IRQ one, and `probe.new` accepts no other kind. The
second half traces too: `luaprobe_release` reaches `unregister_kprobe` through `luaprobe_put` and
`luaprobe_free`, and a set disarms through
[`luaprobe_disarm`](https://github.com/luainkernel/lunatik/blob/13a2b0a09f9b9c703be41cda541321b75c2d2fa2/lib/luaprobe.c#L206)
before its runtimes close.

The constraint in the commit body is the load-bearing claim and it holds in the vendored Lua:
`singlestep` runs `GCScallfin` and calls `GCTM(L)`, reached from `luaC_checkGC` inside `luaV_execute`, so
a finalizer does run inside the handler's `lua_pcall`, and the chain from `__gc` to `unregister_kprobe`
is real. Nothing stops the collector on that path — `lua_gc(L, LUA_GCSTOP)` appears only around
`lunatik_monitor`'s call.

Worth adding to that, if the question ever comes back: `luanotifier_stop` clears the callback key and
lets the handler become a silent no-op. The same move on a probe would clear the key `luaprobe_handler`
reads at
[luaprobe.c:81](https://github.com/luainkernel/lunatik/blob/13a2b0a09f9b9c703be41cda541321b75c2d2fa2/lib/luaprobe.c#L81),
so every later hit would log `couldn't find probe table`, which
[percpu_probe.sh:106](https://github.com/luainkernel/lunatik/blob/13a2b0a09f9b9c703be41cda541321b75c2d2fa2/tests/probe/percpu_probe.sh#L106)
asserts on as a real-bug signal. That is a third reason the notifier shape does not port, beyond the two
the body already gives.

The answer the block documents is the tree's own idiom, not an invention: `armed.lua` and `argument.lua`
both already guard their handlers on a `done` flag. `armed.sh` case 3 is what holds it, green here along
with the rest of the suite.

PR #905, opened 2026-09-15
lunatik_aux: drop the unreachable BUG_ON in lunatik_lookup

### LINKED ISSUE #902: lunatik_aux: the BUG_ON in lunatik_lookup cannot fire
`lunatik_aux.c:105` is `BUG_ON(__lunatik_lookup == NULL)` immediately after
`__lunatik_lookup = lunatik_cfi_entry(kp.addr)` at `:102`. The branch cannot execute:
`register_kprobe` sets `p->addr` from `_kprobe_addr()` after an `IS_ERR` check and returns non-zero
otherwise, which the `return NULL` at `:100` already covers, and `lunatik_cfi_entry` returns either
`addr` or `addr - ENDBR_INSN_SIZE` (`lunatik_cfi.h:24-28`), reducing to the identity macro where
`CONFIG_X86_KERNEL_IBT` is unset. Neither can be NULL for a valid text address.

A `BUG_ON` that cannot fire misstates the contract of the call above it. Its own branch, or a line
in whatever touches `lunatik_aux.c` next.

lneto:
`lunatik_lookup` ends its kprobe dance with `BUG_ON(__lunatik_lookup == NULL)`, a branch that cannot execute: `register_kprobe` sets `p->addr` only from a non-error `_kprobe_addr`, which the `return NULL` above it already covers, and `lunatik_cfi_entry` returns that address or that address minus `ENDBR_INSN_SIZE`. A `BUG_ON` that cannot fire misstates the contract of the call above it.

The trace is in the commit body, read at 6.8 and at the 5.15 floor. Nothing changes at runtime, so there is no new case: what exercises the function is every module init that resolves an unexported symbol, which `lunatik reload` and the suite already run.

Closes #902


lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #905: it can merge as it stands.

@lneto The branch is what the issue asked for and nothing else: the `BUG_ON` goes, the blank line above it goes with it, and no test, since there is no behaviour to cover.

The unreachability was re-derived rather than read off the issue, at 6.8 and at the 5.15 floor. `register_kprobe` takes `addr = _kprobe_addr(...)` and returns `PTR_ERR(addr)` when it is an error pointer (https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L1634), so `p->addr` is never NULL by the time `lunatik_lookup` reads it, and the early `return NULL` above already covers the failure. `lunatik_cfi_entry` returns either the address or the address minus `ENDBR_INSN_SIZE`, and reduces to the identity where `CONFIG_X86_KERNEL_IBT` is unset, so neither arm can produce NULL. The same shape holds at v5.15 through `kprobe_addr`.

Two things recorded rather than raised: `module-conventions.sh` still exits 1 on `lunatik_aux.c` for the comment after `#else /* CONFIG_KPROBES */`, which reproduces identically on master, and `guard-removed.sh` is silent here because `BUG_ON` is not in its guard list, which is right: this is an assertion the branch proves unreachable, not a crash guard.

CI green on the head, and the merge is clean against master.

PR #908, opened 2026-09-15
tests/probe: a hit before a plain runtime is ready is dropped

### LINKED ISSUE #904: tests/probe: a hit before a plain runtime is ready is dropped
`tests/probe/percpu_probe.sh`'s second case, "a call reaching the shared kprobe before the runtimes
are published is dropped", drops its early hits through `lunatik_pin` returning NULL, which is the
publication half of the invariant. The `lunatik_isready` half, a hit on a *plain* runtime between
`register_kprobe` at `lib/luaprobe.c:186` and `lunatik_setready` at `lunatik_core.c:287`, is
exercised by no case.

`tests/probe/percpu_probe_early.lua` already spins to widen exactly that window; running it as a
plain hardirq runtime and asserting zero hits before the body returns, `COUNT` after, and no
"couldn't find probe table" in dmesg is most of the work.

lneto:
`percpu_probe.sh`'s second case drops its early hits through `lunatik_pin` returning NULL, which is the publication half of "a hit before the runtime is up is dropped". The readiness half, a hit on a plain runtime between `register_kprobe` and `lunatik_setready`, was exercised by no case: for a plain runtime `lunatik_pin` returns the object itself, so readiness is the only thing between arming the kprobe and the handler.

The new ninth case runs `percpu_probe_early.lua` as a plain hardirq runtime, triggers while it is still spinning, and asserts zero handler prints then, `COUNT` from the same trigger once the script is loaded, no `couldn't find probe table` in dmesg, and the kprobe armed on load and gone on stop. The early trigger is not pinned and `trigger()` now pins only when it is given a CPU; the commit body carries the measurement behind that. Removing `lunatik_isready` to watch the case fail was not run here, since it reproduces the hardirq re-entry into a `lua_State` that is mid-`lua_pcall` the guard exists to prevent; what was run is the in-test A/B, where moving the trigger after the body returns makes the case fail with three late hits.

Closes #904


lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #908: it can merge as it stands.

@lneto The case lands where the invariant is structural: third in `percpu_probe.sh`, right after the percpu "published" case it completes, reusing `percpu_probe_early.lua` rather than duplicating it, with `ktap_plan` and the skip list moved together and the `tests/README.md` entry extended. No `run.sh` change, so it never contended with #897.

For a plain runtime, readiness is the only way a hit is dropped, which is what makes the assertion exact: `lunatik_run` skips the handler on `lunatik_pin` returning NULL or on `!lunatik_isready`, and `lunatik_pin` returns the object itself for a non-percpu one, so what the case measures is the `ready` half that `percpu_probe`'s second case never reached.

One deviation from the plan, measured rather than argued: the trigger is left unpinned. A pinned trigger waits out the script body that holds the CPU, which flaked 2 in 20 for the author and 8 in 30 when the review reproduced it, against 0 in 40 unpinned; the host runs `CONFIG_PREEMPT_VOLUNTARY` with no `preempt=` override, so the mechanism is the compiled default, and `tests/runtime/percpu_netfilter.sh` already carries the same unpinned shape for the same kind of assertion.

Verified here at the head, through the host lock so nothing else was on the device: probe suite green, and the case passes.

PR #909, opened 2026-09-15
probe: install a post handler only when the script defines one

### LINKED ISSUE #901: probe: post_handler is installed whether the script has one or not
`luaprobe_new` sets `.post_handler = luaprobe_post_handler` unconditionally
(`lib/luaprobe.c:356`), so every lunatik kprobe asks the kernel for a mode most scripts do not use.
In the kernel that is not free, and none of the three consequences is a performance adjective:

- `kernel/kprobes.c:685-686`, "kprobes with 'post_handler' can not be optimized": the probe stays a
  breakpoint instead of becoming a jump.
- `kernel/kprobes.c:1104` and `:1134`, `bool ipmodify = (p->post_handler != NULL);` picks
  `kprobe_ipmodify_ops` over `kprobe_ftrace_ops` when arming through ftrace, so lunatik takes an
  `FTRACE_OPS_FL_IPMODIFY` reservation on the function. That reservation is what conflicts with
  livepatch, so a `pre`-only Lua probe on a live-patched function is refused where it need not be.
- `kernel/kprobes.c:1266`, `add_new_kprobe` calls `unoptimize_kprobe(ap, true)` when the new probe
  has a `post_handler`, so a lunatik probe added at an address someone else optimized de-optimizes
  their probe.

Setting `spec.kp.post_handler` from the handlers table at `probe.new` removes all three for a
`pre`-only script. It is not a free change and that is why it is an issue rather than a patch:

- `kp->pre_handler != NULL` is the binding's "still registered" sentinel (`lib/luaprobe.c:136` in
  `luaprobe_delete`, `:265` in `luaprobe_enable`), so `pre_handler` cannot be made conditional
  without replacing that sentinel first;
- `luaprobe_delete` carries "disarm picks the ftrace_ops from post_handler; clear handlers only
  after unregister" (`:137`), so the teardown order is already keyed on that field;
- it narrows the contract. Today the handlers table is read on every hit, so a `post` added after
  `probe.new` starts firing. Afterwards it would not. That is a behaviour change to decide here,
  not inside a fix for #844.

Not observable on the maintainer's aarch64 host: `CONFIG_KPROBES_ON_FTRACE`, `CONFIG_OPTPROBES` and
`CONFIG_LIVEPATCH` are unset, and arm64 selects neither `HAVE_KPROBES_ON_FTRACE` nor
`HAVE_OPTPROBES` in 6.8. Every consequence above is x86-only on the kernels available there, so a
branch for this has to say "unverified on this host".

When it is done it wants a `tests/probe/handlers.sh` row asserting that a `post` added after
`probe.new` does not fire, which is why it follows #897 rather than sharing it.

lneto:
`luaprobe_new` set `.post_handler` unconditionally, so every lunatik kprobe asked the kernel for a mode most scripts do not use. The kernel reads that field in three places, the same at 6.8 and at the 5.15 floor: it leaves a probe that has one unoptimized, it picks `kprobe_ipmodify_ops` over `kprobe_ftrace_ops` when arming through ftrace, taking the `FTRACE_OPS_FL_IPMODIFY` reservation that `__ftrace_hash_update_ipmodify` refuses beside livepatch's, and it unoptimizes an aggregate that a probe with one joins. The second is not a cost but a refusal: a `pre`-only Lua probe on a live-patched function was turned away for a mode the script never asked for. The commit body carries the line references.

`probe.new` now reads the handlers table once, where it registers, and installs the post handler only when the table holds a `post` function. That narrows the contract asymmetrically, which the doc block states: a `post` added to the table afterwards no longer fires, a `pre` added afterwards still does, because `kp->pre_handler != NULL` is the binding's "still registered" sentinel and cannot become conditional without a teardown change on the hardirq path. A percpu set keeps one kprobe and one handlers table per runtime, so `luaprobe_share` refuses a set whose runtimes disagree instead of letting the first registration decide for the rest. `tests/probe/handlers` gains a row for each. None of the three consequences is reachable on this host, where arm64 at 6.8 selects `HAVE_KPROBES` alone and `CONFIG_LIVEPATCH` is unset, so they are unverified here; what is verified is the behaviour change, since arm64 does call `post_handler`, and both new rows fail against the old `luaprobe_new` and pass against this one.

Based on #897, which creates `tests/probe/handlers.sh`. Closes #901


lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #909: it can merge as it stands.

@lneto Rebased onto master, the duplicate of #897 dropped by patch id, and the review's one finding is folded: `handlers_late.lua` held only the negative half of the asymmetry the new doc block states, and the row now holds both, since the table is empty at `probe.new` and both handlers are added afterwards.

Every kernel citation in the commit body was re-derived at 6.8 and at the 5.15 floor: `kprobes with 'post_handler' can not be optimized` (https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L685), `bool ipmodify = (p->post_handler != NULL)` in both `arm_kprobe_ftrace` and `disarm_kprobe_ftrace` (https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L1104), and `add_new_kprobe`'s `unoptimize_kprobe` for a new probe carrying one (https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L1266). The refusal of a disagreement inside a percpu set is structural rather than a preference: `post_handler` cannot change after `register_kprobe`, since the ftrace ops are picked from it, so the only order-independent options are installing one always, which this removes, or refusing.

Stack and error paths hold: `lua_getfield` is popped before `lunatik_newobject` pushes the object, so `lunatik_registerobject(L, 2, object)` still names the handlers table, and the new arm raises before anything is acquired, with the repeat-inside-one-runtime message keeping precedence.

Verified here at this head, through the host lock: probe suite green with the six `handlers` rows, and `examples/systrack/device` run on the rebased tree, which matters because the example was rewritten on master after the first round proved this branch on the old one (https://github.com/luainkernel/lunatik/commit/f8542934983284ab87e1fdd40303926b0da236a4): 304 kprobes armed over an idle list, counts by name, all given back on stop, clean dmesg.

One state this makes reachable that nothing covers yet: an aggregate with mixed post-ness, a `pre`-only probe and a `post` one at the same address, which the kernel upgrades itself. It is the same asymmetry the tree already documents for a repeated target, so it is coherent rather than a gap, and it belongs in `tests/probe/aggregate.sh` once #911 is in.

PR #911, opened 2026-09-15
tests/probe: a second probe on one address runs both handlers

### LINKED ISSUE #903: tests/probe: a plain runtime probing one address twice runs both handlers
#898's `probe.new` block now states that a target another kprobe already holds is aggregated by the
kernel and every handler attached to that address runs on a hit. Nothing asserts it.
`tests/probe/percpu_probe.sh` case 3 holds the percpu refusal, and
`tests/probe/kprobe_concurrent.lua:16-21` does register 306 probes on 305 addresses, so the
aggregate registration and teardown are exercised, but it asserts only that the stop returns and
that dmesg is clean.

The case is small: one plain hardirq runtime, two `probe.new` calls on the same address, one
trigger, both handlers counted once, two kprobes on the list at that address, none after stop. The
*behaviour* was verified by hand during the review of #898, with a scratch script that registered
two probes on `systab["personality"]`: the kprobe list showed two entries at the same address, one
`setarch` printed both handlers, and the stop left zero. What is missing is the assertion, not the
knowledge.

It is deferred rather than folded because it needs its own `.sh` and a line in
`tests/probe/run.sh`, and that line is the one #897 also edits: adding it while both were open
would have put two pull requests in conflict over a guard on kernel behaviour the binding cannot
easily break. Once #897 and #898 are both in, it is a clean standalone commit.

lneto:
`probe.new`'s block states that a target another kprobe already holds is aggregated by the kernel and that every handler attached to that address runs on a hit, and nothing in the tree asserts it. `percpu_probe.sh` case 3 holds the percpu refusal, and `kprobe_concurrent.lua` does register 306 probes over 305 addresses on this host, since `lookup_dcookie` and `nfsservctl` resolve to one, but it asserts only that the stop returns within 30 s and that dmesg is clean.

`tests/probe/aggregate` is that assertion: one plain hardirq runtime, two `probe.new` calls on `syscall.table["personality"]`, one `setarch`. It holds that the second `probe.new` succeeds, that both handlers print once, that the kprobe list grows by two lines over one more address, and that the stop gives all of it back. The target is the same address twice, not two names the kernel resolves to one, since that collision is arm64-specific and needs a call to a syscall the kernel does not implement; the commit body carries the kernel path and its line references.

Based on #897, which edits the same `for t in ...` line in `tests/probe/run.sh`. Closes #903


lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #911: it can merge as it stands.

@lneto Rebased onto master and retargeted: the merge conflicted in `tests/probe/run.sh`, the branch's own commit cherry-picks clean, so it is one commit and no fixup.

The aggregation path was re-derived rather than read off the body, at 6.8 and at the 5.15 floor: `register_kprobe` finds `old_p = get_kprobe(p->addr)` and hands over to `register_aggr_kprobe` (https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L1657), which calls `init_aggr_kprobe` (https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L1280) to set `ap->pre_handler = aggr_pre_handler`, and that walks `p->list` on a hit (https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L1198). The second `probe.new` escapes `warn_kprobe_rereg` because `__get_valid_kprobe` looks for *this* `struct kprobe` in the aggregate's list (https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L1503) and a fresh `luaprobe_t` is not there. Same shape at v5.15, `:1629`, `:1326`, `:1166`.

The listing reads the way the two counters assume: `report_probe` prints `%px` of `p->addr` as the first column and `show_kprobe_addr` reports every member of an aggregate with the same address, so two probes on one target is lines +2 over addresses +1, which is exactly what the case asserts. Counting by distinct first column rather than by symbol is right for a reason the header gives and that the source confirms: on arm64 the target is `SYSCALL_DEFINE1(arm64_personality, ...)` (`arch/arm64/kernel/sys.c:31`), so the name in the listing is `__arm64_sys_arm64_personality`, and a grep on the syscall's name would be the arch-specific thing the case set out to avoid.

The wrapped `for t in ...` line has a precedent on master, `tests/netlink/run.sh:14-16` wraps the same list the same way.

Verified here at this head, through the host lock so nothing else was on the device: `ok 1 a second probe on one address is aggregated, and one call runs both handlers`, totals pass:1 fail:0 skip:0, dmesg clean.

One thing recorded rather than raised, since it does not fire here and folding it is not what the issue asked for: the address counter depends on `kallsyms_show_value`. `report_probe` sets `addr = NULL` when it fails (https://github.com/torvalds/linux/blob/v6.8/kernel/kprobes.c#L2775), so under `kptr_restrict >= 2` every first column is zeros, `sort -u` collapses the whole list to one address, and the assertion passes only while the host is idle and false-fails once any other kprobe is armed. This host is `kptr_restrict = 1` and runs the suite as root, so it was not reproduced. `kprobes()` counts lines and is immune. A one-line guard that skips the address half when the first column is all zeros would close it; I can open it as a follow-up if you want it covered.

PR #916, opened 2026-09-15
tests/crypto: skip comp where the binding is not built

### LINKED ISSUE #912: tests/crypto: comp is exercised on kernels where the binding is compiled out
On Linux 6.15 and later the COMP binding is compiled out on purpose, but the suite still runs it, so `crypto/comp` fails on every current kernel.

## The gate exists on the C side, and only there

`lib/luacrypto_comp.c` guards its whole implementation:

```c
#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0))
#warning "crypto_comp API was removed in Linux 6.15, skip COMP module"
#else
...
#endif
```

`lib/luacrypto.h` gates the prototype and the class the same way, and `lib/luacrypto_core.c` drops the entry from both `luacrypto_lib` and `luacrypto_classes`. So on >= 6.15 `require("crypto").comp` is `nil`, exactly as intended.

`tests/crypto/run.sh` has no matching gate:

```sh
TESTS="shash skcipher aead rng hkdf comp"
```

and `tests/crypto/comp.lua` opens every case with `local c = comp"lz4"`, so all six raise.

## Reproducing

On a kernel >= 6.15 (here 7.2.6):

```sh
sudo make install
sudo lunatik test crypto
```

```
ok 5 crypto/hkdf
# FAIL	COMP compress empty string (error)	/lib/modules/lua/tests/crypto/comp.lua:10: attempt to call a nil value (upvalue 'comp')
#      	/lib/modules/lua/util.lua:47: in local 'test'
#      	/lib/modules/lua/tests/crypto/comp.lua:9: in main chunk
[... the same for the other five cases ...]
not ok 6 crypto/comp
# Totals: pass:5 fail:1 skip:0
```

It is the only failure in a full `lunatik test` run here: `pass:165 fail:1 skip:16`.

## Suggestion

Skip rather than fail, the way the BTF-dependent suites already do:

```sh
# tests/xdp/test_xdp.sh:41, tests/tc/test_tc.sh:57, tests/sched/run.sh:45
[ -f /sys/kernel/btf/$MODULE ] || skip_all "$MODULE built without BTF (make btf_install, rebuild)"
```

The equivalent precondition here is whether the binding is present at all, so dropping `comp` from `TESTS` (or reporting a skip from `comp.lua`) when `require("crypto").comp` is `nil` would keep the suite honest on both sides of 6.15. Porting the binding to the `acomp` API would be the other way out, but that is a separate change and this report is only about the suite going red.

## Environment

- lunatik `f8542934`
- Arch Linux ARM, aarch64, kernel `7.2.6-1-aarch64-ARCH`, gcc 16.1.1
- Unrelated to this report, but needed to build at all on this kernel: `strncpy` has been removed from the kernel, and four call sites remain (`lunatik.h`, `lib/luarcu.c` x2, `lib/luasocket.c`). Happy to send that as a separate PR if it is not already being tracked.

lneto:
`crypto.comp` is nil by design from Linux 6.15 on, where the kernel removed the `crypto_comp` API: `lib/luacrypto_comp.c` compiles to nothing and `lib/luacrypto_core.c` drops the entry from the library and the class list. `tests/crypto/run.sh` has no matching gate, so all six cases of the comp test raise and the suite is red on every current kernel.

The suite asks the runtime for the binding, through a `hascomp` probe, and reports a skip when it is absent, the way the suites that need BTF or sched_ext already do. The gate keys on the binding being there rather than on a kernel version, so porting the binding to `acomp` needs no change here.

The cleanup commit comes first because the probe depends on it: every case stays registered once it has run, so a second run of the suite failed all six with "is already running", and the probe would have read that as a missing binding. Closes #912.


lneto (commented):

**Verdict: ready to merge.** @lneto, the branch is now the one commit #912 still needs, https://github.com/luainkernel/lunatik/commit/35f3a534360104b04ec1384dc448c269e40c3252, rebased onto master.

Master already carries the first commit, the suite's cleanup, in https://github.com/luainkernel/lunatik/commit/117cdf22cdb7e2c3469a32b6f18b8bd19389291e, which does the same across more suites. GitHub reported the old head as mergeable, but the merge kept both `cleanup` definitions in `tests/crypto/run.sh`, and master's, defined second, would have replaced the one that stops `hascomp`: where `crypto.comp` is built, the probe would stay registered and a second run of the suite would fail it as already running. The branch drops that commit, and master's `cleanup` now stops the probe too.

On 7.2, `lunatik test crypto` passes five with `crypto/comp` skipped, and `run.sh` run twice in a row without an unload gives the same. The path where `comp` is built, before 6.15, was not run here.

Issue #924, opened 2026-09-15
examples: an unbounded receive in a spawned thread body cannot be stopped


lneto:
AGENTS.md, "Kernel threads": an unbounded blocking call in a kernel thread makes it unstoppable,
because the body runs holding the runtime lock and `stop()` waits for it to return, which a
`sock:receive()` with no timeout never does. Three spawned bodies do exactly that.

- `examples/echod/worker.lua:20`, `local message = session:receive(1024)`
- `examples/shared.lua:45`, `local request = session:receive(size)`
- `examples/cpuexporter.lua:122`, `local request, err = session:receive(1024)`

`echod` binds TCP 127.0.0.1:1337 and `shared` binds TCP 127.0.0.1:90, neither of which is gated by
anything: any local user can connect and send nothing, and the worker never returns. `lunatik stop`
then waits in `kthread_stop()` and its caller goes to D state, which only a reboot clears.

`cpuexporter` is the milder one. Its socket comes out `srwxr-xr-x root:root`, and `unix_find_bsd()`
requires `path_permission(&path, MAY_WRITE)`, so only root reaches it. `tests/socket/unix/stream_server.lua:25`
has the same shape and is reached only by the test's own client.

Source-traced, not reproduced: reproducing it wedges `/dev/lunatik` on the machine it runs on.

## Suggestion

The fix each site wants is the one AGENTS.md names, a bounded call plus a poll of `shouldstop()`:
`setsockopt(sk.sol.SOCKET, sk.so.RCVTIMEO_NEW, ...)` on the accepted session, as
`tests/socket/setsockopt.lua` and `tests/socket/unix/abstract.lua` do, or `MSG_DONTWAIT`. It is one
change over four files rather than a finding on any one of them.

Found while reviewing #921, whose own test carried the same shape until it was bounded.

## Environment

- lunatik `claude_sockaddr` head, Arch Linux ARM, aarch64, kernel 7.2.6

Issue #923, opened 2026-09-15
data: data.new() hands Lua uninitialized kernel heap


lneto:
`luadata_lnew()` (`lib/luadata.c:363`) allocates with `lunatik_checkalloc`, not
`lunatik_checkzalloc`, so a fresh `data` object carries whatever the slab last held and
`data:getstring()` past what the script wrote returns it.

```lua
local d = data.new(1024)
local s = d:getstring(0, 1024)
```

```
data.new(1024) nonzero=351
```

`examples/shared` does exactly that: it stores a value with `slot:setstring(0, value)` into a
`data.new(size)` of 1024 bytes and answers a lookup with `slot:getstring(0, size)`, so the reply
carries the value and then the rest of the allocation. Driven over TCP 127.0.0.1:90:

```
key=value
key
-> 1025 bytes, b'value\x00\xff\xff\x801...', 414 non-zero bytes past the value
```

Earlier runs of the same example returned audit log text, packet bytes and ELF program headers.
Loading the script is privileged; the peer that reads the reply is not.

## Suggestion

`lunatik_checkzalloc`. `data.new`'s doc block (`lib/luadata.c:191`) says nothing about
initialization, so the alternative is to document that reading past what was written is the caller's
error, which leaves the shipped example wrong.

Found while running `examples/shared` for the review of #921.

## Environment

- lunatik `claude_sockaddr` head, Arch Linux ARM, aarch64, kernel 7.2.6

PR #925, opened 2026-09-15
examples/shared: answer with the value stored, and survive what a peer can do

### LINKED ISSUE #917: examples/shared: a GET of an unset key, a client that hangs up, and a reply carrying the whole slot
`examples/shared.lua` is the in-memory key-value store the README walks through, and three separate
defects show up in one session with it. All three reproduce on `f8542934`, with the example spawned
as the README says.

## A GET of a key that was never set kills the daemon

```
sudo lunatik spawn examples/shared
python3 -c "import socket; s=socket.create_connection(('127.0.0.1',90)); s.sendall(b'nokey\n')"
```

```
luathread: [00000000a68b7b21] /lib/modules/lua/examples/shared.lua:57: attempt to index a nil value (field '?')
```

Line 57 is `local value = shared[key]:getstring(0, size)`: `shared[key]` is `nil` for a key that was
never assigned, and indexing it raises out of the thread body.

## A client that closes right after its request kills the daemon

```
luathread: [000000005a52f53e] ECONNRESET
```

`handle()` is called from `daemon()` without a `pcall`, so an error out of `session:receive()` leaves
the accept loop for good. After this the port is dead: the next connection times out.

## A GET answers with the whole slot, not with the value

`shared[key] = data.new(size)` with `size = 1024`, and the read is `getstring(0, size)`, so the reply
is the value followed by the rest of the buffer. `data.new()` allocates through
`lunatik_checkalloc()`, which does not zero, so what follows the value is whatever was in that
allocation:

```
foo=bar
foo
-> b'bar\x00\x04\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00h\x02...'
```

That is about 1020 bytes of kernel heap per GET, to a client on port 90. The example only binds
`inet.localhost`, but it is the file a reader copies from.

## Suggestion

The first two are a nil check and a `pcall` around `handle()`. The third is the example keeping the
length it stored, rather than reading the slot's full size, so that the reply is the value; sizing
the slot to the value is the same fix from the other end.

## Environment

lneto:
`examples/shared` answers a GET of a key nothing ever assigned by indexing the nil `rcu.table` returns for a missing entry, and a peer that hangs up with its reply still unread turns its close into a reset the daemon's next receive raises on. Either raise leaves the spawned thread body for good and the port stays bound with nobody in `accept()`, so the next client's connect succeeds and its read times out. A GET that does answer sends the whole 1024-byte slot rather than the value, and what follows the value is whatever the allocation held.

Three commits fix the request loop where each defect is: an absent slot answers with an empty line, `handle()` is called through `pcall` and the session is closed after it, and a SET stores the value in a slot of its own size instead of rewriting one the table has already published. A new `tests/examples` suite drives the spawned daemon over its own port and asserts each reply byte for byte; the peer that resets is a userspace one, since a lunatik socket shuts down before it releases and so always says goodbye with a FIN.

The uninitialized heap `data.new()` hands out is #923 and the unbounded receive in the body is #924, neither of them touched here. Two further commits repair the suite list the first one edits: `sched` was missing from it, and `lunatik test io` could not resolve the one suite whose entry point was not named `run.sh`. Closes #917.


lneto (commented):

**Verdict: ready to merge.** @lneto, the branch is now the three commits #917 still needs, rebased onto master, up to https://github.com/luainkernel/lunatik/commit/a32c772bd7316ff8ea375a26521789b58b0aeed5.

Master already carries the last two commits, the suite list naming `sched` in https://github.com/luainkernel/lunatik/commit/35119e83ec4d9c8853b662b6be1d08a9a129b572 and the io runner named `run.sh` in https://github.com/luainkernel/lunatik/commit/f440af91bea0330267b165b4d53e206a18e06f5b, so the branch drops them. The one conflict, the README's suite list, takes master's list with `examples` added.

On 7.2, `lunatik test examples` passes all five cases, twice in a row. Each fix was checked against its case: with master's `examples/shared.lua` the first case fails, with only the `pcall` reverted the reset case fails, and with only the slot sizing reverted the value case fails.

#962, stacked on this branch, was restacked onto it: its four commits change the same lines as before, and the one conflict, in `tests/README.md`, keeps master's `orphan` and `rtnl` entries. It merges after this one.

PR #928, opened 2026-09-16
linux: lookup answers nil for a symbol kallsyms does not carry

### LINKED ISSUE #926: linux: lookup answers a truthy NULL where it documents nil
`linux.lookup` documents that a symbol it cannot find comes back as `nil`, and ships a worked
example that tests for it. The code pushes the NULL pointer instead, and a light userdata is truthy
in Lua, so the documented test never takes the other branch.

```c
/* lib/lualinux.c:172 */
* @treturn lightuserdata symbol address if found, otherwise `nil` (represented as a NULL lightuserdata).
* @usage
*   local addr = linux.lookup("jiffies")
*   if addr then print("Address of jiffies:", addr) end
*/
static int lualinux_lookup(lua_State *L)
{
	const char *symbol = luaL_checkstring(L, 1);

	lua_pushlightuserdata(L, lunatik_lookup(symbol));
	return 1;
}
```

`lua_pushlightuserdata` has no NULL case (`lua/lapi.c:651-656`), and Lua's only false values are
`nil` and `false`. On this host:

```
> return tostring(require("linux").lookup("a_symbol_no_kernel_has"))
userdata: 0000000000000000
```

so the `if addr then` of the `@usage` is always taken. The parenthetical in the `@treturn` line,
"represented as a NULL lightuserdata", reads like someone noticing the contradiction and writing it
into the sentence rather than into the code.

## Why it survived

`git grep "linux\.lookup" -- '*.lua'` returns exactly one hit, the `@usage` above: the function has
no caller in the tree's Lua. `lunatik_lookup` itself (`lunatik_aux.c:93-112`) returns NULL correctly
and its four C callers (`lib/luabpf.c:502`, `lib/luaprobe.c:384`, `lib/luasyscall.c:53`,
`lunatik_aux.c:93`) all check it. Only the Lua boundary drops the information.

## Suggestion

Push `nil` when the lookup returns NULL, and drop the parenthetical. The `@usage` becomes true
rather than aspirational.

The change reaches out-of-tree scripts, which `tools/checks/consumers.sh` cannot see from here
(`LUNATIK_CONSUMERS` is unset). What a consumer can be doing today: `if addr then`, which is broken
now and fixed by this; passing the result to a binding, which raises either way, with a different
message; or comparing against a sentinel built from another failed lookup, which keeps working,
since two missing lookups compare equal now and both become `nil` after.

## Why it is worth filing on its own

`examples/dropreason` (#918) needs to ask the kernel which of two drop functions it exports, and
this is the only API a kernel script has for that question. The alternative, `pcall` around
`probe.new`, cannot tell "this kernel does not export this symbol" from a blacklisted address, an
already-registered probe or ENOMEM: it correlates rather than testing the property.

## Environment

lneto:
`linux.lookup` pushes the NULL that `lunatik_lookup` returns for a symbol kallsyms does not carry as a light userdata, which is truthy in Lua. So the `if addr then` the block ships as its own `@usage` never takes its other branch, and a script asking whether the kernel has a symbol gets an address of zero instead of an answer.

It pushes `nil` now, and the doc drops the parenthetical that wrote the contradiction into the sentence instead of into the code. A new `tests/linux/lookup` covers the nil, the address and the refusal of a non-string argument, skipped without `CONFIG_KPROBES`, which is how the module reaches `kallsyms_lookup_name`. The commit before it gives the suite the cleanup it never had, so a second run does not fail on the runtime the first one left registered.

Scripts written to the documented shape are fixed by this; one that formats the result reads "nil" where it read a zero pointer, and one that uses it as a table key now raises. The six public clones read here (`snihook`, `moontastik`, `openwrt_feed`, `lunatik_packages`, `luaxdp`, `poc-driver`) carry no call to it; the private consumers the review skill names were not on the machine, so they were not read. Closes #926.


lneto (commented):

**Verdict: ready to merge.** @lneto, the branch is now the one commit #926 needs, rebased onto master: https://github.com/luainkernel/lunatik/commit/e00f5d3318440e98639b360d6624ea8b35bd80b9.

Master already carries the first commit, the linux suite's cleanup, in https://github.com/luainkernel/lunatik/commit/117cdf22cdb7e2c3469a32b6f18b8bd19389291e, so the branch drops it. Its two conflicts, the cleanup in `tests/linux/run.sh` and the suite's entry in `tests/README.md`, keep master's side and add this branch's config probe and `lookup` entry.

On 7.2, `lunatik test linux` passes all three, twice in a row, and with master's `lib/lualinux.c` both nil cases fail with `expected nil, got userdata: 0000000000000000`. No script in the tree calls `linux.lookup`, and of the consumer clones on the machine only one loads `linux`, without calling `lookup`.

#929 and #933, which carry this branch's commits under their own, were restacked onto it and keep their own changed lines. Both target master, so they merge after this one.

PR #929, opened 2026-09-16
examples/dropreason: probe the symbol the running kernel carries

### LINKED ISSUE #918: examples/dropreason: the probed symbol is gone on current kernels
`examples/dropreason/monitor.lua` does not load on a current kernel: the function it probes is no
longer a symbol.

```
sudo lunatik run examples/dropreason/monitor hardirq
/lib/modules/lua/examples/dropreason/monitor.lua:63: failed to register probe (-2)
```

`-2` is `-ENOENT` from `register_kprobe()`, and `kfree_skb_reason` is absent from `/proc/kallsyms`
here. In `include/linux/skbuff.h` on this kernel it is an inline wrapper:

```c
void sk_skb_reason_drop(struct sock *sk, struct sk_buff *skb,
                        enum skb_drop_reason reason);

static inline void
kfree_skb_reason(struct sk_buff *skb, enum skb_drop_reason reason)
{
        sk_skb_reason_drop(NULL, skb, reason);
}
```

`sk_skb_reason_drop` is the out-of-line function that remains (`ffff8000815168f0 T` in
`/proc/kallsyms`).

## What the fix has to decide

Two things change together, so a one-word substitution is wrong: the target becomes
`sk_skb_reason_drop`, and the reason moves one argument to the right, since `sk` goes in front, so
the `local REASON <const> = 1` of `monitor.lua:30` is `2` there.

The open question is the range. This tree supports 5.15 and later, and `kfree_skb_reason` is a real
symbol on the older half of that range, so an example that simply switches targets stops working on
the kernels where it works today. What the example needs is a way to probe whichever of the two the
running kernel exports, with the argument index that goes with it. Whether that belongs in the
example, or is a gap in what `probe` offers a script, is the thing to settle: a script cannot ask
`probe` which symbols exist, so today it cannot choose.

I have not established in which release the change landed; the reading above is from this host's
headers and `/proc/kallsyms` only.

## Environment

lneto:
`examples/dropreason/monitor.lua` does not load on a current kernel: it probes `kfree_skb_reason` by name, which v6.11 turned into a static inline over `sk_skb_reason_drop`, so kallsyms carries no such symbol and `register_kprobe` answers `-ENOENT`. Substituting the name is not the fix, because `kfree_skb_reason` is the real symbol from v5.17 to v6.10 and the new one takes the socket first, so the reason moves from argument 1 to argument 2 with the target, and reading the wrong one raises nothing, it counts the drops under a name their reason never carried.

The example now asks `linux.lookup` which of the two the running kernel has, newest first, and takes the argument index that goes with it; when neither answers it names both, which is what a kernel below v5.17 gets, where the drop path takes no reason at all. `tests/probe/dropreason` pins the same pair, since the argument half fails silently: it checks the symbol the script chose against `/proc/kallsyms`, and that the handler counts the `NO_SOCKET` drops of a batch of datagrams sent to a closed UDP port, which a handler reading the wrong argument counts none of.

The branch sits on #928 and merges after it. On master `linux.lookup` pushes the NULL as a light userdata, which is truthy, so the first candidate always wins and the example would pick `sk_skb_reason_drop` on the v5.17 to v6.10 kernels it exists for. Closes #918.


lneto (commented):

**Verdict: ready to merge.** @lneto, the branch is one commit on master, https://github.com/luainkernel/lunatik/commit/a87ad27175f54d247d64098d791d6652e84dfaea, restacked after #928 merged.

Master still probes `kfree_skb_reason`, which a 7.2 kernel does not carry, so the example does not load there. The tree took a 6.6 floor after the branch was written, in https://github.com/luainkernel/lunatik/commit/afffc7a755498290ec0329de3c5177c9187416e8, so the commit now drops what the README, the script's header and the test's skip message said about kernels before v5.17; both symbols stay in range, `kfree_skb_reason` through v6.10 and `sk_skb_reason_drop` from v6.11.

On 7.2, `lunatik test probe` passes 27, the two `dropreason` cases among them, and with `sk_skb_reason_drop`'s reason read from argument 1 the second case fails. The example was run as the README shows it: it probed `sk_skb_reason_drop`, three datagrams to a closed port read back as 3 `NO_SOCKET` from the REPL, and the stop left nothing registered. The v6.6 to v6.10 path, `kfree_skb_reason`, was not run here.

PR #932, opened 2026-09-16
lunatik: one spelling for the registry lookup by class

lneto:
`lunatik.h` reached the registry by class pointer in two spellings, and the lookup that raises was written twice after `lunatik_checkmetatable` arrived copying `lunatik_setclass`'s first three lines.

`lunatik_pushmetatable` is that lookup, in the spelling `lunatik_hasclass` and `lunatik_getregistry` already use. `lunatik_checkmetatable` is it plus a pop, `lunatik_setclass` is it plus the setmetatable it had, and `lunatik_newclass` takes the write side and stops carrying the key under the table it builds.

No call site changes and no behaviour changes; the convention is recorded in `AGENTS.md`.

https://claude.ai/code/session_01LQZ4D5bHwdQ7hVy4c15tCs


lneto (commented):
_Agent review (Claude Code), posted from @lneto's account._ Verdict on #932: it can merge as it stands.

@lneto Traced against the vendored Lua 5.5.0 in this tree, the rewrite is behaviour-neutral. `lua_rawgetp` (`lua/lapi.c:786-793`) builds its key with the same `setpvalue` as `lua_pushlightuserdata` (`lua/lapi.c:651-656`, `lua/lobject.h:485`) and returns the same `novariant(tag)` through `finishrawget` (`lua/lapi.c:748-754`), so the `== LUA_TNIL` raise is the condition it was. `lua_rawsetp` pops one where `lua_rawset` popped two, through the same `aux_rawset` (`lua/lapi.c:931-953`), and everything `lunatik_newclass` runs between the `lua_newtable` and the write is top-relative: `luaL_setfuncs` with nup 0 writes at -2 (`lua/lauxlib.c:971-985`), `lunatik_monitorobject` (`lunatik_obj.c:161-172`), `lunatik_hasindex`. So the key it no longer carries was never read, the net stack effect of each helper is what it was, and no call site had to move.

The sweep is complete: outside `lua/` no `lua_rawget` or `lua_rawset` reaches the registry, and every surviving `lua_pushlightuserdata` pushes a value or a `lua_pcall` argument. No new test is owed either: all three helpers run on every runtime creation (`lunatik_core.c:225` and `:233`, `lunatik_obj.c:25`), so a wrong key or a stack imbalance lands on the first script of the first suite.

Two findings were raised, both prose, both folded at this head. The `AGENTS.md` bullet opened "The Lua registry is keyed by pointer", a universal this tree contradicts: `lunatik_ebpf.h` keys by ref on purpose, `luaL_ref` at `:78`, `lua_rawgeti` at `:55`, `luaL_unref` at `:86`, four lines from a `lunatik_register` that is `lua_rawsetp`, and `LUA_RIDX_GLOBALS` and `LUA_RIDX_MAINTHREAD` are integer keys (`lua/lua.h:84-85`). It now conditions on the key instead. The commit body listed `lunatik_getregistry` among the readers keyed by class, though it takes an arbitrary key and none of its nineteen callers passes one, and it counted four call sites where there are five, missing the two in `lunatik_newclasses`.

Verified here at this head, through the host lock: full suite 185 pass, 0 fail, 3 skip (no sched_ext on this kernel), dmesg clean.

PR #937, opened 2026-09-16
skills: the review workflow takes its machine from args

lneto:
A review on any machine but the one `review.js` was written on cannot run it: the workflow names one checkout, one sudo password, one token path, one kernel and one consumer list, and its GitHub calls assume the `gh` CLI.

This session hit that and wrote a review pipeline of its own instead, which lost what the workflow already knew: the checkpoint that keeps findings on disk as they close, so an agent dying costs one phase, and the rule against rebuilding a head the suite already passed.

The machine is arguments now, each defaulting to the value it had, so a run that passes none of them behaves as it always did.


lneto:
Superseded by f06436f4 and e0a24a47 on master, which solve the same problem better: the checkout and the sudo prefix default to values that name no machine (`$(git rev-parse --show-toplevel)`, plain `sudo`), and the host, the kernel reference, the token and the consumer list are read from the environment at run time rather than passed as arguments with one machine's paths as defaults. Closing.

PR #951, opened 2026-09-17
data: a new buffer holds no byte the script did not write

### LINKED ISSUE #923: data: data.new() hands Lua uninitialized kernel heap

lneto:
`data.new()` allocated through `lunatik_checkalloc()`, so a fresh buffer held whatever the slab last served and `getstring()` past what the script wrote returned it. `examples/shared` answers a lookup with `slot:getstring(0, size)` over TCP, so a peer that never loaded the script reads the rest of that allocation, which in the reported runs was audit log text, packet bytes and ELF program headers. `data:resize()` leaves the same hole on the bytes a growth adds, two Lua lines from `data.new()`.

Both allocate zeroed now, through `lunatik_checkzalloc()` on creation and a `memset` over the bytes a growth adds, and the doc blocks say so. No supported kernel writes those bytes for us: `krealloc()` copies the old size and zeroes the spare only under `want_init_on_alloc()`, and `lunatik_alloc()`'s `kvmalloc` arm copies `min(osize, nsize)`. The views `luadata_reset()` hands out, `skb:data()` and the XDP and TC arguments, keep their packet: `data.new()` is the only producer of an owned buffer, and `data:resize()` refuses one it does not own before it reaches the `memset`. A first commit documents the `"shared"` and `"single"` mode `data.new()` already took, which its block omitted.

`tests/data/zeroed` poisons a batch of buffers and frees it so the next round allocates over them, on both allocator arms, in the `single` mode and from an atomic runtime; a shrink followed by a growth back into the block it kept needs no such luck. The case skips on a kernel that zeroes every allocation itself, where a fixed `data.new()` and a broken one read the same. Closes #923.


tests/data/run.sh
```diff
@@ -29,8 +39,20 @@ cleanup
 ktap_header
 ktap_plan $TOTAL
 
+# init_on_alloc is an early_param read with kstrtobool, so the command line decides and the config is the default
+zeroesallocations() {
+	case " $(cat /proc/cmdline) " in
+		*" init_on_alloc="[yYtT1]*|*" init_on_alloc="[oO][nN]*) return 0 ;;
+		*" init_on_alloc="[nNfF0]*|*" init_on_alloc="[oO][fF]*) return 1 ;;
```
lneto, line 46: (posted by an agent, not by @lneto)

`kstrtobool` reads `f` and `F` as false too ([lib/kstrtox.c at v6.6](https://github.com/torvalds/linux/blob/v6.6/lib/kstrtox.c#L363-L364)), so `init_on_alloc=f` fell through to the config default and skipped a run that could tell the cases apart. Folded in the squash; the fixup was https://github.com/luainkernel/lunatik/commit/78b1284891067f4a624c12903be11872ef16f74e

lneto (commented):

@lneto Mergeable as it stands: the fixup on `tests/data/run.sh` is folded, and the commit body and the description now say what `krealloc()` does with the old block.

Ran the head on its own base and, cherry-picked, on master (5fbdf0988), each time after checking every loaded module against the installed one: `data/bounds` and `data/resize_atomic` pass and `data/zeroed` skips, since this host's kernel has `CONFIG_INIT_ON_ALLOC_DEFAULT_ON=y`; `examples/echod` echoes and `examples/shared` answers a lookup with the value and 1021 NULs. Skipping the whole case is the right call: the regrow cell looked as if it could still tell the cases apart under `init_on_alloc` before 6.12, but Ubuntu's 6.8.0 carries the `want_init_on_alloc` memset of [`__do_krealloc`](https://github.com/torvalds/linux/blob/v6.12/mm/slab_common.c#L1217-L1223), and with the resize memset removed that cell passed here. So no test reaches the fix on a distribution kernel, and CI only builds.

Issue #967, opened 2026-09-17
lunatik: every errno name is (efault) below 6.7


lneto:
On every kernel below 6.7, `lunatik_pusherrname` pushes the string `(efault)` instead of the errno's
name, so every comparison a script makes against a name is false there.

```c
/* lunatik_aux.c:78 */
err = abs(err);
#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 7, 0)
	const char *name = errname(err);
	lua_pushstring(L, name ? name : "unknown");
#else
	char buf[LUAL_BUFFERSIZE];
	snprintf(buf, sizeof(buf), "%pE", ERR_PTR(-err));
	lua_pushstring(L, buf);
#endif
```

`%pE` is not the errno-name specifier. In `lib/vsprintf.c` at v5.15, `case 'E'` is
`escaped_string()` and `case 'e'` is the errno name; the two are separate specs. Worse, the argument
never reaches the escaping: `check_pointer_msg()` returns `"(efault)"` for a pointer that satisfies
`IS_ERR_VALUE`, and `ERR_PTR(-err)` is precisely that. So the branch yields a constant.

The guard itself is right about the kernel: `lib/errname.c` carries no `EXPORT_SYMBOL` at v5.15 or
v6.6 and one at v6.7, which is why the fallback exists at all.

## What it reaches

Every `== "EAGAIN"` in the tree, on 5.15 through 6.6. `examples/echod/daemon.lua`,
`examples/shared.lua` and `examples/cpuexporter.lua` all branch on that name in their accept loops,
and so does the arm #962 adds to each of them. An out-of-tree consumer does the same:
`snihook/quic.lua:82` matches `"ETIME"`.

On this host the question cannot be reproduced: 7.2.6 takes the `errname()` arm and answers
correctly. What is above is read from the kernel's own source at v5.15 and v6.6.

## Suggestion

`%pe`, and the `-` that spec keeps for a negative error has to be skipped: `errname()` returns
`"-EAGAIN"` shaped output through that path (`lib/errname.c`), while the 6.7 arm above returns
`EAGAIN`, so the two arms must agree on the spelling or a script gets one name on one kernel and
another on the next.

Whether this survives at all depends on #963: a floor at 6.7 or above deletes the whole `#else`
arm. A floor below it keeps the defect, so the two decisions are linked.

Issue #965, opened 2026-09-17
ci: the matrix builds no current kernel, and asserts one module of twenty-eight


lneto:
The matrix builds 5.15 and 6.12 (`.github/workflows/build.yaml:13`). Nothing builds a current
kernel, and this cycle produced three defects that only a current one shows:

- #913: `<linux/string.h>` no longer declares `strncpy`, and four call sites stopped the build
  outright. A build job on a current series catches this one.
- #912: `crypto_comp` is gone since 6.15, so `lib/luacrypto_comp.c` compiles to nothing and every
  case of `tests/crypto/comp.lua` fails. A build job does not catch this; the suite does.
- #918: `kfree_skb_reason()` became a `static inline` in 6.11, so `examples/dropreason` fails at
  `register_kprobe` with `-ENOENT`. Neither a build nor the suite catches this; running the example
  does.

All three were found by hand on a 7.2 host. Covering 7.x is the first half of this issue and the
second half is what a job can prove at all.

## What the build job asserts today

```yaml
- name: Verify modules built
  run: |
    test -f lunatik.ko || { echo "Build failed" ; exit 1; }
```

One module of the twenty-eight, and the build runs with `KBUILD_MODPOST_WARN=1`, which turns an
unresolved symbol into a warning. A binding that fails to link, or one that compiles to an empty
translation unit the way `luacrypto_comp.c` does past 6.15, leaves the job green.

## The matrix is tied to what upstream still maintains

The series is resolved from `releases.json` and the step fails when the series is EOL: `test -n
"$VERSION" || { echo "no maintained release for kernel series $SERIES"; exit 1; }`. The 5.15 line
goes end of life in December 2026, so that job breaks on its own schedule. #963 moves the floor,
which is the other end of the same matrix; this issue is the top of it.

## Suggestion

Add a current series and assert what the build actually produced: every `.ko` the configuration
asks for, rather than `lunatik.ko` alone. Whether CI can go further than compiling is the open
question, since the suite needs a booted kernel with the modules loaded, which a
`modules_prepare` tree cannot give; if that is out of reach, saying so in the workflow is worth
more than leaving the gap implicit, because a green matrix currently reads as "it works on those
kernels" and means "it compiled on them".

PR #969, opened 2026-09-17
build: take a 6.6 kernel floor and drop the 5.x line

### LINKED ISSUE #963: take a 6.x kernel floor and drop the 5.x line
The tree declares 5.15 and later, in `README.md:10`, in `AGENTS.md:16` and in the CI matrix
(`.github/workflows/build.yaml:13`, which builds 5.15 and 6.12). The decision is to drop the 5.x
line and take a 6.x floor. This issue is that decision and the work it entails.

## What the floor costs today, measured

The tree carries 24 `LINUX_VERSION_CODE` conditionals. **Not one has a boundary inside 5.x**: the
lowest is `>= 6.0`, so every `#else` arm serves 5.15 *and* a stretch of early 6.x, and "drop 5.x"
and "delete the compatibility code" are different changes. What each candidate floor clears:

| | 6.1 | 6.6 | 6.12 |
|---|---|---|---|
| conditionals removed outright | 0 | 9 | 16 |
| reduced to a plain `#ifdef` | 1 | 1 | 1 |
| remaining, of 24 | 23 | 14 | 7 |
| other compat shims removed | 1 | 1 | 1 |
| mechanisms removed | 0 | 0 | 1 (`autogen.lua`'s `include()`) |
| files deleted | 0 | 0 | 0 |
| lines removed, C and Lua | 3 | 40 | 83 |
| CI jobs | 1 of 2 drops at every floor | | |

Seven of the 24 survive every floor considered, because they are ceilings and not floors: 6.15
removed `crypto_comp` and renamed `is_endbr`, and 6.19 changed `sockaddr_unsized`. A third of the
version debt is not floor debt.

Two conditionals are worth naming separately. `lib/luanetfilter.c:152-154` gates `hook_ops_type` on
`>= 6.5` and compensates for nothing: the field is present at `v5.14`. And the `>= 6.4` gate in
`lunatik_ebpf.h:130-160` decides whether `xdp`, `tc` and `sched` are real modules or empty stubs, so
a floor at 6.6 removes a whole supported configuration rather than an arm.

## What it costs outside the tree

RHEL 9 ships 5.14 with maintenance to 2032, already below the declared floor. Ubuntu 22.04 ships
5.15 until April 2027. `README.md:75` sends users to the project's own OpenWrt feed on the
`openwrt-23.05` branch, which is 5.15 and whose last commit is more recent than the 24.10 one. The
5.10 and 5.15 upstream LTS lines both go end of life in December 2026.

## What the change has to carry

The floor is declared in three places and enforced in a fourth, and they move together: `README.md`,
`AGENTS.md`, the CI matrix, and the conditionals themselves. Which number to take is the first
question, and it is a support-policy decision rather than a simplification: at 6.1 the change
deletes three lines.

The full analysis, with every site, its gate, what it compensates for and the floor that clears it,
is the one this issue quotes; it also records what could not be established, including the purpose
of `#include <linux/l2tp.h>` in `lib/luasocket.c` and the exact upstream change the `>= 6.4` eBPF
gate depends on.

lneto:
Below 6.4, `lunatik_ebpf.h` gives `luaxdp` and `luatc` an empty class list, an init that
returns 0 and a lib table with no `attach` or `detach`, while `README.md` advertises `xdp`
and `tc` with no caveat. So on 6.1, an LTS inside the 5.15-and-later range the tree
declares, `require("xdp")` answers a table that cannot attach.

This takes 6.6 as the floor, in `README.md`, in `AGENTS.md` and in the CI matrix, and drops
the eight version conditionals and three older accommodations that only a kernel below it
needed. 6.6 is the lowest series kernel.org still maintains at or above 6.4, and it costs
nothing upstream, since 6.1 and 6.6 are projected to the same month; 6.12 would clear 43
more lines at the price of OpenWrt 24.10. The matrix becomes `[6.6, 7.2]`, which takes the
top of #965 along with the floor, because the arms only a kernel at or above 6.15 or 6.19
compiles are built by neither job today. `README.md`'s OpenWrt link moves off the feed's
`openwrt-23.05` branch, which is a 5.15 build.

Closes #963. #966 removes the one version gate this floor leaves dead and is not stacked.
What the number settles for the open work: #967's pre-6.7 `errname` branch stays a fix
rather than a deletion, and #965 keeps its second half, asserting every `.ko` and getting
past `KBUILD_MODPOST_WARN`. Two follow-ups this does not carry: the feed still has a
pre-5.15 accommodation on its default branch, an issue for that repository, and `snihook`,
the one consumer clone on this machine that loads `xdp`, `socket` and `device`, needs no
source change but must now build on 6.6 or later.


README.md
```diff
@@ -199,9 +199,9 @@ The table below lists the available kernel Lua modules:
 | `skb` | Socket buffer (`sk_buff`): inspect and modify packets |
 | `xdp` | XDP (eXpress Data Path) hooks |
 | `tc` | TC (Traffic Control) hooks |
-| `sched` | sched_ext (extensible scheduler) hooks: dispatch queue and slice from Lua |
+| `sched` | sched_ext (extensible scheduler) hooks: dispatch queue and slice from Lua; 6.12 and later, with `CONFIG_SCHED_CLASS_EXT` |
```
lneto, line 202: (posted by an agent, not by @lneto)

With line 10 now reading 6.6 and later, this row and the crypto row two below promise what a 6.6 kernel does not give: sched needs CONFIG_SCHED_CLASS_EXT, which arrived in 6.12 (https://github.com/torvalds/linux/blob/v6.12/init/Kconfig#L1072, absent at v6.11; lib/luasched.c:27 builds the stub without it), and compression is not built at 6.15 or later (https://github.com/torvalds/linux/blob/v6.14/include/linux/crypto.h#L436 is the last struct crypto_comp; lib/luacrypto_comp.c:20). The commit body says both and the rows stay silent, the caveat gap this PR opens with for xdp and tc. If the note is meant as a promise for every row, the fixup adds "6.12 and later, with CONFIG_SCHED_CLASS_EXT" here and "(below 6.15)" to compression; if it is meant only as the build floor, drop it. Fixup: https://github.com/luainkernel/lunatik/commit/eec71b4449c3d387ac0a495c6dfcad51c805860a

lneto (commented):

COMMENT: this merges as it stands, after #971 or after its rebase onto master lands; the one finding is a documentation fixup already on the branch, and the rest are notes for the squash.

@lneto, read at c9a0066fd on claude_msrq (b7fb62911, #971); the rebase onto master had not landed when it was read, and origin/master (f9cdd41c9) touches AGENTS.md, README.md and lib/luasocket.c among this PR's files without conflict, so the fixup below rides the rebase.

Built and run on 6.8.0-138 arm64 by an independent agent and read again here: make C=2 at the head and at its parent identical but for a two-line shift of the same four pre-existing sparse warnings in luatc.c and luaxdp.c; install and reload with the running core at the worktree's srcversion and all 29 installed modules byte identical to the build; suite 224 pass, 0 fail, 3 skip (sched, no sched_ext on 6.8), tc 6 of 6 and xdp 7 of 7 rather than skipping. Examples: filter/sni (xdp) and sniclassify (tc, through its setup.sh and cleanup.sh) on a veth with TLS ClientHellos, both verdicts logged; tap, systrack and spyglass through the watchdog, each device node created and read with its hook firing; echod, shared, cpuexporter and lldpd spawned and driven; ifquarantine not run.

Every removed guard rests on the floor and not on the old gate: __bpf_kfunc is at https://github.com/torvalds/linux/blob/v6.3/include/linux/btf.h#L81 and https://github.com/torvalds/linux/blob/v6.6/include/linux/btf.h#L84, register_btf_kfunc_id_set is exported at https://github.com/torvalds/linux/blob/v6.6/kernel/bpf/btf.c#L7925, so the 6.4 gate served nothing between 6.3 and 6.4 that the code needs; class_create(name) at https://github.com/torvalds/linux/blob/v6.4/include/linux/device/class.h#L230 against the macro at https://github.com/torvalds/linux/blob/v6.3/include/linux/device/class.h#L273; the const devnode at https://github.com/torvalds/linux/blob/v6.2/include/linux/device/class.h#L63; CPUTIME_FORCEIDLE at https://github.com/torvalds/linux/blob/v6.0/include/linux/kernel_stat.h#L31-L32; unsafe_memcpy at https://github.com/torvalds/linux/blob/v5.19/include/linux/fortify-string.h#L68; FS_ERROR at https://github.com/torvalds/linux/blob/v5.16/include/linux/fsnotify_backend.h#L46 and FS_RENAME at https://github.com/torvalds/linux/blob/v5.17/include/linux/fsnotify_backend.h#L66; always-y from https://github.com/torvalds/linux/blob/v5.6/scripts/Makefile.lib#L7 with no $(always) read at https://github.com/torvalds/linux/blob/v5.15/scripts/Makefile.build#L18. Both CI series resolve today (6.6.157 longterm, 7.2.6 stable, 7.1 EOL); the 6.6 job compiles one arm of every surviving conditional and the 7.2 job the other, the single-armed >= 6.5 store excepted, which #966 removes. The 7.2 job was read, not run: nothing on this arm64 host compiles the x86 IBT block.

Two notes for the squash, on the body only. The l2tp sentence can now say what it served: the include supplied IPPROTO_L2TP for luasocket's own IPPROTO table (https://github.com/luainkernel/lunatik/commit/5460fec38a4794959ad10f7b962a446c03c3ca16), which in.h carries only from https://github.com/torvalds/linux/blob/v6.1/include/uapi/linux/in.h#L72-L73 (https://github.com/torvalds/linux/blob/v5.15/include/uapi/linux/l2tp.h#L16 before), and it has served nothing since https://github.com/luainkernel/lunatik/commit/6bbd0ef7ab18463abf6e0e86ce7c0d2a1b17e7f3 moved the constants to autogen. And "compiles every arm that one did" is loose by luanetfilter.c:152, a single-armed gate 5.15 skipped and 6.6 takes, while "luacrypto's comp module" names the module where the >= 6.15 arm is its #warning stub.

Seen on the way, not this PR's: every load of an xdp or tc script logs "no callback attached (cpu N)" once per CPU, from attach calling detach first (lib/luaxdp.c:235 into lunatik_ebpf.h:47), which leaves as #1017. The "< 6.19" inversion idioms.sh names at luasocket.c:274 and :332 is already flipped on master by https://github.com/luainkernel/lunatik/commit/336e9414f286a8e337814299ca1a87213f95aa85, so the rebase resolves it. doc/design/fsnotify/kernel-notes.md:190 still says 5.15 and later, a design note left as such.

PR #971, opened 2026-09-17
core: call the CET MSR accessors by the name 7.2 kept

lneto:
`lunatik_cfi.h` calls `rdmsrl` and `wrmsrl`, which
[v7.2](https://github.com/torvalds/linux/blob/v7.2/arch/x86/include/asm/msr.h) no longer has:
at [v7.0](https://github.com/torvalds/linux/blob/v7.0/arch/x86/include/asm/msr.h#L330-L331)
and v7.1 they were aliases of `rdmsrq` and `wrmsrq` kept for the transition, and v7.2 dropped
them. So the tree fails to build on a 7.2 x86 kernel with `CONFIG_X86_KERNEL_IBT` set, with
`implicit declaration of function 'rdmsrl'`.

It now calls `rdmsrq` and `wrmsrq`, which
[v6.16](https://github.com/torvalds/linux/blob/v6.16/arch/x86/include/asm/msr.h#L194) added,
and aliases them to the old pair below 6.16. Nothing on an aarch64 host compiles this block,
so only a CI job on a 7.x x86 kernel proves it.

Depends on nothing. #969 is stacked on this, because the 7.2 job it adds to the matrix is
what found this.


lneto (commented):

COMMENT: mergeable as it stands; no finding on the code, and the notes below are wording for the squash, not changes to the diff.

@lneto, the claim was checked against the kernel source at pinned tags, since this host (6.8.0-138 aarch64) has no CONFIG_X86_KERNEL_IBT and preprocesses the whole block away. rdmsrq and wrmsrq arrive at v6.16-rc1 (https://github.com/torvalds/linux/blob/v6.16-rc1/arch/x86/include/asm/msr.h#L194-L200; the paravirt arm at https://github.com/torvalds/linux/blob/v6.16/arch/x86/include/asm/paravirt.h#L210-L217), so every 6.16.0 LINUX_VERSION_CODE carries them and the < 6.16 shim never redefines a kernel macro. rdmsrl and wrmsrl are aliases of the new pair from v6.16 (https://github.com/torvalds/linux/blob/v6.16/arch/x86/include/asm/msr.h#L329-L331) through v7.1 (https://github.com/torvalds/linux/blob/v7.1/arch/x86/include/asm/msr.h#L329-L331) and are gone at v7.2-rc1 (https://github.com/torvalds/linux/blob/v7.2-rc1/arch/x86/include/asm/msr.h#L290-L306 is the file's end, without them). Below 6.16 the shim resolves to the definitions the block called before: v6.6 https://github.com/torvalds/linux/blob/v6.6/arch/x86/include/asm/msr.h#L257-L263 and https://github.com/torvalds/linux/blob/v6.6/arch/x86/include/asm/paravirt.h#L222-L230, v6.15 https://github.com/torvalds/linux/blob/v6.15/arch/x86/include/asm/msr.h#L259-L265. No stable series below 6.16 backports the new names (v6.6.157, v6.12.110 and v6.15.11 carry only the old pair), so the version test is the guard the rule asks for, not the #ifndef it reserves for a backported macro. The other names the block uses did not move between v6.6 and v7.2: __is_endbr and ENDBR_INSN_SIZE (https://github.com/torvalds/linux/blob/v6.15-rc1/arch/x86/include/asm/ibt.h#L68, https://github.com/torvalds/linux/blob/v7.2/arch/x86/include/asm/ibt.h#L68), MSR_IA32_S_CET and CET_ENDBR_EN (https://github.com/torvalds/linux/blob/v6.6/arch/x86/include/asm/msr-index.h#L448-L451, https://github.com/torvalds/linux/blob/v7.2/arch/x86/include/asm/msr-index.h#L547-L550), X86_FEATURE_IBT (https://github.com/torvalds/linux/blob/v6.6/arch/x86/include/asm/cpufeatures.h#L423, https://github.com/torvalds/linux/blob/v7.2/arch/x86/include/asm/cpufeatures.h#L436), cpu_feature_enabled (https://github.com/torvalds/linux/blob/v6.6/arch/x86/include/asm/cpufeature.h#L139, https://github.com/torvalds/linux/blob/v7.2/arch/x86/include/asm/cpufeature.h#L71).

CI: this commit's own runs are master's matrix, Build 5.15 (linux-5.15.221, where CONFIG_X86_KERNEL_IBT does not exist yet, it enters Kconfig at v5.18, https://github.com/torvalds/linux/blob/v5.18/arch/x86/Kconfig#L1855) and Build 6.12 (linux-6.12.110, lunatik_aux.o compiled through the shim arm), https://github.com/luainkernel/lunatik/actions/runs/35236573528. The 7.2 proof is #969's: on 42f8da5fc, its first push, whose parent is 8f7a44f49 and which does not carry this commit, Build 7.2 on linux-7.2.6 fails at lunatik_cfi.h:37:9: error: implicit declaration of function 'rdmsrl'; did you mean 'rdmsrq'? (https://github.com/luainkernel/lunatik/actions/runs/35235965210/job/105251812971); on eec71b444, which carries b7fb62911, Build 7.2 and Build 6.6 (linux-6.6.157) are green (https://github.com/luainkernel/lunatik/actions/runs/35455430895). The failing run is also what shows the block compiles under CI's x86_64 defconfig, X86_KERNEL_IBT being def_bool y there (https://github.com/torvalds/linux/blob/v7.2/arch/x86/Kconfig#L1742-L1747, the same at v6.6, https://github.com/torvalds/linux/blob/v6.6/arch/x86/Kconfig#L1823-L1830). Not built here: the host has no x86_64-linux-gnu-gcc and no prepared x86 header tree, and on aarch64 the tree builds the #else arm of the header, which the change does not touch; runtime on IBT hardware was exercised nowhere, the expansion at each version being the same native_read_msr and native_write_msr, or paravirt, call the block made before, traced and not measured.

Two wording notes for the squash: the body says the old names were aliases "at v7.0 and v7.1"; they were aliases from v6.16 on, and the pull request body says the same. And "Only a CI job on a 7.x x86 kernel builds this block": the 6.6 and 6.12 x86 jobs build it too, through the shim; what only a job on 7.2 or later does is prove the kept names, since 6.16 to 7.1 accept both.

A separate item, not this pull request's: the is_endbr guard beside the new shim runs the other way, the call site keeping the older kernel's name and the guard mapping it to the current one from 6.15, which the version-guard rule on master names as the inversion. A three-line cleanup that calls __is_endbr and shims it below 6.15, the shape the rdmsrq lines have, is the second commit on this branch, https://github.com/luainkernel/lunatik/commit/78dabe3de209ee23e764c067ed924b6619d706f3; the 6.12 job compiles its shim arm, and only #969's 7.2 job the other.

Issue #1001, opened 2026-09-19
notifier: a registration from inside a netdevice callback deadlocks on RTNL


lneto:
`notifier.netdevice(callback)` called from inside a netdevice callback deadlocks the task that raised the event: `register_netdevice_notifier` takes `rtnl_lock` (https://github.com/torvalds/linux/blob/v6.8/net/core/dev.c#L1796-L1810), and the netdevice chain runs under RTNL, so the registration waits on the lock its own task holds. `luanotifier_new` is shared by every chain (`LUANOTIFIER_NEWCHAIN`), and the kernel has no per-task holder query for RTNL outside lockdep (`rtnl_is_locked` is `mutex_is_locked`), so the guard has to be the binding's own: a per-chain flag set around the callback's dispatch, in the shape #851 gives `fsnotify`'s path resolution, refusing a registration from inside that chain's callback with an error that names it.

Read from the source, not reproduced; found by the independent review of #851. Pre-existing on master, outside that pull request.

PR #1005, opened 2026-09-19
notifier: refuse a netdevice registration from inside its own callback

lneto:
A script that calls `notifier.netdevice` from inside a netdevice callback blocks on a lock its own task already holds: the chain is called under RTNL, and the registrar takes the namespace rwsem for write and RTNL before replaying REGISTER and UP on the registering task. The task stays in D state holding both, and only a reboot clears it, reported as #1001.

The constructor refuses that call now. A boolean in the Lua registry, written around the one `lua_pcall` the live chain and the replay share, says whether a callback of this runtime is on the stack, and the registry is the state's own, so a coroutine resumed from the callback reads it too. Only `netdevice` is refused: the keyboard and vt registrars take the chain's own spinlock, which their callbacks do not hold, so refusing them would give up a capability that works. The flag is per state, so a runtime created, resumed or stopped from inside the callback is not covered.

It depends on nothing outside master, and the out-of-tree scripts that use this binding register from their script body, where the flag is clear. A second commit corrects five lines of the suite's prose that named a registrar the binding does not call.


tests/notifier/inside.lua
```diff
@@ -0,0 +1,53 @@
...
+		report("stop")
+		stopper:stop()
+	end
+	return notify.OK
+end
+
+notifier.netdevice(cb)
+stopper = notifier.netdevice(stop)
+loading = false
+report("after " .. probe())
```
lneto, line 64: (posted by an agent, not by @lneto)

No case here makes a callback raise, so nothing separates the clear's position from the one the function's old shape invites: if (lua_pcall(...) != LUA_OK) { pr_err_ratelimited(...); return NOTIFY_OK; } luanotifier_setincallback(L, false); compiles, keeps the guard, and leaves the flag set for the life of the runtime, after which every netdevice registration is refused with the very message this test asserts. The hoisted status is what avoids that, and it is the one non-obvious line of the change.

The fixup adds a third notifier that raises once from its replay, with error(msg, 0) so the message carries no position and check_dmesg stays clean; this accepted registration then pins the clear on the error path as well. From the run: luanotifier: notifier inside test: raised followed by notifier inside test: after registered. The defective side was not built, so the separation is structural rather than measured: with the clear past the error return, this line prints the refusal instead of registered. Fixup: https://github.com/luainkernel/lunatik/commit/e25fd8fdd3f9409cd9961b1c0eb15b1ac51fa518

lib/luanotifier.c
```diff
@@ -157,10 +178,18 @@ static int luanotifier_netdevice_handler(lua_State *L, void *data)
...
-* @raise if called from a percpu runtime
+* @raise if called from a percpu runtime, or from a netdevice callback
 * @within notifier
 */
-LUANOTIFIER_NEWCHAIN(netdevice, &luanotifier_process_class);
+static int luanotifier_netdevice(lua_State *L)
+{
+	/* the registrar waits on the namespace rwsem and RTNL this task already holds */
+	if (luanotifier_incallback(L))
+		luaL_error(L, "not allowed from a netdevice callback");
```
lneto: (posted by an agent, not by @lneto)

The flag is written around every chain's pcall, so a keyboard or vt callback of a hardirq runtime sets it too, and this guard runs before lunatik_checkruntime. notifier.netdevice from inside a keyboard callback therefore answers "not allowed from a netdevice callback" where master answers "runtime context mismatch": the message names a callback that did not run, and it reads as though moving the call out of the callback would help, which in a hardirq runtime it would not. No capability is lost, since the process class was already refused there, and the suite is green either way.

"not allowed from a notifier callback" is true wherever the guard can fire and stays exact in a process runtime, where netdevice is the only process-class chain, and it is the condition the predicate tests; the @raise and inside.sh's REFUSAL follow it, and the notifier suite ran 26 of 26 on it. Fixup: https://github.com/luainkernel/lunatik/commit/9c434687a0a0f2bc0dade25f87306a73846ca99d

lneto (commented):

COMMENT: the guard is sound and this can merge once the two fixups below are squashed.

@lneto, built and run at the author head 08233a938, at the fixup head e25fd8fdd and at the fixup head 9c434687a, installed and run under the host lock on 6.8.0-138 arm64 with the running core at srcversion 7A881A956F958C559DD7C4E. Notifier suite 26 of 26 on every head (context_mismatch 2, init_dispatch 2, chain_continues 3, netns_scope 6, replay 7, inside 6) and the full suite pass:263 fail:0 skip:11 on the first two, with no not ok anywhere. Two of the implementer's three full runs saw tests/netlink/nl80211_station.sh fail with ENETDOWN on the same build; neither of the reviewer's two runs did, which measures those runs and not the flake, and it leaves as #1010.

The trace holds against the v6.8 source: register_netdevice_notifier takes down_write(&pernet_ops_rwsem) then rtnl_lock() and replays REGISTER and UP on the registering task with both held (https://github.com/torvalds/linux/blob/v6.8/net/core/dev.c#L1741-L1752), the chain call asserts RTNL (https://github.com/torvalds/linux/blob/v6.8/net/core/dev.c#L1957), and unregister takes the same two (https://github.com/torvalds/linux/blob/v6.8/net/core/dev.c#L1788-L1789). There is no per-task query to key on instead: rtnl_is_locked() is mutex_is_locked() (https://github.com/torvalds/linux/blob/v6.8/net/core/rtnetlink.c#L162-L165) and lockdep_rtnl_is_held() exists only under CONFIG_PROVE_LOCKING (https://github.com/torvalds/linux/blob/v6.8/net/core/rtnetlink.c#L174-L180). Leaving keyboard and vt unguarded is right: atomic_notifier_chain_register takes only the chain's spinlock (https://github.com/torvalds/linux/blob/v6.8/kernel/notifier.c#L155-L157). The three Lua claims hold at the pinned submodule: LUA_REGISTRYINDEX resolves to G(L)->l_registry (https://github.com/luainkernel/lua/blob/d7ca49b3a9c44278078b4e12a7cb26bf4c9e30fd/lapi.c#L72), lua_newthread copies the extra space at creation (https://github.com/luainkernel/lua/blob/d7ca49b3a9c44278078b4e12a7cb26bf4c9e30fd/lstate.c#L291-L292), and luaD_throw with no handler calls abort() (https://github.com/luainkernel/lua/blob/d7ca49b3a9c44278078b4e12a7cb26bf4c9e30fd/ldo.c#L144). The shape is #851's, key, setter, predicate and seed, and the seed sits in the shared constructor, which every chain needs since the dispatcher writes the flag for every chain; that the two bindings now carry the same pair is #1011.

A note for the squash, on the first commit's last sentence and the same sentence of this body. The flag is per state, so "a runtime created or stopped from inside the callback is not covered" has a third door: runtime:resume runs the target state's code on the calling task with no context refusal (lunatik_lresume, lunatik_core.c), and that state has its own registry, so a netdevice callback of A doing B:resume(...) where B registers reaches the registrar on the RTNL-holding task; "created, resumed or stopped" completes it. The out-of-tree consumer reaches its notifier.netdevice through exactly that resume, from its daemon's body and not from a callback, so nothing there changes. Those cells, and the unregister family, getmetatable(n).__gc(n) and B:stop() on a runtime holding a netdevice block, both reaching unregister_netdevice_notifier on the RTNL-holding task, stay uncovered and unguarded, since each wedges the host that exercises it; they leave as #1009.

Examples, from tools/checks/examples-touched.sh: spyglass ran and its hook fired, with keycodes 30, 48 and 46 injected through a synthetic uinput keyboard and /dev/spyglass returning abc, so the two new lua_rawsetp writes executed inside a hardirq keyboard callback with interrupts off and dmesg stayed clean; keylocker ran the same way at the implementer's hands, with the konami sequence toggling it. That is the only exercise the hardirq arm gets, since the suite has no keyboard stimulus. ifquarantine was not run: it quarantines the loopback and the uplink on the replay and takes the network off the machine that runs it; read instead, its only constructor call is in the script body and its callback writes an rcu.table and prints, registering nothing.

PR #1012, opened 2026-09-19
core: the registry flag of fsnotify and notifier is one pair

### LINKED ISSUE #1011: core: the registry flag helpers of fsnotify and notifier are one pair
`lib/luafsnotify.c` (#851) and `lib/luanotifier.c` (#1005) each carry a static key, a setter and a predicate for the same fact, a callback of this state is on the stack: `lua<lib>_incallback_key`, `lua<lib>_setincallback(L, on)` and `lua<lib>_incallback(L)`, byte for byte the same but for the prefix, plus the same seed line in each constructor, `setincallback(L, incallback(L))`, which keeps the registry node in place so the dispatcher's write outside any protected call never allocates.

The key stays per binding: the fact is per binding, so fsnotify's callback must not set notifier's flag, and #851's determination put it in the registry under the binding's own key for that reason. What can move to the core is the mechanism: a pair in `lunatik.h` that takes the key, in the shape and naming of the helpers already there (`lunatik_getregistry`, `lunatik_register` and `lunatik_unregister`), with the seed as a third member or folded into the setter's contract, so a binding declares its key and calls the pair, and a third binding with the same fact adds a key and nothing else. The cost is a core API for two users, which #851 weighed at one and declined; at two it is the repetition the C style rules collapse into one helper.

Depends on #1005, which introduces the second copy: the change stacks on it and rebases onto master once it merges. Out of scope: the home of the fact, settled in #851, and the per-state limit of #1009.

lneto:
`lib/luafsnotify.c` and `lib/luanotifier.c` each carry a static key, a setter and a predicate for the same fact, a callback of this state is on the stack, byte for byte the same but for the prefix, plus the same seed line in each constructor.

`lunatik_setflag`, `lunatik_getflag` and `lunatik_seedflag` take the key, so a binding declares its key and calls them, and a third one adds a key and nothing else. The key stays the binding's: the fact is per binding, and fsnotify's callback must not set notifier's flag. It subsumes the two copies, and the seed keeps a name of its own because the dispatcher writes the flag outside any protected call, where a first insertion can abort the state.

Merges after #1005, which introduces the second copy. Closes #1011.


doc/capi.md
```diff
@@ -423,6 +423,32 @@ Pushes the value stored in `LUA_REGISTRYINDEX` at `key` and returns it as a Luna
 registered (an `skb`, a handle) reads it through this rather than through `lunatik_toobject`,
 which does not check.
 
+### lunatik\_setflag
+```C
+void lunatik_setflag(lua_State *L, const void *key, bool on);
+```
+Stores `on` in `LUA_REGISTRYINDEX` at `key`, where a binding keeps a fact its own dispatcher sets
+and its own code reads, as `fsnotify` and `notifier` keep "a callback of this state is on the
```
lneto: (posted by an agent, not by @lneto)

The illustration names the two bindings that adopt the pair today, and #1011 asks for a third to need nothing but a key, so the list goes stale on that adoption; the sentence loses nothing without the names. Fixup: https://github.com/luainkernel/lunatik/commit/c55eeb6acb51b3f6a255461445b25bb6ce404b5e

doc/capi.md
```diff
@@ -423,6 +423,32 @@ Pushes the value stored in `LUA_REGISTRYINDEX` at `key` and returns it as a Luna
...
+bool lunatik_getflag(lua_State *L, const void *key);
+```
+Returns the flag stored at `key`, `false` where nothing is stored, and leaves the stack as it found
+it.
+
+### lunatik\_seedflag
+```C
+void lunatik_seedflag(lua_State *L, const void *key);
+```
+Writes the flag back to itself, so the registry node exists without what it says changing. A
```
lneto: (posted by an agent, not by @lneto)

The restriction sits on the wrong entry: what a dispatcher's author has to know is that their write must not be the first at that key, and what they read is lunatik_setflag, where nothing said so. The fixup gives each entry its own: setflag ends with the dispatcher seeding the key in its constructor, and seedflag says it is itself the write that can insert, so a constructor calls it where a protected call is on the stack. Fixup: https://github.com/luainkernel/lunatik/commit/7d901357777ea4a74a08eef40eb5a6c72d1c6184

doc/capi.md
```diff
@@ -423,6 +423,32 @@ Pushes the value stored in `LUA_REGISTRYINDEX` at `key` and returns it as a Luna
 registered (an `skb`, a handle) reads it through this rather than through `lunatik_toobject`,
 which does not check.
 
+### lunatik\_setflag
+```C
+void lunatik_setflag(lua_State *L, const void *key, bool on);
```
lneto, line 428: (posted by an agent, not by @lneto)

The three new entries type key as const void *, which is what lua_rawgetp and lua_rawsetp take (https://github.com/luainkernel/lua/blob/d7ca49b3a9c44278078b4e12a7cb26bf4c9e30fd/lua.h#L287, https://github.com/luainkernel/lua/blob/d7ca49b3a9c44278078b4e12a7cb26bf4c9e30fd/lua.h#L304) and what both keys are; lunatik_getregistry thirteen lines above, and lunatik_register, lunatik_unregister and lunatik_getregistryobject in the header, still say void * (https://github.com/luainkernel/lunatik/blob/68773e04c3a1c498645c519a1d19327efbda25b9/lunatik.h#L463). The trio is the right side of that split, and widening the other three builds clean with no call site changed, measured on this host; that is master's cleanup, #1013, not this branch's.

lunatik.h
```diff
@@ -484,6 +484,26 @@ static inline void lunatik_unregisterobject(lua_State *L, lunatik_object_t *obje
 	lunatik_unregister(L, object); /* remove object, now it might be GC'ed */
 }
 
+static inline void lunatik_setflag(lua_State *L, const void *key, bool on)
```
lneto, line 487: (posted by an agent, not by @lneto)

lunatik_setflag reads beside lunatik_setstring and lunatik_setinteger in this header (https://github.com/luainkernel/lunatik/blob/68773e04c3a1c498645c519a1d19327efbda25b9/lunatik.h#L420), which copy a named table field into a struct field; flag not being a Lua type name is what keeps it apart, and getflag and seedflag have no such neighbour, so the trio as named is defensible. If the reading bothers you, lunatik_setregistryflag and lunatik_getregistryflag sit beside lunatik_getregistryobject, at eight call sites and three doc entries; a naming call, left to you.

lneto (commented):

COMMENT: mergeable after #1005, with the two doc/capi.md fixups folded; nothing to fix in the C.

@lneto, a behaviour-preserving move: the four call sites per binding are the same writes and reads in the same places, the seed still preserves rather than clears, and the key stays each binding's static const char, so fsnotify's callback cannot answer for notifier's. Built and run through the host lock on 6.8.0-138 arm64 with the installed lunatik.ko, luafsnotify.ko and luanotifier.ko byte identical to the worktree and the running core at srcversion 4D567204E45AA944AAD8010: fsnotify 61 pass 0 fail 8 skip (kinds, no mount kind before 6.10), notifier 26 pass 0 fail, whole tree 263 pass 0 fail 11 skip, dmesg clean. spyglass and keylocker ran through tools/watchdog.sh with key events from a uinput device, which is the one place lunatik_setflag runs with interrupts off and no suite reaches; ifquarantine was not run, since it quarantines the loopback and the uplink of the host that loads it.

The hazard the seed answers holds at the pinned Lua: luaH_pset writes in place and returns HOK whenever the value is present, and only an absent one reaches luaH_newkey (https://github.com/luainkernel/lua/blob/d7ca49b3a9c44278078b4e12a7cb26bf4c9e30fd/ltable.h#L76-L79), so after the seed no write at that key allocates; the protected call a constructor counts on is on every route into one, the script body included (https://github.com/luainkernel/lunatik/blob/68773e04c3a1c498645c519a1d19327efbda25b9/lunatik_core.c#L276). A seed in each opener instead of each constructor was weighed and left: fsnotify's opener is hand written since #852, but notifier's is LUNATIK_NEWLIB's, so that shape costs notifier a hand-written opener to gain one line.

Nothing is owed on tests: between tests/notifier/inside.sh and tests/fsnotify/inside.sh every operation of the trio runs, set and clear on the returning and the raising arm, the read from a callback, from a coroutine of it and from a script body, and both of fsnotify's resolution paths. Four cells stay uncovered and this change opens none of them, all uncovered on master and on #1005 as well: fsnotify's clear at its own call site, its uncached read through attachmark, and the seed's preserve in each binding. The const void * split with the rest of the registry family leaves as #1013. comment-style.sh reports lib/luanotifier.c:100, the two-line note in luanotifier_release, which predates the branch (03e91d9bf) and sits outside every hunk.

After #1005 is squashed this rebases onto master with --onto, and lib/luafsnotify.c will conflict with #852's hunks around the same pcall and opener; that resolution is the rebase's, not a change here.

Issue #1010, opened 2026-09-19
tests: nl80211_station fails with ENETDOWN on some full-suite runs


lneto:
Seen on the 6.8.0-138 arm64 host while #1005 was under test: `tests/netlink/nl80211_station.sh` reported `not ok 1 Lua error in script` with `# ENETDOWN` on two of three full-suite runs of the same build, and passed on the third and on two later runs of the same modules. The script loads no notifier, so the variable is not that change; a symptom that comes and goes from the same inputs is state on the host, and the failing run's log was not kept, so the mechanism below is a hypothesis.

`nl80211_station.lua` brings the AP interface up with `RTM_SETLINK` and issues `START_AP` right after, and both `NL80211_CMD_START_AP` and `NL80211_CMD_NEW_STATION` carry `NL80211_FLAG_NEED_NETDEV_UP`, which `nl80211_pre_doit` answers with `-ENETDOWN` while `wdev_running(wdev)` is false (net/wireless/nl80211.c at v6.8). What the script does not read is whether the set link left the interface running; if the link set can fail or complete after the ack, that is the window.

First step: capture a failing run's log (`dmesg` from the mark and the KTAP output), then read the set link's answer in the script before theorising further.

lneto:
(posted by an agent, not by @lneto)

[...]

PR #1016, opened 2026-09-19
core: the registry key is const void * across the family

### LINKED ISSUE #1013: core: the registry key is const void * across the family
`lunatik_register`, `lunatik_unregister` and `lunatik_getregistryobject` in `lunatik.h`, and the prototype `doc/capi.md` gives `lunatik_getregistry`, type the registry key as `void *`, while `lua_rawgetp` and `lua_rawsetp`, which they wrap, take `const void *`, and the keys the bindings declare are `static const char`. The flag trio #1012 adds already takes `const void *`, so one family now types one argument two ways, and `lunatik_ebpf_env_key` is a non-const `static char` only because `lunatik_register` cannot take a const one.

Widening the three to `const void *key`, and the `capi.md` prototype with them, builds clean with no call site changed (measured on the 6.8 host during #1012's review), and lets `lunatik_ebpf_env_key` become `static const char` like the two flag keys. A change of its own, after #1012.

lneto:
The registry family types one argument two ways. `lunatik_register`, `lunatik_unregister` and `lunatik_getregistryobject` take the key as `void *`, while `lunatik_getregistry` is a macro over `lua_rawgetp` and so takes what Lua takes, `const void *`. A binding can read a `static const char` key through the macro and cannot write the same key through `lunatik_register`, which is the only reason `lunatik_ebpf_env_key` is a non-const `static char`.

The three helpers now take `const void *key`, along with the prototypes `doc/capi.md` gives for `lunatik_getregistry` and `lunatik_getregistryobject`, and `lunatik_ebpf_env_key` becomes `static const char`, like the key `lib/luafsnotify.c` declares. No call site changes and none gains or loses a cast: widening a parameter rejects nothing that was accepted before.

#1012 adds `lunatik_setflag`, `lunatik_getflag` and `lunatik_seedflag`, which already take `const void *key`, so this relates to it but does not depend on it: the two merge in either order, and a three way merge of both against their merge base reports no conflict in `lunatik.h`, `doc/capi.md` or `lunatik_ebpf.h`.
Closes #1013


lunatik.h
```diff
@@ -194,7 +194,7 @@ static inline lunatik_opt_t lunatik_checkcontext(lua_State *L, int ix)
 }
 
 #define lunatik_setruntime(L, libname, priv)	((priv)->runtime = lunatik_checkruntime((L), lua##libname##_class.opt))
-#define lunatik_monitormt(class, monitor)	((monitor) ? (void *)&(class)->opt : (void *)(class))
+#define lunatik_monitormt(class, monitor)	((monitor) ? (const void *)&(class)->opt : (const void *)(class))
```
lneto, line 197: (posted by an agent, not by @lneto)

Both arms cast a const lunatik_class_t * to void * for callees that take const void *, the one registry key left spelled the old way once the three helpers widen; both casts stay, since the arms point at different types. Fixup: https://github.com/luainkernel/lunatik/commit/e9eecee9505e5794e8d4b2ce458a22b2f821b915

lneto (commented):

COMMENT: this can merge with the fixup below folded and the docs commit kept; a comment rather than an approval only because the posting account authored the pull request.

@lneto, the six lines are the whole change and they do what the subject says. Built at 9a261d07c on 6.8.0-138 arm64, installed and run under the host lock with all 29 modules byte identical to the worktree and the running core at srcversion BFD81E1E0AEFD3FB397E467: the whole suite pass:332 fail:0 skip:11, the skips being tests/sched (no /sys/kernel/sched_ext) and tests/fsnotify/kinds (no mount kind before 6.10); tests/tc ran 6 of 6 and tests/xdp 7 of 7, so the two sites that carry the change, lunatik_ebpf.h:80 and :88, ran rather than only compiled. make C=2 at this head and again with master's two headers over the same worktree gives the same 6367 diagnostics on each side, so the change adds none. The premise holds on the tree: master's lunatik.h with this branch's lunatik_ebpf.h gives exactly two -Wdiscarded-qualifiers, at :80 and :88, in each of the three units that include the header. Run again at 2dd605fd4, with the fixup and the docs commit: 28 modules and the core byte identical to the worktree, core 47E37CE5CD08AE852593CB0, pass:328 fail:1 skip:11, the one failure being tests/netlink/nl80211_station.sh with ENOENT, which passed 4 of 4 on a rerun of that suite against the same install and is the flake #1010 tracks; nothing in this branch reaches it.

The 38 call sites are unchanged and none casts; the only other key a binding declares, luafsnotify_incallback_key, is already static const char. One measured effect, the only observable one: const moves the key from .bss to .rodata, so luatc.ko and luaxdp.ko gain an add in the address computation while every other module's .text stays byte identical; the key's identity is untouched, since the offset is distinct from every other object in the module's .rodata and the module compile flags carry no -fmerge-all-constants. lua_rawgetp and lua_rawsetp take the key as const void * at the pinned submodule (https://github.com/luainkernel/lua/blob/d7ca49b3a9c44278078b4e12a7cb26bf4c9e30fd/lua.h#L287, https://github.com/luainkernel/lua/blob/d7ca49b3a9c44278078b4e12a7cb26bf4c9e30fd/lua.h#L304).

One key producer was left spelled the old way: lunatik_monitormt discarded const in both arms, from the const lunatik_class_t * its three callers hold, for lua_rawgetp and lua_rawsetp called directly (https://github.com/luainkernel/lunatik/blob/9a261d07cd83ed29de444999da0235699d61fd2f/lunatik.h#L197, callers at L219, L300 and L355); the fixup inline makes the subject literally true of the header. doc/capi.md had no entry for lunatik_register or lunatik_unregister, so two of the three widened helpers had no prototype to update; the second commit on the branch adds them, in the shape of the entries beside them.

On the order: #1013 says after #1012 and the branch is off master, which leaves the choice open. git merge-tree of this head and 7d9013577 against their merge base leaves lunatik.h and doc/capi.md clean, and #1012's three flag helpers already take const void *; the one conflict the cross merge shows, lib/luafsnotify.c, appears identically for origin/master against 7d9013577, so it is #1012's rebase over #852, not this branch's.

PR #1018, opened 2026-09-19
checks: a run of run.sh is the command, a failure is read before it is posted, and the disk is checked before a build

lneto:
Three things one afternoon of reviews paid for twice: two command guards read a path to a suite's `run.sh` inside a `git diff` or a check as a suite run and refused it while another session's suite held the device; a suite failure was answered by a rerun that passed and posted as the intermittent failure #1010 tracks, when the journal around the failing run had NetworkManager and wpa_supplicant taking the AP interface the test had just brought up; and `make` died on a full disk halfway through autogen, with five gigabytes of the checkout in scratch worktrees nobody removed.

`commands.sh` reads a run as the command word, bare or under `sudo`, `bash` or `sh`, and the lock and the crash guard source it. `untraced.sh` refuses the word for a failure nobody read in a review, a comment or a pull request body, through `review-post-guard.sh` and `pr-body.sh`, and `tools/journal.sh` prints every unit's lines around the last kernel line matching a pattern, the read the rule asks for; the lunatik-cycle skill orders it before any rerun and the review-pr skill records where an inline comment anchors and that a review's worktree goes when the review posts. `disk.sh` fails below a free-space threshold naming the largest scratch worktrees, and `tools/lunatik-host` runs it as a warning before every cycle.

Each check was run on the inputs that failed and the ones that passed today, and depends on nothing outside master.


tools/checks/commands.sh
```diff
@@ -0,0 +1,11 @@
+# Sourced by the command guards, lunatik-lock.sh and crash-guard.sh: what a tool command runs,
+# read from the raw hook input, which embeds the command verbatim.
+
+# a suite's run.sh as the command, bare or under sudo, bash or sh; a mention of the file, a path
+# handed to git or to a check, is not a run, and the bare substring read every one of those as one
+runs_suite() {
+	# the command field with its escapes undone: a line of its own is a command, an escaped quote is not
+	printf '%s' "$1" | sed -E 's/^.*"command"[[:space:]]*:[[:space:]]*"//; s/\\n/\n/g' |
+		grep -Eq '(^|[;&|(])[[:space:]]*((sudo|bash|sh)([[:space:]]+-[^[:space:]]+)*[[:space:]]+)*([^[:space:]]*/)?run\.sh([[:space:]]|$|[;&|)"\\])'
```
lneto, line 9: (posted by an agent, not by @lneto)

The first shape read the raw payload, where a shell quote arrives as \" and a newline as \n, and it cut both ways: " in the prefix class gave a quoted path the anchor back, so git log --oneline -- "tests/fsnotify/run.sh", grep -n "run.sh" tools/checks/lunatik-lock.sh and echo "tests/x/run.sh" were refused with a suite running, and \n was in no class, so a run on the second line of a command, or under sudo -n, was not read at all. Any token ending in run.sh matched too, bash tools/dryrun.sh among them. The fixup cuts the command field out of the payload and undoes its escapes before the match, so the pattern is about shell rather than JSON; the trade is that a heredoc writing a run line reads as a run, the safe direction for a lock. Two shapes stay outside it: a run behind tools/lunatik-host, which takes the lock itself, and one inside $(...). Fixup: https://github.com/luainkernel/lunatik/commit/0cefe919af2d214fa1857f17fbc83d318e4e3dba

tools/checks/lunatik-lock.sh
```diff
@@ -9,11 +9,13 @@
...
 input=$(cat)
 
+. "$(dirname "$0")/commands.sh"
+
 case "$input" in
 	*"lunatik test"*|*"lunatik reload"*|*"lunatik load"*|*"lunatik unload"*|\
 	*"lunatik run"*|*"lunatik spawn"*|*"lunatik stop"*|*"make install"*|\
-	*"bash tests/"*|*"/run.sh"*|*"watchdog.sh"*) ;;
-	*) exit 0 ;;
+	*"bash tests/"*|*"watchdog.sh"*) ;;
```
lneto, line 17: (posted by an agent, not by @lneto)

The bare substring survives for the other tokens on this line: with a suite running, git diff origin/master..HEAD -- tools/watchdog.sh and grep -n 'lunatik reload' AGENTS.md are refused, the same afternoon's defect with a different word. Closing it is a predicate per guarded token that reads through sudo and env, wider than this commit; it leaves as #1019.

tools/checks/untraced.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+# A failure that comes and goes is read in the journal and named by its mechanism, or carried
+# as a hypothesis; "flake" is the word for a symptom nobody read, and a review that used it for
+# tests/netlink/nl80211_station passed over NetworkManager taking the test's interface (#1010).
+# Takes text files, a review body, a comment, a pull request body or a hand-back; prints each
+# line that names a failure that way and exits 1, silent otherwise.
+
+status=0
+for file in "$@"; do
+	hits=$(grep -niE '\bflak(e|y|es|ed|iness)\b' "$file")
```
lneto, line 10: (posted by an agent, not by @lneto)

The match is the word anywhere in the file, so the check cannot separate a use from a mention: a link whose path carries it fails, and so does a review that names the rule or quotes an author to disagree, which is why this review is written around it. snowflake and flakeless pass, so the boundaries are right. It sits before the marker in review-post-guard.sh with no override, unlike LUNATIK_LOCK_OK, CRASH_AB_OK, CONSUMERS_OK and NETWORK_LOSS_OK; that is the gate as designed, the way machine-leak.sh is hard, and a text that needs the word says the mechanism instead.

tools/checks/pr-body.sh
```diff
@@ -20,6 +21,7 @@ for file in "$@"; do
 		echo "$file: carries a Test plan section"
 		status=1
 	fi
+	bash "$(dirname "$0")/untraced.sh" "$file" || status=1
```
lneto, line 24: (posted by an agent, not by @lneto)

The header enumerated three rules and the code applied four; the fixup names the fourth. Fixup: https://github.com/luainkernel/lunatik/commit/792dc42b085e7230fd0366792ab9e46e899190a6

tools/journal.sh
```diff
@@ -0,0 +1,27 @@
...
+# SPDX-FileCopyrightText: (c) 2026 Ring Zero Desenvolvimento de Software LTDA
+# SPDX-License-Identifier: MIT OR GPL-2.0-only
+#
+# Prints every unit's journal lines, not only the kernel's, around the last
+# kernel line that matches a pattern. The window a test's own prints sit in is
+# where a host process acting on what the test created shows up: NetworkManager
+# and wpa_supplicant taking the AP interface tests/netlink/nl80211_station had
+# just brought up (#1010) were there, and nowhere in the KTAP output. The suite
+# clears the ring buffer at every test, so dmesg is not the record; the journal is.
+# Needs the journal's read permission: the adm group grants it, sudo otherwise.
```
lneto, line 12: (posted by an agent, not by @lneto)

"which is sudo here" did not hold on this host: ubuntu is in adm, and the same call without sudo returned the same 260 lines and nine units. The sentence and the skill's command line now say the adm group grants it, sudo otherwise. Fixup: https://github.com/luainkernel/lunatik/commit/5518db2e2c9b5ce8b1a930ee40016e6deb222a90. Separately, this file, untraced.sh and disk.sh carried a shebang at mode 644 where every other script in tools/ is 755; two fixups, since the files arrive in two commits: https://github.com/luainkernel/lunatik/commit/83ff054a6662adfcb42ad73b051007f6210cbda7 and https://github.com/luainkernel/lunatik/commit/5936ceb07f4b3773586e4496867b6d65b2a4b97b

tools/journal.sh
```diff
@@ -0,0 +1,27 @@
...
+pattern=$1
+span=${2:-3}
+
+[ -n "$pattern" ] || { echo "usage: $0 <pattern> [seconds]"; exit 1; }
+
+last=$(journalctl -q -k -o short-iso --grep="$pattern" 2>/dev/null | tail -1 | cut -d' ' -f1)
+[ -n "$last" ] || { echo "no kernel line matches $pattern"; exit 1; }
+at=$(date -d "$last" +%s)
+
+echo "journal: the last match is at $last; a rerun after the failure puts the failing run before it" >&2
```
lneto, line 25: (posted by an agent, not by @lneto)

The window is the last match's, and nothing said which timestamp it anchored on, so once a rerun has happened, the habit the rule is fighting, the dump is the rerun's and reads clean; the reviewer hit it, needing 13:56:54 and getting 13:58:16. This line names the anchor on stderr and says what a rerun does to it. Fixup: https://github.com/luainkernel/lunatik/commit/f65405191b2482118e930a06c3fc8bc14404f021

tools/checks/disk.sh
```diff
@@ -0,0 +1,23 @@
...
+min=${1:-1024}
+common=$(git rev-parse --path-format=absolute --git-common-dir 2>/dev/null)
+project=${common:+$(dirname "$common")}
+project=${project:-${CLAUDE_PROJECT_DIR:-$PWD}}
+
+free=$(df -Pm "$project" | awk 'NR == 2 { print $4 }')
+[ "$free" -ge "$min" ] && exit 0
+
+# every worktree git knows: they sit under scratch/wt, under scratch/review and beside the checkout
+sizes=$(git -C "$project" worktree list --porcelain 2>/dev/null | awk -v main="$project" '/^worktree / && $2 != main { print $2 }' |
```
lneto, line 16: (posted by an agent, not by @lneto)

The glob was where one session's worktrees sit, not where they are: of the checkout's 95 worktrees (7338 MB) it saw the 66 under scratch/wt and printed five trees of the same 133 MB. git worktree list is the ground truth AGENTS.md prefers over a structure that encodes it; the fixup reads it, skips the main checkout, and prints the five largest with the count and the total, in 0.33 s here, and the wording in AGENTS.md and the skill goes with it. Fixup: https://github.com/luainkernel/lunatik/commit/568b47491e02fe6cad9bb0acd53b914eededff35

AGENTS.md
```diff
@@ -102,12 +102,17 @@ A worktree named for a task may belong to another session on the same machine. C
...
-end. A `git checkout` carries what is uncommitted onto the new HEAD, where an edit made against the
-old base reads as a change to the new one: switch branches in a tree with nothing pending, or read
-the other branch in a worktree of its own. `git stash` has no place here at all: the stack belongs
-to the repository and not to the worktree, so `stash@{0}` is usually another session's work and the
-pop that follows spreads its conflicts over your tree. Set work aside with a commit, and compare two
-revisions with a worktree or `git show <ref>:<path>`.
+end, with `rm -rf` and `git worktree prune`, since `git worktree remove` refuses a tree with
+submodules. Each one left behind carries a build on the disk every session shares, and a build that
+stops halfway on a full disk leaves the previous install in place for the suite to measure:
+`tools/checks/disk.sh` fails below a free-space threshold and names the largest worktrees, and
```
lneto, line 108: (posted by an agent, not by @lneto)

The sentence was appended to the line the paragraph already carried, leaving 156 characters where the file wraps at 100; the fixup rewraps the paragraph word for word. Fixup: https://github.com/luainkernel/lunatik/commit/5e56d23cb06e6ee8bf5fa5a4917d5d395b9c3a27

lneto (commented):

COMMENT, the only state GitHub takes on a pull request this account authored; in substance it asked for changes, and they are on the branch: merge with the eight fixups folded.

@lneto, an independent agent drove both guards the way Claude Code drives them, a JSON payload on stdin, over twenty-two commands with the device idle and then with a suite-shaped process holding it, and two of the three mechanisms were wrong on inputs a session types every day. commands.sh read a mention as a run whenever the path was quoted, so git log --oneline -- "tests/fsnotify/run.sh" exited 2 with the device busy, and it missed a run written on a second line or under sudo -n, the direction that wedges the device; disk.sh globbed scratch/wt and named five of the ninety-five worktrees this checkout holds, all the same size, missing 939 MB under scratch/review and about 1.6 GB beside the checkout. Both are fixed on the branch and every one of the twenty-two shapes now lands where it should. untraced.sh was run over eight word cases, this pull request's body and the branch's commit messages; tools/journal.sh on a pattern that matches and one that does not; disk.sh above and below its threshold, from inside a worktree and from outside a repository; tools/lunatik-host on a lock of its own, where the warning prints and the wrapper keeps the command's status. .claude/settings.json needs no change: the branch adds no hook, and both new checks are reached through guards already wired.

The mechanism the Findings bullet records holds on this host: the journal still has the run that printed "added" and "duplicate add raises" and never "authorized", with NetworkManager registering the AP interface and wpa_supplicant deinit'ing lunatikap0 in the same second, and mark_dmesg in tests/lib.sh is dmesg -C before every script, so the ring buffer is no record.

Two things the branch does not settle. untraced.sh matches the word wherever it appears, so it cannot tell a use from a mention, and it sits before the marker in review-post-guard.sh with no override; that is the gate as designed, and a review writes around the word, as this one does. The bare substrings survive for watchdog.sh and the CLI verbs on the same case lines, so a mention of those is still refused as an operation while a suite runs; closing it is a predicate per token that reads through sudo and env, and it leaves as #1019.

